PR #3, opened 2024-03-28
add backward for softmax

iclementine:
Add triton kernel for softmax backward(as an example for registering a kernel for autograd key). This is than by registering a function that support autograd to `softmax.int`. This may not be the best practice.

Other alternatives are:
1. register a kernel for `_softmax_backward_data`.
2. register a kernel for `_softmax`(so as not to implement backward for softmax).

### DIFF
--- a/src/gems/softmax.py
+++ b/src/gems/softmax.py
@@ -55,7 +55,52 @@ def softmax_kernel(
     tl.store(output_ptrs, softmax_output, mask=mask)
 
 
-def softmax(x, dim=1, dtype=None, out=None):
+@libentry()
+@triton.autotune(
+    configs=[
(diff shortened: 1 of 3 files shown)


src/gems/softmax.py
```diff
@@ -86,3 +130,58 @@ def softmax(x, dim=1, dtype=None, out=None):
...
+
+    assert dim >= -out.ndim and dim < out.ndim, "Invalid dim"
+    dim = dim % out.ndim
+    M = 1
+    N = out.shape[dim]
+    for i in range(dim):
+        M *= out.shape[i]
+
+    # NOTE: ensure contiguous?
+    in_grad = torch.empty_like(out)
```
StrongSpoon, line 147: Actually not. Need argument memory_format to be `torch.contiguous_format`.

PR #5, opened 2024-04-08
add codegen for general pointwise kernels

iclementine:
1. add codegen utilities for general pointwise kernels( both dynamic & static)
   - dynamic: speciallize to the rank of the task space;
   - static: speciallize to the shape & strides of each input tensor;
2. use pointwise_dynamic to replace abs, gelu, reciprocals, relu, rsqrt, silu and implement sigmoid

### DIFF
--- a/src/flag_gems/gelu.py
+++ b/src/flag_gems/gelu.py
@@ -1,7 +1,6 @@
 import torch
 import triton
 import triton.language as tl
-from .__libentry__ import libentry
 
 from flag_gems.utils.pointwise_dynamic import pointwise_dynamic
 
--- a/src/flag_gems/utils/code_cache.py
+++ b/src/flag_gems/utils/code_cache.py
@@ -3,6 +3,7 @@ import os
 from pathlib import Path
 import shutil
 
+@functools.cache
 def cache_dir_path() -> Path:
     """Return the cache directory for generated files in flaggems."""
     _cache_dir = os.environ.get("FLAGGEMS_CACHE_DIR")
@@ -12,7 +13,6 @@ def cache_dir_path() -> Path:
         _cache_dir = Path(_cache_dir)
     return _cache_dir
 
-@functools.lru_cache(None)
 def cache_dir() -> Path:
     """Return cache directory for generated files in flaggems. Create it if it does not exist."""
     _cache_dir = cache_dir_path()
--- a/src/flag_gems/utils/inliner.py
+++ b/src/flag_gems/utils/inliner.py
@@ -2,7 +2,7 @@ from typing import List
 import ast
 
 from triton.runtime import JITFunction
-from torch.fx.graph import _Namespace
+from flag_gems.utils.code_utils import NameSpace
 
 
 class Rename(ast.NodeTransformer):
@@ -58,7 +58,7 @@ class NameCollector(ast.NodeVisitor):
 
 
 def inline_function(f: JITFunction, input_names: List[str],
-                    output_names: List[str], namespace: _Namespace):
+                    output_names: List[str], namespace: NameSpace):
     nc = NameCollector()
     ast_tree = ast.parse(f.src)
     nc.visit(ast_tree)
@@ -71,7 +71,7 @@ def inline_function(f: JITFunction, input_names: List[str],
 
     # replace temporaries with auto-named names to avoid conflicts
     for name in nc.names:
-        mapping[name] = namespace.create_name(name, None)
+        mapping[name] = namespace.create_name(name)
 
     rename = Rename(mapping)
     renamed_tree = rename.visit(ast_tree)
--- a/src/flag_gems/utils/shape_utils.py
+++ b/src/flag_gems/utils/shape_utils.py
@@ -95,10 +95,8 @@ def volume(shape: Shape) -> int:
 def is_valid_perm(perm: Perm) -> bool:
     r = len(perm)
     sorted_axes = sorted(perm)
-    if sorted_axes[0] != 0:
-        return False
-    for i in range(1, r):
-        if sorted_axes[i] != (sorted_axes[i - 1] + 1):
+    for i in range(r):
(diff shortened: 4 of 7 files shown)


src/flag_gems/gelu.py
```diff
@@ -3,115 +3,37 @@
 import triton.language as tl
 from .__libentry__ import libentry
```
StrongSpoon: this line could be deleted

src/flag_gems/sigmoid.py
```diff
@@ -0,0 +1,39 @@
+import torch
+import triton
+import triton.language as tl
+from flag_gems.utils.pointwise_dynamic import pointwise_dynamic
+
+@pointwise_dynamic
+@triton.jit
+def sigmoid_forward(x):
+    return 1 / (1 + tl.exp(-x.to(tl.float32)))
```
tongxin: Isn't it preferable to convert exp to math.exp2?

src/flag_gems/utils/shape_utils.py
```diff
@@ -0,0 +1,122 @@
+from typing import Tuple, Iterable
+import functools
+import operator
+
+Shape = Tuple[int]
+Stride = Tuple[int]
+MultiIndex = Tuple[int]
+Perm = Tuple[int]
+
+
```
tongxin, line 10: This whole shape util could be slow due to Python interpreter. I would suggest calculating with numpy arrays.

src/flag_gems/sigmoid.py
iclementine: Ok, done.

src/flag_gems/utils/shape_utils.py
```diff
@@ -0,0 +1,122 @@
...
+    d = r2 - r1
+    new_stride = [0 for _ in range(r2)]
+    for i in range(r1):
+        new_stride[d + i] = 0 if (shape[i] == 1 and new_shape[d + i] > 1) else stride[i]
+    return tuple(new_stride)
+
+def volume(shape: Shape) -> int:
+    return functools.reduce(operator.mul, shape, 1)
+
+def is_valid_perm(perm: Perm) -> bool:
```
StrongSpoon, line 95: simplify this fn

src/flag_gems/utils/code_cache.py
```diff
@@ -0,0 +1,27 @@
...
+def cache_dir_path() -> Path:
+    """Return the cache directory for generated files in flaggems."""
+    _cache_dir = os.environ.get("FLAGGEMS_CACHE_DIR")
+    if _cache_dir is None:
+        _cache_dir = Path.home() / ".flaggems"
+    else:
+        _cache_dir = Path(_cache_dir)
+    return _cache_dir
+
+@functools.lru_cache(None)
```
tongxin: I would suggest using functools.cached_property, for cache_dir should be invariant once initiated.

src/flag_gems/utils/code_cache.py
```diff
@@ -0,0 +1,27 @@
...
+def cache_dir() -> Path:
+    """Return cache directory for generated files in flaggems. Create it if it does not exist."""
+    _cache_dir = cache_dir_path()
+    os.makedirs(_cache_dir, exist_ok=True)
+    return _cache_dir
+
+def clear_cache():
+    """Clear the cache directory for code cache."""
+    _cache_dir = cache_dir_path()
```
tongxin, line 24: then here simply querying cache_dir() could be quicker

src/flag_gems/utils/inliner.py
```diff
@@ -0,0 +1,84 @@
...
+    mapping = {}
+    # replace arguemnt names with input names
+    # TODO: handle non-tensor arguments to scalar function
+    for p, input_name in zip(f.arg_names, input_names):
+        mapping[p] = input_name
+
+    # replace temporaries with auto-named names to avoid conflicts
+    for name in nc.names:
+        mapping[name] = namespace.create_name(name, None)
+
```
tongxin: There could come name conflicts with the rewriting. For example, 

foo (a, b):
   x = ...
   y = 

If rewriting arg a, b with input name x, y, it will result in name aliasing.

src/flag_gems/utils/shape_utils.py
iclementine: Done.  Now the perm is sorted and test if sorted[i] = i for all i in [0, r).

src/flag_gems/gelu.py
iclementine: done

src/flag_gems/utils/code_cache.py
iclementine: I use `functools.cache` since it is a standalone function.
Also `functools.cache` is more light-weight than `functools.lrucache`/

src/flag_gems/utils/code_cache.py
iclementine: Done.

src/flag_gems/utils/inliner.py
iclementine: Input names to the inlined function are populated into the namespace before inlining, so in this case, `x`, `y` would be renamed to `x_1`, `y_1`.

Also, now all the args names and local names of the jit function, into which the inlined function is inlined, are populated into the namespace brefore inlining.

src/flag_gems/utils/shape_utils.py
iclementine: done

src/flag_gems/gelu.py
iclementine: done

src/flag_gems/utils/shape_utils.py
iclementine: Some operations used in shape_utils are not simple array operation. So I prefer using python tuples & list. We  may find a better implementation after benchmarking them.

PR #8, opened 2024-05-10
[Operators] vector_norm and var_mean

StrongSpoon:
1. vector_norm: support all legal inputs
    - ord: 2, inf, -inf, 0, other float/int
    - dim: None, list of int
    - keepdim: True, False
    - dtype: float16, float32, bfloat16
2. var_mean: support all legal inputs
    - dim: None, list of int
    - correction: None, float/int
    - keepdim: True, False
3. group_norm by Adam
    - forward finished
    - little error with bf16 weight grad
4. mean_dim: support all legal inputs
    - dim: None, list of int
    - keepdim: True, False

### DIFF
--- a/src/flag_gems/groupnorm.py
+++ b/src/flag_gems/groupnorm.py
@@ -2,7 +2,6 @@ import torch
 import triton
 import triton.language as tl
 from .__libentry__ import libentry
-import math
 
 
 @libentry()
@@ -222,6 +221,7 @@ def weight_bias_backward_kernel(
     dW,
     dB,
     num_groups,
+    group_size,
     N,
     C,
     HW,
@@ -246,7 +246,7 @@ def weight_bias_backward_kernel(
         order=(1, 0),
     )
     mean_ptr = tl.make_block_ptr(
-        Mean + pid % num_groups,
+        Mean + pid // group_size,
         shape=(N,),
         strides=(num_groups,),
         offsets=(0,),
@@ -254,7 +254,7 @@ def weight_bias_backward_kernel(
         order=(0,),
     )
     rstd_ptr = tl.make_block_ptr(
-        Rstd + pid % num_groups,
+        Rstd + pid // group_size,
         shape=(N,),
         strides=(num_groups,),
         offsets=(0,),
@@ -366,11 +366,12 @@ class GroupNorm(torch.autograd.Function):
             weight_grad,
             bias_grad,
             num_groups,
+            group_size,
             N,
             C,
             HW,
-            BLOCK_N = triton.next_power_of_2(N),
-            BLOCK_HW = triton.next_power_of_2(HW),
+            BLOCK_N=triton.next_power_of_2(N),
+            BLOCK_HW=triton.next_power_of_2(HW),
         )
         return x_grad, weight_grad, bias_grad, None, None, None, None, None
 
--- a/tests/flag_gems/op_accu_test.py
+++ b/tests/flag_gems/op_accu_test.py
@@ -363,33 +363,6 @@ def test_accuracy_gelu(shape, dtype):
 )
 @pytest.mark.parametrize("dtype", [torch.float16, torch.float32, torch.bfloat16])
 def test_accuracy_groupnorm(N, C, H, W, num_groups, dtype):
-    def group_norm_torch_manual(x, num_groups, weight=None, bias=None, eps=1e-5):
-        N, C, H, W = x.shape
-        group_size = C // num_groups
-        # Reshape
-        x = x.view(N, num_groups, group_size, H, W)
-        x = x.reshape(N * num_groups, -1)
-
-        # Mean and Rstd
-        mean = x.mean(dim=1, keepdim=True)
-        var = x.var(dim=1, unbiased=False, keepdim=True)
-        rstd = 1 / torch.sqrt(var + eps)
-
-        # Norm
-        x = (x - mean) * rstd
-        x = x.view(N, C, H, W)
-        if weight is not None:
-            weight = weight.view(1, C, 1, 1)
-            x = x * weight
-        if bias is not None:
-            bias = bias.view(1, C, 1, 1)
-            x = x + bias
-
-        # Reshape
-        mean = mean.reshape(N, num_groups)
-        rstd = rstd.reshape(N, num_groups)
-        return x, mean, rstd
-
     HW = H * W
     inp = torch.randn(size=(N, C, H, W), dtype=dtype, device="cuda", requires_grad=True)
     weight = torch.randn(size=(C,), dtype=dtype, device="cuda", requires_grad=True)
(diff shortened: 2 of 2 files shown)


src/flag_gems/vector_norm.py
```diff
@@ -0,0 +1,170 @@
...
+        shape = list(x.shape)
+        N = 1
+        order = list(range(x.ndim))
+        for i in dim:
+            i = i % x.ndim
+            order.remove(i)
+            order.append(i)
+            N *= shape[i]
+            shape[i] = 1
+        x = x.permute(order)
```
iclementine: I would suggest construct the permutation in a more efficient way, since list.remove involves comparing the value to remove with values in the list. 

In this case, you can just pop the value at a specific index and append it at the back.

```python
order.append(order.pop(i))
```

or just construct the permutation a-fresh without init-and-modify.

```python
order = (i for i in range(x.dim) if i not in dim) + tuple(dim)
```

BTW, should dim be sorted beforehand to avoid harder permutation?

src/flag_gems/vector_norm.py
```diff
@@ -0,0 +1,170 @@
...
+        mask = row_mask and col_mask
+
+        a = tl.load(X + cols, mask, other=0.0).to(tl.float32)
+        _sum += tl.math.pow(tl.abs(a), ord)
+    sum = tl.sum(_sum, axis=1)
+    out = tl.math.pow(sum, 1 / ord)[:, None]
+    tl.store(Out, out, row_mask)
+
+
+def vector_norm(x, ord=2, dim=None, keepdim=False, dtype=None):
```
iclementine, line 128: `dim` should also accept int, in addition to tuple or list of int, according to pytorch convention.

```python
import torch
import flag_gems
x = torch.randn(1, 1, 16, 1, device="cuda")
flag_gems.vector_norm(x, ord=2, dim=2, keepdim=False)
```

src/flag_gems/vector_norm.py
```diff
@@ -0,0 +1,170 @@
...
+    elif ord == float("inf"):
+        max_norm_kernel[grid](x, out, M, N)
+    elif ord == -float("inf"):
+        min_norm_kernel[grid](x, out, M, N)
+    elif ord == 0:
+        l0_norm_kernel[grid](x, out, M, N)
+    else:
+        v_norm_kernel[grid](x, out, M, N, ord)
+    if not keepdim:
+        out = out.squeeze()
```
iclementine: When not `keepdim` only the axis specified in dim should be squeeze out. Or else, other axis with size equals 1 would also be squeezed out.

Use the following snippet to test

```python
import torch
import flag_gems
x = torch.randn(1, 1, 16, 1, device="cuda")
print(torch.linalg.vector_norm(x, ord=2, dim=(2, ), keepdim=False).shape)
print(flag_gems.vector_norm(x, ord=2, dim=(2, ), keepdim=False).shape)
```

src/flag_gems/vector_norm.py
```diff
@@ -0,0 +1,170 @@
...
+@libentry()
+@triton.autotune(configs=cfggen(), key=["M", "N"])
+@triton.jit
+def l0_norm_kernel(X, Out, M, N, BLOCK_M: tl.constexpr, BLOCK_N: tl.constexpr):
+    pid = tl.program_id(0) * BLOCK_M + tl.arange(0, BLOCK_M)[:, None]
+    X = X + pid * N
+    Out = Out + pid
+    row_mask = pid < M
+
+    _sum = tl.zeros([BLOCK_M, BLOCK_N], dtype=tl.float32)
```
iclementine, line 93: I have a concern about data type here. What if the input data is double, then the accumulator type should at least be double. What sbout adding a tl.constexpr parameter `acc_type`?

src/flag_gems/vector_norm.py
StrongSpoon: integer dim is precessed as a list of length 1

src/flag_gems/vector_norm.py
StrongSpoon: this is a common problem for all operators. there are too many expressions using fixed datatype, which is unfriendly for datatype expansion.

src/flag_gems/var_mean.py
```diff
@@ -0,0 +1,97 @@
...
+        dim = 0
+        M = 1
+        N = x.numel()
+        shape = [1] * x.ndim
+        x = x.flatten()
+    else:
+        shape = list(x.shape)
+        N = 1
+        order = list(range(x.ndim))
+        for i in dim:
```
iclementine: same case with vector_norm

src/flag_gems/vector_norm.py
iclementine: Yes, but 
图片

src/flag_gems/vector_norm.py
iclementine: What about just unsqueeze the reduce axes?

src/flag_gems/vector_norm.py
iclementine: I see, flag_gems.vector_norm is not called directly. It is called by overriding aten::linalg_vector_norm. 
Resolved.

src/flag_gems/vector_norm.py
StrongSpoon: done. I think it doesn't matter whether dim is sorted or not, cause input tensor doesn't have to be contiguous.

src/flag_gems/vector_norm.py
StrongSpoon: done

src/flag_gems/var_mean.py
StrongSpoon: done

src/flag_gems/vector_norm.py
iclementine: Yes, the input tensor may not be contiguous.

src/flag_gems/groupnorm.py
```diff
@@ -0,0 +1,317 @@
...
+        Y_val += bias
+
+    tl.store(Y_block_ptr, Y_val.to(X_val.dtype), boundary_check=(0, 1))
+    tl.store(rstd_block_ptr, rstd.to(X_val.dtype))
+    tl.store(mean_block_ptr, mean.to(X_val.dtype))
+
+
+@libentry()
+@triton.jit
+def group_norm_backward_kernel(
```
iclementine, line 118: The backward kernel is launched with grid `(N * num_groups, 1, 1)`. Then the gradient of W & B should not be computed like this.

A block of the kernel loads a block of size `(BLOCK_GROUP_SIZE,)` from W or B `(size: num_groups, group_size)`， while  X, Y & Z(normalized X) have shape (N, num_groups, group_size, H, W).

The gradien of W should be the gradient of Y reduced sum by axes( N, H, W). In a task partitioning with grid `(N* num_groups, 1, 1`, it would require a reduce across blocks to accompolish the reduction.

I suggest the use a different task partitioning scheme for computing the gradient of W & B.

PR #12, opened 2024-05-17
[Operators] v2 operators addition

StrongSpoon:
1. reorganize the source code. single operators and fused operators are placed in two directories, in order to add operators from silicon flow.
2. fix the bug of backward error. group_norm, tanh and sigmoid can pass all test cases new.
3. implement binary pointwise ops using codegen, including div, mul, pow, bitwise_and, bitwise_or.

### DIFF
--- a/tests/flag_gems/op_accu_test.py
+++ b/tests/flag_gems/op_accu_test.py
@@ -160,7 +160,7 @@ def test_accuracy_bitwiseand(shape, dtype):
     with flag_gems.use_gems():
         res_out = torch.bitwise_and(inp1, inp2)
 
-    torch.testing.assert_close(res_out, ref_out, atol=0, rtol=0)
+    assert torch.equal(res_out, ref_out), f"ref_out: {ref_out}, res_out: {res_out}"
 
 
 @pytest.mark.parametrize(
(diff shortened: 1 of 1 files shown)


tests/flag_gems/op_accu_test.py
```diff
@@ -143,6 +143,26 @@ def test_accuracy_addmm(M, N, K, alpha, beta, dtype):
...
+    )
+    inp2 = torch.randint(
+        low=-0x7FFF, high=0x7FFF, size=shape, dtype=dtype, device="cuda"
+    )
+
+    ref_out = torch.bitwise_and(inp1, inp2)
+    with flag_gems.use_gems():
+        res_out = torch.bitwise_and(inp1, inp2)
+
+    torch.testing.assert_close(res_out, ref_out, atol=0, rtol=0)
```
tongxin: why not testing eq?

src/flag_gems/ops/tanh.py
```diff
@@ -21,9 +21,9 @@ class Tanh(torch.autograd.Function):
     @staticmethod
     def forward(ctx, A):
         logging.debug("GEMS TANH FORWARD")
-        O = tanh_forward(A)
+        O = tanh_forward(A.to(torch.float32))
         ctx.save_for_backward(O)
```
tongxin, line 25: do we always need to do save for backward regardless gradient context?

tests/flag_gems/op_accu_test.py
StrongSpoon: because I forgot orz

src/flag_gems/ops/tanh.py
StrongSpoon: no, we only save variants needed in backward

tests/flag_gems/op_accu_test.py
StrongSpoon: done

PR #15, opened 2024-05-20
add log_softmax && outer

FatJhon:
add ops of log_softmax，outer，detail functions are as follows：

log_softmax(tensor, dim, dtype)  :bfloat16 is not supported
outer(tensor,tensor)

### DIFF
--- a/src/flag_gems/ops/__init__.py
+++ b/src/flag_gems/ops/__init__.py
@@ -94,6 +94,5 @@ __all__ = [
     "var_mean",
     "vector_norm",
     "log_softmax",
-    "outer"
-    
+    "outer",
 ]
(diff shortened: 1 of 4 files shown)


src/flag_gems/outer.py
```diff
@@ -0,0 +1,49 @@
...
+        if __debug__:
+            print("GEMS OUTER VJP")
+        assert out_grad.ndim == 2, "invalide out_grad shape"
+        out_grad_trans = torch.transpose(out_grad, 0, 1)
+        inp, weight = ctx.saved_tensors
+
+        inp_shape = inp.shape
+        weight = weight[None, :]
+        weight = weight.contiguous()
+        inp_grad_mid = mm(weight, out_grad_trans)
```
iclementine: Using `mm(out_grad, weight[:, None])` is more efficient here because a transposition of out_grad(a matrix) is not needed.

FatJhon: thanks for your optimization suggestions，done

PR #23, opened 2024-05-21
Memory efficient dropout

tongxin:
The Philox RNG algorithm is solely determined by its input state, a 4 int32 tuple (c0, c1, c2, c3). Pytorch uses c0 for seed, c1 for thread counter, and c2 for intra-thead counter (aka offset), whereas in Triton c0 used for seed, and c1 used for offset. The crucial issue is Triton RNG using int32 offsets lacks randomness. This may have negative impact on large model trainings. Triton does allow an int64 offset, but then it won't split it up in two int32 counters in the algorithm. This discrepancy makes it hard for us to align the dropout implementation with Pytorch.

### DIFF
--- a/src/flag_gems/utils/random_utils.py
+++ b/src/flag_gems/utils/random_utils.py
@@ -1,0 +1,15 @@
+import torch
+
+# This function returns the current state of the default Philox RNG in seed and offset and
+# updates the next offset by adding `increment`.
+def philox_cuda_seed_offset(increment, device=None):
+    device = device or torch.cuda.current_device()
+    gen = torch.cuda.default_generators[device]
+    state_copy = gen.get_state()
+    c0, c1, c2, _ = state_copy.view(torch.int32)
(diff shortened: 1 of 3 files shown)


src/flag_gems/utils/random_utils.py
```diff
@@ -0,0 +1,15 @@
+import torch
+
+# This function returns the current state of the default Philox RNG in seed and offset and
+# updates the next offset by adding `increment`.
+def philox_cuda_seed_offset(increment, device=None):
+    device = device or torch.cuda.current_device()
+    gen = torch.cuda.default_generators[device]
+    state_copy = gen.get_state()
+    c0, c1, c2, _ = state_copy.view(torch.int32)
```
iclementine, line 9: is `c1` not used here?
It looks like only `c1, c2, c3` is used as counter and `c0` used as key, but only c2 is updated.

PR #24, opened 2024-05-21
[Operator] add all, any, sum.dim, amax.dim

GwokHiujin:
1. Added all, any operators to support the definitions in the API documentation.
2. Implemented *dim* for sum and amax, which can support int or int[] inputs.

### DIFF
--- a/src/flag_gems/ops/all.py
+++ b/src/flag_gems/ops/all.py
@@ -62,7 +62,7 @@ def all_kernel_1(
     inp_val = tl.load(inp_ptrs, mask=mask, other=1.0)
     all_val = tl.min(tl.abs(inp_val), axis=0)
     mid_ptr = mid + pid
-    mid_mask = mid_ptr < mid_size
+    mid_mask = pid < mid_size
     tl.store(mid_ptr, all_val, mask=mid_mask)
 
 
@@ -96,29 +96,25 @@ def all(inp):
 
 def all_dim(inp, dim=None, keepdim=False): 
     logging.debug("GEMS all_dim")
-    if dim is None:
-        dim = list(range(inp.ndim))
-    else:
-        dim = [dim]
-    assert ((i >= -inp.ndim and i < inp.ndim) for i in dim), "Invalid dim" 
-    dtype = inp.dtype
-
     shape = list(inp.shape)
-    dim = [d % inp.ndim for d in dim]
-    order = [i for i in range(inp.ndim) if i not in dim] + dim
-    inp = inp.permute(order).contiguous()
-    N = 1
-    for i in dim:
-        N *= shape[i]
-        shape[i] = 1
-    M = inp.numel() // N
+    if dim is None:
+        out = all(inp).reshape(shape)
+        dim = shape
+    else:
+        assert (dim >= -inp.ndim and dim < inp.ndim) , "Invalid dim" 
 
-    out = torch.empty(shape, dtype=torch.bool, device=inp.device)
+        dim = dim % inp.ndim
+        inp = inp.contiguous()
+        N = shape[dim]
+        shape[dim] = 1
+        M = inp.numel() // N
 
-    grid = lambda meta: (
-        triton.cdiv(M, meta["BLOCK_M"]),
-    )
-    all_kernel_dim[grid](inp, out, M, N)
+        out = torch.empty(shape, dtype=torch.bool, device=inp.device)
+
+        grid = lambda meta: (
+            triton.cdiv(M, meta["BLOCK_M"]),
+        )
+        all_kernel_dim[grid](inp, out, M, N)
     if not keepdim:
         out = out.squeeze(dim=dim)
     return out
--- a/src/flag_gems/ops/any.py
+++ b/src/flag_gems/ops/any.py
@@ -96,29 +96,25 @@ def any(inp):
 
 def any_dim(inp, dim=None, keepdim=False):
     logging.debug("GEMS any_dim")
-    if dim is None:
-        dim = list(range(inp.ndim))
-    else:
-        dim = [dim]
-    assert ((i >= -inp.ndim and i < inp.ndim) for i in dim), "Invalid dim" 
-    dtype = inp.dtype
-
     shape = list(inp.shape)
-    dim = [d % inp.ndim for d in dim]
-    order = [i for i in range(inp.ndim) if i not in dim] + dim
-    inp = inp.permute(order).contiguous()
-    N = 1
-    for i in dim:
-        N *= shape[i]
-        shape[i] = 1
-    M = inp.numel() // N
+    if dim is None:
+        out = any(inp).reshape(shape)
+        dim = shape
+    else:
+        assert (dim >= -inp.ndim and dim < inp.ndim) , "Invalid dim" 
 
-    out = torch.empty(shape, dtype=torch.bool, device=inp.device)
+        dim = dim % inp.ndim
+        inp = inp.contiguous()
+        N = shape[dim]
+        shape[dim] = 1
+        M = inp.numel() // N
 
-    grid = lambda meta: (
-        triton.cdiv(M, meta["BLOCK_M"]),
-    )
-    any_kernel_dim[grid](inp, out, M, N)
+        out = torch.empty(shape, dtype=torch.bool, device=inp.device)
+
+        grid = lambda meta: (
+            triton.cdiv(M, meta["BLOCK_M"]),
+        )
+        any_kernel_dim[grid](inp, out, M, N)
     if not keepdim:
         out = out.squeeze(dim=dim)
     return out
--- a/src/flag_gems/ops/sum.py
+++ b/src/flag_gems/ops/sum.py
@@ -60,16 +60,15 @@ def sum_kernel(
     out = out + pid
     row_mask = pid < M
 
-    _all = tl.zeros([BLOCK_M, BLOCK_N], dtype=tl.float32)
+    _sum = tl.zeros([BLOCK_M, 1], dtype=tl.float32)
     for off in range(0, N, BLOCK_N):
         cols = off + tl.arange(0, BLOCK_N)[None, :]
         col_mask = cols < N
         mask = row_mask and col_mask
 
         a = tl.load(inp + cols, mask, other=0.0).to(tl.float32)
-        _all += a
-    all = tl.sum(_all, axis=1)[:, None]
-    tl.store(out, all, row_mask)
+        _sum += tl.sum(a, axis=1)[:, None]
+    tl.store(out, _sum, row_mask)
 
 
 def sum(inp, *, dtype=None):
--- a/tests/flag_gems/op_accu_test.py
+++ b/tests/flag_gems/op_accu_test.py
@@ -1192,7 +1192,7 @@ def test_accuracy_sum(shape, dtype):
 )
 @pytest.mark.parametrize("keepdim", [True, False])
 @pytest.mark.parametrize("dim", [[0, 1], 0, 1, None])
-@pytest.mark.parametrize("dtype", [torch.float32])
+@pytest.mark.parametrize("dtype", [torch.float16, torch.float32, torch.bfloat16])
 def test_accuracy_sum_dim(shape, dim, keepdim, dtype):
     inp = torch.randn(shape, dtype=dtype, device="cuda")
 
@@ -1358,10 +1358,10 @@ def test_accuracy_all_dim(shape, dim, keepdim):
 
 @pytest.mark.parametrize(
     "shape",
-    [(1024, 1024, 16), (16, 1024, 256), (16, 128, 64, 64), (20, 320, 15)],
+    [(1024, 1024), (16, 1024, 256), (16, 128, 64, 64), (20, 320, 15)],
 )
+@pytest.mark.parametrize("dim", [-1, 0, 1, None, [1, 0]])
 @pytest.mark.parametrize("keepdim", [True, False])
-@pytest.mark.parametrize("dim", [[1, 0], [1, 2], 0, 1])
 def test_accuracy_all_dims(shape, dim, keepdim):
     inp = torch.rand(shape, device="cuda") < 0.5
 


src/flag_gems/__init__.py
```diff
@@ -23,8 +24,6 @@ def enable(lib=aten_lib):
     lib.impl("isinf", isinf, "CUDA")
     lib.impl("isnan", isnan, "CUDA")
     lib.impl("native_layer_norm", layer_norm, "AutogradCUDA")
-    lib.impl("rms_norm", rms_norm, "CUDA")
```
FatJhon, line 26: misdeletion？

GwokHiujin: Sorry, it looks like there may have been a mistake during a branch merge at some point. It has been fixed now.

src/flag_gems/ops/all.py
```diff
@@ -0,0 +1,154 @@
...
+    )
+    all_kernel_dim[grid](inp, out, M, N)
+    if not keepdim:
+        out = out.squeeze(dim=dim)
+    return (out == 0.)
+    
+
+def all_dims(inp, dim=None, keepdim=False):
+    logging.debug("GEMS all_dims")
+    assert (dim is None) or ((i >= -inp.ndim and i < inp.ndim) for i in dim), "Invalid dim"
```
tongxin: Scalar dim values can fail. Convert dim into tuple or list first.

src/flag_gems/ops/all.py
```diff
@@ -0,0 +1,154 @@
...
+    assert (dim is None) or ((i >= -inp.ndim and i < inp.ndim) for i in dim), "Invalid dim" 
+    dtype = inp.dtype
+
+    if dim is None:
+        dim = list(range(inp.ndim))
+
+    shape = list(inp.shape)
+    dim = [d % inp.ndim for d in dim]
+    order = [i for i in range(inp.ndim) if i not in dim] + dim
+    inp = inp.permute(order).contiguous()
```
tongxin: This can be heavy.  Considering A with dims (0, 1, 2), all(A, dim=(2, 1)) equals all(A, dim=(1, 2)). However, the former is causing a memory copy.

src/flag_gems/ops/all.py
```diff
@@ -0,0 +1,154 @@
...
+
+    _all = tl.zeros([BLOCK_M, BLOCK_N], dtype=tl.float32)
+    for off in range(0, N, BLOCK_N):
+        cols = off + tl.arange(0, BLOCK_N)[None, :]
+        col_mask = cols < N
+        mask = row_mask and col_mask
+
+        a = tl.load(inp + cols, mask, other=1.0).to(tl.float32)
+        _all += (a == 0)
+    all = tl.max(_all, axis=1)[:, None]
```
tongxin: The kernel result is NOT ALL. It then gets negated after the kernel is called. Can we simply the two steps into one?

tests/flag_gems/op_accu_test.py
```diff
@@ -1288,3 +1251,116 @@ def test_accuracy_outer(shape, dtype):
...
+@pytest.mark.parametrize(
+    "shape",
+    [(4096, i * 64) for i in range(1, 20)],
+)
+
+@pytest.mark.parametrize("dtype", [torch.float16, torch.float32, torch.bfloat16, torch.bool])
+def test_accuracy_all(shape, dtype):
+    if (dtype == torch.bool):
+        inp = torch.randn(shape, dtype=torch.float32, device="cuda") > 0
+    else:
```
tongxin: This test case has a strong bias towards all() == False

GwokHiujin: Randomly generated boolean cases all have similar issues, let's try to address that by adding options for all True or all False cases.

src/flag_gems/ops/all.py
GwokHiujin: Okay, originally I thought the *.dims interface only accepted input in the form of a dim tuple. This has been done.

tests/flag_gems/op_accu_test.py
```diff
@@ -523,43 +523,6 @@ def test_accuracy_layernorm(shape, dtype):
     allclose_with_dtype(res_bias_grad, ref_bias_grad, dtype, reduce_dim=M)
 
 
-@pytest.mark.parametrize(
```
StrongSpoon: delete by mistake?

src/flag_gems/ops/all.py
GwokHiujin: Oops, there is indeed redundant operation here... It should be verified whether the order is equal to the original dim to avoid possible additional permutation. It will be fixed soon. 
However, I reviewed the repository and found some other operators related to dims also have this issue (such as var_mean, amax, etc.), do you think they should be fixed together now?

src/flag_gems/ops/sum.py
```diff
@@ -93,37 +87,36 @@ def sum(inp, *, dtype=None):
     mid = torch.empty((mid_size,), dtype=dtype, device=inp.device)
     out = torch.empty([], dtype=dtype, device=inp.device)
 
-    sum_kernel_1[(mid_size, 1, 1)](inp, mid, M, block_size)
-    sum_kernel_2[(1, 1, 1)](mid, out, mid_size, block_mid)
+    sum_kernel_1[(mid_size, 1)](inp, mid, M, block_size)
+    sum_kernel_2[(1, 1)](mid, out, mid_size, block_mid)
```
StrongSpoon, line 87: grid tuple shorter than 3 will be filled with 1 at runtime. it doesn't matter whether it's length is 3 or not.

src/flag_gems/ops/any.py
```diff
@@ -0,0 +1,157 @@
...
+        dim = list(range(inp.ndim))
+    else:
+        dim = [dim]
+    assert ((i >= -inp.ndim and i < inp.ndim) for i in dim), "Invalid dim" 
+    dtype = inp.dtype
+
+    shape = list(inp.shape)
+    dim = [d % inp.ndim for d in dim]
+    order = [i for i in range(inp.ndim) if i not in dim] + dim
+    inp = inp.permute(order).contiguous()
```
StrongSpoon: I think it's a bit lazy to process integer dim the same way as tuple. integer dim is much easier than tuple, there must be a better method.

src/flag_gems/ops/any.py
```diff
@@ -0,0 +1,157 @@
...
+    BLOCK_SIZE: tl.constexpr,
+):
+    pid = tl.program_id(0)
+    offset = pid * BLOCK_SIZE + tl.arange(0, BLOCK_SIZE)
+    inp_ptrs = inp + offset
+    mask = offset < n_elements
+    inp_val = tl.load(inp_ptrs, mask=mask, other=0.0).to(tl.float32)
+    any_val = tl.min(inp_val == 0., axis=0)
+    mid_ptr = mid + pid
+    tl.store(mid_ptr, any_val)
```
StrongSpoon: mask is needed here

src/flag_gems/ops/any.py
```diff
@@ -0,0 +1,157 @@
...
+    BLOCK_N: tl.constexpr,
+):
+    # Map the program id to the row of inp it should compute.
+    pid = tl.program_id(0) * BLOCK_M + tl.arange(0, BLOCK_M)[:, None]
+    inp = inp + pid * N
+    out = out + pid
+    row_mask = pid < M
+
+    _all = tl.zeros([BLOCK_M, BLOCK_N], dtype=tl.float32)
+    for off in range(0, N, BLOCK_N):
```
StrongSpoon, line 38: the goal of splitting N with BLOCK_N is to save memory for each block, thus variant _all doesn't have to load the whole row. so if you set BLOCK_N as the next power of 2 of N, there is no need to loop on N. setting BLOCK_N as a constant value is okay for now.

StrongSpoon: same for other operators

src/flag_gems/ops/all.py
StrongSpoon: it takes large cost to permute tensor. but I think it's a complex problem to tell if permuting is needed or not. I don't have a perfect solution yet.

src/flag_gems/ops/any.py
```diff
@@ -0,0 +1,157 @@
...
+def any(inp):
+    logging.debug("GEMS any")
+    n_elements = inp.numel()
+    block_size = triton.next_power_of_2(math.ceil(math.sqrt(n_elements)))
+    mid_size = triton.cdiv(n_elements, block_size)
+    block_mid = triton.next_power_of_2(mid_size)
+    dtype = inp.dtype
+
+    mid = torch.empty((mid_size,), dtype=dtype, device=inp.device)
+    out = torch.empty([], dtype=dtype, device=inp.device)
```
StrongSpoon, line 89: why not initialize out tensor as bool type?

src/flag_gems/ops/any.py
```diff
@@ -0,0 +1,157 @@
...
+    out = out + pid
+    row_mask = pid < M
+
+    _all = tl.zeros([BLOCK_M, BLOCK_N], dtype=tl.float32)
+    for off in range(0, N, BLOCK_N):
+        cols = off + tl.arange(0, BLOCK_N)[None, :]
+        col_mask = cols < N
+        mask = row_mask and col_mask
+
+        a = tl.load(inp + cols, mask, other=1.0).to(tl.float32)
```
StrongSpoon: upcast is unnecessary

src/flag_gems/ops/any.py
```diff
@@ -0,0 +1,157 @@
...
+        _all += (a == 0)
+    all = tl.min(_all, axis=1)[:, None]
```
StrongSpoon: need to rewrite this kernel

tests/flag_gems/op_accu_test.py
```diff
@@ -1288,3 +1251,116 @@ def test_accuracy_outer(shape, dtype):
...
     allclose_with_dtype(res_in1_grad, ref_in1_grad, dtype)
     allclose_with_dtype(res_in2_grad, ref_in2_grad, dtype)
+
+
+@pytest.mark.parametrize(
+    "shape",
+    [(4096, i * 64) for i in range(1, 20)],
+)
+
+@pytest.mark.parametrize("dtype", [torch.float16, torch.float32, torch.bfloat16, torch.bool])
```
StrongSpoon: you don't have to test all & any with float type. initialize input tensor using torch.rand with high and low boundary. and ensure that test cases could generate both true and false result.

src/flag_gems/ops/all.py
GwokHiujin: I've temporarily used this solution: sort the dim, then check if it is the same as the order, based on which decide whether to permute.

src/flag_gems/ops/all.py
GwokHiujin: fixed

src/flag_gems/ops/any.py
GwokHiujin: fixed

src/flag_gems/ops/any.py
GwokHiujin: fixed

src/flag_gems/ops/any.py
GwokHiujin: fixed

tests/flag_gems/op_accu_test.py
GwokHiujin: Okay. Because torch's all and any can be used for non-boolean type tensors, I included these types in the test cases before. I'll fix these test cases.

src/flag_gems/ops/any.py
GwokHiujin: fixed

src/flag_gems/ops/any.py
GwokHiujin: fixed

src/flag_gems/ops/any.py
GwokHiujin: The reason why it is processed in the same way as a tuple is mainly because *dim* can also accept None as input. This can solve both scenarios at the same time. But maybe there is a better way...

src/flag_gems/ops/all.py
```diff
@@ -0,0 +1,157 @@
...
+    BLOCK_M: tl.constexpr,
+    BLOCK_N: tl.constexpr,
+):
+    # Map the program id to the row of inp it should compute.
+    pid = tl.program_id(0) * BLOCK_M + tl.arange(0, BLOCK_M)[:, None]
+    inp = inp + pid * N
+    out = out + pid
+    row_mask = pid < M
+
+    all = tl.full([BLOCK_M, 1], value=1., dtype=tl.float32)
```
StrongSpoon, line 37: you can initialize all with "dtype=tl.int1", cause int1 is acceptable by tl.max/tl.min.

tests/flag_gems/op_accu_test.py
```diff
@@ -1179,7 +1179,7 @@ def test_accuracy_min_dim(shape, dim, keepdim, dtype):
 def test_accuracy_sum(shape, dtype):
     inp = torch.randn(shape, dtype=dtype, device="cuda")
 
-    ref_out = torch.sum(inp.to(torch.float64))
+    ref_out = torch.sum(inp, dtype=dtype)
```
StrongSpoon: why?

StrongSpoon: using low precision as reference might be incorrect

tests/flag_gems/op_accu_test.py
```diff
@@ -1188,27 +1188,27 @@ def test_accuracy_sum(shape, dtype):
 
 @pytest.mark.parametrize(
     "shape",
-    [(4096, i * 64) for i in range(1, 20)],
+    [(4096, i * 64) for i in range(1, 2)],
```
StrongSpoon: large tensor is more convincing

tests/flag_gems/op_accu_test.py
```diff
@@ -1188,27 +1188,27 @@ def test_accuracy_sum(shape, dtype):
...
 )
 @pytest.mark.parametrize("keepdim", [True, False])
-@pytest.mark.parametrize("dim", [0, 1])
+@pytest.mark.parametrize("dim", [[0, 1], 0, 1, None])
 @pytest.mark.parametrize("dtype", [torch.float16, torch.float32, torch.bfloat16])
 def test_accuracy_sum_dim(shape, dim, keepdim, dtype):
     inp = torch.randn(shape, dtype=dtype, device="cuda")
 
-    ref_out = torch.sum(inp.to(torch.float64), dim=dim, keepdim=keepdim)
+    ref_out = torch.sum(inp, dim=dim, dtype=dtype, keepdim=keepdim)
```
StrongSpoon: ditto

tests/flag_gems/op_accu_test.py
```diff
@@ -1188,27 +1188,27 @@ def test_accuracy_sum(shape, dtype):
...
+    allclose_with_dtype(res_out, ref_out, dtype)
 
 
 @pytest.mark.parametrize(
     "shape",
     [(4096, i * 64) for i in range(1, 20)],
 )
 @pytest.mark.parametrize("keepdim", [True, False])
-@pytest.mark.parametrize("dim", [0, 1, None])
+@pytest.mark.parametrize("dim", [[0, 1], [1, 0]])
```
StrongSpoon: keep None and integer reserved

tests/flag_gems/op_accu_test.py
```diff
@@ -1327,6 +1327,94 @@ def test_accuracy_outer(shape, dtype):
     allclose_with_dtype(res_in2_grad, ref_in2_grad, dtype)
 
 
+@pytest.mark.parametrize(
+    "shape",
+    [(4096, i * 64) for i in range(1, 20)],
+)
+def test_accuracy_all(shape):
+    inp = torch.rand(shape) < 0.5
```
StrongSpoon: device="cuda"!

src/flag_gems/ops/all.py
StrongSpoon: I suggest undo this commit. simply sorting might be inefficient for non-contiguous tensor. we'll come up with a better solution soon.

src/flag_gems/ops/all.py
GwokHiujin: After testing I found it is not feasible, because boolType input will be converted to fp32 in the triton kernel. You need a same fp32 type tensor to do the comparing job.

tests/flag_gems/op_accu_test.py
GwokHiujin: It was a mistake, sorry for that. Fixed.

tests/flag_gems/op_accu_test.py
GwokHiujin: fixed

tests/flag_gems/op_accu_test.py
GwokHiujin: done

tests/flag_gems/op_accu_test.py
GwokHiujin: done

PR #26, opened 2024-05-22
[Operators] discard tl.make_block_ptr

StrongSpoon:
Function 'tl.make_block_ptr' is not supported by triton 2.0.0 and not stable in triton 2.2.0. This may bring compilation fault for migration of other hardware and cause accuracy error of operators. This pull request discards all tl.make_block_ptr, using common pointers instead.
Moreover, clamp which accepts Tensor as input is supported by codegen.

### DIFF
--- a/src/flag_gems/ops/groupnorm.py
+++ b/src/flag_gems/ops/groupnorm.py
@@ -43,10 +43,10 @@ def group_norm_kernel(
     Y_ptr = Y + xy_offset
 
     X_val = tl.load(X_ptr, mask=xy_mask, other=0.0).to(tl.float32)
-    mean = tl.sum(X_val / num_elements)
+    mean = tl.sum(X_val) / num_elements
     x = tl.where(xy_mask, X_val - mean, 0.0)
 
-    var = tl.sum(x * x / num_elements)
+    var = tl.sum(x * x) / num_elements
     rstd = tl.math.rsqrt(var + eps)
     x_hat = x * rstd
 
@@ -109,7 +109,7 @@ def group_norm_backward_kernel(
     grad_var = grad_std * -(0.5 * rstd * rstd * rstd) / (HW * group_size)
     grad_distance = 2 * x * grad_var
     grad_centered_mean = dx_hat * rstd + grad_distance
-    grad_mean = -tl.sum(grad_centered_mean / num_elements)
+    grad_mean = -tl.sum(grad_centered_mean) / num_elements
     grad_X = grad_centered_mean + grad_mean
     tl.store(dX_ptr, grad_X, mask=xy_mask)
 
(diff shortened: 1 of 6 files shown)


src/flag_gems/ops/clamp.py
```diff
@@ -0,0 +1,35 @@
+import triton
+import triton.language as tl
+import logging
+from ..utils import pointwise_dynamic
+
+
+@pointwise_dynamic
+@triton.jit
+def clamp_func(x, mini, maxi):
+    return tl.minimum(maxi, tl.maximum(mini, x.to(tl.float32)))
```
tongxin, line 10: do we need to cast result dtype?

src/flag_gems/ops/groupnorm.py
```diff
@@ -23,95 +23,40 @@ def group_norm_kernel(
...
+    xy_mask = wb_offset[:, None] < C and hw_offset[None, :] < HW
+
+    Mean_ptr = Mean + pid
+    Rstd_ptr = Rstd + pid
+
+    X_ptr = X + xy_offset
+    Y_ptr = Y + xy_offset
+
+    X_val = tl.load(X_ptr, mask=xy_mask, other=0.0).to(tl.float32)
+    mean = tl.sum(X_val / num_elements)
```
tongxin: if num_elements is scalar, how about moving the divide out of tl.sum?

src/flag_gems/ops/groupnorm.py
```diff
@@ -131,85 +76,42 @@ def group_norm_backward_kernel(
...
+    Rstd_ptr = Rstd + pid
+    X_ptr = X + xy_offset
+    dY_ptr = grad_y + xy_offset
+    dX_ptr = grad_x + xy_offset
+
+    rstd = tl.load(Rstd_ptr)
+    mean = tl.load(Mean_ptr)
+    dY_val = tl.load(dY_ptr, mask=xy_mask, other=0.0)
+    X_val = tl.load(X_ptr, mask=xy_mask, other=0.0)
+    weight = tl.load(W_ptr, mask=wb_mask, other=0.0)[:, None]
```
tongxin: somehow I suspect we may need upcasting here.

src/flag_gems/ops/clamp.py
StrongSpoon: no. tl.store will cast automatically, but sometimes.

src/flag_gems/ops/groupnorm.py
StrongSpoon: I'm concerned that sum may overflow.

src/flag_gems/ops/groupnorm.py
StrongSpoon: done

src/flag_gems/ops/clamp.py
tongxin: That's fine.

src/flag_gems/ops/groupnorm.py
tongxin: Yeah that's an issue but normally fine for sum.

StrongSpoon: done

PR #27, opened 2024-05-22
Dropout using Philox cuda state

tongxin:
This PR is a dropout implementation using the default philoxcudastate as the source of inputs to Triton RNG. 
It differs from Pytorch native dropout in two ways:
- offset is set to nelems rounded up to 4's multiples.
- seed and offset are downcast to uint32 in the kernel for not all vendors support long long in their devices.
Performance was not carefully tuned yet.


src/flag_gems/ops/dropout.py
```diff
@@ -129,24 +94,26 @@ def forward(ctx, x, p, train):
...
-        Mask = torch.empty(x.shape, dtype=torch.bool, device=x.device)
         N = x.numel()
         grid_fn = lambda meta: (triton.cdiv(N, meta["N_BLOCK_SIZE"]),)
-        dropout_forward_kernel[grid_fn](x, O, Mask, N, p)
-        ctx.save_for_backward(Mask)
+        # (TODO) Using Triton autotuner makes kernel parameters opaque to the caller,
+        # hence we cannot obtain the per thread offset as in Pytorch.
+        increment = N
+        philox_seed, philox_offset = philox_cuda_seed_offset(increment)
+        dropout_forward_kernel[grid_fn](x, O, N, p, philox_seed, philox_offset)
```
iclementine, line 103: Should offset be increased after the kernel call?
Or is there some convention for this?

tongxin: It follows aten's convention, specifically CUDAGenerator's philox_cuda_state utility function. 

https://github.com/pytorch/pytorch/blob/f0366de4140c1b0a03a6b8f4f6da0ad585b23bcf/aten/src/ATen/cuda/CUDAGeneratorImpl.cpp#L452

tongxin: We certainly can separate this function in two steps, but in this case, the offset increment can be determined upfront so we spare a separate update.

PR #28, opened 2024-05-23
add cross_entropy_loss

FatJhon:
add cross_entropy_loss, support as followed:
all value is set default, except input and target.
CrossEntropyLoss(input, target, weight=None, size_average=None, ignore_index=-100, reduce=None, reduction='mean', label_smoothing=0.0)

### DIFF
--- a/src/flag_gems/ops/cross_entropy_loss.py
+++ b/src/flag_gems/ops/cross_entropy_loss.py
@@ -36,7 +36,7 @@ def log_softmax_and_mul_kernel(
     output_ptr,
     input_ptr,
     target_ptr,
-    all_mean,
+    mean_num,
     M,
     N,
     K,
@@ -56,7 +56,7 @@ def log_softmax_and_mul_kernel(
     denominator = tl.sum(numerator, axis=1)[:, None]
     softmax_output = tl.log(numerator / denominator)
     target = tl.load(target_ptr + offset, mask=mask, other=0.0)
-    out = softmax_output * target / (all_mean) * -1
+    out = softmax_output * target / (-mean_num)
     output_ptrs = output_ptr + offset
     tl.store(output_ptrs, out, mask=mask)
 
@@ -92,7 +92,7 @@ def softmax_and_sub_kernel(
     input_ptr,
     target_ptr,
     out_grad,
-    all_mean,
+    mean_num,
     M,
     N,
     K,
@@ -115,7 +115,7 @@ def softmax_and_sub_kernel(
     target_ptrs = target_ptr + offset
     target = tl.load(target_ptrs, mask=mask, other=0.0)
     out_grad_value = tl.load(out_grad)
-    out = out_grad_value * (softmax_output - target) / all_mean
+    out = out_grad_value * (softmax_output - target) / mean_num
     output_ptrs = output_ptr + offset
 
     tl.store(output_ptrs, out, mask=mask)
@@ -133,7 +133,7 @@ class CrossEntropyLoss(torch.autograd.Function):
 
         shape = list(input.shape)
         shape[dim] = 1
-        all_mean = target.numel()
+        mean_num = target.numel()
         target = torch.zeros_like(input).scatter(dim, target.view(shape), 1)
 
         M = 1
@@ -152,7 +152,7 @@ class CrossEntropyLoss(torch.autograd.Function):
             out,
             inp,
             target,
-            all_mean,
+            mean_num,
             M,
             N,
             K,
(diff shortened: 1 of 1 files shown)


src/flag_gems/ops/cross_entropy_loss.py
```diff
@@ -25,27 +139,66 @@ def forward(ctx, input, target, weight, reduction, ignore_index, label_smoothing
...
-        input_dim = ctx.dim
+        dim = ctx.dim
         all_mean = ctx.mean
-        in_grad = (
-            out_grad * sub(softmax(input, dim=input_dim), target) / all_mean
-        )  # todo: sub & softmax in one kernel
-        return in_grad, None, None, None, None, None
+
+        assert dim >= -input.ndim and dim < input.ndim, "Invalid dim"
+        dim = dim % input.ndim
```
StrongSpoon: is this necessary? in forward pass, dim is initialized as a legal value.

src/flag_gems/ops/cross_entropy_loss.py
```diff
@@ -10,6 +10,120 @@
...
+    offset = m_offset[:, None] * N * K + n_offset[None, :] * K + pid_k
+    mask = m_offset[:, None] < M and n_offset[None, :] < N
+    input_ptrs = input_ptr + offset
+    inp = tl.load(input_ptrs, mask=mask, other=-float("inf"))
+    row_minus_max = inp - tl.max(inp, axis=1)[:, None]
+    numerator = tl.exp(row_minus_max)
+    denominator = tl.sum(numerator, axis=1)[:, None]
+    softmax_output = tl.log(numerator / denominator)
+    target = tl.load(target_ptr + offset, mask=mask, other=0.0)
+    out = softmax_output * target / (all_mean) * -1
```
StrongSpoon: as each row in target has only one and zero, why not use mask instead of multiplication?
and you could even input -all_mean to save multiplication by one time.

src/flag_gems/ops/cross_entropy_loss.py
```diff
@@ -0,0 +1,205 @@
...
+            M,
+            N,
+            K,
+        )
+        return out, None, None, None, None, None
+
+
+def cross_entropy_loss(
+    input, target, weight=None, reduction="Mean", ignore_index=-100, label_smoothing=0.0
+):
```
iclementine, line 202: In the `native_functions.yaml` of pytorch. The schema(signature) of `cross_entropy` is

```yaml
- func: cross_entropy_loss(Tensor self, Tensor target, Tensor? weight=None, int reduction=Mean, SymInt ignore_index=-100, float label_smoothing=0.0) -> Tensor
  python_module: nn
  dispatch:
    CompositeImplicitAutograd: cross_entropy_loss_symint
```

the type of reducetion is `int`. We should be consistent with the signature.

Although there is some enum values defined in torch's c++ code, these enum values are not exported to python. 

However, torch provides a utility function to convert parameter from `string` to the integer, which is passed to the internal aten function. 

```python
def get_enum(reduction: str) -> int:
    if reduction == 'none':
        ret = 0
    elif reduction == 'mean':
        ret = 1
    elif reduction == 'elementwise_mean':
        warnings.warn("reduction='elementwise_mean' is deprecated, please use reduction='mean' instead.")
        ret = 1
    elif reduction == 'sum':
        ret = 2
    else:
        ret = -1  # TODO: remove once JIT exceptions support control flow
        raise ValueError(f"{reduction} is not a valid value for reduction")
    return ret
```

https://github.com/pytorch/pytorch/blob/ff63e8bac8e8ef00c3f368acc18e2f1fd57c3aa8/torch/nn/_reduction.py#L7

And the python wrapper `torch.nn.functional` already uses this.

```python
    return torch._C._nn.cross_entropy_loss(input, target, weight, _Reduction.get_enum(reduction), ignore_index, label_smoothing)
```

https://github.com/pytorch/pytorch/blob/ff63e8bac8e8ef00c3f368acc18e2f1fd57c3aa8/torch/nn/functional.py#L3103

So we can just use int as the parameter type of reduction and set the default value to `1`(which correspond to kMean).

FatJhon: done

src/flag_gems/ops/cross_entropy_loss.py
FatJhon: done

src/flag_gems/ops/cross_entropy_loss.py
FatJhon: done

PR #29, opened 2024-05-23
[Codegen] refactor codegen for poinwise

iclementine:
1. support 0d task space;
2. support specifying output dtypes;
3. support non-tensor arguments to the scalar function.

and remove pointwise_static.py

### DIFF
--- a/src/flag_gems/utils/pointwise_dynamic.py
+++ b/src/flag_gems/utils/pointwise_dynamic.py
@@ -39,8 +39,6 @@ class OPDesc:
     _num_outputs: int
     _output_dtypes: List[torch.dtype]
 
-    scalar_fn: JITFunction
-
     def __init__(
         self,
         *,


src/flag_gems/utils/pointwise_dynamic.py
```diff
@@ -1,172 +1,346 @@
...
+                    code.writeline(f"out{num_output_tensor_index} = torch.empty(shape, dtype=in0.dtype, device=in0.device)")
+                else:
+                    code.writeline(f"out{num_output_tensor_index} = torch.empty(shape, dtype={type_name(op_desc.dtypes[i])}, device=in0.device)")
+                num_output_tensor_index += 1
+
+        # call destination_passing_func
+        call_str = f"{output_ref_for_wrapper(op_desc)} = {destination_passing_func_name}({parameter_ref_for_wrapper(op_desc, include_outputs=True)})"
+        code.writeline(call_str)
+
+        return_str = f"return {output_ref_for_wrapper(op_desc)}"
```
StrongSpoon: what about storing the return value of output_ref_for_wrapper?

src/flag_gems/utils/pointwise_dynamic.py
```diff
@@ -1,172 +1,346 @@
...
+    # task partitioning, 1d task indexing
+    tile_size = 512
+    num_warps = 4
+    if rank == 0:
+        tile_size = 32
+        num_warps = 1
+
+    with code.indent():
+        # docstring
+        wrapper_docstring = docstring_for_wrapper(op_desc)
```
StrongSpoon: is it redundant?

src/flag_gems/utils/pointwise_dynamic.py
```diff
@@ -1,172 +1,346 @@
...
+                output_tensor_index += 1
+    return ", ".join(parameters)
+
+def docstring_for_wrapper(op_desc: OPDesc):
+    doc = f'"""Generated destination passing style wrapper function with {op_desc.num_input_tensors()} input tensor(s), {op_desc.num_output_tensors()} output tensor(s), {op_desc.num_non_tensor_args()} non tesnor input(s)."""'
+    return doc
+
+
+def generate_imports(code: IndentedBuffer) -> IndentedBuffer:
+    code.writeline("import math")
```
StrongSpoon, line 230: what is "import math" used for?

iclementine: Triton kernel code can contain invocations of functions and use of variables defined in `math`. So it is included for these uses.

src/flag_gems/utils/pointwise_dynamic.py
iclementine: How to sort the return value? 
I return them by the order they are defined in the scalar function.

src/flag_gems/utils/pointwise_dynamic.py
iclementine: The 2 wrapper both have a docstring.

src/flag_gems/utils/pointwise_dynamic.py
StrongSpoon: aren't they of the same order?

src/flag_gems/utils/pointwise_dynamic.py
StrongSpoon: yes but the contents are the same

src/flag_gems/utils/pointwise_dynamic.py
StrongSpoon: get

src/flag_gems/utils/pointwise_dynamic.py
iclementine: Done, I have change that.

src/flag_gems/utils/pointwise_dynamic.py
iclementine: Sorry，I misred storing as sorting

PR #30, opened 2024-05-24
[Operators] add all, any, sum.dim, amax.dim

GwokHiujin:
1. Added all, any operators to support the definitions in the API documentation.
2. Implemented dim for sum and amax, which can support int or int[] inputs.

### DIFF
--- a/src/flag_gems/ops/vector_norm.py
+++ b/src/flag_gems/ops/vector_norm.py
@@ -137,12 +137,9 @@ def vector_norm(x, ord=2, dim=None, keepdim=False, dtype=None):
     if dim is None:
         dim = list(range(x.ndim))
     shape = list(x.shape)
-    dim = sorted([d % x.ndim for d in dim])
+    dim = [d % x.ndim for d in dim]
     order = [i for i in range(x.ndim) if i not in dim] + dim
-    if order == shape:
-        x = x.contiguous()
-    else:
-        x = x.permute(order).contiguous()
+    x = x.permute(order).contiguous()
     N = 1
     for i in dim:
         N *= shape[i]


src/flag_gems/ops/all.py
```diff
@@ -0,0 +1,154 @@
...
+def all_kernel_dim(
+    inp,
+    out,
+    M,
+    N,
+    BLOCK_M: tl.constexpr,
+    BLOCK_N: tl.constexpr,
+):
+    # Map the program id to the row of inp it should compute.
+    pid = tl.program_id(0) * BLOCK_M + tl.arange(0, BLOCK_M)[:, None]
```
tongxin: The naming is kind of misleading... I think that one would normally consider pid to be a single integer for a block program.

src/flag_gems/ops/all.py
```diff
@@ -0,0 +1,154 @@
...
+    row_mask = pid < M
+
+    _all = tl.full([BLOCK_M, BLOCK_N], value=1, dtype=tl.int1)
+    for off in range(0, N, BLOCK_N):
+        cols = off + tl.arange(0, BLOCK_N)[None, :]
+        col_mask = cols < N
+        mask = row_mask and col_mask
+
+        a = tl.load(inp + cols, mask, other=1.0)
+        _all = _all and (a != 0)
```
tongxin, line 50: Previous tuning experience tells me looping is likely inferior than direct reduction using tl.reduce.

src/flag_gems/ops/all.py
StrongSpoon: this is my fault. "pid" is used improperly in many other operators. I'll fix them later.

src/flag_gems/ops/all.py
StrongSpoon: direct reduction requires BLOCK_N to be the next power of 2 of N, which might be beyond the shared memory bound of some gpu platforms.

src/flag_gems/ops/all.py
```diff
@@ -0,0 +1,155 @@
...
+    BLOCK_SIZE: tl.constexpr,
+):
+    pid = tl.program_id(0)
+    offset = pid * BLOCK_SIZE + tl.arange(0, BLOCK_SIZE)
+    inp_ptrs = inp + offset
+    mask = offset < n_elements
+    inp_val = tl.load(inp_ptrs, mask=mask, other=1.0)
+    all_val = tl.reduce(inp_val != 0, axis=0, combine_fn=reduce_all)
+    mid_ptr = mid + pid
+    mid_mask = pid < mid_size
```
StrongSpoon: mid_mask is useless

StrongSpoon: so does mid_size

src/flag_gems/ops/all.py
GwokHiujin: Yes, I have repeated this error before: in some cases next_power_of_2(N) will exceed the Triton maximum of 2^20.

src/flag_gems/ops/all.py
GwokHiujin: done

src/flag_gems/ops/vector_norm.py
```diff
@@ -137,9 +137,12 @@ def vector_norm(x, ord=2, dim=None, keepdim=False, dtype=None):
...
         dim = list(range(x.ndim))
     shape = list(x.shape)
-    dim = [d % x.ndim for d in dim]
+    dim = sorted([d % x.ndim for d in dim])
     order = [i for i in range(x.ndim) if i not in dim] + dim
-    x = x.permute(order).contiguous()
+    if order == shape:
+        x = x.contiguous()
+    else:
+        x = x.permute(order).contiguous()
```
StrongSpoon: undo this

GwokHiujin: done

PR #39, opened 2024-05-30
[Tests] Reconstruct test files

StrongSpoon:
1. reorganized test files in examples, benchmark, tests respectively
2. divide operators into pointwise, reduction, bias and so on
3. implement v1 bitwise ops add, sub, mul, div and pow by codegen

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -24,11 +24,9 @@ FlagGems
 │   │   ├──ops: single operators
 │   │   ├──fused: fused operators
 │   │   ├──__init__.py
-├── tests
-│   ├──flag_gems
-│   │   ├──model_bert_test.py: test for BERT model running with flag_gems
-│   │   ├──op_accu_test.py: test for accuracy of operators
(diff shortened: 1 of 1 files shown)


benchmark/performance_utils.py
```diff
@@ -0,0 +1,196 @@
...
+    def __init__(self, op_name, torch_op, arg_func, dtype, batch, sizes):
+        self.op_name = op_name
+        self.torch_op = torch_op
+        self.arg_func = arg_func
+        self.dtype = dtype
+        self.batch = batch
+        self.sizes = sizes
+
+    def profile(self, op, *args):
+        if CPU_MODE:
```
iclementine, line 22: Whether cpu_mode or not, does the kernel always run on gpu? It is only the method of timing: by inserting device synchronzation or by cuda events?

StrongSpoon: yes, benchmark only compares gpu kernels. cpu mode records end2end latency instead of coda time.

PR #41, opened 2024-05-30
add reduction of sum and  none for CrossEntropyLoss 

FatJhon:
1. modify inaccurate variable names
2. add reduction of sum and none for CrossEntropyLoss

### DIFF
--- a/src/flag_gems/ops/cross_entropy_loss.py
+++ b/src/flag_gems/ops/cross_entropy_loss.py
@@ -2,12 +2,12 @@ import torch
 import triton
 import triton.language as tl
 import logging
-from enum import Enum
+from enum import IntEnum
 from ..utils import libentry
 from .sum import sum, sum_dim
 
 
-class Reduction(Enum):
+class Reduction(IntEnum):
     NONE = 0
     MEAN = 1
     SUM = 2
@@ -194,6 +194,7 @@ class CrossEntropyLoss(torch.autograd.Function):
     @staticmethod
     def forward(ctx, input, target, weight, reduction, ignore_index, label_smoothing):
         logging.debug("GEMS CrossEntropyLoss")
+        assert reduction in Reduction._value2member_map_, "Invalid reduction"
         assert isinstance(input, torch.Tensor), "input is not a tensor"
         if input.ndim >= 2:
             dim = 1


src/flag_gems/ops/cross_entropy_loss.py
```diff
@@ -56,7 +56,7 @@ def log_softmax_and_mul_kernel(
     denominator = tl.sum(numerator, axis=1)[:, None]
     softmax_output = tl.log(numerator / denominator)
     target = tl.load(target_ptr + offset, mask=mask, other=0.0)
-    out = softmax_output * target / (-mean_num)
+    out = softmax_output * target / (mean_num)
```
tongxin, line 66: Can we unify these fused log softmax and blah functions

src/flag_gems/ops/cross_entropy_loss.py
```diff
@@ -130,10 +192,12 @@ def forward(ctx, input, target, weight, reduction, ignore_index, label_smoothing
             dim = 1
         else:
             dim = 0
-
+        if reduction != 1:
```
tongxin: Can we explicitly define an enum type for reduction?

src/flag_gems/ops/cross_entropy_loss.py
```diff
@@ -157,11 +221,15 @@ def forward(ctx, input, target, weight, reduction, ignore_index, label_smoothing
             N,
             K,
         )
-        out_result = sum(out)
+        if reduction:
```
tongxin: Please be explicit on the reduction types

src/flag_gems/ops/cross_entropy_loss.py
FatJhon: I will try. The difference is that when reduction is none，out_grad is a tensor,  reduction is sum or mean, out_grad is a constant, it is difficult to separate in triton.jt

src/flag_gems/ops/cross_entropy_loss.py
FatJhon: done

src/flag_gems/ops/cross_entropy_loss.py
FatJhon: done

src/flag_gems/ops/cross_entropy_loss.py
tongxin: Got it.

src/flag_gems/ops/cross_entropy_loss.py
```diff
@@ -2,8 +2,15 @@
 import triton
 import triton.language as tl
 import logging
+from enum import Enum
 from ..utils import libentry
-from .sum import sum
+from .sum import sum, sum_dim
+
+
+class Reduction(Enum):
```
tongxin: Use IntEnum then its members are also integers.

FatJhon: Thanks  for your correction. The interface usage will be more precise.

PR #42, opened 2024-05-30
[CI/CD] Add python-test config yaml for Unit Test

Bowen12992:
Add `.github/workflows/python-test.yaml` for deployment of CI/CD

### DIFF
--- a/.github/workflows/python-test.yaml
+++ b/.github/workflows/python-test.yaml
@@ -1,0 +1,35 @@
+
+# This workflow will install Python dependencies, run tests and lint with a single version of Python
+# For more information see: https://docs.github.com/en/actions/automating-builds-and-tests/building-and-testing-python
+
+name: flag-gems-test
+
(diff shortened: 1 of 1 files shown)


.github/workflows/python-test.yaml
```diff
@@ -0,0 +1,37 @@
...
+  pull_request:
+    branches: [ "master" ]
+
+jobs:
+  container-test-job:
+    runs-on: [self-hosted, docker]
+    container:
+      image: localhost:5000/flag-gems-ci:v1.0
+      env:                                         
+        NODE_ENV: development
```
iclementine: What is this environment `NODE_ENV` for?

Bowen12992: It is a  redundant env from the template file which i forgot to delete. BTW,  we should use `PYTHON_ENV: production` for this project if an env needed.

PR #45, opened 2024-06-02
[CI/CD] Upgrading actions/checkout to v4

Bowen12992:
- upgrading actions/checkout to v4

- split unit task into different CUDA devices

- add timeout config


StrongSpoon (approved):
lg

tongxin (approved):
LG

PR #48, opened 2024-06-03
cambricon: Cambricon merge 0603 and adopt mlu

functionxu123:
merge origin/master (4ab2d92) into cambricon, fix adopt mlu problems


sethbrin:
LGTM

PR #49, opened 2024-06-03
Support llava inference example

jokmingwong:
Considering 2 special cases and fix bugs: 
1. boolean computation on cumsum, eq, mul, sum 
2. negative index value on CrossEntropyLoss

### DIFF
--- a/src/flag_gems/ops/eq.py
+++ b/src/flag_gems/ops/eq.py
@@ -13,6 +13,11 @@ def eq_func(x, y):
 
 
 def eq(A, B):
+    if A.device != B.device:
+        if A.device.type == "cuda":
+            B = B.to(A.device)
+        else:
+            A = A.to(B.device)
     logging.debug("GEMS EQ")
     return eq_func(A, B)
 
--- a/README.md
+++ b/README.md
@@ -175,6 +175,7 @@ Operators will be implemented according to [OperatorList.md](./OperatorList.md).
 
 - Bert-base-uncased
 - Llama-2-7b
+- Llava-1.5-7b
 
 ## Supported Platforms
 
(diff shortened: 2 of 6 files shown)


src/flag_gems/ops/eq.py
```diff
@@ -13,7 +13,14 @@ def eq_func(x, y):
 
 def eq(A, B):
     logging.debug("GEMS EQ")
-    O = eq_func(A, B)
+    if A.dim() == 0 and B.dim()==0:
+        O = eq_func_scalar(A.item(), B.item())
+    elif A.dim() == 0:
```
StrongSpoon: can't 0-dim tensor broadcast to high-dimension?

StrongSpoon (dismissed):

src/flag_gems/ops/eq.py
StrongSpoon: pointwise_dynamic can deal with 0-dim tensor. while the problem here is that equal may receive a cuda tensor and a cpu tensor. the solution is cast cpu tensor to the same device as cuda tensor.

src/flag_gems/ops/eq.py
```diff
@@ -13,6 +13,11 @@ def eq_func(x, y):
 
 
 def eq(A, B):
+    if A.device != B.device:
+        if A.device.type == "cuda":
+            B = B.to(A.device)
+        else:
+            A = A.to(B.device)
     logging.debug("GEMS EQ")
     return eq_func(A, B)
```
Bowen12992, line 22: should we add device guard here？

StrongSpoon: device guard is inside code generation

Bowen12992 (approved):
LGTM, we can add model llava test into `.github/workflows/python-test.yaml`

PR #52, opened 2024-06-05
Add where op

Bowen12992:
- Add where op

Some perf result of where Op (torch.float32) on NVIDIA GeForce RTX 3090

| Size | Torch Latency (ms) | Gems Latency (ms) |
|------|---------------------|-------------------|
| 64   | 0.003072            | 0.003072          |
| 192  | 0.004096            | 0.003136          |
| 320  | 0.004096            | 0.003072          |
| 448  | 0.004864            | 0.003168          |
| 576  | 0.004096            | 0.003200            |
| 704  | 0.004096            | 0.003296          |
| 832  | 0.004096            | 0.003136          |
| 960  | 0.004192            | 0.003296          |
| 1088 | 0.004096            | 0.003168          |
| 1216 | 0.004096            | 0.00384           |

TODO：
- where_scalar with pointwise codegen (output Type)
- where(condition)


src/flag_gems/ops/where.py
```diff
@@ -0,0 +1,16 @@
...
+from ..utils import pointwise_dynamic
+
+
+@pointwise_dynamic
+@triton.jit
+def where_func(condition, x, y):
+    return tl.where(condition, x, y)
+
+
+def where(condition, x, y):
```
tongxin: where_self?

tongxin: Are we missing where(Tensor condition) in this PR?

src/flag_gems/__init__.py
```diff
@@ -67,6 +67,7 @@ def enable(lib=aten_lib):
     lib.impl("triu", triu, "CUDA")
     lib.impl("var_mean.correction", var_mean, "CUDA")
     lib.impl("linalg_vector_norm", vector_norm, "CUDA")
+    lib.impl("where.self", where, "CUDA")
```
StrongSpoon: there are interfaces like where.Scalar, where.ScalarOther and so on. I suggest implementing them all.

src/flag_gems/ops/where.py
Bowen12992: yep，all cases will be finished in the following commit

src/flag_gems/__init__.py
Bowen12992: yes，same as above comment

src/flag_gems/__init__.py
```diff
@@ -67,7 +67,10 @@ def enable(lib=aten_lib):
     lib.impl("triu", triu, "CUDA")
     lib.impl("var_mean.correction", var_mean, "CUDA")
     lib.impl("linalg_vector_norm", vector_norm, "CUDA")
-
+    lib.impl("where.self", where, "CUDA")
+    lib.impl("where.self_out", where, "CUDA")
```
StrongSpoon: we don't need to support op_out api

src/flag_gems/__init__.py
```diff
@@ -67,7 +67,10 @@ def enable(lib=aten_lib):
...
+    lib.impl("where.ScalarSelf", where, "CUDA")
```
StrongSpoon: better separate them as different interfaces

src/flag_gems/__init__.py
Bowen12992: OK

tests/test_binary_pointwise_ops.py
```diff
@@ -600,3 +600,48 @@ def test_accuracy_sub_scalar_tensor(shape, scalar, alpha, dtype):
...
+    inp1 = torch.randn(shape, dtype=dtype, device="cuda")
+    inp2 = torch.randn(shape, dtype=dtype, device="cuda")
+    ref_inp1 = to_reference(inp1, True)
+    ref_inp2 = to_reference(inp2, True)
+
+    ref_out = torch.where(ref_inp1 > 0, ref_inp1, ref_inp2)
+    with flag_gems.use_gems():
+        res_out = torch.where(inp1 > 0, inp1, inp2)
+
+    gems_assert_close(res_out, ref_out, dtype)
```
StrongSpoon: I think gems_assert_equal without upcasting is better. there is no actual computation on data, result and reference can be equal.

benchmark/performance_utils.py
```diff
@@ -99,3 +99,10 @@ def ternary_args(dtype, batch, size):
     inp2 = torch.randn([batch, size], dtype=dtype, device="cuda")
     inp3 = torch.randn([batch, size], dtype=dtype, device="cuda")
     return inp1, inp2, inp3
+
+
+def where_args(dtype, batch, size):
```
StrongSpoon: you can define where_args in test_perf_where. it's more clear for users.

src/flag_gems/__init__.py
Bowen12992: done

tests/test_binary_pointwise_ops.py
Bowen12992: it is indeed

benchmark/performance_utils.py
Bowen12992: this file has conflict with other PR which separate the Op_args

benchmark/test_pointwise_perf.py
```diff
@@ -405,3 +405,23 @@ def test_perf_triu(dtype):
         sizes=SIZES,
     )
     bench.run()
+
+
+def where_args(dtype, batch, size):
```
StrongSpoon, line 410: I prefer defining arg-gen function inside test function

src/flag_gems/ops/where.py
```diff
@@ -0,0 +1,41 @@
...
+    return tl.where(condition, self, other)
+
+
+def where_scalar_self(condition, self, other):
+    logging.debug("GEMS WHERE_SCALAR_SELF")
+    O = where_scalar_self_func(other, condition, self)
+    return O
+
+
+@pointwise_dynamic(is_tensor=[True, True, False])
```
tongxin, line 32: What about having the parameters ordered the same way as in condition, self, other?

tongxin (approved):
LG

StrongSpoon (approved):
lgtm

src/flag_gems/ops/where.py
Bowen12992: because the `pointwise_dynanamic`  decorator will make the output type identical with the type of arg1 tensor，so i put the non-scalar dtype between `self` and `other` left-hand side

benchmark/test_pointwise_perf.py
Bowen12992: OK，i will fix it next PR

PR #53, opened 2024-06-05
[Document] README Feature: automatic code generation

StrongSpoon:
update test cases as well


Bowen12992 (approved):
LGTM

README.md
```diff
@@ -7,6 +7,45 @@ FlagGems is a high-performance general operator library implemented in [OpenAI T
...
+```python
+@pointwise_dynamic(is_tensor=[True, True, False], dtypes=[None, None, float])
+@triton.jit
+def add_func(x, y, alpha):
+    return x + y * alpha
+```
+
+#### Output Data Type
+
+FlagGems' automatic code generation defaults that output tensors are with the same data type as the first input tensor. When developers need to use other data types, they can specify this by passing a list of data types to the parameter `output_dtypes`. For example, in the following code, the output tensor type is specified as `torch.bool`.
```
iclementine: By default, all output tensors have the same data type as the first input tensor, but it can also be customized by providing a list of data types. For example, in the following code, the output tensor type is specified as `torch.bool`.

iclementine (dismissed):

README.md
StrongSpoon: done

iclementine (approved):

Bowen12992 (approved):

PR #56, opened 2024-06-06
pointwise_codegen: now each process generate its own code to avoid re…

iclementine:
pointwise_codegen: now each process generate its own code to avoid read & write conflict when used with multiprocessing


StrongSpoon (approved):
works on distributed training.

tongxin (approved):
LGTM

PR #58, opened 2024-06-06
cambricon: fix layernorm, improve groupnorm

tututuski:
For groupnorm, add a constant （DEAL_NUM） to mark the number of elements processed at a time. It can also be set to 0, which will make the behavior the same as before the merge.

For layernorm, fixed the problem that BLOCK_ROW_SIZE can only be 1 (taking 2/4/8... will cause calculation errors). After the fix, BLOCK_ROW_SIZE can be 1/2/4/8/...

### DIFF
--- a/src/flag_gems/ops/layernorm.py
+++ b/src/flag_gems/ops/layernorm.py
@@ -125,15 +125,15 @@ def layer_norm_backward_kernel(
         mask = row_mask and col_mask
         dy = tl.load(dY + cols[None, :], mask).to(tl.float32)
         x = tl.load(X + cols[None, :], mask).to(tl.float32)
-        x = tl.where(col_mask, x - mean, 0.0)
+        x = tl.where(mask, x - mean, 0.0)
         x_hat = x * rstd
         w = tl.load(W + cols, mask=cols < N).to(tl.float32)
         dx_hat = dy * w
         dx_part2 += dx_hat
         dx_part3 += dx_hat * x_hat
 
-    dx_2 = tl.sum(dx_part2)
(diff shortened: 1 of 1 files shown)


src/flag_gems/ops/layernorm.py
```diff
@@ -125,15 +125,15 @@ def layer_norm_backward_kernel(
         mask = row_mask and col_mask
         dy = tl.load(dY + cols[None, :], mask).to(tl.float32)
         x = tl.load(X + cols[None, :], mask).to(tl.float32)
-        x = tl.where(col_mask, x - mean, 0.0)
+        x = tl.where(mask, x - mean, 0.0)
```
sethbrin, line 128: 这个是不是原生 layernorm的问题，如果是的话，可以提一个PR直接到main分支哈

src/flag_gems/ops/groupnorm.py
```diff
@@ -130,33 +138,46 @@ def weight_bias_backward_kernel(
     HW,
     BLOCK_N: tl.constexpr,
     BLOCK_HW: tl.constexpr,
+    DEAL_NUM: tl.constexpr,
 ):
```
sethbrin: 这个是不是直接改调用侧代码呢？直接引入一层循环来处理`BLOCK_HW`?

sethbrin (commented):
others, LGTM

tututuski:
放弃 groupnorm的改动， layernorm 的修改直接合入到master

src/flag_gems/ops/layernorm.py
tongxin: Just saw it.

PR #59, opened 2024-06-07
[bugfix] fix error in layer_norm_backward_kernel

tututuski:
layer_norm_backward_kernel does not work correctly with BLOCK_ROW_SIZE > 1. This PR fixes this bug.

### DIFF
--- a/src/flag_gems/ops/layernorm.py
+++ b/src/flag_gems/ops/layernorm.py
@@ -125,15 +125,15 @@ def layer_norm_backward_kernel(
         mask = row_mask and col_mask
         dy = tl.load(dY + cols[None, :], mask).to(tl.float32)
         x = tl.load(X + cols[None, :], mask).to(tl.float32)
-        x = tl.where(col_mask, x - mean, 0.0)
+        x = tl.where(mask, x - mean, 0.0)
         x_hat = x * rstd
         w = tl.load(W + cols, mask=cols < N).to(tl.float32)
         dx_hat = dy * w
         dx_part2 += dx_hat
         dx_part3 += dx_hat * x_hat
 
-    dx_2 = tl.sum(dx_part2)
-    dx_3 = tl.sum(dx_part3)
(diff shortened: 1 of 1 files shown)


StrongSpoon (commented):
My mistake. In LayerNorm backward, I distributed M rows in M block but forgot to remove the parameter BLOCK_ROW_SIZE.

src/flag_gems/ops/layernorm.py
```diff
@@ -125,15 +125,15 @@ def layer_norm_backward_kernel(
...
         x_hat = x * rstd
         w = tl.load(W + cols, mask=cols < N).to(tl.float32)
         dx_hat = dy * w
         dx_part2 += dx_hat
         dx_part3 += dx_hat * x_hat
 
-    dx_2 = tl.sum(dx_part2)
-    dx_3 = tl.sum(dx_part3)
+    dx_2 = tl.sum(dx_part2, axis=1, keep_dims=True)
+    dx_3 = tl.sum(dx_part3, axis=1, keep_dims=True)
```
StrongSpoon: it seems tl.sum in triton2.2 has no argument named keep_dims.

tututuski: The way to maintain dimensions has been modified. Or should we delete the BLOCK_ROW_SIZE parameter? That is, only support the case where it is equal to 1

StrongSpoon: removing keep_dims is okay. I think dx_2 and dx_3 with 1-dimension could be automatically broadcast in the following computation.

StrongSpoon (approved):
lgtm

src/flag_gems/ops/layernorm.py
tongxin: It misses a broadcasting in case of no keep_dims.
dx_2 = dx_2[:, None] suffices.

tongxin:
LG

tongxin (approved):
LG

PR #60, opened 2024-06-11
[Test] Reconstruct benchmark & support fused ops

StrongSpoon:
1. move specific arg-gen functions into corresponding tests
2. add test_fused_perf.py (including gelu&mul, silu&mul, skip_layernorm, skip_rmsnorm)

### DIFF
--- a/benchmark/test_fused_perf.py
+++ b/benchmark/test_fused_perf.py
@@ -1,0 +1,134 @@
+import torch
+import pytest
+import flag_gems
+from .performance_utils import *
+
+
+@pytest.mark.parametrize("dtype", FLOAT_DTYPES)
+def test_perf_gelu_and_mul(dtype):
+
+    def torch_op(x, y):
+        return torch.mul(torch.nn.functional.gelu(x), y)
+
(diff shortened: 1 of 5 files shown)


benchmark/test_fused_perf.py
```diff
@@ -0,0 +1,134 @@
...
+import pytest
+import flag_gems
+from .performance_utils import *
+
+
+@pytest.mark.parametrize("dtype", FLOAT_DTYPES)
+def test_perf_gelu_and_mul(dtype):
+
+    def torch_op(x, y):
+        return torch.mul(torch.nn.functional.gelu(x), y)
```
tongxin, line 11: Looks like `torch_op` is not used in the test right? This is just for reference purpose?

StrongSpoon: yes. fused operators can't be called by torch api. so the compared operators are torch_op and gems_op.

tongxin: I missed the baseline... It looks good.

tongxin (approved):
LG

iclementine (approved):

PR #62, opened 2024-06-12
cambricon: add config of autotune to argmax, max and min

tututuski:
Add some configuration to ensure a working kernel


functionxu123:
LGTM

PR #65, opened 2024-06-13
[format] Add pre-commit and foramat all the code

Bowen12992:
Add pre-commit and format all the code

However some flake8 rules connot apply:

   -  \# F405 : Name may be undefined, or defined from star imports: module
   - \# E731 : Do not assign a lambda expression, use a def
   - \# F403 : 'from module import *' used; unable to detect undefined names
   - \# W503 : Line break before binary operator
   - \# E722 : Do not use bare 'except'
   - \# E203 : Whitespace before ':'


.pre-commit-config.yaml
```diff
@@ -0,0 +1,28 @@
...
+  - id: flake8
+    args: ["--ignore=F405,E731,F403,W503,E722,E203", --max-line-length=120]
+    # F405 : Name may be undefined, or defined from star imports: module
+    # E731 : Do not assign a lambda expression, use a def
+    # F403 : 'from module import *' used; unable to detect undefined names
+    # W503 : Line break before binary operator
+    # E722 : Do not use bare 'except'
+    # E203 : Whitespace before ':'
+
+- repo: https://github.com/pycqa/isort
```
tongxin, line 18: I hate it...
You'll be very careful letting it messing with your imports. It may break your code

tongxin: But normally it's fine.

tongxin (approved):
Thanks for adding pre-commit !

.pre-commit-config.yaml
Bowen12992: Actually, isort has conflict with black becase they use different code style... but it dose can beautify the code...

.github/workflows/pre-commit.yml
```diff
@@ -0,0 +1,17 @@
...
+    branches: [ "master" ]
+
+jobs:
+  pre-commit:
+    runs-on: ubuntu-latest
+    steps:
+    - uses: actions/checkout@v4
+    - uses: actions/setup-python@v5
+      with:
+        python-version: '3.11'
```
tongxin, line 16: The python 3.11 is not a mandate as of now.

.pre-commit-config.yaml
```diff
@@ -0,0 +1,30 @@
+repos:
+- repo: https://github.com/pre-commit/pre-commit-hooks
+  rev: v2.3.0
+  hooks:
+  - id: check-yaml
+  - id: end-of-file-fixer
+  - id: trailing-whitespace
+  - id: flake8
+    language_version: python3.11
```
tongxin, line 9: We should address Python version first.

iclementine (approved):

.github/workflows/pre-commit.yml
Bowen12992: Because we use py3.11 for CICD environment，so we use the same version for pre-commit

.pre-commit-config.yaml
Bowen12992: OK

PR #68, opened 2024-06-14
[bugfix] fix libentry on argument processing

iclementine:
fix libentry:
1. ensure that decorator cascading is working as expected, i.e. inner decorator can use arguments provided by outer decorator
2. ensure that grid function can use all the arguments provided by decorators(Autotuner & Heuristics)
3. simplify LibEntry, extract captured constant arguments from CompiledKernel, instead of traversing layers of decorator.


StrongSpoon (approved):
lgtm

PR #69, opened 2024-06-14
Optimize softmax

iclementine:
1. use different kernels (inner & non_inner) for softmax forward
   a. inner: for reduction the last dim(and the input is preprocessed to be contiguous)
 a. inner: for reduce along other dimensions(and the input is preprocessed to be contiguous)
3. both have `ONE_TILE_PER_CTA` static condition
   a. when `ONE_TILE_PER_CTA` is True,  load only one tile per cta without looping over reduction dim
   b. when `ONE_TILE_PER_CTA` is False, use online softmax normalizer algorithm to save one swipe over the input.

We can leave other optimizations(optimize accordign to input layout or two-pass-reduction) for future PRs

### DIFF
--- a/tests/test_reduction_ops.py
+++ b/tests/test_reduction_ops.py
@@ -535,8 +535,8 @@ def test_accuracy_skip_rmsnorm(shape, dtype):
 
 @pytest.mark.parametrize("shape", REDUCTION_SHAPES)
 @pytest.mark.parametrize("dtype", FLOAT_DTYPES)
-def test_accuracy_softmax(shape, dtype):
-    dim = 1
+@pytest.mark.parametrize("dim", [0, 1])
+def test_accuracy_softmax(shape, dtype, dim):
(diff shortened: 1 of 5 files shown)


StrongSpoon:
need to rebase after merging pr68

tests/test_reduction_ops.py
```diff
@@ -535,8 +535,8 @@ def _torch_rms_norm(x, residual, weight, eps):
 
 @pytest.mark.parametrize("shape", REDUCTION_SHAPES)
 @pytest.mark.parametrize("dtype", FLOAT_DTYPES)
-def test_accuracy_softmax(shape, dtype):
-    dim = 1
+@pytest.mark.parametrize("dim", [0, 1])
```
StrongSpoon, line 538: DIM_LIST

iclementine: "dim" passed to the function here is an integer.
Do you mean we should add test for non-inner reduction in general?

PR #72, opened 2024-06-17
Develop Embedding[SiliconFlow]

MARD1NO:
测试环境：4090

测试case，Embedding=4096 * 4096 ，输入index大小为32

前向：
triton 耗时 2.82us, torch耗时3.78us
image

反向：
triton耗时 14.78us，torch耗时 3.78us
image
image

torch的反向机制是利用了 warp ballot，一个block读取所有index存到smem，如果有遇到相同index的，则逐出warp，只让一个warp来负责加，triton似乎不存在这种机制


StrongSpoon:
why not register in src/flag_gems/__init__.py

src/flag_gems/ops/embedding.py
```diff
@@ -0,0 +1,188 @@
...
+
+        if ctx.scale_grad_by_freq:
+            embedding_grad_scale_kernel[ctx.M,](
+                grad_inputs, indice_freq, ctx.num_weights, ctx.N, BLOCK_SIZE
+            )
+        return grad_inputs, None, None, None, None
+
+
+def embedding(
+    indices, weight, padding_idx=None, scale_grad_by_freq=False, sparse=False
```
StrongSpoon: please reconstruct the interface in the same order as aten::embedding
`func: embedding(Tensor weight, Tensor indices, SymInt padding_idx=-1, bool scale_grad_by_freq=False, bool sparse=False) -> Tensor`

src/flag_gems/ops/embedding.py
```diff
@@ -0,0 +1,188 @@
...
+import triton
+import triton.language as tl
+
+from ..utils import libentry
+
+
+@libentry()
+@triton.jit
+def embedding_kernel(
+    Y,  # pointer to the output
```
Bowen12992:
```suggestion
    out_ptr,
```
This is not a necessary modification suggestion, but I don’t think this is an informative code comment. Why not reflect the meaning of the comment in the variable name?

src/flag_gems/ops/embedding.py
```diff
@@ -0,0 +1,188 @@
...
+    row_idx = tl.load(X).to(tl.int32)
+    W += row_idx * N
+    embedding_weight = tl.load(W + cols, mask, other=0.0)
+    tl.store(Y + cols, embedding_weight, mask)
+
+
+@libentry()
+@triton.jit
+def indice_freq_kernel(
+    indices_freq,  # indice frequency
```
Bowen12992, line 36: Also not an informative code comment here

Bowen12992:
```suggestion
    indices_freq,
```
Also not an informative code comment here

src/flag_gems/ops/embedding.py
```diff
@@ -0,0 +1,188 @@
...
+
+class Embedding(torch.autograd.Function):
+    @staticmethod
+    def forward(
+        ctx, weight, indices, padding_idx=-1, scale_grad_by_freq=False, sparse=False
+    ):
+        logging.debug("GEMS EMBEDDING FORWARD")
+        assert not sparse, "Currently do not support sparse format"
+
+        M = math.prod(indices.shape)
```
Bowen12992, line 114: Import tl.math from different paths for different triton version
Ref：https://github.com/FlagOpen/FlagGems/pull/104

tests/test_special_ops.py
```diff
@@ -156,3 +157,42 @@ def test_apply_rotary_pos_emb(
...
+    )
+    embedding = torch.randn(
+        (EmbeddingSize, N), device="cuda", dtype=dtype, requires_grad=True
+    )
+    ref_embedding = to_reference(embedding)
+
+    res_out = torch.nn.functional.embedding(
+        indices, embedding, padding_idx, scale_grad_by_freq=scale_grad_by_freq
+    )
+    ref_out = flag_gems.embedding(
```
StrongSpoon: you could call it by torch.nn.functional.embedding with operator registered in aten library

tests/test_special_ops.py
```diff
@@ -156,3 +157,42 @@ def test_apply_rotary_pos_emb(
...
+    )
+
+    out_grad = torch.randn_like(ref_out)
+    ref_grad = to_reference(out_grad)
+
+    (ref_in_grad,) = torch.autograd.grad(ref_out, ref_embedding, ref_grad)
+    (res_in_grad,) = torch.autograd.grad(res_out, embedding, out_grad)
+
+    res_out = to_reference(res_out)
+    res_in_grad = to_reference(res_in_grad)
```
StrongSpoon: to_reference is not needed here

src/flag_gems/ops/embedding.py
```diff
@@ -0,0 +1,188 @@
...
+
+    index_element = tl.load(indices + offsets, mask=mask).to(tl.int32)
+    tl.atomic_add(indices_freq + index_element, 1, mask=mask)
+
+
+@libentry()
+@triton.jit(do_not_specialize=["padding_idx"])
+def embedding_backward_kernel(
+    GradIn,  # pointer to the gradient input
+    GradOut,  # pointer to the gradient output
```
Bowen12992: `grad_out`  Unified naming style

src/flag_gems/ops/embedding.py
```diff
@@ -0,0 +1,188 @@
...
+    ):
+        logging.debug("GEMS EMBEDDING FORWARD")
+        assert not sparse, "Currently do not support sparse format"
+
+        M = math.prod(indices.shape)
+        N = weight.shape[-1]
+
+        BLOCK_SIZE = triton.next_power_of_2(N)
+        indices = indices.contiguous()
+        weight = weight.contiguous()
```
Bowen12992, line 119: Is it necessary to serialize here first? because it may cause memory copy overhead

Bowen12992 (commented):
I think we can replace `torch.nn.functional.embedding` in `src/flag_gems/__init__.py` so we can use it with `flag_gems.enable()` or `with flag_gems.use_gems():` ; I wonder the performance of this op and whether we should do some related tuning work, so would you add a benchmark for this OP?

src/flag_gems/ops/embedding.py
```diff
@@ -0,0 +1,188 @@
...
+        output = torch.empty(
+            (*indices.shape, N), device=indices.device, dtype=weight.dtype
+        )
+
+        embedding_kernel[M,](output, indices, weight, N, BLOCK_SIZE)
```
StrongSpoon: please add `with torch.cuda.device(q_embed.device):` here to guard computation is on the same device as input tensor

src/flag_gems/ops/embedding.py
```diff
@@ -0,0 +1,188 @@
...
+            indice_freq_kernel[indice_grid](
+                indice_freq, ctx.indices, ctx.M, INDICE_BLOCK_SIZE
+            )
+        else:
+            indice_freq = None
+
+        BLOCK_SIZE = triton.next_power_of_2(ctx.N)
+
+        HAS_PADDING_IDX = ctx.padding_idx is not None
+        embedding_backward_kernel[ctx.M,](
```
StrongSpoon: ditto

src/flag_gems/ops/embedding.py
```diff
@@ -0,0 +1,188 @@
...
+        if ctx.scale_grad_by_freq:
+            indice_freq = torch.zeros(
+                (ctx.num_weights,),
+                requires_grad=False,
+                device=grad_outputs.device,
+                dtype=torch.int32,
+            )
+            INDICE_BLOCK_SIZE = 256
+            indice_grid = lambda meta: (triton.cdiv(ctx.M, INDICE_BLOCK_SIZE),)
+            indice_freq_kernel[indice_grid](
```
StrongSpoon: ditto

src/flag_gems/ops/embedding.py
```diff
@@ -0,0 +1,188 @@
...
+    BLOCK_SIZE: tl.constexpr,
+):
+    pid = tl.program_id(0)
+    GradOut += pid * N
+    indices += pid
+
+    mask = tl.arange(0, BLOCK_SIZE) < N
+    cols = tl.arange(0, BLOCK_SIZE)
+
+    row_idx = tl.load(indices).to(tl.int32)
```
StrongSpoon, line 69: is it necessary to cast to tl.int32?

src/flag_gems/ops/embedding.py
```diff
@@ -0,0 +1,188 @@
...
+        tl.atomic_add(GradIn + cols, embedding_grad, mask=mask)
+    else:
+        if row_idx != padding_idx:
+            GradIn += row_idx * N
+            embedding_grad = tl.load(GradOut + cols, mask, other=0.0)
+            tl.atomic_add(GradIn + cols, embedding_grad, mask=mask)
+
+
+@libentry()
+@triton.jit(do_not_specialize=["n_rows", "N"])
```
StrongSpoon: I prefer specializing "N"

src/flag_gems/ops/embedding.py
```diff
@@ -0,0 +1,188 @@
...
+    n_rows,
+    N,
+    BLOCK_SIZE: tl.constexpr,
+):
+    row_start = tl.program_id(0)
+    row_step = tl.num_programs(0)
+
+    for row_idx in range(row_start, n_rows, row_step):
+        embedding_scale = 1.0
+        indice_freq_val = tl.load(indice_freq + row_idx).to(tl.int32)
```
StrongSpoon: ditto, is casting necessary?

StrongSpoon (commented):
1. I suggest implementing performance test as well.
2. Accuracy tests failed with relative difference equals to 1.0 . What about rebasing on the latest master branch and trying again?

Bowen12992:
> 1. I suggest implementing performance test as well.
> 2. Accuracy tests failed with relative difference equals to 1.0 . What about rebasing on the latest master branch and trying again?


We have verified locally that the problem of unit test accuracy can be solved by pulling the latest code

src/flag_gems/ops/embedding.py
MARD1NO: the math.prod use python `math` package not tl.math :D

src/flag_gems/ops/embedding.py
MARD1NO: OK, I just make mistake that we just align the Autograd.function apply interface to aten::embedding, the `embedding` wrapper align to torch.nn.functional.embedding, I will reconstruct it

MARD1NO: I also have another question is https://github.com/pytorch/pytorch/blob/main/aten/src/ATen/native/native_functions.yaml#L2274
here padding_idx default value is -1, but in torch.nn.functional.embedding, its default value is None, which default value should I use here?

src/flag_gems/ops/embedding.py
MARD1NO: I will delete this cast

src/flag_gems/ops/embedding.py
MARD1NO: I will delete it

src/flag_gems/ops/embedding.py
StrongSpoon: we keep up with aten interface instead of torch. cause operator is registered as an implementation choice of aten dispatcher. the default padding_idx value should be -1.

Bowen12992 (approved):
LGTM

PR #73, opened 2024-06-17
add ignore_index for cross_entropy_loss

FatJhon:
add ignore_index for cross_entropy_loss:
default ignore_index is -100, means no need to ignore.
 ignore_index value should be within [0, C-1].

### DIFF
--- a/src/flag_gems/ops/cross_entropy_loss.py
+++ b/src/flag_gems/ops/cross_entropy_loss.py
@@ -125,7 +125,9 @@ def softmax_and_sub_kernel(
     target = tl.load(target_ptrs, mask=mask, other=0.0)
     out_grad_ptr = out_grad + m_offset[:, None] * K + pid_k
     out_grad_value = tl.load(out_grad_ptr)
-    out = out_grad_value * (softmax_output - target) / mean_num
+    # backward formula derivation for value of ingnore index
+    target_sum = tl.sum(target, axis=1)[:, None]
+    out = out_grad_value * (target_sum * softmax_output - target) / mean_num
     output_ptrs = output_ptr + offset
 
     tl.store(output_ptrs, out, mask=mask)
@@ -184,9 +186,10 @@ def softmax_and_sub_reduce_kernel(
     softmax_output = numerator / denominator
     target_ptrs = target_ptr + offset
     target = tl.load(target_ptrs, mask=mask, other=0.0)
-
+    # backward formula derivation for value of ingnore index
+    target_sum = tl.sum(target, axis=1)[:, None]
     out_grad_value = tl.load(out_grad)
-    out = out_grad_value * (softmax_output - target) / mean_num
+    out = out_grad_value * (target_sum * softmax_output - target) / mean_num
     output_ptrs = output_ptr + offset
 
     tl.store(output_ptrs, out, mask=mask)
@@ -198,6 +201,7 @@ class CrossEntropyLoss(torch.autograd.Function):
         logging.debug("GEMS CrossEntropyLoss")
         assert reduction in Reduction._value2member_map_, "Invalid reduction"
         assert isinstance(input, torch.Tensor), "input is not a tensor"
+
         if input.ndim >= 2:
             dim = 1
         else:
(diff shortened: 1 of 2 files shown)


src/flag_gems/ops/cross_entropy_loss.py
```diff
@@ -198,18 +201,29 @@ def forward(ctx, input, target, weight, reduction, ignore_index, label_smoothing
...
+
         if input.ndim >= 2:
             dim = 1
         else:
             dim = 0
         if reduction != Reduction.MEAN.value:
             mean_num = -1
         else:
-            mean_num = -target.numel()
+            if ignore_index != -100:
```
tongxin: It occurs to me -100 is used as the default ignore_index value that indicates that the target class for a particular input is to be ignored.  Here the targets are ignored only when ignore_index is not -100.

tongxin (commented):
ignore_index is handled in a different way than pytorch

src/flag_gems/ops/cross_entropy_loss.py
tongxin: it suffices to remove this condition to make it correct.

src/flag_gems/ops/cross_entropy_loss.py
```diff
@@ -198,18 +201,29 @@ def forward(ctx, input, target, weight, reduction, ignore_index, label_smoothing
...
+                ignore_count = (target == ignore_index).to(torch.int16).sum().item()
```
tongxin: this line causes extra kernel launches

src/flag_gems/ops/cross_entropy_loss.py
```diff
@@ -198,18 +201,29 @@ def forward(ctx, input, target, weight, reduction, ignore_index, label_smoothing
...
+            else:
+                ignore_count = 0
+            mean_num = -target.numel() + ignore_count
         shape = list(input.shape)
         shape[dim] = 1
         target = torch.zeros_like(input).scatter(dim, target.view(shape), 1)
```
tongxin, line 239: target may contain negative values

src/flag_gems/ops/cross_entropy_loss.py
```diff
@@ -198,18 +201,29 @@ def forward(ctx, input, target, weight, reduction, ignore_index, label_smoothing
...
 
+        # set ignore value 0
+        if ignore_index != -100:
```
tongxin: ignore_index can be anything outside [1, C)

src/flag_gems/ops/cross_entropy_loss.py
```diff
@@ -125,7 +125,9 @@ def softmax_and_sub_kernel(
     target = tl.load(target_ptrs, mask=mask, other=0.0)
     out_grad_ptr = out_grad + m_offset[:, None] * K + pid_k
     out_grad_value = tl.load(out_grad_ptr)
-    out = out_grad_value * (softmax_output - target) / mean_num
```
tongxin, line 128: how about replacing all `/ mean_num` with `* reciprocal_count`

src/flag_gems/ops/cross_entropy_loss.py
tongxin: plus, casting to int16 is not necessary

src/flag_gems/ops/cross_entropy_loss.py
```diff
@@ -198,18 +201,48 @@ def forward(ctx, input, target, weight, reduction, ignore_index, label_smoothing
...
             dim = 1
         else:
             dim = 0
+
         if reduction != Reduction.MEAN.value:
             mean_num = -1
         else:
-            mean_num = -target.numel()
+            # get all ingore count of target
+            ignore_count = (target == ignore_index).to(torch.int16).sum().item()
```
tongxin, line 214: int16 might be too small for size count

tongxin:
```suggestion
            ignore_count = (target == ignore_index).to(torch.int32).sum().item()
```

tongxin (approved):
LGTM

src/flag_gems/ops/cross_entropy_loss.py
```diff
@@ -198,18 +201,49 @@ def forward(ctx, input, target, weight, reduction, ignore_index, label_smoothing
...
+
+            mean_num = -target.numel() + ignore_count
+
+        # special mean_num is 0, return 0
+        if mean_num == 0:
+            ctx.save_for_backward(input, target)
+            ctx.dim = dim
+            ctx.mean_num = -mean_num
+            ctx.reduction = reduction
+            return torch.tensor(float("nan"), device=input.device, dtype=input.dtype)
```
StrongSpoon: if mean_num is 0, is there necessity of saving for backward?

StrongSpoon (approved):
lgtm

PR #75, opened 2024-06-18
[Operator] optimization of vector_norm, tanh, sigmoid, layernorm

StrongSpoon:
1. implement two-pass kernels for all reduce vector_norm
2. avoid upcasting of half precision for tanh and sigmoid when gradient is not required
3. expand config tuning space for LayerNorm

### DIFF
--- a/src/flag_gems/ops/sigmoid.py
+++ b/src/flag_gems/ops/sigmoid.py
@@ -27,9 +27,13 @@ class Sigmoid(torch.autograd.Function):
     @staticmethod
     def forward(ctx, A):
         logging.debug("GEMS SIGMOID FORWARD")
-        out = sigmoid_forward(A.to(torch.float32))
-        ctx.save_for_backward(out)
-        return out.to(A.dtype)
+        if A.requires_grad is True:
+            out = sigmoid_forward(A.to(torch.float32))
+            ctx.save_for_backward(out)
+            return out.to(A.dtype)
+        else:
+            out = sigmoid_forward(A)
+            return out
 
     @staticmethod
     def backward(ctx, out_grad):
--- a/src/flag_gems/ops/tanh.py
+++ b/src/flag_gems/ops/tanh.py
@@ -23,9 +23,13 @@ class Tanh(torch.autograd.Function):
     @staticmethod
     def forward(ctx, A):
         logging.debug("GEMS TANH FORWARD")
-        out = tanh_forward(A.to(torch.float32))
-        ctx.save_for_backward(out)
-        return out.to(A.dtype)
+        if A.requires_grad is True:
+            out = tanh_forward(A.to(torch.float32))
+            ctx.save_for_backward(out)
+            return out.to(A.dtype)
+        else:
+            out = tanh_forward(A)
+            return out
 
     @staticmethod
     def backward(ctx, out_grad):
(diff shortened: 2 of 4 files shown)


src/flag_gems/ops/vector_norm.py
```diff
@@ -136,28 +263,53 @@ def vector_norm(x, ord=2, dim=None, keepdim=False, dtype=None):
         dtype = x.dtype
     if dtype not in [torch.float16, torch.float32, torch.bfloat16]:
         raise NotImplementedError(f"vector_norm not implemented for {dtype}")
-    if dim is None:
+
+    if dim is None or len(dim) == x.ndim:
```
iclementine, line 267: In this case(`full reduce`), it is not enasured that the intput is contiguous. So maybe `x=x.contiguous()` is required, or the stage-1 kernel should consider the stride of the input.

StrongSpoon: solved with dim_compress

src/flag_gems/ops/sigmoid.py
```diff
@@ -27,9 +27,13 @@ class Sigmoid(torch.autograd.Function):
...
-        out = sigmoid_forward(A.to(torch.float32))
-        ctx.save_for_backward(out)
-        return out.to(A.dtype)
+        if A.requires_grad is True:
+            out = sigmoid_forward(A.to(torch.float32))
+            ctx.save_for_backward(out)
+            return out.to(A.dtype)
+        else:
+            out = sigmoid_forward(A)
+            return out
```
tongxin, line 36: this may render the results in different precision given grad and no grad context. Shall we just keep the old version?

src/flag_gems/ops/tanh.py
```diff
@@ -23,9 +23,13 @@ class Tanh(torch.autograd.Function):
...
-        out = tanh_forward(A.to(torch.float32))
-        ctx.save_for_backward(out)
-        return out.to(A.dtype)
+        if A.requires_grad is True:
+            out = tanh_forward(A.to(torch.float32))
+            ctx.save_for_backward(out)
+            return out.to(A.dtype)
+        else:
+            out = tanh_forward(A)
+            return out
```
tongxin, line 32: Same issue as sigmoid

tongxin (approved):
My only question is about the castings in tanh and sigmoid.  Otherwise looks good.

src/flag_gems/ops/sigmoid.py
StrongSpoon: input in any data types is upcast to tl.float32 in triton kernel, thus the precision of computing is the same. and for half precision, even if we upcast the input tensor to torch.float32 before calling kernel, it will be downcast when returning.

tongxin (approved):
LG

src/flag_gems/ops/vector_norm.py
```diff
@@ -38,6 +39,31 @@ def l2_norm_kernel(X, Out, M, N, BLOCK_M: tl.constexpr, BLOCK_N: tl.constexpr):
...
+@libentry()
+@triton.jit
+def l2_norm_kernel_1(X, Mid, M, BLOCK_SIZE: tl.constexpr):
+    pid = tl.program_id(0)
+    offset = pid * BLOCK_SIZE + tl.arange(0, BLOCK_SIZE)
+    X = X + offset
+    Mid = Mid + pid
+    mask = offset < M
+
+    x = tl.load(X, mask=mask, other=0.0).to(tl.float32)
```
tongxin, line 51: Pytorch ppl found that adding evict_fist to loads that are only used once may lead to better perf.

PR #79, opened 2024-06-21
[Operator] Type promotion for pointwise Ops

Bowen12992:
- Add Type promotion for pointwise Ops 
- Add trunc & floor mode for divOp
- Move Type promotion into pointwise JIT 
- Support different condition of whereOp
- Add UTs of type promotion for pointwise Ops
- Fix related docs

### DIFF
--- a/src/flag_gems/ops/div.py
+++ b/src/flag_gems/ops/div.py
@@ -6,19 +6,19 @@ import triton
 from ..utils import pointwise_dynamic
 
 
-@pointwise_dynamic(promotion_methods=[[0, 1, "INT_TO_FLOAT"]])
+@pointwise_dynamic(promotion_methods=[(0, 1, "INT_TO_FLOAT")])
 @triton.jit
 def true_div_func(x, y):
     return x / y
 
 
-@pointwise_dynamic(is_tensor=[True, False], promotion_methods=[[0, 1, "INT_TO_FLOAT"]])
+@pointwise_dynamic(is_tensor=[True, False], promotion_methods=[(0, 1, "INT_TO_FLOAT")])
 @triton.jit
 def true_div_func_tensor_scalar(x, y):
     return x / y
 
 
-@pointwise_dynamic(is_tensor=[False, True], promotion_methods=[[0, 1, "INT_TO_FLOAT"]])
+@pointwise_dynamic(is_tensor=[False, True], promotion_methods=[(0, 1, "INT_TO_FLOAT")])
 @triton.jit
 def true_div_func_scalar_tensor(x, y):
     return x / y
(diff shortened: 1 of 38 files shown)


src/flag_gems/utils/type_utils.py
```diff
@@ -0,0 +1,9 @@
+import torch._prims_common as utils
+
+
+def type_promotion(*args, type_promotion: utils.ELEMENTWISE_TYPE_PROMOTION_KIND):
```
iclementine, line 4: Maybe we can use a simpler implementation for type promotion rules.(e.g. use `torch.promote_types` or `torch.result_type` for `ELEMENTWISE_TYPE_PROMOTION_KIND.DEFAULT`, and write our own implementation for other cases).

src/flag_gems/utils/pointwise_dynamic.py
```diff
@@ -50,12 +51,18 @@ def __init__(
         is_tensor: Optional[List[bool]] = None,
         dtypes: Optional[List[Optional[type]]] = None,
         num_outputs: Optional[int] = None,
-        output_dtypes: Optional[List[torch.dtype]] = None,
+        promotion_methods: Optional[List[Tuple[int, ...]]] = None,
```
iclementine, line 54: What about still support `List[torch.dtype]`?

Bowen12992: I thought because `pointwise_dynamic` is for element_wise ops and once we add type promotion it can deal with more complicated input type combination，so we should use promotion to get all kind output type for different type combination. If we still use `List[torch.dtype]` to specify output，on the one hand，it‘s hard to figure out all kind output type for different type combination, on the other hand, it makes  the logic of `pointwise_dynamic` much more complicated and confusing

src/flag_gems/utils/type_utils.py
Bowen12992: Yeah, we can simplify the logic!  but I have two concerns. 1. Will Gems support dynamic-shape with maybe `sympy` in the feature？2. I‘m going to figure out  if there are differences among `torch.promote_types`  `torch.result_type` and  `ELEMENTWISE_TYPE_PROMOTION_KIND.DEFAULT`

src/flag_gems/ops/div.py
```diff
@@ -6,32 +6,106 @@
...
+    elif isinstance(A, torch.Tensor):
+        return floor_div_func_tensor_scalar(A, B)
+    elif isinstance(B, torch.Tensor):
+        return floor_div_func_scalar_tensor(A, B)
+    else:
+        # Both scalar
+        return A // B
+
+
+def div(A, B, rounding_mode=None):
```
StrongSpoon, line 102: aten interface div.Tensor doesn't have argument named rounding_mode. do you want to implement 'div.Tensor_mode'?

StrongSpoon: besides, trunc_divide might come to accuracy error due to low precision, and we didn't solve it last time.

Bowen12992: I use  `div_round_zero` for  `trunc_divide` rather than `O_val = tl.math.trunc([X_val.to](http://x_val.to/)(tl.float64) / Y_val)` here and did some tests, i didn't get accuracy error due to low precision any more

src/flag_gems/utils/type_utils.py
iclementine: We may not support symbolic shape since flaggems is basically designed for processing real tensors, with concrete shapes.

iclementine (approved):
LGTM

PR #82, opened 2024-06-26
[Operator] Add isclose & allclose op

zhzhcookie:
1. Add isclose and allclose op.
2. Add unit tests and performance tests of these two ops.

### DIFF
--- a/src/flag_gems/ops/isclose.py
+++ b/src/flag_gems/ops/isclose.py
@@ -5,6 +5,7 @@ import triton
 import triton.language as tl
 
 from ..utils import pointwise_dynamic
+from .all import all
 
 
 @pointwise_dynamic(is_tensor=[True, True, False, False], output_dtypes=[torch.bool])
@@ -44,10 +45,21 @@ def isclose(
     rtol = 1e-05,
     atol = 1e-08,
     equal_nan : bool = False,
-):
+) -> torch.Tensor:
     logging.debug("GEMS ISCLOSE")
     if equal_nan:
         return isclose_func_equal_nan(A, B, rtol, atol)
     else:
         return isclose_func(A, B, rtol, atol)
+
+
+def allclose(
+    A : torch.Tensor,
+    B : torch.Tensor,
+    rtol = 1e-05,
+    atol = 1e-08,
+    equal_nan : bool = False,
+) -> bool:
+    logging.debug("GEMS ALLCLOSE")
+    return all(isclose(A, B, rtol, atol, equal_nan)).item()
 
--- a/src/flag_gems/ops/isclose.py
+++ b/src/flag_gems/ops/isclose.py
@@ -47,6 +47,12 @@ def isclose(
     equal_nan: bool = False,
 ) -> torch.Tensor:
     logging.debug("GEMS ISCLOSE")
+    if rtol < 0:
+        raise RuntimeError(
+            "rtol must be greater than or equal to zero, but got {}".format(rtol))
+    if atol < 0:
+        raise RuntimeError(
+            "atol must be greater than or equal to zero, but got {}".format(atol))
     if equal_nan:
         return isclose_func_equal_nan(A, B, rtol, atol)
     else:
--- a/src/flag_gems/ops/isclose.py
+++ b/src/flag_gems/ops/isclose.py
@@ -9,8 +9,8 @@ from .all import all
 
 
 @pointwise_dynamic(
-    is_tensor=[True, True, False, False, False, False, False],
-    promotion_methods=[(0, 1, 2, 3, 4, 5, 6, "ALWAYS_BOOL")],
+    is_tensor=[True, True, False, False, False, False, False, False],
+    promotion_methods=[(0, 1, "ALWAYS_BOOL")],
 )
 @triton.jit
 def isclose_func(
@@ -19,49 +19,30 @@ def isclose_func(
     rtol,
     atol,
     float_dtype: tl.constexpr,
+    reduced_dtype: tl.constexpr,
     zero_tol: tl.constexpr,
     equal_nan: tl.constexpr,
 ):
-    if zero_tol:
-        if equal_nan:
-            x_nan = x != x
-            y_nan = y != y
-            return tl.where(
-                x_nan | y_nan,
-                x_nan == y_nan,
-                x == y,
-            )
-        else:
-            return x == y
-    if float_dtype:
-        if x.dtype == torch.float64:
-            x_fp = x
-            y_fp = y
-        else:
-            x_fp = x.to(tl.float32)
-            y_fp = y.to(tl.float32)
-        if equal_nan:
-            x_nan = x_fp != x_fp
-            y_nan = y_fp != y_fp
-            return tl.where(
-                x_nan | y_nan,
-                x_nan == y_nan,
-                tl.where(
-                    tl.math.isinf(x_fp) | tl.math.isinf(y_fp),
-                    x_fp == y_fp,
-                    tl.abs(x - y) <= atol + rtol * tl.abs(y),
-                ),
-            )
-        else:
-            return tl.where(
-                tl.math.isinf(x_fp) | tl.math.isinf(y_fp),
-                x_fp == y_fp,
-                tl.abs(x - y) <= atol + rtol * tl.abs(y),
-            )
+    cast_x = x if x.dtype == tl.float64 else x.to(tl.float32)
+    cast_y = y if x.dtype == tl.float64 else y.to(tl.float32)
+    if x.dtype == tl.bfloat16:
+        close = cast_x == cast_y
+    elif reduced_dtype:
+        close = cast_x == cast_y
     else:
-        x_long = x.to(tl.int64)
-        y_long = y.to(tl.int64)
-        return tl.abs(x_long - y_long) <= atol + rtol * tl.abs(y_long)
+        close = x == y
+    if equal_nan:
+        if float_dtype:
+            close |= (cast_x != cast_x) & (cast_y != cast_y)
+    if zero_tol:
+        return close
+    else:
+        allowed_error = atol + tl.abs(rtol * cast_y)
+        actual_error = tl.abs(cast_x - cast_y)
+        actual_error_finite = (actual_error != float("inf")) & (
+            actual_error != float("-inf")
+        )
+        return close | (actual_error_finite & (actual_error <= allowed_error))
 
 
 def _isclose(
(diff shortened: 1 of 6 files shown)


src/flag_gems/ops/isclose.py
```diff
@@ -0,0 +1,64 @@
...
+@triton.jit
+def isclose_func(x, y, rtol, atol):
+    x_fp = x.to(tl.float64)
+    y_fp = y.to(tl.float64)
+    return tl.where(
+        tl.math.isinf(x.to(tl.float32)) | tl.math.isinf(y.to(tl.float32)),
+        x_fp == y_fp,
+        tl.abs(x_fp - y_fp) <= atol + rtol * tl.abs(y_fp),
+    )
+    # return tl.abs(x - y) <= atol + rtol * tl.abs(y)
```
Bowen12992: useless comment

src/flag_gems/ops/isclose.py
```diff
@@ -0,0 +1,64 @@
...
+import triton.language as tl
+
+from ..utils import pointwise_dynamic
+from .all import all
+
+
+@pointwise_dynamic(is_tensor=[True, True, False, False], output_dtypes=[torch.bool])
+@triton.jit
+def isclose_func(x, y, rtol, atol):
+    x_fp = x.to(tl.float64)
```
iclementine: maybe for integer & floating point & bool, is_close should use different paths(use different JITFunction for different dtypes)?
Is it necesary to cast them all to float64 and then float32?

src/flag_gems/ops/isclose.py
zhzhcookie: done

src/flag_gems/ops/isclose.py
zhzhcookie: > maybe for integer & floating point & bool, is_close should use different paths(use different JITFunction for different dtypes)? Is it necesary to cast them all to float64 and then float32?

Thanks for the reminder, so I implemente a specific integer JITFunction, because there is no need to judge nan and inf when inputs are integers, so the efficiency will be a little higher. 
Floating point numbers must be converted to float64, otherwise the precision is not enough and it cannot be completely consistent with the judgment result of torch.

src/flag_gems/ops/isclose.py
```diff
@@ -0,0 +1,134 @@
...
+
+from ..utils import pointwise_dynamic
+from .all import all
+
+
+@pointwise_dynamic(is_tensor=[True, True, False, False], output_dtypes=[torch.bool])
+@triton.jit
+def isclose_func(x, y, rtol, atol):
+    x_fp = x.to(tl.float32)
+    y_fp = y.to(tl.float32)
```
tongxin: the operands should be compared in their original precision in case of no type promotion, right?

src/flag_gems/ops/isclose.py
```diff
@@ -0,0 +1,134 @@
...
+
+def isclose(
+    A: torch.Tensor,
+    B: torch.Tensor,
+    rtol=1e-05,
+    atol=1e-08,
+    equal_nan: bool = False,
+) -> torch.Tensor:
+    logging.debug("GEMS ISCLOSE")
+    return _isclose(A, B, rtol, atol, equal_nan)
```
tongxin: what's the point of this extra redirection?

src/flag_gems/ops/isclose.py
```diff
@@ -0,0 +1,134 @@
...
+    equal_nan: bool = False,
+) -> torch.Tensor:
+    if A.dtype != B.dtype:
+        raise RuntimeError("{} did not match {}".format(A.dtype, B.dtype))
+    if A.is_quantized or B.is_quantized:
+        raise RuntimeError("isclose is not supported for quantized inputs.")
+    if rtol < 0:
+        raise RuntimeError(
+            "rtol must be greater than or equal to zero, but got {}".format(rtol)
+        )
```
tongxin, line 75: How about just assert rtol >=0 ?

src/flag_gems/ops/isclose.py
```diff
@@ -0,0 +1,134 @@
...
+            "atol must be greater than or equal to zero, but got {}".format(atol)
+        )
+
+    if (
+        A.dtype == torch.int64
+        or A.dtype == torch.int32
+        or A.dtype == torch.int16
+        or A.dtype == torch.int8
+        or A.dtype == torch.bool
+    ):
```
tongxin: A more common Python style is 
```
if A.dtype in (torch.int64, torch.int32, ... ):
```

src/flag_gems/ops/isclose.py
```diff
@@ -0,0 +1,134 @@
...
+        return isclose_func_int(A, B, rtol, atol)
```
tongxin: I think these different versions of Triton JITFunction could be better of implemented in one JITFunction, putting dtype and other constexpr specific logics in static control flow blocks. The Triton JIT compiler is able to take care of kernel versioning and all.

tongxin: You can pass specialization keys as params to the JITFunction. for instance, is_equalnan

src/flag_gems/ops/isclose.py
zhzhcookie: > How about just assert rtol >=0 ?

These inputs checks are consistent with ATen, including error types and messages.

src/flag_gems/ops/isclose.py
iclementine: to allow adding docstring and use the desired parameter name since the generated pointwise dynamic function has a unified naming convention.

src/flag_gems/ops/isclose.py
iclementine: Using some constexpr arguments in JITFunction means that it can generate multiple CompiledKernel w.r.t. the value of the constexpr arguments.

src/flag_gems/ops/isclose.py
zhzhcookie: > what's the point of this extra redirection?

To avoid allclose printing both "GEMS ALLCLOSE" & "GEMS ISCLOSE". Is it necessary?

src/flag_gems/ops/isclose.py
iclementine: for example `ONE_TILE_PER_CTA` is a constexpr in https://github.com/FlagOpen/FlagGems/blob/36ee6e39943d205f99690ec1f23f020ea7214d9b/src/flag_gems/ops/softmax.py#L49C1-L58C3

src/flag_gems/ops/isclose.py
tongxin: One more function call means more interpretation overhead. Unless _isclose is called at multiple sites.

src/flag_gems/ops/isclose.py
```diff
@@ -0,0 +1,99 @@
...
+    if zero_tol:
+        return close
+    else:
+        allowed_error = atol + tl.abs(rtol * cast_y)
+        actual_error = tl.abs(cast_x - cast_y)
+        actual_error_finite = (actual_error != float("inf")) & (
+            actual_error != float("-inf")
+        )
+        return close | (actual_error_finite & (actual_error <= allowed_error))
+
```
tongxin: Code can be simplified a bit. Following is my suggestion. 

```
@triton.jit
def close_kernel(x, y, z, atol, rtol, equal_nan: tl.constexpr, zero_tol: tl.constexpr, BLOCK: tl.constexpr):
    start = tl.program_id(0) * BLOCK
    offset = start + tl.arange(0, BLOCK)
    X = tl.load(x + offset)
    Y = tl.load(y + offset)
    if X.dtype.is_bf16():
        X = X.to(tl.float32)
        Y = Y.to(tl.float32)
    close = X == Y

    if equal_nan:
        close |= (X != X) & (Y != Y)

    if not zero_tol:
        if X.dtype != tl.float64:
            X = X.to(tl.float32)
            Y = Y.to(tl.float32)
        actual = tl.abs(X - Y)
        allowed = atol + tl.abs(rtol * Y)
        isfinite = tl.math.finitef(actual)
        close |= isfinite & (actual < allowed)
 
    tl.store(z + offset, close)
```

src/flag_gems/ops/isclose.py
zhzhcookie: > One more function call means more interpretation overhead. Unless _isclose is called at multiple sites.

OK, I fix it.

src/flag_gems/ops/isclose.py
```diff
@@ -0,0 +1,78 @@
...
+        close = cast_x == cast_y
+    else:
+        close = x == y
+    if equal_nan:
+        close |= (cast_x != cast_x) & (cast_y != cast_y)
+    if not zero_tol:
+        allowed = atol + tl.abs(rtol * cast_y)
+        actual = tl.abs(cast_x - cast_y)
+        actual_finite = (
+            tl.math.isfinited(actual) if x.dtype.is_fp64() else tl.math.finitef(actual)
```
Bowen12992: import `tl.math` from different paths
Ref：https://github.com/FlagOpen/FlagGems/pull/104

src/flag_gems/ops/isclose.py
```diff
@@ -0,0 +1,78 @@
...
+    x,
+    y,
+    rtol,
+    atol,
+    equal_nan: tl.constexpr,
+    zero_tol: tl.constexpr,
+):
+    cast_x = x if x.dtype.is_fp64() else x.to(tl.float32)
+    cast_y = y if x.dtype.is_fp64() else y.to(tl.float32)
+    if x.dtype.is_bf16():
```
iclementine, line 42: Should fp16 also use (cast_x == cast_y) to compute closeness?

src/flag_gems/ops/isclose.py
```diff
@@ -0,0 +1,78 @@
...
+        raise RuntimeError("isclose is not supported for quantized inputs.")
+    if rtol < 0:
+        raise RuntimeError(
+            "rtol must be greater than or equal to zero, but got {}".format(rtol)
+        )
+    if atol < 0:
+        raise RuntimeError(
+            "atol must be greater than or equal to zero, but got {}".format(atol)
+        )
+    zero_tol = (rtol == 0) and (atol == 0)
```
StrongSpoon, line 80: why not call eq when zero_tol is true?

src/flag_gems/ops/isclose.py
zhzhcookie: > Should fp16 also use (cast_x == cast_y) to compute closeness?

This will cause coredump...

src/flag_gems/ops/isclose.py
zhzhcookie: > why not call eq when zero_tol is true?

eq can not handle nan...

src/flag_gems/ops/isclose.py
zhzhcookie: > import `tl.math` from different paths Ref：#104

done.

tongxin (approved):
LG

PR #83, opened 2024-06-27
cambricon: fix log softmax nram

FuncSherl:
add split c kernel for log softmax forward and backward to avoid nram error


sethbrin:
LGTM

PR #89, opened 2024-06-28
[operator] fix libentry to support triton 2.3

StrongSpoon:
1. modified libentry arguments collection, cache key generation and constexpr collection
2. specify dns arguments for operators
3. works for triton 2.2 and 2.3

### DIFF
--- a/src/flag_gems/ops/dropout.py
+++ b/src/flag_gems/ops/dropout.py
@@ -41,7 +41,7 @@ del _offset
         "N",
     ],
 )
-@triton.jit
+@triton.jit(do_not_specialize=["p", "philox_seed", "philox_offset"])
 def dropout_forward_kernel(
     X,
     Y,
@@ -82,7 +82,7 @@ def dropout_forward_kernel(
         "N",
     ],
 )
-@triton.jit
+@triton.jit(do_not_specialize=["p", "philox_seed", "philox_offset"])
 def dropout_backward_kernel(
     DY,
     DX,
--- a/src/flag_gems/utils/libentry.py
+++ b/src/flag_gems/utils/libentry.py
@@ -38,6 +38,13 @@ class LibEntry(triton.KernelInterface):
         for arg in dns_args:
             if hasattr(arg, "data_ptr"):
                 entry_key.append(str(arg.dtype))
+            elif isinstance(arg, int):
+                if -(2**31) <= arg and arg <= 2**31 - 1:
+                    entry_key.append("i32")
+                elif 2**63 <= arg and arg <= 2**64 - 1:
+                    entry_key.append("u64")
+                else:
+                    entry_key.append("i64")
             else:
                 entry_key.append(type(arg))
         # const args passed by position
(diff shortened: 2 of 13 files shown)


src/flag_gems/ops/dropout.py
```diff
@@ -82,7 +82,7 @@ def dropout_forward_kernel(
         "N",
     ],
 )
-@triton.jit
+@triton.jit(do_not_specialize=["p", "philox_seed", "philox_offset"])
```
tongxin, line 85: Does Triton specialize on floats?

src/flag_gems/utils/libentry.py
```diff
@@ -14,47 +16,110 @@ def __init__(
         while not isinstance(fn, triton.runtime.JITFunction):
             fn = fn.fn
         self.jit_function: triton.runtime.JITFunction = fn
-        self.kernel_arg_indices = []
+        self.spec_indices = []
+        self.dns_indices = []
```
tongxin: let's be more precise? do_not_specialize_indices

src/flag_gems/utils/libentry.py
```diff
@@ -14,47 +16,110 @@ def __init__(
...
         for p in self.jit_function.params:
             if not p.is_constexpr:
-                self.kernel_arg_indices.append(p.num)
+                if p.do_not_specialize:
+                    self.dns_indices.append(p.num)
+                else:
+                    self.spec_indices.append(p.num)
```
tongxin: Isn't a bit of list comprehension is more favorable than loops?

src/flag_gems/utils/libentry.py
```diff
@@ -14,47 +16,110 @@ def __init__(
...
-        for arg in args:
+    def key(self, spec_args, dns_args, const_args):
+        entry_key = []
+        for arg in spec_args:
+            if hasattr(arg, "data_ptr"):
+                entry_key.append(str(arg.dtype))
+                entry_key.append(arg.data_ptr() % self.divisibility == 0)
+            else:
+                entry_key.append(type(arg))
+                entry_key.append(arg)
```
tongxin: List comprehension is an immutable fashion for coding loops. It's normally more efficient than the mutable counterpart.

src/flag_gems/ops/dropout.py
StrongSpoon: floats not marked as dns are specialized as False

src/flag_gems/utils/libentry.py
StrongSpoon: done

src/flag_gems/utils/libentry.py
StrongSpoon: done

src/flag_gems/utils/libentry.py
StrongSpoon: done

iclementine:
LGTM

The triton jit runtime is too complicated. 
- It offers several ways to pass parameter(via explicitly passing parameter, using default values, via decorator cascading(autotunner & heuristics).
- It offers several ways to specify how arguments are handled in the jit decorator (do not specialize, ...) or in the function (tl.constexpr). 
- It handles different types of actual arguments differently. (specialization of different types of arguments)
- It has some limitations of argument types, which is not well-documented.
- The parser handles values differntly depending on whether a value is a constexpr or tensor, while these processes are not well-documented.(equal to 1 arguments are handled as if they were constexpr)

After all, for quick understanding of the rationale, a cheatsheet is needed.
图片

iclementine (approved):

PR #93, opened 2024-07-03
add assertion that argnums from OPDesc matches that from scalar_fn

iclementine:
1. add assertion that argnums from OPDesc matches that from scalar_fn
2. fix mismatch in `bitwise_not`.


StrongSpoon:
error fixed in pr #92

PR #99, opened 2024-07-08
add device guard around kernel launch

iclementine:
add device guard around kernel launch，or otherwise the kernel launch would be on device 0.


StrongSpoon:
seems like misspell

StrongSpoon (approved):
lgtm

PR #100, opened 2024-07-08
[runtime] only run once when entry_key doesn't hit kernel cache

StrongSpoon:
avoid redundant execution when entry_key doesn't hit kernel cache.
support atomic operation without autotuning to pass accuracy tests.
and update README as well, with speedup data based on commit 233507bfb83b649686b914c73df86565a8cb4faa


iclementine (approved):
I see, when the key is not in cache, the JITFunction would be run once to get the compiled kernel.

But What about returning the CompiledKernel & constexprs following JITFunction's convention?

pyproject.toml
```diff
@@ -17,7 +17,8 @@ classifiers = [
     "License :: OSI Approved :: Apache Software License",
 ]
 dependencies = [
-    "triton>=2.2.0",
+    "triton>=2.2.0, <3.0.0",
+    "torch>=2.1.2",
     "transformers>=4.40.2"
```
iclementine, line 22: FYI: version spec and version entry in pyproject.toml
https://peps.python.org/pep-0508/
https://peps.python.org/pep-0631/

iclementine (approved):

iclementine (approved):

PR #102, opened 2024-07-08
Develop TopK[SiliconFlow]

MARD1NO:
测试环境4090

元素为4096，topk=2

torch耗时：20.58 + 4.64
triton耗时：2.53 + 2.75 

image

### MicroBenchmark
这里topk的k值选了5

调整了下 topk chunksize 目前性能都能优于torch一倍

image


Bowen12992 (commented):
LGTM. Will this PR support topk with different condition of `dim`,`largest`, `sorted` args?

MARD1NO:
> LGTM. Will this PR support topk with different condition of `dim`,`largest`, `sorted` args?

has supported `largest=False` (equals to ascending)

currently only support topk in last dim(I think it is the common case)

And I think there is no benefit to support `sorted=False`

src/flag_gems/ops/topk.py
```diff
@@ -0,0 +1,295 @@
...
+    # If dim equals to last dim, we set it to -1.
+    if dim == x.ndim - 1:
+        dim = -1
+
+    assert dim == -1, "Currently only support topk in last dimension"
+    assert sorted, "Currently only support sorted == True"
+
+    descending = True
+    if not largest:
+        descending = False
```
StrongSpoon, line 245: is descending redundant?

src/flag_gems/ops/topk.py
```diff
@@ -0,0 +1,295 @@
...
+import triton
+import triton.language as tl
+import triton.language.core as core
+from triton.language.standard import (
+    _get_sort_dim,
+    _indicator,
+    _is_power_of_two,
+    _log2,
+    _take_slice,
+    zeros_like,
```
StrongSpoon: _get_sort_dim, _indicator, _take_slice are deleted in triton3.0. maybe we can re-implement these functions.

src/flag_gems/ops/topk.py
```diff
@@ -0,0 +1,295 @@
...
+    index_ptr += cur_batch * chunk_num * k + cur_chunk_idx * k
+
+    chunk_offset = cur_chunk_idx * CHUNK_SIZE
+    x_ptr += cur_batch * N + chunk_offset
+
+    cols = tl.arange(0, CHUNK_SIZE)
+    mask = (chunk_offset + cols) < N
+
+    mask_val = _get_finfo_val(x_ptr.dtype.element_ty, return_max=not DESCENDING)
+    x_val = tl.load(x_ptr + cols, mask=mask, other=mask_val).to(tl.float32)
```
StrongSpoon, line 69: can't float("inf") and -float("inf") perform the same?

MARD1NO: It seems there exist bug when dtype is float16 / bfloat16, I have to choose use torch finfo

src/flag_gems/ops/topk.py
MARD1NO: no, it control the sort order of topk result

StrongSpoon: I think it's actually the opposite of largest. maybe there is no need to create a new variable.

src/flag_gems/ops/topk.py
MARD1NO: I have fixed it to support both triton2.3.0 and triton3.0.0

Bowen12992 (approved):

PR #98, opened 2024-07-05
[codegen] support one_tile_per_cta, and change the max tile_size to 1024

iclementine:
update code generation:
1. change the way to inline scalar function: It is re-imported into the generated module and called in the generated JITFunction;
2. change the max tile_size back to a moderate value(1024);
3. add a constexpr argument one_tile_per_cta to control whether it is neede to use grid-stride-loop.

PR #104, opened 2024-07-08
try to import tl.math functions from different paths

iclementine:
try to import tl.math functions from different path at various versions of triton

## why
triton moves some of the function in `tl.math` to `tl.extra.cuda.libdevice`, which *breaks* backward compatibility, so we work around it.

## how

Since PR #98 changes the way to inline function, now we can directly *call* a JITFunction in another JITFunction.

but from where should we import the inlined function? well, ... try except

```python
try:
    from triton.language.extra.cuda.libdevice import rsqrt
except ImportError:
    try:
        from triton.language.math import rsqrt
    except ImportError:
        from triton.language.libdevice import rsqrt
```

## Caveats
Never mask a symbol(redefine it with the same name) which is used in JITFunction, since the `DependencyFinder` would find the latest value.

For example, `src/flag_gems/ops/tanh.py`

```python
try:
    from triton.language.extra.cuda.libdevice import tanh as _tanh
except ImportError:
    try:
        from triton.language.math import tanh as _tanh
    except ImportError:
        from triton.language.libdevice import tanh as _tanh


@pointwise_dynamic(promotion_methods=[(0, "INT_TO_FLOAT")])
@triton.jit
def tanh_forward(x):
    return _tanh(x.to(tl.float32))

...

# CAUTION: it would mask tanh if we do not import tanh as _tanh
def tanh(A): 
    return Tanh.apply(A)
```

We import tanh as `_tanh` to avoid masking that name later.

### DIFF
--- a/src/flag_gems/fused/gelu_and_mul.py
+++ b/src/flag_gems/fused/gelu_and_mul.py
@@ -6,12 +6,20 @@ import triton.language as tl
 
 from ..utils import pointwise_dynamic
 
+try:
+    from triton.language.extra.cuda.libdevice import erf, pow, tanh
+except ImportError:
+    try:
+        from triton.language.math import erf, pow, tanh
+    except ImportError:
+        from triton.language.libdevice import erf, pow, tanh
+
 
 @pointwise_dynamic(promotion_methods=[(0, 1, "DEFAULT")])
 @triton.jit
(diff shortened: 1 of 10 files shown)


Bowen12992 (approved):
LGTM

PR #108, opened 2024-07-09
[Operator] Add addmm tuning config

hesse-x:


Bowen12992 (approved):
LGTM，it‘s better to provide some more performance results into PR description

PR #115, opened 2024-07-11
[Runtime] Heuristic expression allowing python3.8

StrongSpoon:
For python before version 3.10.10, lambda expression might cause python inspect error. It is fixed [here](https://github.com/python/cpython/pull/99654) , but requires higher version.
To support python3.8, we redefine all lambda expressions in triton.Heuristics as functions.


Bowen12992 (approved):
LGTM

tongxin (commented):
LG

PR #117, opened 2024-07-12
[Operator] Add flip op, and support kwargs in pointwise_dynamic

zhzhcookie:
1. Support kwargs in pointwise_dynamic.
2. Add flip op as a pointwise op.

### DIFF
--- a/benchmark/test_pointwise_perf.py
+++ b/benchmark/test_pointwise_perf.py
@@ -474,3 +474,35 @@ def test_perf_isfinite_int():
         sizes=SIZES,
     )
     bench.run()
+
+
+def test_perf_flip():
+    def flip_kwargs(dtype, batch, size):
+        return {"dims": [0, 1]}
+
+    bench = Benchmark(
+        op_name="flip",
+        torch_op=torch.flip,
+        arg_func=unary_arg,
+        dtypes=FLOAT_DTYPES,
+        batch=POINTWISE_BATCH,
+        sizes=SIZES,
+        kwargs_func=flip_kwargs,
+    )
+    bench.run()
+
+
+def test_perf_flip_int():
+    def flip_kwargs(dtype, batch, size):
+        return {"dims": [0, 1]}
+
+    bench = Benchmark(
+        op_name="flip",
+        torch_op=torch.flip,
+        arg_func=unary_int_arg,
+        dtypes=INT_DTYPES,
+        batch=POINTWISE_BATCH,
+        sizes=SIZES,
+        kwargs_func=flip_kwargs,
+    )
+    bench.run()
--- a/src/flag_gems/__init__.py
+++ b/src/flag_gems/__init__.py
@@ -95,6 +95,7 @@ def enable(lib=aten_lib):
     lib.impl("cross_entropy_loss", cross_entropy_loss, "AutogradCUDA")
     lib.impl("isclose", isclose, "CUDA")
     lib.impl("allclose", allclose, "CUDA")
+    lib.impl("flip", flip, "CUDA")
 
 
 class use_gems:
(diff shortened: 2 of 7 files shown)


iclementine:
As a principle, `pointwise_dynamic` can be used to generate pointwise operation for nd tensors with strided layout. But flip is not a pointwise operation in this sense.

So if we can find (on construct) a specific strided layout for the input and the output, we can turn flip into a view+copy, where the copy can be implemented with `pointwise_dynamic`.

Unfortunately torch tensors do not support negative strides(while numpy does support it), so there is no torch public api that can create a view for clipped input.

But for triton kernels generated by `pointwise_dynamic` with strides as inputs, negative strides are supported out-of-the-box. Torch tensors as parameters to triton kernels are merely a pointer. So if we can get the pointer to the **first** element of the flipped view.

This is actually not a big issue. For an nd tensor, for each flipped axis, we can shift the pointer to the last position at that axis. Then we get the pointer to the **first** element of the flipped view.

For example:

a is a 1d tensor 
`[1, 2, 3, 4]`
pointer to flipped a's first element is `a[3]` or more precisely(`a.data_ptr() + (a.size(0) - 1) * a.stride(0)`).

a is a 2d tensor 
```text
[0 1 2 3
 4 5 6 7
 8 9 10 11]

pointer to `flipped(a, 0)`'s first element is 
`a.data_ptr() + (a.size(0) - 1) + a.stride(0)`

pointer to `flipped(a, 1)`'s first element is 
`a.data_ptr() + (a.size(1) - 1) + a.stride(1)`

As a general case
pointer to `flipped(a, axes)`'s first element is 

`a.data_ptr() + sum(((a.size(i) - 1) + a.stride(i)) for i in axes)`

Strides for flipped view are the same strides from the input tensor, with strides for the flipped axes set to negative values.

zhzhcookie:
> As a principle, `pointwise_dynamic` can be used to generate pointwise operation for nd tensors with strided layout. But flip is not a pointwise operation in this sense.
> 
> So if we can find (on construct) a specific strided layout for the input and the output, we can turn flip into a view+copy, where the copy can be implemented with `pointwise_dynamic`.
> 
> Unfortunately torch tensors do not support negative strides(while numpy does support it), so there is no torch public api that can create a view for clipped input.
> 
> But for triton kernels generated by `pointwise_dynamic` with strides as inputs, negative strides are supported out-of-the-box. Torch tensors as parameters to triton kernels are merely a pointer. So if we can get the pointer to the **first** element of the flipped view.
> 
> This is actually not a big issue. For an nd tensor, for each flipped axis, we can shift the pointer to the last position at that axis. Then we get the pointer to the **first** element of the flipped view.
> 
> For example:
> 
> a is a 1d tensor `[1, 2, 3, 4]` pointer to flipped a's first element is `a[3]` or more precisely(`a.data_ptr() + (a.size(0) - 1) * a.stride(0)`).
> 
> a is a 2d tensor
> 
> ```
> [0 1 2 3
>  4 5 6 7
>  8 9 10 11]
> ```
> 
> pointer to `flipped(a, 0)`'s first element is `a.data_ptr() + (a.size(0) - 1) + a.stride(0)`
> 
> pointer to `flipped(a, 1)`'s first element is `a.data_ptr() + (a.size(1) - 1) + a.stride(1)`
> 
> As a general case pointer to `flipped(a, axes)`'s first element is
> 
> `a.data_ptr() + sum(((a.size(i) - 1) + a.stride(i)) for i in axes)`
> 
> Strides for flipped view are the same strides from the input tensor, with strides for the flipped axes set to negative values.
> 
> Unfortunately, torch cannot generate a flipped view with negative strides. So we may need to generate code for negative strides properly. We may extend pointwise_dynamic to be more powerful to support this later.

We will reconstruct the pointwise codegen soon.

iclementine (approved):

PR #118, opened 2024-07-15
avoid accessing non-constexpr globals from jit-ed function out of a p…

iclementine:
avoid accessing non-constexpr globals from jit-ed function out of a preference for constexpr locals.

In triton 3.0, accesing a non-constexpr global is not allowed by default, with the following error message.

```text
>       return ast_to_ttir(self.fn, self, context=context, options=options, codegen_fns=codegen_fns)
E       triton.compiler.errors.CompilationError: at 29:39:
E           r0 = uint_to_uniform_float(r0)
E           r1 = uint_to_uniform_float(r1)
E           r2 = uint_to_uniform_float(r2)
E           r3 = uint_to_uniform_float(r3)
E
E           mask0 = r0 > p
E           mask1 = r1 > p
E           mask2 = r2 > p
E           mask3 = r3 > p
E           p = 1.0 / (1.0 - p)
E
E           off_0 = tl.program_id(0) * BLOCK * UNROLL + tl.arange(0, BLOCK)
E                                              ^
E       NameError("Cannot access global variable UNROLL from within @jit'ed function. Triton kernels can only access global variables that are annotated as constexpr (`x: triton.language.constexpr = 42` or `x = triton.language.constexpr(42)`).  Alternatively, set the envvar TRITON_ALLOW_NON_CONSTEXPR_GLOBALS=1, but we do not promise to support this forever.")
```

### DIFF
--- a/src/flag_gems/ops/dropout.py
+++ b/src/flag_gems/ops/dropout.py
@@ -4,9 +4,7 @@ import torch
 import triton
 import triton.language as tl
 
-from ..utils.random_utils import philox_cuda_seed_offset, uint_to_uniform_float
-
-UNROLL = 4
+from flag_gems.utils.random_utils import philox_cuda_seed_offset, uint_to_uniform_float
 
 
 def heur_block(args):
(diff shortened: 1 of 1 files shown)


tongxin (approved):
LG

PR #120, opened 2024-07-15
Avoid calling external library in rhs of assignemnts to tl.constexpr

iclementine:
Avoid calling external library in rhs of assignemnts to tl.constexpr, as disallowed in triton3.0.0, use numeric literal instead.

See https://github.com/triton-lang/triton/issues/4321 for more details.


tongxin (approved):
LG

PR #121, opened 2024-07-16
[Benchmark] rename benchmark for integer operator tests

StrongSpoon:
rename benchmark of operators like bitwise ops, allclose as xx_int.


zhzhcookie (approved):
LGTM

PR #130, opened 2024-07-18
[codegen] Fix out_offset in pointwise_dynamic when rank is 0

zhzhcookie:
1. Fix out_offset in pointwise_dynamic when rank is 0


iclementine (approved):
LGTM
but when rank is 0, flip can just return without launching a kernel.

PR #133, opened 2024-07-23
Develop Fill[SiliconFlow]

MARD1NO:
image


src/flag_gems/ops/fill.py
```diff
@@ -0,0 +1,34 @@
...
+    offset = pid * BLOCK_SIZE + cols
+    tl.store(out_ptr + offset, value, mask=offset < N)
+
+
+def fill(self, value):
+    logging.debug("GEMS FILL")
+    out = torch.empty_like(self)
+    N = out.numel()
+
+    BLOCK_SIZE = 128
```
StrongSpoon: what about setting block_size to 512?

MARD1NO: Fixed

src/flag_gems/__init__.py
```diff
@@ -137,6 +137,7 @@ def enable(lib=aten_lib):
     # lib.impl("gather.out", gather_out, "CUDA")
     lib.impl("isclose", isclose, "CUDA")
     lib.impl("allclose", allclose, "CUDA")
+    lib.impl("fill.Scalar", fill, "CUDA")
```
Bowen12992: how about implement `fill.Tensor`as well?

Bowen12992 (commented):
The performance pic seems broken... maybe we can update it or just use a markdown table

src/flag_gems/__init__.py
MARD1NO: Done

src/flag_gems/ops/fill.py
```diff
@@ -14,11 +14,17 @@ def fill_kernel(
     N,
     value,
     BLOCK_SIZE: tl.constexpr,
+    IS_VALUE_SCALAR: tl.constexpr,
```
Bowen12992: It's better to use two funcs

Bowen12992 (dismissed):
LGTM

src/flag_gems/ops/fill.py
```diff
@@ -0,0 +1,60 @@
...
+
+import torch
+import triton
+import triton.language as tl
+
+from ..utils import libentry
+
+
+@libentry()
+@triton.jit
```
StrongSpoon: recommend to register value_scalar as do_not_specialize

MARD1NO: Fixed

StrongSpoon (approved):
lgtm

PR #134, opened 2024-07-24
[Operator] Fix rand/randn/rand_like grid and seed error

hesse-x:
We observed that the kernel of the rand class operator writes data four times the size of a BLOCK each time, but the grid issued for processing is handled based on the amount of data each kernel processes per BLOCK. Consequently, we emulated the dropout mechanism to modify the grid processing logic, and similarly, we adapted the seed calculation parameters based on the dropout approach.


tongxin (approved):
LGTM

Bowen12992 (approved):
LGTM

PR #135, opened 2024-07-24
[Benchmark] add backward mode for benchmark

AHuskie:
* [benchmark] add argument 'is_backward' to class 'Benchmark', we can set 'is_backward' true to perf backward process of some operation.

### DIFF
--- a/benchmark/performance_utils.py
+++ b/benchmark/performance_utils.py
@@ -14,9 +14,19 @@ torch.backends.cuda.matmul.allow_tf32 = False
 
 class Benchmark:
     def __init__(
-        self, op_name, torch_op, arg_func, dtypes, batch, sizes, kwargs_func=None
+        self,
+        op_name,
+        torch_op,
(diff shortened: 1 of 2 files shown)


benchmark/performance_utils.py
```diff
@@ -14,34 +14,50 @@
...
             for i in range(REPETITION):
-                op(*args, **kwargs)
+                fn()
             torch.cuda.synchronize()
             end = time.time()
             latency = (end - start) / REPETITION * 1000
         else:
             latency = triton.testing.do_bench(
-                lambda: op(*args, **kwargs),
+                fn(),
```
zhzhcookie: Maybe it should be fn ? What's the triton version in your testing environment?

AHuskie: ok. I have changed from fn() to fn. 
My trion version is 3.0.0.

zhzhcookie (approved):
LG

PR #136, opened 2024-07-25
Making libentry thread safe

tongxin:
The present multithreading support in Triton is not yet in a ready state. For example, a simple use case is throwing a bunch of Python threads to run a model on multiple devices in parallel. However, this usage can easily break Triton as its runtime is not properly serialized. As a result, FlagGems using libentry is prone to race conditions when gems are called in parallel threads without proper locking. This PR brings jit serialization into libentry to enable race free invocations of all libentry wrapped functions.


StrongSpoon (dismissed):
lgtm. recommend deleting useless 'print' annotations.

tongxin:
Done.

StrongSpoon (approved):
lgtm

PR #138, opened 2024-07-26
Exponential added.

tongxin:
Added Triton function for inplace operator tensor.exponential_. 
The output value range is (0, \infin)

### DIFF
--- a/src/flag_gems/ops/exponential_.py
+++ b/src/flag_gems/ops/exponential_.py
@@ -29,17 +29,17 @@ def heur_num_warps(args):
         "num_warps": heur_num_warps,
     }
 )
-@triton.jit(do_not_specialize=["philox_seed", "philox_offset"])
+@triton.jit(do_not_specialize=["philox_seed", "philox_offset", "N"])
 def fused_exponential_kernel(
     out_ptr,
     N,
+    is_double,
     lambd,
     eps,
     philox_seed,
     philox_offset,
     BLOCK: tl.constexpr,
 ):
-    UNROLL: tl.constexpr = 4
     philox_seed = philox_seed.to(tl.int64)
     philox_offset = philox_offset.to(tl.int64)
     c0 = (philox_offset & 0xFFFFFFFF).to(tl.uint32)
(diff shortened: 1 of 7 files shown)


src/flag_gems/ops/exponential_.py
```diff
@@ -0,0 +1,99 @@
...
+    is_subnormal = u > 1.0 + eps1
+    log = tl.where(is_subnormal, eps1, tl.math.log(u))
+    v = -1.0 / lambd * log
+    return v
+
+
+UNROLL = 4
+
+
+def exponential_(x, lambd: float = 1.0, gen=None):
```
StrongSpoon: the interface in aten is '_exponential_(Tensor(a!) self, float lambd=1, *, Generator? generator=None)'. suggest adding asterisk between lambd and generator.

tests/test_special_ops.py
```diff
@@ -222,3 +222,12 @@ def test_accuracy_rand_like(shape, dtype):
...
     assert (res_out >= 0.0).all()
+
+
+@pytest.mark.parametrize("shape", POINTWISE_SHAPES)
+@pytest.mark.parametrize("dtype", FLOAT_DTYPES)
+def test_accuracy_exponential_(shape, dtype):
+    x = torch.randn(size=shape, dtype=dtype, device="cuda")
+    with flag_gems.use_gems():
+        res_out = x.exponential_(lambd=0.5)
+    assert res_out.min() > 0
```
StrongSpoon: is it enough to ensure the accuracy? I think we need more checks.

iclementine: Maybe add a K-S test for the distribution?
Testing for positiveness is too minimal
https://en.wikipedia.org/wiki/Kolmogorov%E2%80%93Smirnov_test

src/flag_gems/ops/exponential_.py
```diff
@@ -0,0 +1,99 @@
...
+    tl.store(out_ptr + off_3, r3, mask=off_3 < N, eviction_policy="evict_first")
+
+
+@triton.jit
+def transform_exponential(u, lambd, eps):
+    eps1 = -0.5 * eps
+    is_subnormal = u > 1.0 + eps1
+    log = tl.where(is_subnormal, eps1, tl.math.log(u))
+    v = -1.0 / lambd * log
+    return v
```
iclementine, line 96: The inverse cdf  of exponential distribution is

`x = - log(1-y) / lambda`

y in [0, 1].

First sample y from a Uniform distribution U(0,1) then apply the inverse cdf.

Does this implementation exploits the feature that `1-y` is also a sample from `U(0,1)` to eliminate a subtraction? Then change the formula to 

`x = - log(y) / lambda`?

StrongSpoon: I got it. Uniform is of the same distribution as 1-u.

StrongSpoon: but what is eps used for? and what't the expected data type of is_subnormal?

StrongSpoon (commented):
benchmark to do

tests/test_special_ops.py
tongxin: Thanks and more substantial tests will be on the way.

src/flag_gems/ops/exponential_.py
tongxin: Sure.

tests/test_special_ops.py
tongxin: More tests will be done.

src/flag_gems/ops/exponential_.py
iclementine: For an inplace operation, I think it conventional to add a check that the input has no internal overlapping. Though it would add some overhead.

src/flag_gems/ops/exponential_.py
```diff
@@ -0,0 +1,121 @@
...
+    hi = hi.to(tl.uint64) << 32
+    x = hi | lo.to(tl.uint64)
+    return x
+
+
+@triton.jit
+def transform_exponential(u, lambd, eps):
+    eps1 = -0.5 * eps
+    is_min = u >= 1.0 + eps1
+    log = tl.where(is_min, eps1, tl.math.log(u))
```
iclementine, line 94: Is this really needed? What about just using `log(u)` or `log(1-u)`?

tongxin: This is for enforcing compatibility with Pytorch..

Bowen12992 (approved):
LGTM, added scipy to UT ENV

src/flag_gems/ops/exponential_.py
```diff
@@ -0,0 +1,121 @@
...
+    v = -1.0 / lambd * log
+    return v
+
+
+def exponential_(x, lambd: float = 1.0, *, gen=None):
+    logging.debug("GEMS EXPONENTIAL_")
+    dtype = x.dtype
+    device = x.device
+    inplace = x.is_contiguous()
```
iclementine, line 103: Performing inplace operation on a tensor with internal overlapping should raise a Runtime Exception.

iclementine:
```text
RuntimeError: unsupported operation: more than one element of the written-to tensor refers to a single memory location. Please clone() the tensor before performing the operation.
```

iclementine: Now it would raise a runtime error when copying data back.

```python
import torch
import flag_gems
flag_gems.enable()
x = torch.ones(2, device="cuda")
x = torch.broadcast_to(x, (3, 2))
x.exponential_()
```

iclementine (approved):

src/flag_gems/ops/exponential_.py
tongxin: I'll fix it.

tongxin: Pytorch throws with exactly the same error. We'll just keep the current way.

PR #140, opened 2024-07-29
Develop Pad[SiliconFlow]

MARD1NO:
1. 基于codegen 实现的 constant pad
2. ncu结果显示 triton pad性能是更优的，而且torch其实是拆分成两个kernel，一个fill一个copy：
1024 * 1024： 
triton=4.54us
torch=2.82+5.57us
image
但是 flaggems benchmark显示triton实现特别慢：

image


src/flag_gems/ops/pad.py
```diff
@@ -0,0 +1,455 @@
...
+        code.writeline("assert pad_size % 2 == 0")
+        code.newline()
+        code.writeline("pad_before = [0 for _ in range(ndim)]")
+        code.writeline("pad_after = [0 for _ in range(ndim)]")
+        code.newline()
+        code.writeline("pad_pair = pad_size // 2 ")
+        code.writeline("for i in range(pad_pair): ")
+        with code.indent():
+            code.writeline("pad_before[ndim - i - 1] = pad[2 * i]")
+            code.writeline("pad_after[ndim - i - 1] = pad[2 * i + 1]")
```
iclementine, line 99: Maybe just 

```python
pad_before = pad[::2]
pad_after = pad[1::2]
```

src/flag_gems/ops/pad.py
```diff
@@ -0,0 +1,455 @@
...
+            stride_args = ", ".join(f"valid_dim{j}_end: int" for j in range(rank))
+            code.writeline(f"{stride_args}, # valid dim end")
+
+            code.writeline("in_elem_cnt: tl.constexpr, ")
+            code.writeline("out_elem_cnt: tl.constexpr, ")
+            code.writeline("value, # padding value")
+            code.writeline("IS_CONSTANT, ")
+            code.writeline("IS_REFLECT, ")
+            code.writeline("IS_REPLICATE, ")
+            code.writeline("IS_CIRCULAR, ")
```
iclementine: Shoud these 4 indicator variables for padding mode be `constexpr`?

src/flag_gems/ops/pad.py
```diff
@@ -0,0 +1,455 @@
...
+        code.writeline("block_offset = pid * BLOCK_SIZE")
+        code.writeline("offset = block_offset + tl.arange(0, BLOCK_SIZE)")
+        code.newline()
+
+        code.writeline("remaining = offset ")
+        for i in range(rank):
+            code.writeline(f"idx = remaining // out_strides{i}")
+            code.writeline(f"dst_index_{i} = idx")
+            code.writeline(f"remaining = remaining - idx * out_strides{i}")
+            code.newline()
```
iclementine, line 284: This part to calculate the multi-index on tensor out is not correct.

since `src_index`s are computed from `dst_index`s, where these indices are logical multi-indx for a multi-dimensional tensor. We can also see that when computing memory offsets on tensor in with following code:

```python
src_offset = src_index_0 * in_strides0
src_offset += src_index_1 * in_strides1
```

Since we are unraveling `offset` to get the logical multi-index, it is the tensor shape that should be considered(in divide and modular operations).

A simple test case where the input is not C-contiguous can verify this.

```python
@pytest.mark.parametrize("shape", [(100, 100)])
def test_constant_pad(shape, dtype):
    x = torch.randn(size=shape, dtype=dtype, device="cuda")[::2, ::2]
    ref_x = to_reference(x)

    rank = x.ndim
    pad_params = tuple(np.random.randint(0, 10, rank * 2))

    pad_value = np.random.randint(0, 1024, 1)[0]
    ref_out = torch.nn.functional.pad(x, pad_params, "constant", pad_value)
    with flag_gems.use_gems():
        res_out = torch.nn.functional.pad(ref_x, pad_params, "constant", pad_value)

    gems_assert_equal(ref_out, res_out)
```

tests/test_special_ops.py
```diff
@@ -222,3 +225,41 @@ def test_accuracy_rand_like(shape, dtype):
...
+    with flag_gems.use_gems():
+        x.exponential_(lambd=lambd)
+    expo_cdf = lambda x: np.where(x < 0, 0, 1.0 - np.exp(-lambd * x))
+    pvalue = scipy.stats.kstest(x.cpu().numpy().flatten(), expo_cdf).pvalue
+    assert pvalue > 0.05
+
+
+@pytest.mark.parametrize("shape", POINTWISE_SHAPES)
+@pytest.mark.parametrize("dtype", FLOAT_DTYPES)
+def test_constant_pad(shape, dtype):
```
iclementine: Add more tests since other padding mode is supported?

iclementine: Also, add some cases where the input is not C-contiguous.

src/flag_gems/ops/pad.py
MARD1NO: good catch, fixed it by change `load_cond` to check index is valid for input shape instead of offset

tests/test_special_ops.py
MARD1NO: done

src/flag_gems/ops/pad.py
MARD1NO: yes, add tl.constexpr

src/flag_gems/ops/pad.py
```diff
@@ -0,0 +1,458 @@
...
+            )
+
+        code.writeline(
+            "if_pad = tl.where(cond, if_pad_false_mask, if_pad_true_mask).to(tl.int1)"
+        )
+
+        for i in range(rank):
+            code.writeline(f"src_index_{i} = dst_index_{i} - valid_dim{i}_start ")
+
+        for i in range(rank):
```
iclementine, line 308: Would there be cases where `src_index < 0`?

src/flag_gems/ops/pad.py
```diff
@@ -0,0 +1,458 @@
...
+            code.writeline(
+                f"src_index_{i} = tl.where(src_index_{i} < 0, 0, src_index_{i})"
+            )
```
iclementine, line 308: This would generate code like

```python
    if_pad_false_mask = tl.zeros((BLOCK_SIZE, ), dtype=tl.int32)
    if_pad_true_mask = tl.full((BLOCK_SIZE, ), 1, dtype=tl.int32)
    src_index_0 = tl.zeros((BLOCK_SIZE, ), dtype=tl.int32)
    src_index_1 = tl.zeros((BLOCK_SIZE, ), dtype=tl.int32)
    cond = (dst_index_0 >= valid_dim0_start and dst_index_0 < valid_dim0_end)
    cond &= (dst_index_1 >= valid_dim1_start and dst_index_1 < valid_dim1_end)
    if_pad = tl.where(cond, if_pad_false_mask, if_pad_true_mask).to(tl.int1)
    src_index_0 = dst_index_0 - valid_dim0_start
    src_index_1 = dst_index_1 - valid_dim1_start
    src_index_0 = tl.where(src_index_0 < 0, 0, src_index_0)
    src_index_1 = tl.where(src_index_1 < 0, 0, src_index_1)
```

if seems 

```python
    src_index_0 = tl.zeros((BLOCK_SIZE, ), dtype=tl.int32)
    src_index_1 = tl.zeros((BLOCK_SIZE, ), dtype=tl.int32)
```
is redundant.

src/flag_gems/ops/pad.py
```diff
@@ -0,0 +1,458 @@
...
+                    f"""src_index_{i} = tl.where(dst_index_{i} < valid_dim{i}_start,
+                        valid_dim{i}_start - dst_index_{i}, src_index_{i})"""
+                )
+            for i in range(rank):
+                code.writeline(
+                    f"""src_index_{i} = tl.where(dst_index_{i} >= valid_dim{i}_end,
+                    (x_shape{i} + valid_dim{i}_start - 1) * 2 - dst_index_{i} - valid_dim{i}_start, src_index_{i})"""
+                )
+
+        code.newline()
```
iclementine, line 324: There is a problem when padding size is larger than or equals to the input size for REFLECT padding mode. Or it raises

```text
RuntimeError: Argument #4: Padding size should be less than the corresponding input dimension, but got: padding (5, 6) at dimension 1 of input [4, 6]
```

Try the following snippet to see the Error.

```python
x = torch.arange(24).reshape(4, 6).cuda()
torch.nn.functional.pad(x, (5, 6), mode="reflect")
```



This should be raised before launching the kernel.

src/flag_gems/ops/pad.py
```diff
@@ -0,0 +1,458 @@
...
+            for i in range(rank):
+                code.writeline(
+                    f"""src_index_{i} = tl.where(dst_index_{i} < valid_dim{i}_start,
+                        dst_index_{i} + x_shape{i} - valid_dim{i}_start, src_index_{i})"""
+                )
+            for i in range(rank):
+                code.writeline(
+                    f"""src_index_{i} = tl.where(dst_index_{i} >= valid_dim{i}_end,
+                        dst_index_{i} - valid_dim{i}_end, src_index_{i})"""
+                )
```
iclementine, line 348: Also, torch does not allow padding size(either before or after) to be larger that the input size.

Or it raises `RuntimeError: Padding value causes wrapping around more than once.`

Try the following snippet to see the error.

```python
x = torch.arange(24).reshape(4, 6).cuda()
torch.nn.functional.pad(x, (6, 7), mode="circular")
```

iclementine (changes requested):
1. reflect padding requires `padding size <= input size - 1`
2. replicate padding requires that `padding size <= input size`.

Bowen12992 (commented):
Please fix the new conflict and update the code to make all the checks work

src/flag_gems/ops/pad.py
MARD1NO: good catch, fixed.

src/flag_gems/ops/pad.py
MARD1NO: Add torch style check info

src/flag_gems/ops/pad.py
MARD1NO: Fixed

iclementine (approved):
LGTM

PR #144, opened 2024-07-31
[Operator] masked_fill

GwokHiujin:
We have completed the development of the masked_fill operator. Specifically: 

- The corresponding aTen operator is `masked_fill`
- Added accuracy test and perf test

### DIFF
--- a/src/flag_gems/ops/masked_fill.py
+++ b/src/flag_gems/ops/masked_fill.py
@@ -1,0 +1,64 @@
+import logging
+
+import torch
+import triton
+import triton.language as tl
+
+from ..utils import broadcastable_to, libentry
+
+
+def cfggen():
+    block_m = [1, 2, 4, 8]
+    configs = [
+        triton.Config({"BLOCK_M": m, "BLOCK_N": 128}, num_warps=4) for m in block_m
+    ]
+    return configs
+
+
+@libentry()
+@triton.autotune(configs=cfggen(), key=["M", "N"])
+@triton.jit
+def masked_fill_kernel(
+    inp, expand_mask, value, out, M, N, BLOCK_M: tl.constexpr, BLOCK_N: tl.constexpr
+):
+    pid = tl.program_id(0)
+    rows_offset = pid * BLOCK_M + tl.arange(0, BLOCK_M)[:, None]
+    rows_mask = rows_offset < M
+
+    for off in range(0, N, BLOCK_N):
+        cols_offset = off + tl.arange(0, BLOCK_N)[None, :]
+        cols_mask = cols_offset < N
+        block_mask = rows_mask and cols_mask
+
+        offsets = rows_offset * N + cols_offset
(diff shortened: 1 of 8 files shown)


tests/test_reduction_ops.py
```diff
@@ -670,3 +670,19 @@ def test_accuracy_vectornorm(shape, ord, dim, keepdim, dtype):
         res_out = torch.linalg.vector_norm(inp, ord, dim, keepdim)
 
     gems_assert_close(res_out, ref_out, dtype)
+
+
+@pytest.mark.parametrize("shape", REDUCTION_SHAPES)
```
iclementine: It seem that test for masked_fill should go into pointwise operation.

src/flag_gems/ops/masked_fill.py
```diff
@@ -0,0 +1,64 @@
...
+@triton.autotune(configs=cfggen(), key=["M", "N"])
+@triton.jit
+def masked_fill_kernel(
+    inp, expand_mask, value, out, M, N, BLOCK_M: tl.constexpr, BLOCK_N: tl.constexpr
+):
+    pid = tl.program_id(0)
+    rows_offset = pid * BLOCK_M + tl.arange(0, BLOCK_M)[:, None]
+    rows_mask = rows_offset < M
+
+    for off in range(0, N, BLOCK_N):
```
iclementine: Why loop over dimension `N` if there is no actual dependencies between iterations?

src/flag_gems/ops/masked_fill.py
```diff
@@ -0,0 +1,64 @@
...
+        cols_offset = off + tl.arange(0, BLOCK_N)[None, :]
+        cols_mask = cols_offset < N
+        block_mask = rows_mask and cols_mask
+
+        offsets = rows_offset * N + cols_offset
```
iclementine: This assumes c-contiguous input and 
```
tl.load(expand_mask + offsets...
tl.load(expand_mask + offsets...
tl.load(inp + offsets...
```
If they are not contiguous, the result would be incorrect.

iclementine (changes requested):
I suggest copy the tensor contiguous before invoking the kernel to ensure that the result in correct.

src/flag_gems/ops/masked_fill.py
GwokHiujin: I understand your point after reading the comments in PR#145. I'll try using a 2D grid. Good advice :)

src/flag_gems/ops/masked_fill.py
GwokHiujin: Thanks, it definitely needs thinking about. Already updated the relevant code.

tests/test_reduction_ops.py
GwokHiujin: Already moved the relevant tests into the pointwise category.

iclementine (approved):

PR #145, opened 2024-07-31
[Operator] index_add

GwokHiujin:
We have completed the development of the index_add operator. Specifically:

- The corresponding aTen operator is `index_add`
- Added accuracy test and perf test


src/flag_gems/ops/index_add.py
```diff
@@ -0,0 +1,83 @@
...
+    for off in range(0, N, BLOCK_N):
+        cols_offsets = off + tl.arange(0, BLOCK_N)
+        index_mask = cols_offsets < N
+        block_mask = rows_mask and index_mask
+
+        cur_indices = tl.load(index + cols_offsets, mask=index_mask, other=0)
+        inp_off = rows_offsets * inp_len + cur_indices[None, :]
+        cur_inp = tl.load(inp + inp_off, mask=block_mask, other=0.0).to(tl.float32)
+        src_off = rows_offsets * N + cols_offsets[None, :]
+        cur_src = tl.load(src + src_off, mask=block_mask, other=0.0).to(tl.float32)
```
tongxin: Possibly lose precision for fp64 src and inputs?

src/flag_gems/ops/index_add.py
```diff
@@ -0,0 +1,83 @@
...
+
+    dim = dim % inp.ndim
+    inp_len = inp.size(dim)
+    N = index.numel()
+    M = src.numel() // N
+    inp = dim_compress(inp, dim)
+    src = dim_compress(src, dim)
+    out = inp.clone()
+
+    grid = lambda meta: (triton.cdiv(M, meta["BLOCK_M"]),)
```
iclementine, line 76: Input is permuted into shape`(..., M)` where `product(...) == inp_len`, src is permuted into shape `(..., M)` where `product(...) == N`.

The task is to scatter `N` subtensors (src, viewd as N subtensors) into inp.
So the grid is along the M dimension instead of `N` dimension?

iclementine: The input & src  is permuted into shapes
input: Shape(M, ...) where product(...) == inp_len
src: Shape(M, ...) where product(...) == N
and contiguous.

So we can view then as 
input: Shape(M, inp_len)
src: Shape(M, N)
index: (N, )

Then the task is partitioned along the `M` dimension in tile size of `BLOCK_M`, while the `N` dimension is looped in tiles of size `BLOCK_N`.

src/flag_gems/ops/index_add.py
```diff
@@ -0,0 +1,83 @@
...
+    pid = tl.program_id(0)
+    rows_offsets = pid * BLOCK_M + tl.arange(0, BLOCK_M)[:, None]
+    rows_mask = rows_offsets < M
+
+    for off in range(0, N, BLOCK_N):
+        cols_offsets = off + tl.arange(0, BLOCK_N)
+        index_mask = cols_offsets < N
+        block_mask = rows_mask and index_mask
+
+        cur_indices = tl.load(index + cols_offsets, mask=index_mask, other=0)
```
iclementine: This would requires index to be contiguous to be correct.

src/flag_gems/ops/index_add.py
```diff
@@ -0,0 +1,83 @@
...
+    alpha,
+    inp_len,
+    BLOCK_M: tl.constexpr,
+    BLOCK_N: tl.constexpr,
+):
+    pid = tl.program_id(0)
+    rows_offsets = pid * BLOCK_M + tl.arange(0, BLOCK_M)[:, None]
+    rows_mask = rows_offsets < M
+
+    for off in range(0, N, BLOCK_N):
```
iclementine: Maybe a 2d grid can be used, since there is no dependencies between iterations in the loop.
If `M` is small and `N` is large, this task partition scheme may not saturate a gpu's SMs.

src/flag_gems/ops/index_add.py
iclementine: What about just keep src and inp as-is without casting?

src/flag_gems/ops/index_add.py
iclementine: Though it is hard to figure out a general solution now, but permuting the tensor to make the inp_len & N dimensional to be contiguous is not always good.

For example,

input & src are both 2d tensors, now index_add along axis 0, then the permutations are actually not needed to make index_add easier.

iclementine (approved):
There are some spaces for optimization, but LGTM.

iclementine (changes requested):
Some suggestions
1. Ensure index is contiguous, or consider its stride;
2. keep the data loaded from src as-is to avoid down-cast;
3. (Maybe) use some heuristics to make a better task partitioning & avoid unnecessary data permutations.

src/flag_gems/ops/index_add.py
GwokHiujin: > Possibly lose precision for fp64 src and inputs?

I've encountered precision loss issues in some data types (like bf16 and float32). Ignoring casting might lead to problems. I'll implement the suggested changes below and see if they resolve the issue.

src/flag_gems/ops/index_add.py
GwokHiujin: > Maybe a 2d grid can be used, since there is no dependencies between iterations in the loop. If `M` is small and `N` is large, this task partition scheme may not saturate a gpu's SMs.

! I get what you mean now; that does make sense. I'll solve this part.

src/flag_gems/ops/index_add.py
GwokHiujin: Yes, this is a key issue I constantly consider(Since it actually occurs in other operations, too). As a temporary solution, I set conditional judgments, such as: if the input dimension equals (self.ndim - 1), I don't perform the permutation. I'm uncertain if this approach is effective. 

BTW Performance testing revealed that permutations can increase latency by about 7 times compared to Torch, making the reduction of unnecessary permutations crucial... ; (

iclementine (approved):

iclementine (approved):
LGTM

PR #146, opened 2024-08-01
Use int64 indexing when needed & fix argmax

iclementine:
1. fix amax, armax and triu, use int64 indexing when the largest tens…or's size_in_bytes exceed int32's max;
2. change the tiling scheme for argmax to loop in the reduction dimension, instead of data-size-dependent-tile-size


tongxin (dismissed):
LG

zhzhcookie (approved):
LGTM after def can_use_int32_index

PR #149, opened 2024-08-02
[Operator] Add unique op

zhzhcookie:
1. Add unique op
2. Add tests and benchmark


tests/test_special_ops.py
```diff
@@ -245,3 +248,91 @@ def test_accuracy_exponential_pvalue(shape, dtype, lambd):
...
+
+
+@pytest.mark.parametrize(
+    "shape",
+    POINTWISE_SHAPES
+    + [
+        (8191,),
+        (1280, 1280 * 21),
+        (8192, 65536 + 8192 + 11),
+    ],
```
StrongSpoon: suggest adding these shapes into POINTWISE_SHAPES, thus corresponding tests could perform on all operators

tests/test_special_ops.py
```diff
@@ -245,3 +248,91 @@ def test_accuracy_exponential_pvalue(shape, dtype, lambd):
...
+)
+@pytest.mark.parametrize("dtype", INT_DTYPES)
+@pytest.mark.parametrize("sorted", [True])
+@pytest.mark.parametrize("return_inverse", [True, False])
+@pytest.mark.parametrize("return_counts", [False, True])
+def test_accuracy_unique(shape, dtype, sorted, return_inverse, return_counts):
```
StrongSpoon, line 265: dim not supported?

src/flag_gems/ops/unique.py
```diff
@@ -0,0 +1,846 @@
...
+        )
+    else:
+        sorted_data, _ = torch.sort(in0.ravel(), stable=False)
+        data_out, inverse_indices, counts = sorted_quick_unique_flat(
+            sorted_data, return_counts
+        )
+    return data_out, inverse_indices, counts
+
+
+def unique(
```
tongxin: The aten function to implement is _unique2, shall we keep this name?

benchmark/test_special_perf.py
```diff
@@ -84,3 +87,67 @@ def embedding_kwargs(dtype, batch, size):
...
+    bench = Benchmark(
+        op_name="unique",
+        torch_op=torch.unique,
+        arg_func=unary_int_arg,
+        dtypes=INT_DTYPES,
+        batch=POINTWISE_BATCH,
+        sizes=WIDE_RANGE_SIZES,
+        kwargs_func=unique_kwargs,
+    )
+    bench.run()
```
StrongSpoon: in my view, testing the most simple case in benchmark is enough.

src/flag_gems/ops/unique.py
tongxin: Also, I don't see why we need a redirected call to unique_flat. Their function prototypes are literally the same.

StrongSpoon (commented):
does isin need other unique functions?

src/flag_gems/ops/unique.py
```diff
@@ -0,0 +1,846 @@
...
+    a = tl.load(sorted_data_ptr + i0, mask=mask)
+    i0_prev = tl.where(i0 > 0, i0 - 1, 0)
+    b = tl.load(sorted_data_ptr + i0_prev, mask=mask)
+
+    # ne & cumsum
+    ne_result = tl.where(i0 > 0, a != b, 0)
+    cumsum = tl.cumsum(ne_result)
+
+    # unique_size
+    unique_size = tl.sum(ne_result)
```
tongxin: This would involve an extra reduction.

src/flag_gems/ops/unique.py
```diff
@@ -0,0 +1,846 @@
...
+    tl.store(unique_size_ptr, unique_size)
```
tongxin: Simply storing the last cumsum may do the work.

```suggestion
    tl.store(unique_size_ptr + i0*0, cumsum, mask=i0 == (tile_size-1))
```

src/flag_gems/ops/unique.py
```diff
@@ -0,0 +1,846 @@
...
+            inverse_indices,
+            idx,
+            unique_size,  # out
+            return_inverse,
+            return_counts,
+            num_tasks,
+            tile_size=triton.next_power_of_2(num_tasks),
+            num_warps=8,
+        )
+    out_size = unique_size.item() + 1
```
tongxin, line 794: Pulling data back to CPU could be avoided by passing the result tensor as parameter to a tensor constructor.

src/flag_gems/ops/unique.py
```diff
@@ -0,0 +1,846 @@
...
+    counts = None
+    if return_counts:
+        idx = idx[:out_size]
```
tongxin, line 797:
```suggestion
        idx = idx.resize_(unique_size + 1)
```

tests/test_special_ops.py
zhzhcookie: > dim not supported?

Currently replaces _unique2 native_function, but not unique_dim yet.

src/flag_gems/ops/unique.py
```diff
@@ -0,0 +1,846 @@
...
+        counts = torch.empty_like(idx)
```
tongxin, line 798: what about allocating `counts` in advance and then do a resize_ to reduce memory allocation overhead?

benchmark/test_special_perf.py
zhzhcookie: > in my view, testing the most simple case in benchmark is enough.

I will change it to common SIZES.

tests/test_special_ops.py
zhzhcookie: > suggest adding these shapes into POINTWISE_SHAPES, thus corresponding tests could perform on all operators

(8192, 65536 + 8192 + 11) is too large for other op, but is necessary for unique implement...

zhzhcookie:
> does isin need other unique functions?

No, isin needs _unique2(return_inverse=True / False, return_counts=False) only.

src/flag_gems/ops/unique.py
zhzhcookie: > Also, I don't see why we need a redirected call to unique_flat. Their function prototypes are literally the same.

Because isin depends on unique_flat, I initially split it into two functions. But now that view_as is not expensive, I will merge it into one function.

src/flag_gems/ops/unique.py
zhzhcookie: > This would involve an extra reduction.

Because I could'n found a more efficient method to get it. For example, unique_size = tl.sum(tl.where(r == tile_size - 1, cumsum, 0))

src/flag_gems/ops/unique.py
zhzhcookie: > Simply storing the last cumsum may do the work.

Oh I see

src/flag_gems/ops/unique.py
zhzhcookie: > 
I implement this way, but the performance decrease ...

src/flag_gems/ops/unique.py
tongxin: I see.

src/flag_gems/ops/unique.py
zhzhcookie: > what about allocating `counts` in advance and then do a resize_ to reduce memory allocation overhead?

I don't know how to? Now the size of counts is already out_size when allocating.

StrongSpoon:
I guess there might be a problem when merging from master.

benchmark/test_special_perf.py
```diff
@@ -73,3 +80,35 @@ def resolve_conj_arg(dtype, batch, size):
         sizes=SIZES,
     )
     bench.run()
+
+
+def test_perf_unique_return_inverse():
```
StrongSpoon: suggest adding performance test of unique with options as default value

src/flag_gems/ops/unique.py
```diff
@@ -0,0 +1,838 @@
...
+    data_out_ptr: tl.tensor,
+    inverse_indices_ptr: tl.tensor,
+    idx_ptr: tl.tensor,
+    unique_size_ptr: tl.tensor,  # out
+    return_inverse: tl.constexpr,
+    return_counts: tl.constexpr,
+    num_tasks: int,
+    tile_size: tl.constexpr,
+):
+    i0 = tl.arange(0, tile_size)
```
StrongSpoon, line 22: suggesting renaming i0 as offset

src/flag_gems/ops/unique.py
```diff
@@ -0,0 +1,838 @@
...
+    i0_prev = tl.where(i0 > 0, i0 - 1, 0)
+    b = tl.load(sorted_data_ptr + i0_prev, mask=mask)
+
+    # ne & cumsum
+    ne_result = tl.where(i0 > 0, a != b, 0)
+    cumsum = tl.cumsum(ne_result)
+
+    # unique_size
+    unique_size_mask = i0 == tile_size - 1
+    tl.store(unique_size_ptr + tl.zeros_like(i0), cumsum, mask=unique_size_mask)
```
StrongSpoon, line 36: Can Triton ensure to store data one by one? I'm concerned that the final unique_size might not be the last element of cumsum tensor.

StrongSpoon (commented):
review ongoing

src/flag_gems/ops/unique.py
zhzhcookie: > Can Triton ensure to store data one by one? I'm concerned that the final unique_size might not be the last element of cumsum tensor.

There is a mask to ensure it.

src/flag_gems/ops/unique.py
zhzhcookie: > suggesting renaming i0 as offset

The var name "i0" is consistent with the naming in pointwise_dynamic, and offset which is global_pid * tile_size has  another meaning...

benchmark/test_special_perf.py
zhzhcookie: > suggest adding performance test of unique with options as default value

OK, I just added it.

StrongSpoon (approved):
lgtm

PR #151, opened 2024-08-02
[Test] Test for op

Bowen12992:
Do test according to specific ops.
usage：
```
python3.11 test_specific_ops.py --name add      # for a specific op "add"
python3.11 test_specific_ops.py --all           # for all ops in the op list
```


zhzhcookie (approved):
LGTM

PR #155, opened 2024-08-07
[Test] Add bfloat16 test for  rand, randn & uniform

Bowen12992:
In k-s test, scipy do not support bfloat16, Since now we have seperate k-s test to another file, bfloat16 should be tested in  rand, randn & uniform.


StrongSpoon (approved):
lg

PR #156, opened 2024-08-07
[Test] Add more tests in named op test

Bowen12992:
Add more tests in named op test for perf


tongxin (approved):
This update per say is OK, but its worthwhile to look into how to structure it more flexible.

Bowen12992:
> This update per say is OK, but its worthwhile to look into how to structure it more flexible.

OK

PR #162, opened 2024-08-12
[Operator] dropout accuracy check

StrongSpoon:
1. Assert accuracy of dropout when reference in on cpu without comparing with reference output
2. Decorating cellos with libentry by the way


Bowen12992 (approved):
LGTM

PR #167, opened 2024-08-16
Refactor code generation for pointwise operation & PointwiseDynamicFunction

iclementine:
1. Now we generate code with nd tiles and 1d grid with grid-stride-loop, where the n is the ndim of the task space;
2. Add some simple logic to simplify task space(when all operand have the same shape and same stride, and all of them are non overlapping and dense, we simplify the task space into a 1d space, although we can use better policy but we leave it for future work);
3. Use a smarter policy for output layout inference:(the output will follow the stride order of the first tensor that has the same shape as the broadcasted shape, pre-defined ouputs has higher priority than all input tensors; otherwise, the output in c-contiguous);
4. make tile size and grid size in generated code configurable;
5. work around the problem that save to block pointer does not automatically cast the value to the pointer's dtype;
6. work around the problem that values loaded from a pointer to bool is int8 and block pointer from pointer to bool has dtype int8;
7. fix the bitwise-* operators without those work arounds, and add test-cases with bool inputs & outputs for them;
8. add TypedPtr and StridedBuffer as drop-in replament for torch.Tensor to be used in generated triton kernel & wrappers, which allows some unsafe reinterpretation of Tensors(dtype, shape, stride, offset), which cannot be done by torch APIs;
9. fix a bug in flip op where the flipped view(shifted data pointer and negative strides) from input in applied to the output.


src/flag_gems/utils/pointwise_dynamic.py
```diff
@@ -102,13 +123,25 @@ def __init__(
             self._num_outputs = num_outputs
             _check_sized_list(promotion_methods, num_outputs)
         else:
+            self._num_outputs = 1
```
sethbrin: This line is meaningless.

src/flag_gems/utils/pointwise_dynamic.py
```diff
@@ -132,627 +168,536 @@ def num_output_tensors(self) -> int:
...
-        if op_desc._is_tensor[i]:
-            parameters.append(f"in{input_tensor_index}")
-            input_tensor_index += 1
+    def gen_decorators(self, code):
+        code.writeline("@libentry()")
+        num_non_tensor_args = self.fx.num_non_tensor_args()
+        if num_non_tensor_args > 0:
+            # we do not specialize non tensor args since they are passed into the inlined function
+            # which means that their values may not deserve specialization
+            non_specialize_arg_names = [f"val{i}" for i in range(num_non_tensor_args)]
```
sethbrin, line 264: Why don't all non-tensor operations specialize, while Triton officially defaults to specializing value equal to 1, which can be beneficial for continuity analysis, especially for stride value. 

Similarly, on our hardware, if the stride of the lowest dimension is not 1 (a runtime variable), performance might degrade to single-element loads. Hardware has certain continuity requirements for the lowest dimension. 

In most of cases, stride of the lowest dimension is eqaual to 1, so can we keep `equal_to_1` specialization?

src/flag_gems/utils/pointwise_dynamic.py
```diff
@@ -132,627 +168,536 @@ def num_output_tensors(self) -> int:
...
-    with code.indent():
-        # docstring
-        wrapper_docstring = docstring_for_functional_wrapper(op_desc)
-        code.writeline(wrapper_docstring)
-
-        shapes_str = ", ".join(
-            f"in{i}.shape" for i in range(op_desc.num_input_tensors())
-        )
-        code.writeline(f"shape = broadcast_shapes([{shapes_str}])")
+            # signature: output ptrs
```
sethbrin, line 302: Would it be more intuitive to place the output at the beginning?

tests/test_pointwise_dynamic.py
```diff
@@ -0,0 +1,428 @@
...
+
+
+def test_dynamic_function_without_non_tensor_args():
+    @pointwise_dynamic(num_inputs=2, promotion_methods=[(0, 1, "DEFAULT")])
+    @triton.jit
+    def add(x, y):
+        return x + y
+
+    SIZE = 2
+    for ndim in range(10):
```
sethbrin: Maybe at most 8 rank is enough?

src/flag_gems/utils/pointwise_dynamic.py
```diff
@@ -761,64 +706,202 @@ class PointwiseDynamicFunction:
...
+            return self.overloads[ndim]
+
+        key = str(ndim)
+        code = IndentedBuffer()
+
+        module_gen = ModuleGenerator(
+            self.fx,
+            self._scalar_fn,
+            ndim,
+            "_kernel",
```
sethbrin: Can the name of triton jit kernel and wrapper include `scalar_fn.__name__`? This would enable the identification of the specific operator during the profiling of the entire network, as the device kernel names would clearly indicate which operation is being performed.

src/flag_gems/utils/pointwise_dynamic.py
```diff
@@ -132,627 +168,536 @@ def num_output_tensors(self) -> int:
...
+
+            # signature: strides, for each tensor arguments
+            ndim = self.ndim
+            if ndim > 0:
+                # strides for inputs
+                for i in range(schema.num_input_tensors()):
+                    stride_args = _cs(f"in{i}_stride{j}: int" for j in range(ndim))
+                    code.writeline(f"{stride_args}, # strides for in{i}")
+                    stride_order_args = _cs(
+                        f"in{i}_stride_order{j}: tl.constexpr" for j in range(ndim)
```
sethbrin: If `input transpose` is not currently supported, `stride_order` can just be set in load/store blockptr, just 

```
bptr = tl.make_blockptr(inptr, ... order=(rank-1, rank-2, .., 0))
```

src/flag_gems/utils/pointwise_dynamic.py
```diff
@@ -132,627 +168,536 @@ def num_output_tensors(self) -> int:
...
-                    code.writeline(f"val{non_tensor_index},")
-                function_ns.create_name(f"val{non_tensor_index}")
-                non_tensor_index += 1
+                    params.append(f"{self.input_name(i)}")
+        # NOTE: [the wrapper's signature and rules for passing parameters ]
+        # input params: must be passed by position, since the names are renamed to
+        # in0, in1, val0, val1, ..., So passing these parameters by keyword is wierd
+        # So we enforce that these parameters must be passed by position.
+        # maybe we can fix it later
+        # output parameters: must be passed by keywore, since the scalar function
```
sethbrin: nit: keywore -> keyword

src/flag_gems/utils/pointwise_dynamic.py
```diff
@@ -132,627 +168,536 @@ def num_output_tensors(self) -> int:
...
-    non_tensor_index = 0
-    for i in range(op_desc.num_inputs()):
-        if op_desc._is_tensor[i]:
-            parameters.append(f"in{input_tensor_index}")
-            input_tensor_index += 1
+    def gen_decorators(self, code):
+        code.writeline("@libentry()")
+        num_non_tensor_args = self.fx.num_non_tensor_args()
+        if num_non_tensor_args > 0:
+            # we do not specialize non tensor args since they are passed into the inlined function
```
sethbrin, line 262: add support for auto tunning that we can pass tuning configs and tuning prune function?

sethbrin (commented):
other, LGTM

src/flag_gems/utils/pointwise_dynamic.py
iclementine: We actually specialize strides and shape arguments to the kernel. Arguments excluded from specialization are those that are passed to the scalar function. For example, 

```python
@triton.jit
def axpy(x, y, alpha):
    return a * x + y
```

where alpha is an integer. In most of the cases, alpha is a floating point number, we actually do not need to exclude it from specialization, but when it is an integer, we want to avoid specialization.

Another reason for doing this is that earlier design of `Libentry` creates a map from `features of input arguments -> compiledkernel`, where feature of integer arguemnt is the value itself, as to avoid specialization computation. But it is changed now, so we may remove this, too. I will check that it is true.

tests/test_pointwise_dynamic.py
iclementine: It is a test, so we test a lot. But in all these cases, the input is contiguous and the task is simplified to 1d.

There is another case (`test_dynamic_function_with_nd_buffer`)that we test for cases where we cannot simplify the task indexing space to 1d.

I also tried implementing tile with this code generation mechanism, where the task indexing space is 8 dimensional and it worked.

src/flag_gems/utils/pointwise_dynamic.py
iclementine: Thx.

src/flag_gems/utils/pointwise_dynamic.py
iclementine: We can have a more detailed discussion on how to generate the configs and pruning methods for configs. Maybe add a few methods to generate them.

src/flag_gems/utils/pointwise_dynamic.py
iclementine: We now argsort each tensor's strides to get the real stride order. Is this well-supported?

src/flag_gems/utils/pointwise_dynamic.py
iclementine: Yes, we can. I will modify this.

For now the generate module, the source code of the scalar function is included as a track for searching.

```python
"""Quoted source of copy_func:
def copy_func(x):
    return x
"""
```

but addding the scalar function's name as a part of the generated functions look better for profiling and debugging.

src/flag_gems/utils/pointwise_dynamic.py
iclementine: The convention now is to follow the order of the arguments in the scalar function, and then append outputs. I personally prefer this order, but either order is okay.

tests/test_pointwise_dynamic.py
sethbrin: Certainly, the main reason for bringing this up is that our internal operator library currently supports a maximum of 8 dimensions. If the dimensions exceed 8, the tests may not pass, necessitating device-specific judgment. If there is no special testing requirement for dimensions exceeding 8 here, could we consider changing it to 8?

src/flag_gems/utils/pointwise_dynamic.py
sethbrin: Yes. Is there test coverage for scenarios similar to input transpose, such as when the input has shapes: (8, 16) and strides: (1, 8)?

src/flag_gems/utils/pointwise_dynamic.py
iclementine: Yes, the previous design does not require promotion methods, when promotion methods is None, it assumes that there is only 1 output. I will fix this, thank you.

tests/test_pointwise_dynamic.py
iclementine: I see, I will change that.

src/flag_gems/utils/pointwise_dynamic.py
iclementine: There is no such test cases now. I will add some cases with different stride orders.

src/flag_gems/utils/pointwise_dynamic.py
iclementine: Done

src/flag_gems/utils/pointwise_dynamic.py
iclementine: Done

src/flag_gems/utils/pointwise_dynamic.py
iclementine: Done

src/flag_gems/utils/pointwise_dynamic.py
zhzhcookie, line 1: L119 Maybe we should change OPDesc to FunctionSchema in the error log.

src/flag_gems/utils/pointwise_dynamic.py
```diff
@@ -761,64 +705,205 @@ class PointwiseDynamicFunction:
...
+        # output dtype promotions
+        outputs_dtypes_for_allocation = []
+        for i in outputs_that_need_allocation:
+            *arg_indices, method = schema._promotion_methods[i]
+            promote_args = (args[j] for j in arg_indices)
+            _, dtype = type_promotion(*promote_args, type_promotion=method)
+            outputs_dtypes_for_allocation.append(dtype)
+
+        tensors = out_tensors + in_tensors
+        if self.use_fast_path(tensors):  # dimension collapse & use physical ordering
```
sethbrin, line 1139: Theremay better add a canonicalize shape, such as [size, size, size] + [1, 1, size]. Currently, a 3D blockptr is being generated. Here, you can normalize the shape first, like [size*size, size] + [1, size]. This can reduce the size of the tuning space.

src/flag_gems/utils/pointwise_dynamic.py
```diff
@@ -132,627 +167,536 @@ def num_output_tensors(self) -> int:
...
+
+                # number of tasks, used to compute mask
+                code.writeline("num_tasks: int,")
+
+            # tile size & tiles_per_cta, gsl style
+            if ndim > 0:
+                code.writeline("tiles_per_cta: int,")
+                tile_sizes = _cs(f"tile_size{i}: tl.constexpr" for i in range(ndim))
+                code.writeline(f"{tile_sizes},")
+                code.writeline("one_tile_per_cta: tl.constexpr,")
```
sethbrin, line 346: why here need `one_tile_per_cta`? In which scenarios does the presence of the tile loops impact performance?

src/flag_gems/utils/pointwise_dynamic.py
iclementine: I have once implemented this `partial dimension collapsing` mechanism but removed it later since it adds more complexity to output layout inference & task simplification(mainly due to development complexity, since I tried to generate the code for output layout inference & task simplification instead of adding a function that handles the general cases for pointwise operation. As a result, runtime overhead is not yet profiled since I finally removed this feature). 

Task simplification involves dimension reordering & collapsing, which requires some analyses of shape & strides of each tensor operand, mainly sorting.

We can try adding it back & profile to see whether it is beneficial.

src/flag_gems/utils/pointwise_dynamic.py
iclementine: Done

src/flag_gems/utils/pointwise_dynamic.py
iclementine: I did not test that(using `one_tile_per_cta=False` when a cta only needs to process a tile) but since tiles_per_cta is not a constexpr, using `one_tile_per-cta=True` saves a branch instruction.

iclementine: I will test it.

src/flag_gems/utils/pointwise_dynamic.py
sethbrin: It may not have a significant impact on performance, but if decide to enable tuning, it could affect the size of the tuning space due to the addition of an extra dimension.

iclementine: I see. Now we have a fast_that that is used when (all operands have the same shape & all non overlapping & the same stride order). This condition can be checked with relatively low cost.

We will add dimension reordering and collapsing back, since it would not add overhead if the fast path is used.

.gitignore
```diff
@@ -21,3 +21,4 @@ build/
 # Project files, i.e. `.project`, `.actionScriptProperties` and `.flexProperties`
 # should NOT be excluded as they contain compiler settings and other important
 # information for Eclipse / Flash Builder.
+playground/
```
Bowen12992: where is `playground` folder

iclementine: Oh, it is a folder where I put some temporary test scripts that are not included in the git. Actually I have a lot of them.

src/flag_gems/utils/tensor_wrapper.py
```diff
@@ -0,0 +1,79 @@
+import math
+
+import torch
+
+
+class TypedPtr:
+    """This is a minimal requirement for a type to be treated as a tensor in triton jit
+    function. Basically it is ia typed pointer, withou knowning the device, size, shape,
```
MARD1NO: small typo: withou knowning -> without knowning

src/flag_gems/utils/tensor_wrapper.py
```diff
@@ -0,0 +1,79 @@
...
+
+    def __init__(
+        self, base: torch.Tensor, shape=None, strides=None, dtype=None, offset=0
+    ):
+        self._base = base
+        self.dtype = dtype or base.dtype
+        if offset == 0:
+            self._data_ptr = self._base.data_ptr()
+        else:
+            offset = self.dtype.itemsize * offset
```
MARD1NO, line 53: The offset param is only used in flip? we have shape and stride info to compute offset

iclementine: For now this is only used in flip, since we cannot construct a tensor view with negative strides via torch APIs. In many other cases, we can use torch's view ops to construct a view.

src/flag_gems/utils/tensor_wrapper.py
iclementine: Thx~

src/flag_gems/utils/pointwise_dynamic.py
iclementine: As tested, `one_tile_per_cta` has little or no peformance effect for pointwise operations.

```text
# one_tile_per_cta: True
test_pointwise_perf.py Operator abs Performance Test (torch.float16)
Size        Torch Latency (ms)   Gems Latency (ms)
--------------------------------------------------
1024                  0.009216            0.009216
6144                  0.034816            0.034816
11264                 0.059392            0.059392
16384                 0.084992            0.084992
21504                 0.109568            0.109568
26624                 0.134144            0.134144
31744                 0.159744            0.159744
36864                  0.18432             0.18432
41984                  0.20992            0.208896
47104                 0.234496            0.234496
52224                 0.259072            0.259072
57344                 0.283648             0.28384
62464                 0.309248            0.309248
67584                 0.333824            0.333824
72704                   0.3584            0.359424
77824                 0.382976            0.385024
```
```text
# one_tile_per_cta = False
test_pointwise_perf.py Operator abs Performance Test (torch.float16)
Size        Torch Latency (ms)   Gems Latency (ms)
--------------------------------------------------
1024                  0.009216            0.009216
6144                  0.034816            0.034816
11264                 0.060416            0.060416
16384                 0.084992            0.084992
21504                 0.109568            0.109568
26624                 0.134144            0.134144
31744                 0.159744            0.159744
36864                  0.18432             0.18432
41984                  0.20992            0.208896
47104                 0.234496            0.234496
52224                 0.259072            0.259072
57344                 0.283648            0.284448
62464                 0.309248            0.309248
67584                 0.333824            0.333824
72704                   0.3584            0.359424
77824                 0.382976            0.385024
```

sethbrin: The performance differences caused by branching are roughly similar; theoretically, the more complex the pointwise operation, the smaller this impact becomes. Is it necessary to remove the `one_tile_per_cta` option?

iclementine: We can probably remove it but retaining it now does not hurt performance. We may remove it after more tests.

Bowen12992 (approved):
LGTM

StrongSpoon (approved):
lgtm

PR #168, opened 2024-08-16
[Operator] Add randn_like, fix trunc_div and floor_div

Bowen12992:
tl.math.div_rz only support float32, type cast will cause diff with torch, so we only do float32 test for now


src/flag_gems/ops/randn_like.py
```diff
@@ -0,0 +1,29 @@
...
+
+UNROLL = 4
+
+
+def randn_like(
+    x, *, dtype=None, layout=None, device=None, pin_memory=None, memory_format=None
+):
+    logging.debug("GEMS RANDN_LIKE")
+    if device is None:
+        device = x.device
```
tongxin: Using device.index is slightly faster.

tongxin (approved):
LG

src/flag_gems/ops/randn_like.py
Bowen12992: ok

iclementine (approved):

PR #174, opened 2024-08-22
ensure scalars are float32 in test cases for dropout

iclementine:
Ensure scalars are float32 in test cases for dropout. Now dropout is computed in the following formula:

`out = tensor * mask / (1-p)`

In our test cases, the reference is also compute in thay.

But in some cases, `1-p` have different byte representations depending on how it is computed.

1. `(1 - p)` when `p` is a native python floating point number then cast to float32;
2. `(1 - p)` when `p` and `1` are both numpy 32bit floating point number then cast to float32;

Thus, `tensor / (1-p)` may have have different results when `1-p` have different representations.

### DIFF
--- a/tests/test_special_ops.py
+++ b/tests/test_special_ops.py
@@ -1,5 +1,6 @@
 from typing import Optional
 
+import numpy as np
 import pytest
 import torch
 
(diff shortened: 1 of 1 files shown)


tongxin (approved):
LG

Bowen12992 (approved):
LGTM

PR #179, opened 2024-08-26
[test]: fix embedding/topk test cpu mode and celoss inf bug

FuncSherl:
1. fix device error in test:    
 pytest -sv test_special_ops.py  --device cpu
2. fix possible numerical accuracy overflow in cross entropy loss, if inp.dtype is f16 and sum(out) > max(f16), the cvt may cause inf, then the multiply result is inf;  but multiply mean_num first may avoid inf


tongxin (approved):
LGTM

StrongSpoon (approved):
lgtm

PR #184, opened 2024-08-27
[CI/CD ] Add safe directory

Bowen12992:


iclementine (approved):
LGTM

A step to enable using git commands on a local repository that is not owned by the current user.

PR #187, opened 2024-08-28
use a singleton to avoid generating code every time

iclementine:
Use a singleton to avoid generating code every time tile or pad is called


MARD1NO (dismissed):

StrongSpoon (approved):
lgtm

zfu82 (approved):

Bowen12992 (approved):
LGTM

PR #190, opened 2024-08-29
Suppress "Warning only once for all operators" during PyTorch dispatch

2niuhe:
This commit implements a warning filter to ignore the "Warning only once for all operators, other operators may also be overrided." message generated during PyTorch dispatch calls. This change helps maintain cleaner logs by preventing this specific warning from appearing, which can occur when multiple operators are overridden.


Bowen12992:
Thanks for your contribution，but I wonder if there is a more helpful way to suppress this warning but can also indicates that the replacement of flag_gems succeeded ？Also，Please use pre-commit to format your codes to make them the same style with the code repo

2niuhe:
Hi @Bowen12992,

Thank you for your feedback. I understand your concerns, but I currently cannot find an **elegant** solution that both suppresses the warning and indicates that the patch was successful. Therefore, I have decided to close this PR for now.

If I come up with a better approach in the future, I will revisit this issue.

PR #202, opened 2024-09-06
[Bugfix] scatter&gather's accuracy test

GwokHiujin:
This PR increases the BLOCK_M sizes in the `offset_calculator` function to prevent excessively small BLOCK_M values, which may resolve the correctness issues encountered during local testing.

---

If CI runs smoothly, I can speculate on the potential causes of this confusing issue...🧐

The correctness problem was *likely* due to insufficient block size during tensor processing, resulting in incomplete or inaccurate memory accesses and inconsistent results. By increasing the BLOCK_M size, we ensure that larger data blocks are processed more consistently, reducing the risk of edge-case errors in tensor indexing.


src/flag_gems/utils/shape_utils.py
```diff
@@ -212,37 +212,48 @@ def add_on_kernel(
...
+
+    # Ensure grid configuration is correct
+    grid = lambda meta: (
+        triton.cdiv(M, meta["BLOCK_M"]),
+        triton.cdiv(N, meta["BLOCK_N"]),
+    )
+
+    for d in range(ndim):
+        # Ensure correct size for add_on
         add_on = torch.zeros_like(inp, dtype=torch.int32, device=inp.device)
```
tongxin, line 229: Since idx is in int64, int32 add_on may trigger overflow. could it be the cause?

GwokHiujin: I tend to say not because I've already considered this reason and made the corresponding modifications before, without fixing the original issue.

tongxin (commented):
The following test case fails. It may require a more robust autotuning. 
```
import torch
import flag_gems
x = torch.ones(1000000000, dtype=torch.float16, device='cuda')
idx = torch.arange(1000000000, device='cuda')
y = flag_gems.ops.gather(x, 0, idx)
```

GwokHiujin:
> The following test case fails. It may require a more robust autotuning.
> 
> ```
> import torch
> import flag_gems
> x = torch.ones(1000000000, dtype=torch.float16, device='cuda')
> idx = torch.arange(1000000000, device='cuda')
> y = flag_gems.ops.gather(x, 0, idx)
> ```



> The following test case fails. It may require a more robust autotuning.
> 
> ```
> import torch
> import flag_gems
> x = torch.ones(1000000000, dtype=torch.float16, device='cuda')
> idx = torch.arange(1000000000, device='cuda')
> y = flag_gems.ops.gather(x, 0, idx)
> ```

So it seems that we just cannot allow the kernel grid's row numbers to be too large! However, since we are planning to use codegen for `scatter`, `gather`, and related operators, the offset calculations will shift into the kernels, potentially making this version out-of-date. Should we proceed with this patch now?

tongxin:
> > The following test case fails. It may require a more robust autotuning.
> > ```
> > import torch
> > import flag_gems
> > x = torch.ones(1000000000, dtype=torch.float16, device='cuda')
> > idx = torch.arange(1000000000, device='cuda')
> > y = flag_gems.ops.gather(x, 0, idx)
> > ```
> 
> So it seems that we just cannot allow the kernel grid's row numbers to be too large! However, since we are planning to use codegen for `scatter`, `gather`, and related operators, the offset calculations will shift into the kernels, potentially making this version out-of-date. Should we proceed with this patch now?

Let's shift to the fused kernel, shall we?

PR #205, opened 2024-09-07
[Operator] Add radix-2 fft

Bowen12992:
### PR Category

Operator
### Type of Change

New Feature
### Description

This PR implement a native FFT with Cooley-Tukey method，current it only support radix-2
The performance is far worse than pytorch `aten` library (which use `CuFFT` library actually) for some reasons
- triton-lang does not support complex dtype so we need to separate the imag and real part
- triton-lang does not support `__syncthreads` so we must launch much more kernels
- A fft plan is needed for differnet kinds of cases 
- sharedMem should be used for small size FFT

### Progress

### Performance

```
benchmark/test_special_perf.py Operator fft.fft Performance Test (torch.complex64)
Size    Torch Latency (ms)    Gems Latency (ms)    Gems Speedup
---------------------------------------------------------------
1024               0.00512              0.04096           0.125
11264             0.012288             0.082944           0.148
21504              0.01536             0.102144            0.15
31744             0.014336             0.100352           0.143
41984              0.01536              0.11776            0.13
52224             0.017408             0.115712            0.15
62464             0.016384              0.11776           0.139
72704             0.017408             0.131072           0.133
82944             0.014496             0.131072           0.111
93184             0.018432             0.131136           0.141
```


src/flag_gems/experimental/fft.py
```diff
@@ -0,0 +1,111 @@
...
+    indices = torch.arange(n, device=x.device)
+    assert x.is_cuda and indices.is_cuda
+    n_elements = indices.numel()
+    grid = lambda meta: (triton.cdiv(n_elements, meta["BLOCK_SIZE"]),)
+    bit_reversal_indices_kernel[grid](indices, n_elements, num_bits, BLOCK_SIZE=1024)
+    return indices
+
+
+# TODO(Bowen12992): support more cases
+def rad2_fft(x: torch.Tensor, n=None, dim=-1, norm=None):
```
tongxin: Shall we use the same name as the aten function to be substituted?

src/flag_gems/experimental/fft.py
```diff
@@ -0,0 +1,111 @@
...
+    logging.debug("GEMS FFT")
+    n_elements = x.numel()
+    assert n is None, "Not support assign `signal length` currently."
```
tongxin: Check the grammar please.

src/flag_gems/experimental/fft.py
```diff
@@ -0,0 +1,111 @@
...
+
+    tl.store(xr_ptr + k, xkr, mask=mask)
+    tl.store(xi_ptr + k, xki, mask=mask)
+
+    tl.store(xr_ptr + j, xjr, mask=mask)
+    tl.store(xi_ptr + j, xji, mask=mask)
+
+
+def bit_reversal_indices(x: torch.Tensor, n):
+    num_bits = int(np.log2(n))
```
tongxin:
```suggestion
    num_bits = int(np.ceil(np.log2(n)))
```

src/flag_gems/experimental/fft.py
```diff
@@ -0,0 +1,111 @@
...
+    indices = torch.arange(n, device=x.device)
+    assert x.is_cuda and indices.is_cuda
```
tongxin: We should be careful that checking cuda device only works on NV cards.

src/flag_gems/experimental/fft.py
```diff
@@ -0,0 +1,111 @@
+import cmath
+import logging
+
+import numpy as np
+import torch
+import triton
+import triton.language as tl
+
+
+@triton.jit
```
tongxin: Redundant jitting is better to be avoided.

```suggestion
@triton.jit(do_not_specialize(['n_elements', 'num_bits']))
```

src/flag_gems/experimental/fft.py
```diff
@@ -0,0 +1,111 @@
...
+    block_start = pid * BLOCK_SIZE
+    offsets = block_start + tl.arange(0, BLOCK_SIZE)
+    mask = offsets < n_elements
+
+    indices = offsets
+    rev = tl.zeros_like(offsets)
+
+    for j in range(num_bits):
+        rev = tl.where(offsets & (1 << j), rev | (1 << (num_bits - 1 - j)), rev)
+    indices = rev
```
tongxin: We'd consider making num_bits constexpr because this is a fine grained loop and `1 << (num_bits - 1 - j)` be better off with a constant operand.

tongxin: Also, can we remove tl.where predicate by right shifting then left shifting the input bits?
 ```suggestion
    for j in range(num_bits):
        offsets = tl.lshr(offsets, 1)
        rev = rev | tl.shl(offsets & 0x1, num_bits - 1 - j)
    indices = rev
```

File: src/flag_gems/experimental/fft.py [hunk 7]
```diff
@@ -0,0 +1,111 @@
...
+    assert norm is None, "Not support assign `normalization mode` currently."
+
+    indices = bit_reversal_indices(x, len(x))
+
+    if x.dtype is not torch.cfloat:
+        xr = x[indices]
+        xi = torch.zeros_like(x)
+    else:
+        xr = x.real[indices]
+        xi = x.imag[indices]
```
tongxin: Is it possible to fuse slicing into the index or fft kernels?

src/flag_gems/experimental/fft.py
```diff
@@ -0,0 +1,111 @@
...
+    rev = tl.zeros_like(offsets)
+
+    for j in range(num_bits):
+        rev = tl.where(offsets & (1 << j), rev | (1 << (num_bits - 1 - j)), rev)
+    indices = rev
+
+    tl.store(output_ptr + offsets, indices, mask=mask)
+
+
+@triton.jit
```
tongxin: do not specialize n_elements and step

Bowen12992: done

src/flag_gems/experimental/fft.py
Bowen12992: i don‘t quite understand why those`shifting` can replace `where`, and i didn't find `tl.lshr` in triton 2.2

src/flag_gems/experimental/fft.py
Bowen12992: yes, we do not need to specialize `n_elements`, but  specialize `num_bits` will bring better performance

src/flag_gems/experimental/fft.py
Bowen12992: thx

src/flag_gems/experimental/fft.py
Bowen12992: this algorithm only support radix-2 fft ，so we don't need use `ceil` currently

src/flag_gems/experimental/fft.py
Bowen12992: Here we need a `fft plan` to manage the different `shapes`， This PR only support radix-2，if we use radix-3  radix-5...  we need that plan to tell us how to do fft and if we need to do padding

src/flag_gems/experimental/fft.py
Bowen12992: yes，but this will return a view  and takes litte time

src/flag_gems/ops/fft.py
```diff
@@ -0,0 +1,109 @@
...
+    indices = bit_reversal_indices(x, len(x))
+
+    if x.dtype is not torch.cfloat:
+        xr = x[indices]
+        xi = torch.zeros_like(x)
+    else:
+        xr = x.real[indices]
+        xi = x.imag[indices]
+
+    assert xi.is_cuda and xr.is_cuda
```
zhzhcookie, line 100: Can we check cuda in gems code?

src/flag_gems/ops/fft.py
```diff
@@ -0,0 +1,109 @@
...
+    n_elements,
+    num_bits,
+    BLOCK_SIZE: tl.constexpr,
+):
+    pid = tl.program_id(axis=0)
+    block_start = pid * BLOCK_SIZE
+    offsets = block_start + tl.arange(0, BLOCK_SIZE)
+    mask = offsets < n_elements
+
+    indices = offsets
```
zhzhcookie, line 22: Seems not necessary to init indices?

src/flag_gems/ops/fft.py
```diff
@@ -0,0 +1,109 @@
...
+    tl.store(xr_ptr + k, xkr, mask=mask)
+    tl.store(xi_ptr + k, xki, mask=mask)
+
+    tl.store(xr_ptr + j, xjr, mask=mask)
+    tl.store(xi_ptr + j, xji, mask=mask)
+
+
+def bit_reversal_indices(x: torch.Tensor, n):
+    num_bits = int(np.log2(n))
+    indices = torch.arange(n, device=x.device)
```
zhzhcookie, line 78: Why not `indices = torch.empty_like(x, dtype=int)`

src/flag_gems/ops/fft.py
```diff
@@ -0,0 +1,109 @@
...
+):
+    pid = tl.program_id(axis=0)
+    block_start = pid * BLOCK_SIZE
+    offsets = block_start + tl.arange(0, BLOCK_SIZE)
+    mask = offsets < n_elements
+
+    i = offsets
+    starts = offsets
+    j = offsets
+    k = offsets
```
zhzhcookie, line 48: Why are these 4 vars initialized here?

src/flag_gems/ops/fft.py
```diff
@@ -0,0 +1,109 @@
...
+    block_start = pid * BLOCK_SIZE
+    offsets = block_start + tl.arange(0, BLOCK_SIZE)
+    mask = offsets < n_elements
+
+    indices = offsets
+    rev = tl.zeros_like(offsets)
+
+    for j in range(num_bits):
+        rev = tl.where(offsets & (1 << j), rev | (1 << (num_bits - 1 - j)), rev)
+    indices = rev
```
tongxin, line 27: I think there's an error here. If num_bits is zero, the default indices will be reset to zeros.

src/flag_gems/ops/fft.py
```diff
@@ -0,0 +1,109 @@
...
+
+import numpy as np
+import torch
+import triton
+import triton.language as tl
+
+
+@triton.jit(do_not_specialize=["n_elements"])
+def bit_reversal_indices_kernel(
+    output_ptr,
```
tongxin, line 12: Here `output_ptr` is only used for storing the output indices. However, to generalize bit reversal to arbitrary input, I think we need to pass in both input and output which could certainly be aliased but it's fine.  That implies that the initial `indices` should be loaded to be distinguished with the tile offsets.

Bowen12992:
Since fft is difficult for triton now, i checkout a new branch named `fft_dev` and close this PR

PR #208, opened 2024-09-12
[Tests] Fix tests 1st ed.

zhzhcookie:
### PR Category
[OP Test]

### Type of Change
[Tests]

### Description
1. Modify test cases, reduce similar type test cases, increase test cases of different dimensions and size ranges, and try to unify the test shapes of similar operators.
2. Fix --device=cpu.

### Progress

### Performance
image

### DIFF
--- a/tests/accuracy_utils.py
+++ b/tests/accuracy_utils.py
@@ -26,37 +26,44 @@ sizes_one = [1]
 sizes_pow_2 = [2**d for d in range(4, 11, 2)]
 sizes_noalign = [d + 17 for d in sizes_pow_2]
 sizes_1d = sizes_one + sizes_pow_2 + sizes_noalign
-sizes_2d_nc = [1, 16, 64, 1000]
-sizes_2d_nr = [1, 5, 1024]
+sizes_2d_nc = [1] if TO_CPU else [1, 16, 64, 1000]
+sizes_2d_nr = [1] if TO_CPU else [1, 5, 1024]
 
 UT_SHAPES_1D = list((n,) for n in sizes_1d)
 UT_SHAPES_2D = list(itertools.product(sizes_2d_nr, sizes_2d_nc))
-POINTWISE_SHAPES = [(1024, 1024), (16, 1024, 256), (16, 128, 64, 64), (20, 320, 15)]
+POINTWISE_SHAPES = [(2, 19, 7)] if TO_CPU else [(1,), (1024, 1024), (20, 320, 15), (16, 128, 64, 60), (16, 7, 57, 32, 29)]
+SPECIAL_SHAPES = [(2, 19, 7)] if TO_CPU else [(1,), (1024, 1024), (20, 320, 15), (16, 128, 64, 1280), (16, 7, 57, 32, 29)]
 DISTRIBUTION_SHAPES = [(20, 320, 15)]
-REDUCTION_SHAPES = [(4096, 256 * i) for i in range(1, 10, 2)]
-MNK_SHAPES = [15, 160, 1024]
-REDUCTION_MNK_SHAPES = [(15, 160, 1024), (16, 1025, 255)]
-ONE_DIM_SHAPES = [(256 * i + 7,) for i in range(1, 10, 2)]
+REDUCTION_SHAPES = [(2, 32)] if TO_CPU else [(1, 2), (4096, 256), (200, 40999, 3)]
+REDUCTION_SMALL_SHAPES = [(1, 32)] if TO_CPU else [(1, 2), (4096, 256), (200, 2560, 3)]
+NONZERO_SHAPES = [(2, 32)] if TO_CPU else REDUCTION_SHAPES + [(2637,)]
+CUMSUM_SHAPES = [(2, 32)] if TO_CPU else REDUCTION_SHAPES + [(2637,), (16, 1025, 255)]
+MN_SHAPES = [(1, 32)] if TO_CPU else [(1, 32), (160, 1024), (5333, 497)]
+MNK_SHAPES = [(1, 1, 32)] if TO_CPU else [(1, 1, 32), (15, 160, 1024), (495, 5333, 71)]
 
-DIM_POINTWISE_SHAPES = [
-    (1024, 1024, 1),
-    (16, 1024, 256),
-    (16, 7, 128, 64, 64),
-    (20, 320, 15),
-]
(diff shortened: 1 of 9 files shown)


tests/accuracy_utils.py
```diff
@@ -26,37 +26,83 @@
 sizes_pow_2 = [2**d for d in range(4, 11, 2)]
 sizes_noalign = [d + 17 for d in sizes_pow_2]
 sizes_1d = sizes_one + sizes_pow_2 + sizes_noalign
-sizes_2d_nc = [1, 16, 64, 1000]
-sizes_2d_nr = [1, 5, 1024]
+sizes_2d_nc = [1] if TO_CPU else [1, 16, 64, 1000]
+sizes_2d_nr = [1] if TO_CPU else [1, 5, 1024]
```
iclementine, line 30: `TO_CPU` indicates to compute the reference value on CPU, so use less and smaller sizes to reduce the runtime?

tests/accuracy_utils.py
```diff
@@ -26,37 +26,83 @@
...
 
 UT_SHAPES_1D = list((n,) for n in sizes_1d)
 UT_SHAPES_2D = list(itertools.product(sizes_2d_nr, sizes_2d_nc))
-POINTWISE_SHAPES = [(1024, 1024), (16, 1024, 256), (16, 128, 64, 64), (20, 320, 15)]
+POINTWISE_SHAPES = (
+    [(2, 19, 7)]
+    if TO_CPU
+    else [(1,), (1024, 1024), (20, 320, 15), (16, 128, 64, 60), (16, 7, 57, 32, 29)]
```
iclementine, line 37: Add a 0-d shape for pointwise operation to test that 0-d works as expected? Although we now flatten them as vectors of shape `(1, )`.

tests/accuracy_utils.py
```diff
@@ -26,37 +26,83 @@
...
-    (16, 1024, 256),
-    (16, 7, 128, 64, 64),
-    (20, 320, 15),
-]
-DIMS = [[0], [-2], [2], [0, 2], [2, 1], [0, -1, 1]]
-TILE_DIMS = [(2,), (2, 0), (0, 2), (2, 2), (2, 2, 2), (2, 2, 2, 2)]
-REPEAT_SIZES = [(2, 3, 4, 5), (2, 0, 4, 5)]
+REDUCTION_SHAPES = [(2, 32)] if TO_CPU else [(1, 2), (4096, 256), (200, 40999, 3)]
+REDUCTION_SMALL_SHAPES = [(1, 32)] if TO_CPU else [(1, 2), (4096, 256), (200, 2560, 3)]
+NONZERO_SHAPES = [(2, 32)] if TO_CPU else REDUCTION_SHAPES + [(2637,)]
```
Bowen12992: There are many `OP_SHAPES` we defined here but only used once for the certain op, how about move them close to the file/lines where we use them

Bowen12992 (commented):
LGTM

Bowen12992 (approved):

Bowen12992 (dismissed):
LGTM. Looking forward the next PR

tests/test_named_ops.py
```diff
@@ -18,8 +18,7 @@
 
 reduction_ops_ut_map = {
     "all": (
-        "test_accuracy_all",
-        "test_accuracy_all_dim",
+        "test_accuracy_all_without_dim",
```
iclementine, line 21: We can use pytest mark to group test cases that test a same operator. 
https://docs.pytest.org/en/stable/how-to/mark.html

tests/accuracy_utils.py
zhzhcookie: > `TO_CPU` indicates to compute the reference value on CPU, so use less and smaller sizes to reduce the runtime?

Yes, we just pass the cpu tests, and will add cpu tests to ci tests soon. We will test all shapes when releasing a version.

tests/accuracy_utils.py
zhzhcookie: > Add a 0-d shape for pointwise operation to test that 0-d works as expected? Although we now flatten them as vectors of shape `(1, )`.

OK. Now there are some bugs when use scalar in tests code. We will fix it soon.

tests/accuracy_utils.py
zhzhcookie: > There are many `OP_SHAPES` we defined here but only used once for the certain op, how about move them close to the file/lines where we use them

Done.

tests/test_unary_pointwise_ops.py
```diff
@@ -303,26 +303,44 @@ def test_accuracy_isfinite(shape, dtype):
...
-@pytest.mark.parametrize("dtype", ALL_FLOAT_DTYPES + ALL_INT_DTYPES)
-@pytest.mark.parametrize("dims", DIMS)
-def test_accuracy_flip(shape, dtype, dims):
+def get_max_ndim(shape, dims):
+    max_ndim = max(len(shape), len(dims))
+    for dim in dims:
+        dim = abs(dim) + 1
+        if dim > max_ndim:
+            max_ndim = dim
+    return max_ndim
```
iclementine: Is this used to make a tensor have at least N dimensions, such that each axis in dims is a valid axis in the tensor?

Then if axis is not negative, the ndim is required to be `ndim >= axis + 1`; and when the axis not positive, the ndim is requires to be `ndim >= abs(axis)`.



For example, a tensor is only requires to have at least 3d to get its axis `-3`.

```text
(0, 1, 2, 3)
(-4, -3, -2, -1)
```

Bowen12992 (dismissed):

tests/test_unary_pointwise_ops.py
zhzhcookie: > Is this used to make a tensor have at least N dimensions, such that each axis in dims is a valid axis in the tensor?
> 
> Then if axis is not negative, the ndim is required to be `ndim >= axis + 1`; and when the axis not positive, the ndim is requires to be `ndim >= abs(axis)`.
> 
> For example, a tensor is only requires to have at least 3d to get its axis `-3`.
> 
> ```
> (0, 1, 2, 3)
> (-4, -3, -2, -1)
> ```

Done.

tests/test_named_ops.py
zhzhcookie: > We can use pytest mark to group test cases that test a same operator. https://docs.pytest.org/en/stable/how-to/mark.html

Thanks, pytest mark will be used at 2nd ed.

Bowen12992 (approved):

PR #232, opened 2024-10-10
[fix] Fix the input shape of gather's perf test

GwokHiujin:
As title, fixed the test data shape in the gather performance test that could lead to errors.


StrongSpoon (approved):
lgtm

PR #238, opened 2024-10-21
[bugfix] upcast input tensor to fp64 for cpu reference

StrongSpoon:
### PR Category

OP Test
### Type of Change

Bug Fix
### Description

upcast input tensor to fp64 for cpu reference. so that weight_norm and upsample_bicubic2d_aa could run fp16 tests in cpu mode.
### Progress


iclementine:
I wonder why the original `upcast` value for these tests were `False`?
I thought non-upcasting only applies to some integer related operarions?

iclementine (approved):

PR #264, opened 2024-10-29
[Optimization] CrossEntropyLoss

StrongSpoon:
### PR Category

Operator
### Type of Change

Performance Optimization
### Description

apply fusion on celoss kernel and sum kernel of weight.
avoid memory copy from device to host by disabling context saving when input tensor doesn't require gradient.
achieve over 200% relative performance on average compared to aten (only 13% before)

A100-benchmark performance
截屏2024-11-20 17 48 52

### Progress

### DIFF
--- a/src/flag_gems/ops/cross_entropy_loss.py
+++ b/src/flag_gems/ops/cross_entropy_loss.py
@@ -5,7 +5,6 @@ import triton
 import triton.language as tl
 
 from ..utils import libentry
-from .sum import sum
 
 
 @libentry()
@@ -16,6 +15,7 @@ from .sum import sum
         for d in [1, 4, 16]
     ],
     key=["C", "D"],
+    reset_to_zero=["out_sum_ptr", "w_tgt_ptr"],
 )
 @triton.jit(do_not_specialize=["ignore_index"])
 def celoss_indice_kernel(
@@ -23,6 +23,7 @@ def celoss_indice_kernel(
     tgt_ptr,
     w_ptr,
     out_ptr,
+    out_sum_ptr,
     w_tgt_ptr,
     ignore_index,
     N,
@@ -39,12 +40,11 @@ def celoss_indice_kernel(
     tgt_mask = offset_d < D
     tgt = tl.load(tgt_ptrs, mask=tgt_mask, other=0)
 
-    ignore_mask = not (tgt == ignore_index)
+    ignore_mask = not (tgt == ignore_index) and tgt_mask
 
-    w_ptrs = w_ptr + tgt
-    w_tgt = tl.load(w_ptrs, mask=tgt_mask, other=0).to(tl.float32)
-    w_tgt_ptrs = w_tgt_ptr + pid_n * D + offset_d
-    tl.store(w_tgt_ptrs, w_tgt, mask=tgt_mask and ignore_mask)
+    w_tgt = tl.load(w_ptr + tgt, mask=ignore_mask, other=0).to(tl.float32)
+    w_tgt_sum = tl.sum(w_tgt)
+    tl.atomic_add(w_tgt_ptr, w_tgt_sum)
 
     tmp_max = tl.zeros([BLOCK_C, BLOCK_D], dtype=tl.float32)
     tmp_sum = tl.zeros([BLOCK_C, BLOCK_D], dtype=tl.float32)
@@ -67,7 +67,9 @@ def celoss_indice_kernel(
 
     out = (final_sum + final_max - inp_tgt) * w_tgt
     out_ptrs = out_ptr + pid_n * D + offset_d
-    tl.store(out_ptrs, out, mask=tgt_mask and ignore_mask)
+    tl.store(out_ptrs, out)
+    out_sum = tl.sum(out)
+    tl.atomic_add(out_sum_ptr, out_sum)
 
 
 @libentry()
@@ -78,6 +80,9 @@ def celoss_indice_kernel(
         for d in [1, 4, 16]
     ],
     key=["C", "D"],
+    reset_to_zero=[
+        "out_sum_ptr",
+    ],
 )
 @triton.jit(do_not_specialize=["label_smoothing"])
 def celoss_probability_kernel(
@@ -85,6 +90,7 @@ def celoss_probability_kernel(
     tgt_ptr,
     w_ptr,
     out_ptr,
+    out_sum_ptr,
     label_smoothing,
     N,
     C,
@@ -121,15 +127,18 @@ def celoss_probability_kernel(
         w_ptrs = w_ptr + offset_c
         w_mask = offset_c < C
         inp = tl.load(inp_ptrs, mask, other=0).to(tl.float32)
-        tgt = tl.load(tgt_ptrs, mask, other=1).to(tl.float32)
+        tgt = tl.load(tgt_ptrs, mask, other=0).to(tl.float32)
         tgt = tgt * (1.0 - label_smoothing) + label_smoothing / C
         w = tl.load(w_ptrs, w_mask, other=0).to(tl.float32)[:, None]
         log = final_sum + final_max - inp
         _sum += w * log * tgt
 
     out = tl.sum(_sum, axis=0)
+    out = tl.where(offset_d < D, out, 0)
     out_ptrs = out_ptr + pid_n * D + offset_d
     tl.store(out_ptrs, out, mask=offset_d < D)
+    out_sum = tl.sum(out)
+    tl.atomic_add(out_sum_ptr, out_sum)
 
 
 @libentry()
@@ -140,6 +149,7 @@ def celoss_probability_kernel(
         for d in [1, 4, 16]
     ],
     key=["C", "D"],
+    reset_to_zero=["out_sum_ptr", "w_tgt_ptr"],
 )
 @triton.jit(do_not_specialize=["ignore_index", "label_smoothing"])
 def celoss_indice_smooth_kernel(
@@ -147,6 +157,7 @@ def celoss_indice_smooth_kernel(
     tgt_ptr,
     w_ptr,
     out_ptr,
+    out_sum_ptr,
     w_tgt_ptr,
     ignore_index,
     label_smoothing,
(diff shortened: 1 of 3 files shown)


src/flag_gems/ops/cross_entropy_loss.py
```diff
@@ -16,6 +15,7 @@
         for d in [1, 4, 16]
     ],
     key=["C", "D"],
+    reset_to_zero=["w_tgt_ptr"],
 )
 @triton.jit(do_not_specialize=["ignore_index"])
 def celoss_indice_kernel(
```
tongxin: typo: indices, or just index

src/flag_gems/ops/cross_entropy_loss.py
```diff
@@ -121,13 +120,14 @@ def celoss_probability_kernel(
...
         inp = tl.load(inp_ptrs, mask, other=0).to(tl.float32)
-        tgt = tl.load(tgt_ptrs, mask, other=1).to(tl.float32)
+        tgt = tl.load(tgt_ptrs, mask, other=0).to(tl.float32)
         tgt = tgt * (1.0 - label_smoothing) + label_smoothing / C
         w = tl.load(w_ptrs, w_mask, other=0).to(tl.float32)[:, None]
         log = final_sum + final_max - inp
         _sum += w * log * tgt
 
     out = tl.sum(_sum, axis=0)
+    out = tl.where(offset_d < D, out, 0)
```
tongxin: I guess this is redundant.

src/flag_gems/ops/cross_entropy_loss.py
StrongSpoon: renamed

src/flag_gems/ops/cross_entropy_loss.py
StrongSpoon: deleted.

src/flag_gems/ops/cross_entropy_loss.py
```diff
@@ -121,7 +120,7 @@ def celoss_probability_kernel(
         w_ptrs = w_ptr + offset_c
         w_mask = offset_c < C
         inp = tl.load(inp_ptrs, mask, other=0).to(tl.float32)
-        tgt = tl.load(tgt_ptrs, mask, other=1).to(tl.float32)
+        tgt = tl.load(tgt_ptrs, mask, other=0).to(tl.float32)
```
iclementine, line 123: Since `w` is also masked on `C` dimension, and tgt only affects the result by pointwise-multiplying w, this change has no effect on outputs?

src/flag_gems/ops/cross_entropy_loss.py
```diff
@@ -302,7 +304,6 @@ def celoss_probability_bwd(
     inp_grad_ptr,
     label_smoothing,
     mean_num,
-    N,
     C,
```
iclementine, line 316: There is possibility that `N` is critical to the choice of `BLOCK_C` and `BLOCK_D` and thus should be used in autotunner key.

src/flag_gems/ops/cross_entropy_loss.py
```diff
@@ -462,11 +462,47 @@ def celoss_indice_smooth_bwd(
...
+        offset = off + tl.arange(0, BLOCK_N)
+        inp_ptrs = inp_ptr + offset
+        mask = offset < N
+        inp_vals = tl.load(inp_ptrs, mask=mask, other=0.0)
+        mid_sum += inp_vals
+    out_val = tl.sum(mid_sum)
+    if ifscale == 1:
+        out_val /= scale
+    elif ifscale == 2:
+        scale_val = tl.load(scale)
```
iclementine: This branching makes the type of scale ambiguous, it is either a number, or a pointer.

src/flag_gems/ops/cross_entropy_loss.py
```diff
@@ -488,47 +524,77 @@ def forward(ctx, inp, target, weight, reduction, ignore_index, label_smoothing):
...
         else:
-            w_tgt = torch.zeros(shape, dtype=torch.float32, device=inp.device)
+            w_tgt = torch.zeros([], dtype=torch.float32, device=inp.device)
             with torch.cuda.device(inp.device):
-                celoss_indice_smooth_kernel[grid](
-                    inp, tgt, weight, out, w_tgt, ignore_index, label_smoothing, N, C, D
+                celoss_indices_smooth_kernel[grid](
+                    inp,
+                    tgt,
+                    weight,
```
iclementine, line 591: Maybe we can optimize weight here. 
The jit function can handle the case that weight is None.(Triton can handle `None` as argument, althought it also involves some kind of overload becase at that case, the type of weight is not a pointer.)

So we do not need to really generate a tensor of ones. Also we do not need to load them in that case.

src/flag_gems/ops/cross_entropy_loss.py
```diff
@@ -488,47 +524,77 @@ def forward(ctx, inp, target, weight, reduction, ignore_index, label_smoothing):
         inp = inp.contiguous()
         tgt = target.contiguous()
         weight = weight.contiguous()
-        out = torch.zeros(shape, dtype=torch.float32, device=inp.device)
+        out = torch.empty(shape, dtype=torch.float32, device=inp.device)
         grid = lambda meta: (N, triton.cdiv(D, meta["BLOCK_D"]))
```
iclementine: I suggest reversing the order of the 2 dimensions, since when CTAs in a grid gets scheduled in an Fortran order. (Iterate grid.x, then grid.y, then grid.z).

src/flag_gems/ops/cross_entropy_loss.py
```diff
@@ -39,12 +39,12 @@ def celoss_indice_kernel(
...
+    ignore_mask = not (tgt == ignore_index) and tgt_mask
 
-    w_ptrs = w_ptr + tgt
-    w_tgt = tl.load(w_ptrs, mask=tgt_mask, other=0).to(tl.float32)
-    w_tgt_ptrs = w_tgt_ptr + pid_n * D + offset_d
-    tl.store(w_tgt_ptrs, w_tgt, mask=tgt_mask and ignore_mask)
+    w_tgt = tl.load(w_ptr + tgt, mask=ignore_mask, other=0).to(tl.float32)
+    if reduction == 1:
+        w_tgt_sum = tl.sum(w_tgt)
+        tl.atomic_add(w_tgt_ptr, w_tgt_sum)
```
iclementine: We can also sum the weight of each example in another kernel(sum_and_scale) to avoid using `atomic_add`. Atomic operations may hurt performance but we need some comparison to verify that.

iclementine: Or we can gather the weight for each example and write it out to an ((N,D) tensor. Then in sum_and_scale we can load them more effectively.

src/flag_gems/ops/cross_entropy_loss.py
```diff
@@ -67,7 +67,7 @@ def celoss_indice_kernel(
 
     out = (final_sum + final_max - inp_tgt) * w_tgt
     out_ptrs = out_ptr + pid_n * D + offset_d
-    tl.store(out_ptrs, out, mask=tgt_mask and ignore_mask)
+    tl.store(out_ptrs, out)
```
iclementine: Why not use mask here? Is it possible that the tile exceeds D dimension's size.

iclementine: Or it would accesss illegal memory.

src/flag_gems/ops/cross_entropy_loss.py
```diff
@@ -208,8 +210,9 @@ def celoss_indice_smooth_kernel(
         _sum += log * smooth * w[:, None]
 
     out = tl.sum(_sum, axis=0)
+    out = tl.where(ignore_mask, out, 0)
     out_ptrs = out_ptr + pid_n * D + offset_d
-    tl.store(out_ptrs, out, mask=tgt_mask and ignore_mask)
+    tl.store(out_ptrs, out)
```
iclementine: Also, I think masking is needed here, or it would access illegal memory.

src/flag_gems/ops/cross_entropy_loss.py
StrongSpoon: actually there is no effect. setting the default value as 0 is more elegant in my view ;)

src/flag_gems/ops/cross_entropy_loss.py
```diff
@@ -549,14 +615,14 @@ def backward(ctx, out_grad):
...
             )
         elif label_smoothing == 0:
-            celoss_indice_bwd[grid](
-                out_grad, inp, tgt, weight, inp_grad, ignore_index, mean_num, N, C, D
+            celoss_indices_bwd[grid](
+                out_grad, inp, tgt, weight, inp_grad, ignore_index, mean_num, C, D
             )
         else:
-            celoss_indice_smooth_bwd[grid](
+            celoss_indices_smooth_bwd[grid](
```
iclementine, line 659: We also need the gradient of weight, if it exists.

src/flag_gems/ops/cross_entropy_loss.py
```diff
@@ -488,47 +524,77 @@ def forward(ctx, inp, target, weight, reduction, ignore_index, label_smoothing):
...
+            ctx.save_for_backward(inp, tgt, weight)
+            ctx.N = N
+            ctx.C = C
+            ctx.D = D
+            ctx.ignore_index = ignore_index
+            ctx.label_smoothing = label_smoothing
+            ctx.shape = shape
+            ctx.mean_num = 1
+            if reduction == 1:
+                ctx.mean_num = N * D if tgt.ndim == dim else w_tgt.item()
```
iclementine: Maybe we can avoid `.item()` here, just save the tensor for backward as-is. Or it would conflict with cudagraph.

src/flag_gems/ops/cross_entropy_loss.py
StrongSpoon: replace parameter "ifscale" with "scalebyw" which means scale by weight

StrongSpoon: and scale is a data pointer when scalebyw is true, and scalar value while false.

src/flag_gems/ops/cross_entropy_loss.py
StrongSpoon: cross entropy loss doesn't accept weight requiring gradient.

src/flag_gems/ops/cross_entropy_loss.py
StrongSpoon: done.

src/flag_gems/ops/cross_entropy_loss.py
StrongSpoon: done

src/flag_gems/ops/cross_entropy_loss.py
StrongSpoon: done

src/flag_gems/ops/cross_entropy_loss.py
StrongSpoon: done

src/flag_gems/ops/cross_entropy_loss.py
StrongSpoon: done.

iclementine (dismissed):
LGTM

iclementine (approved):
LGTM

PR #271, opened 2024-11-01
Add Script to Calculate Summary Information for Benchmark Results

kiddyjinjin:
### Purpose
This PR introduces a new script to calculate summary information for benchmark results, aimed at providing insights into the average speedup for each operation categorized by data type.

### Changes Made
- Added a new script `summary_for_plot.py` that processes benchmark log files.
- Implemented functionality to calculate and display average speedups for different data types.
- Included conditional logic to handle query mode and provide backward compatibility notices.

### Usage
To generate summary information from benchmark logs, run the following command:
```bash
python summary_for_plot.py path_to_your_log_file.log


StrongSpoon (commented):
I suggest adding a shell script to ensure the order of operator benchmarks.

tongxin (approved):
LGTM. But can we also show speedups over input sizes?

kiddyjinjin:
> I suggest adding a shell script to ensure the order of operator benchmarks.

Since the command for generating benchmarks is quite flexible (it might specify op_name, a specific file, or simply run all operations without any specification), it would be challenging to pre-sort the operations to be benchmarked. However, sorting the operations in the benchmark results after the run is a great suggestion.

StrongSpoon (approved):
lgtm

Issue #275, opened 2024-11-05
[Benchmark Feature] Add Method to Calculate TFLOPS for BLAS Operations


kiddyjinjin:
### Description
Please update the implementation of the `get_tflops` function in `benchmark/test_blas_perf.py` under the `benchmark` directory. The function should be fully implemented to calculate TFLOPS accurately for the following operations:
- `addmm`
- `mm`
- `mv`
- `bmm`

### Requirements
- The `get_tflops` function should handle inputs for 2D, 3D, and 4D tensor scenarios.
- Ensure compatibility with the listed operations to provide a consistent TFLOPS calculation across different input dimensions and types.

shuailong616:
shuailong616 已认领

PR #290, opened 2024-11-14
blas tflops support

shuailong616:
### PR Category
 Benchmark 

### Type of Change
New Feature

### Description

The total flops is calculated by judging the shape information by the op name, the time is calculated by triton dobench, and then the TFLOPS is calculated, and finally the TFLOPS is printed out by modifying the attri_util.py
-------------------------------------------------
op mm
shape(m,n)(n,p)  total_flops= mxnx2k
-------------------------------------------------
op addmm
shape(m,n)(n,p)  total_flops= mxpx(2n+1)
-------------------------------------------------
op bmm
shape(b,n,m), (b,m,p) total_flops= bxnxpx2m
-------------------------------------------------
op mv
shape(n,m)(m,) total_flops =  n*2m
-------------------------------------------------
### Issue
#275 

### Performance
data


benchmark/attri_util.py
```diff
@@ -170,12 +170,20 @@ class BenchmarkResult:
     result: List[BenchmarkMetrics]
 
     def __str__(self) -> str:
-        header = (
-            f"\nOperator: {self.op_name}  Performance Test (dtype={self.dtype}, mode={self.mode}, level={self.level})\n"
-            f"{'Size':<10} {'Torch Latency (ms)':>20} {'Gems Latency (ms)':>20} {'Gems Speedup':>20}"
-            f"{'Size Detail':>20}\n"
-            f"{'-' * 90}\n"
-        )
+        if self.op_name in ["mv","mm","addmm","bmm"]:
```
kiddyjinjin: This code is tailored to specific operators, so it would require updates each time we add new ones. Since BenchmarkResult is a general data class, I suggest improving the implementation to avoid relying on specific operator names here.

benchmark/test_blas_perf.py
```diff
@@ -47,12 +50,33 @@ def set_more_shapes(self):
...
         """This method is currently not really implemented and serves as a placeholder.
         A proper implementation will be developed in the future."""
-        from torch.utils.flop_counter import FlopCounterMode
-
         fn = lambda: op(*args, **kwargs)
-        with FlopCounterMode(display=False) as flop_counter:
-            fn()
-        tflops = flop_counter.get_total_flops()
+        #time(s)
+        latency = triton.testing.do_bench(
```
kiddyjinjin: Since latency is already handled in performance_util.py, it would be more appropriate to calculate only total_flops here and then compute tflops in the base Benchmark class (using the latency result). The function name get_tflops may have caused some confusion, and we plan to rename it to clarify this.

shuailong616: Ok, I've noticed the problem, but I've calculated the time again because of the function name. I'll revise it next

benchmark/attri_util.py
shuailong616: ok

shuailong616 (commented):
solve conficts ,and use result value to choose is or no print tflops. latency is use result value

kiddyjinjin (approved):
fine

shuailong616:
pelase tell me some details, I don't konw why some model-test and op-unit-test no run

tongxin:
> pelase tell me some details, I don't konw why model-test and op-unit-test no run

It requires approval running CI for external PRs. Sometimes the requests just skipped our radar.

PR #291, opened 2024-11-14
fix min and max, use loop_kernel when reduction size is larger than 16384

iclementine:
fix min and max, use loop_kernel when reduction size is larger than 16384

### PR Category

Operator

### Description

Add an loop-style reduction kernel for min and max. Use the loop-style kernel when reduction size is larger than 16k, and choose a tile size <= 8k to avoid triton tensors larger than triton's limit.

### Progress


iclementine:
I close it since the changes are move to PR #360.

PR #295, opened 2024-11-18
[bugfix] mark eps as dns arg instead of constexpr

awayzjj:
### Description

same issue as https://github.com/FlagOpen/FlagGems/pull/235. norm_kernel and norm_bwd_kernel with eps as dns argument and constexpr at the same time doesn't work on triton 2.1.
since eps is not necessary for kernel compilation, here mark it as dns.

### Issue
Issue from Iluvatar

### Progress


StrongSpoon (approved):
lg! we should've done it TAT

PR #327, opened 2024-11-26
add program_id & num_programs that returns tensor in tl.int64 

iclementine:
### PR Category

Other

### Type of Change

Bug Fix

### Description

add program_id & num_programs that returns tensor in tl.int64 to avoid integer overflow

### Progress

### Performance

Offset calculating and indexing is triton is usually done in int32 if not coded carefully. For example:

```python
offset1 = pid1 * TILE_SIZE1 + tl.arange(0, TILE_SIZE1)
offsets = offset1 * stride1 + offset2 * stride2 + ...
x = tl.load(ptr + offsets)
```

It is possible that `offsets` exceeds the limit of int32 and overflows into some very large negative number, and that `ptr+ offsets` points to some illegal memory addresses.

Since most indexing code begins at `tl.program_id` and `tl.num_programs`, if we make them `int64`, consecutive indexing code would be in `int64`, thus avoids overflow.

It is also possible to avoid this by writing indexing code carefully

```python
x = tl.load(ptr + offset1 * stride1 + offset2 * stride2 + ...)
```

instead of

```python
x = tl.load(ptr + (offset1 * stride1 + offset2 * stride2 + ...))
```

But rewriting a lot of indexing code is tedious, so we add a module to collect some utility triton.jitted fucntions to do this.

Import the module by
`from flag_gems.utils import triton_lang_extension as tle`
and use `tle.program_id` and `tle.num_programs` instead of `tl.program_id` and `tl.num_programs` will work.

### DIFF
--- a/src/flag_gems/ops/masked_fill.py
+++ b/src/flag_gems/ops/masked_fill.py
@@ -5,6 +5,7 @@ import triton
 import triton.language as tl
 
 from ..utils import broadcastable_to, libentry
+from ..utils import triton_lang_extension as tle
 
 
 def cfggen():
@@ -20,7 +21,7 @@ def cfggen():
 @triton.autotune(configs=cfggen(), key=["N"])
 @triton.jit
 def masked_fill_kernel(inp, expand_mask, value, out, N, BLOCK_SIZE: tl.constexpr):
-    pid = tl.program_id(axis=0)
+    pid = tle.program_id(axis=0)
     offsets = pid * BLOCK_SIZE + tl.arange(0, BLOCK_SIZE)
     mask = offsets < N
 
@@ -36,7 +37,7 @@ def masked_fill_kernel(inp, expand_mask, value, out, N, BLOCK_SIZE: tl.constexpr
 @triton.autotune(configs=cfggen(), key=["N"])
 @triton.jit
 def masked_fill_kernel_self(inp, expand_mask, value, out, N, BLOCK_SIZE: tl.constexpr):
-    pid = tl.program_id(axis=0)
+    pid = tle.program_id(axis=0)
     offsets = pid * BLOCK_SIZE + tl.arange(0, BLOCK_SIZE)
     mask = offsets < N
 
--- a/src/flag_gems/ops/vector_norm.py
+++ b/src/flag_gems/ops/vector_norm.py
@@ -6,6 +6,7 @@ import triton
 import triton.language as tl
 
 from ..utils import dim_compress, libentry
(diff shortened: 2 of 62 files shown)


src/flag_gems/utils/triton_lang_extension.py
```diff
@@ -0,0 +1,28 @@
+"""
```
tongxin, line 1: Is there any reason why we don't shadow triton.language entirely?

iclementine: I don't want to hack `tl.program_id` or `tl.num_programs` since hacking may introduce some surprises.

src/flag_gems/ops/vector_norm.py
```diff
@@ -50,7 +51,7 @@ def l2_norm_kernel(X, Out, M, N, BLOCK_M: tl.constexpr, BLOCK_N: tl.constexpr):
 @libentry()
 @triton.jit
 def l2_norm_kernel_1(X, Mid, M, BLOCK_SIZE: tl.constexpr):
-    pid = tl.program_id(0)
+    pid = tle.program_id(0).to(tl.int64)
```
tongxin, line 54: Now its redundant upcast.

tongxin (dismissed):

tongxin (approved):
LGTM

src/flag_gems/ops/vector_norm.py
iclementine: I removed that now.

tongxin (approved):
LG

src/flag_gems/ops/masked_fill.py
```diff
@@ -33,8 +34,8 @@ def masked_fill_kernel(inp, expand_mask, value, out, N, BLOCK_SIZE: tl.constexpr
 @libentry()
 @triton.autotune(configs=cfggen(), key=["N"])
 @triton.jit
-def masked_fill_kernel_self(inp, expand_mask, value, N, BLOCK_SIZE: tl.constexpr):
-    pid = tl.program_id(axis=0)
+def masked_fill_kernel_self(inp, expand_mask, value, out, N, BLOCK_SIZE: tl.constexpr):
```
StrongSpoon, line 37: argument out was added by mistake. please remove it and submit a bug fix pr.

iclementine: I checkout from master from before masked_fill is merged, where there used to be an out parameter.

PR #332, opened 2024-11-27
[Runtime] solve the problem that llama frequently calls autotuner

StrongSpoon:
### PR Category

Runtime
### Type of Change

New Feature
### Description

implement LibTuner to take place of Autotuner.
it stores tuned configs in database and preloads them before calling.
### Progress

### DIFF
--- a/src/flag_gems/ops/log_softmax.py
+++ b/src/flag_gems/ops/log_softmax.py
@@ -10,8 +10,31 @@ from ..utils import libentry
 from ..utils import triton_lang_extension as tle
 
 
+def heur_block_m(args):
+    return triton.next_power_of_2(triton.cdiv(256, args["N"]))
+
+
+def heur_block_n(args):
+    return triton.next_power_of_2(args["N"])
+
+
+def heur_num_warps(args):
+    if args["N"] <= 1024:
+        return 4
+    elif args["N"] <= 2048:
+        return 8
+    else:
+        return 16
+
+
 @libentry()
-@triton.autotune(configs=runtime.get_triton_config("log_softmax"), key=["M", "N"])
+@triton.heuristics(
+    {
+        "BLOCK_M": heur_block_m,
+        "BLOCK_N": heur_block_n,
+        "num_warps": heur_num_warps,
+    }
+)
 @triton.jit
 def log_softmax_kernel(
     output_ptr,
--- a/src/flag_gems/ops/log_softmax.py
+++ b/src/flag_gems/ops/log_softmax.py
@@ -10,31 +10,7 @@ from ..utils import libentry
 from ..utils import triton_lang_extension as tle
 
 
-def heur_block_m(args):
-    return triton.next_power_of_2(triton.cdiv(256, args["N"]))
-
-
-def heur_block_n(args):
-    return triton.next_power_of_2(args["N"])
-
-
-def heur_num_warps(args):
-    if args["N"] <= 1024:
-        return 4
-    elif args["N"] <= 2048:
-        return 8
-    else:
-        return 16
-
-
 @libentry()
-@triton.heuristics(
-    {
-        "BLOCK_M": heur_block_m,
-        "BLOCK_N": heur_block_n,
-        "num_warps": heur_num_warps,
-    }
-)
 @triton.jit
 def log_softmax_kernel(
     output_ptr,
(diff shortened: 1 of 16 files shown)


src/flag_gems/ops/log_softmax.py
```diff
@@ -8,6 +8,10 @@
 from ..utils import triton_lang_extension as tle
 
 
+def heur_block_m(args):
+    return triton.next_power_of_2(triton.cdiv(256, args["N"]))
+
+
 def heur_block_n(args):
     return triton.next_power_of_2(args["N"])
```
tongxin: may result in large tiles.

src/flag_gems/utils/libentry.py
```diff
@@ -1,12 +1,130 @@
...
+            for k, v in cfg_ls[:-4]:
+                config.kwargs[k] = eval(v)
+            config.num_warps = int(cfg_ls[-4][1])
+            config.num_ctas = int(cfg_ls[-3][1])
+            config.num_stages = int(cfg_ls[-2][1])
+            config.enable_fp_fusion = eval(cfg_ls[-1][1])
+
+            self.cache[key] = config
+
+        connect.close()
```
tongxin: Probably we could leave the connection open and shared because we need to fetch stored params for all gems ops. A try catch block is possibly necessary for external IO operations.

tongxin: But I think its fine at this stage. Let's improve it later.

src/flag_gems/utils/code_cache.py
```diff
@@ -22,6 +22,18 @@ def cache_dir() -> Path:
...
 
 
+def code_cache_dir() -> Path:
+    _code_cache_dir = cache_dir() / "code_cache"
+    os.makedirs(_code_cache_dir, exist_ok=True)
+    return _code_cache_dir
+
+
+def config_cache_dir() -> Path:
+    _config_cache_dir = cache_dir() / "config_cache"
```
iclementine, line 32: Suggestion for name: "tunning_cache"

src/flag_gems/utils/libentry.py
```diff
@@ -1,12 +1,130 @@
...
+            restore_value,
+            pre_hook,
+            post_hook,
+            prune_configs_by,
+            warmup,
+            rep,
+            use_cuda_graph,
+        )
+        self.__name__ = self.base_fn.__name__
+        self.cache_dir = config_cache_dir() / f"{self.__name__}.db"
```
iclementine: It is not a dir

iclementine: I prefer a single database for all operators thta uses autotuner.

StrongSpoon: done

src/flag_gems/utils/libentry.py
```diff
@@ -1,12 +1,149 @@
...
+            if major_version == 2:
+                config.enable_warp_specialization = eval(cfg_ls[attrs + 3][1])
+                config.enable_persistent = eval(cfg_ls[attrs + 4][1])
+            else:
+                config.maxnreg = eval(cfg_ls[attrs + 3][1])
+
+            self.cache[tuple(key)] = config
+
+        connect.close()
+
```
kiddyjinjin, line 98: I think the trigger conditions for `store` should be more specific. it would be better to trigger an `update_db` after each real `autotune` . This approach allows shapes encountered during real model runs to be updated in the database.

src/flag_gems/ops/log_softmax.py
StrongSpoon: update to master branch

src/flag_gems/utils/libentry.py
StrongSpoon: done. add length check before store.

src/flag_gems/utils/libentry.py
StrongSpoon: there is not always a explicit exit of flag_gems. I'm not sure what time to close the connection if keep it open.

src/flag_gems/utils/code_cache.py
StrongSpoon: what's the difference?

kiddyjinjin (approved):

PR #335, opened 2024-11-28
support zero-sized tensor for PointwiseDynamic

iclementine:
### PR Category

Other

### Type of Change

Bug Fix

### Description

Support zero-sized tensor for Pointwise Dynamic.
Previously, pointwise functions would launch a kernel to do the job. The task partition logic involves a function that try to find a tile size with numel limit and prefer larger tile size for inner dimensions.

```python
def heuristics_for_tile_size(max_tile_size, *sizes):
    ndim = len(sizes)
    tile_sizes = [0 for _ in range(ndim)]
    for i in range(ndim):
        size = sizes[ndim - 1 - i]
        tile_size = min(max_tile_size, triton.next_power_of_2(size))
        tile_sizes[ndim - 1 - i] = tile_size
        max_tile_size = max(1, max_tile_size // tile_size)
    return tuple(tile_sizes)
```

It has a lot of divisions, when the size is 0, it would raise `DivisionByZero`s.

We fix it by skipping kernel launch and task partitioning when the number of tasks(numel of the brodcasted shape) is zero.

For example

```python
ef axpy_wrapper_rank_1(in0: Union[torch.Tensor, StridedBuffer], in1: Union[torch.Tensor, StridedBuffer], val0, /, *, out0: Union[torch.Tens\
or, StridedBuffer]):
    """Generated wrapper function with Pointwise: StridedBuffer, StridedBuffer, scalar, StridedBuffer(a1!) -> StridedBuffer(a1!)"""
    assert in0.shape == in1.shape == out0.shape, 'operand shapes mismatch'
    # task partitioning
    shape = out0.shape
    num_tasks = out0.numel()
    if num_tasks == 0:
        return out0
    #  task partitioning & kernl launch here ...
    return out0
```

### Issue

https://github.com/FlagOpen/FlagGems/issues/302

### Progress

### DIFF
--- a/src/flag_gems/utils/pointwise_dynamic.py
+++ b/src/flag_gems/utils/pointwise_dynamic.py
@@ -830,6 +830,9 @@ class WrapperGenerator:
         else:
             code.writeline("shape = out0.shape")
             code.writeline("num_tasks = out0.numel()")
+            code.writeline("if num_tasks == 0:")
+            with code.indent():
+                self.gen_return(code)
             max_tile_size = self.config.max_tile_size
             code.writeline(
                 f"tile_sizes = heuristics_for_tile_size({max_tile_size}, *shape)"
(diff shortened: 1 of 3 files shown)


tongxin (approved):
LG

PR #336, opened 2024-11-29
Remove INT_DTYPES from isfinite operation in benchmark and tests

kiddyjinjin:
### PR Category

Benchmark

### Type of Change

Other

### Description
This PR removes the benchmarking for `INT_DTYPE` in the `isfinite` operator's performance tests. 
since PyTorch does not support `inf` or `nan` values for integer types, the `isfinite` function for `INT_DTYPE` will always return `True`. As a result, testing this data type is unnecessary and does not provide meaningful information.


Bowen12992 (approved):
LGTM

PR #337, opened 2024-11-29
[bugfix] replace tle with tl for operators using philox

StrongSpoon:
### PR Category

Operator
### Type of Change

Bug Fix
### Description

remove tle for those operators using philox.
uniform function doesn't support int64 as offset temporarily.
### Progress

### DIFF
--- a/src/flag_gems/utils/libentry.py
+++ b/src/flag_gems/utils/libentry.py
@@ -122,7 +122,11 @@ class LibEntry(triton.KernelInterface):
                         raise RuntimeError("Invalid Runtime Function")
                     fn = fn.fn
                 for p in self.jit_function.params:
-                    if p.is_constexpr and p.name not in constexprs:
+                    if (
+                        p.is_constexpr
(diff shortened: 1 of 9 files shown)


src/flag_gems/utils/libentry.py
```diff
@@ -122,7 +122,11 @@ def run(self, *args, **kwargs):
                         raise RuntimeError("Invalid Runtime Function")
                     fn = fn.fn
                 for p in self.jit_function.params:
-                    if p.is_constexpr and p.name not in constexprs:
+                    if (
+                        p.is_constexpr
+                        and p.name not in constexprs
+                        and (p.default is not inspect._empty)
+                    ):
```
iclementine, line 129: Did we trigger any error because of this?

StrongSpoon: yes. when we pass constexpr as a key-word argument, it will be covered by the default empty in metadata, which causes error in grid function.

PR #338, opened 2024-11-29
[Operator] Fix full & full_like op

zhzhcookie:
### PR Category
Operator

### Type of Change
Bug Fix

### Description
1. When dtype is not provided, dtype should be set according to fill_value.
2. When dtype is provided, fill_value should be check for overflow.
3. If dtype is not bool and fill_value is bool, fill_value should be converted to int.
4. If dtype is float, the truncation of fill_value should be noted.
5. The default device should be "cpu".

### Progress

### DIFF
--- a/src/flag_gems/ops/full.py
+++ b/src/flag_gems/ops/full.py
@@ -4,35 +4,75 @@ import torch
 import triton
 import triton.language as tl
 
-from flag_gems.utils.shape_utils import volume
-
 from ..utils import triton_lang_extension as tle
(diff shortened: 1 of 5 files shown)


iclementine (approved):
LGTM, triton jit functions treat all float argument, we have to work around it by wrapping a value into a tensor that can hold it.

Bowen12992 (dismissed):
LGTM

Bowen12992 (approved):

PR #339, opened 2024-12-02
Add cummin op

awayzjj:
### Issue
Closes: https://github.com/FlagOpen/FlagGems/issues/321

### Progress

### Performance

On NV A10:
image
update(tl.load with inf, use tle.minimum_with_index_tie_break_right as combine_fn)

image
image

update(changes after disscution with iclementine, tl_min with global_index rather than offset, use tle.program_id):
image
image


src/flag_gems/ops/cummin.py
```diff
@@ -0,0 +1,341 @@
...
+    else:
+        inp_vals = inp_vals.to(tl.float32)
+    result, cummin_indices = tl_cummin(inp_vals, offset, axis=0)
+
+    # tl.min do not support min_indices_tie_break_right
+    part_min_via_min, part_min_indices_via_min = tl_min_tie_break_right(
+        inp_vals, axis=0
+    )
+    # part_min_via_min = result[BLOCK_SIZE - 1]
+    # part_min_indices_via_min = indices[BLOCK_SIZE - 1]
```
awayzjj: Why I can not index from `result`, I want to get `part_min_via_min` directly from `result` rather than get min from `inp_vals`
image

StrongSpoon:
triton 2.2 doesn't support associative_scan with two input tensors. operator cummax also faces this problem. we are trying to solve it.

src/flag_gems/ops/cummin.py
iclementine: Triton does not support indexing.

src/flag_gems/ops/cummin.py
```diff
@@ -0,0 +1,341 @@
...
+
+from ..utils import libentry
+
+
+@triton.jit
+def _min_combine(a, idxa, b, idxb):
+    if a < b:
+        return a, idxa
+    else:
+        return b, idxb
```
iclementine: We can use tl.where here to avoid divergence.

```python
mask = a < b
out = tl.where(mask, a, b)
index = tl.where(mask, idxa, idxb)
```

iclementine: Or we can adopt from inductor here since it handles Nans in a desired way.

```python
@triton.jit
def minimum_with_index(a_value, a_index, b_value, b_index):
    mask = a_value < b_value
    equal = a_value == b_value
    if is_floating(a_value):
        a_isnan = a_value != a_value
        b_isnan = b_value != b_value
        mask |= a_isnan and not b_isnan
        # Consider NaNs as equal
        equal |= a_isnan and b_isnan

    # Prefer lowest index if values are equal
    mask |= equal & (a_index < b_index)
    return tl.where(mask, a_value, b_value), tl.where(mask, a_index, b_index)
```

src/flag_gems/ops/cummin.py
```diff
@@ -0,0 +1,341 @@
...
+    # TODO(all): tune on target board
+    BLOCK_SIZE = 1024
+    if n_ele <= 1024 * 4:
+        BLOCK_SIZE = triton.next_power_of_2(n_ele)
+    part_num = math.ceil(n_ele / BLOCK_SIZE)
+    partial_min = torch.empty(part_num, dtype=dtype, device=inp.device)
+    partial_min_indices = torch.empty(part_num, dtype=torch.int64, device=inp.device)
+
+    grid = (part_num,)
+    with torch.cuda.device(inp.device):
```
iclementine, line 121: Maybe we can specialize a kernel for `part_num==1` from the beginning, where recursion, `partial_min` and `partial_min_indices` are not needed

src/flag_gems/ops/cummin.py
iclementine: We can add it to utis/triton_language_extension.py

src/flag_gems/ops/cummin.py
```diff
@@ -0,0 +1,341 @@
...
+    partial_min_indices,
+    n_elements,
+    BLOCK_SIZE: tl.constexpr,
+):
+    pid = tl.program_id(0)
+    offset = pid * BLOCK_SIZE + tl.arange(0, BLOCK_SIZE)
+    mask = offset < n_elements
+
+    inp_ptrs = inp + offset
+    inp_vals = tl.load(inp_ptrs, mask=mask)
```
iclementine: Since this kernel for cummin, I think inp_vales should be loaded with padding value `inf` to avoid wrong result.

src/flag_gems/ops/cummin.py
awayzjj: OK, I will change `minimum_with_index` a little, since the cummin need highest rather than lowest index if values are equal

src/flag_gems/ops/cummin.py
awayzjj: Do it will help to improve the performance?

src/flag_gems/ops/cummin.py
awayzjj: Got it!

awayzjj:
@iclementine  I have already made some changes according to your suggestions, please review my PR again, thank you!

src/flag_gems/ops/cummin.py
iclementine: Yes, you are right

src/flag_gems/ops/cummin.py
```diff
@@ -0,0 +1,334 @@
...
+import triton.language as tl
+import triton.language.core as core
+
+from ..utils import libentry
+from ..utils import triton_lang_extension as tle
+
+
+@triton.jit
+def tl_cummin(input, index, axis=0):
+    return core.associative_scan(
```
iclementine: Please use `tl.associative_scan` to access this API.

src/flag_gems/ops/cummin.py
```diff
@@ -0,0 +1,334 @@
...
+        (input, index), axis, tle.minimum_with_index_tie_break_right
+    )
+
+
+@triton.jit
+def _argmin_combine_tie_break_right(value1, index1, value2, index2):
+    tie = value1 == value2 and index1 > index2
+    lt = value1 < value2 or tie
+    value_ret = core.where(lt, value1, value2)
```
iclementine: Also for reduce, where, etc, use `tl.xxx` to access them.

src/flag_gems/ops/cummin.py
```diff
@@ -0,0 +1,334 @@
...
+
+@triton.jit
+def tl_cummin(input, index, axis=0):
+    return core.associative_scan(
+        (input, index), axis, tle.minimum_with_index_tie_break_right
+    )
+
+
+@triton.jit
+def _argmin_combine_tie_break_right(value1, index1, value2, index2):
```
iclementine: Is thi redundant when `tle.minimum_with_index_tie_break_right` is used？

src/flag_gems/ops/cummin.py
```diff
@@ -0,0 +1,334 @@
...
+    if (
+        tl.constexpr(inp_vals.dtype.is_int64())
+        or tl.constexpr(inp_vals.dtype.is_uint64())
+    ) or tl.constexpr(inp_vals.dtype.is_fp64()):
+        inp_vals = inp_vals
+    elif tl.constexpr(inp_vals.dtype.is_int()):
+        inp_vals = inp_vals.to(tl.int32)
+    else:
+        inp_vals = inp_vals.to(tl.float32)
+    result, cummin_indices = tl_cummin(inp_vals, offset, axis=0)
```
iclementine: I have an idea to make the code easier to read and potentially improves the performance.

Let's take `M=N=1` as an example, the digram of kernels is 


There `scan_part_min_kernel` maybe called `N` times and util ther is only one block in the grid. Then call `scan_part_min_kernel` `N-1` times in a reverse order(the recusrsion here).

After each call to scan_part_min_kernel, you get an output(cummin values  & indices）whose size has been reduced by a factor of `1024`). But the indices are the indices at the original array.

Since you have used `reducetion` and `scan` with a combiner that takes care of the index, it may not be a arange from 0 to some extent. It is like a kind of argmin.

So I suggest not using `offset` in `tl_cummin(inp_vals, offset, axis=0)`. Instead load the input indices and `tl_cummin(inp_vals, input_indices, axis=0)`. Since after each call to this kernel, the input is shrinked, and the indices are not contiguous.

Loading the input indices maks this more staright-forward, without needing to reconstruct `true_indices` beblow. I believe that the reason why you need to reconstruct true_out_indices is that the `indices` used in `tl_cummin`ares not the indices in the original indices, instead, there are contiguous `offsets` in the new kernel. 

In a consequence, you have to load from `out_indices` (which is used also as input indices when you load from them). But as the `cummin_indices` is the result of `tl_cummin`, indirect memory access is needed, and the resulting addresses to load from are not single strided view.

awayzjj:
@iclementine Please review my changes again, thanks!

iclementine (approved):

iclementine (approved):
LGTM

PR #361, opened 2024-12-13
muti_backend_part_3 enhancement of ability

Galaxy1458:
### PR Category
Other

### Type of Change
New Feature 

### Description
Enhance the expression ability of operator library and user interaction ability

```python
import flag_gems as gems

 # Gets the names of all registered operators
gems.all_ops() 

#The user specifies that some operators are disabled
gems.use_gems(unused=['xxx'])
gems.enable(unused=['xxx'])
```

### Progress

### DIFF
--- a/src/flag_gems/__init__.py
+++ b/src/flag_gems/__init__.py
@@ -10,12 +10,13 @@ from .runtime.register import Register
 __version__ = "2.1"
 device = runtime.device.name
 aten_lib = torch.library.Library("aten", "IMPL")
+registar = Register
+current_work_registar = None
 
 
-def enable(lib=aten_lib, unused=None):
-    if unused is None:
-        unused = []
-    Register(
+def enable(lib=aten_lib, unused=None, registar=registar):
+    global current_work_registar
+    current_work_registar = registar(
         (
             ("abs", abs, Autograd.disable),
             ("add.Tensor", add, Autograd.disable),
@@ -201,20 +202,38 @@ def enable(lib=aten_lib, unused=None):
             ("logical_xor", logical_xor, Autograd.disable),
             ("logical_not", logical_not, Autograd.disable),
         ),
-        user_unused_ops_list=unused,
+        user_unused_ops_list=[] if unused is None else unused,
         lib=lib,
     )
 
 
 class use_gems:
-    def __init__(self):
+    def __init__(self, unused=None):
         self.lib = torch.library.Library("aten", "IMPL")
+        self.unused = [] if unused is None else unused
+        self.registar = Register
 
     def __enter__(self):
-        enable(lib=self.lib)
+        enable(lib=self.lib, unused=self.unused, registar=self.registar)
 
     def __exit__(self, exc_type, exc_val, exc_tb):
+        global current_work_registar
         del self.lib
+        del self.unused
+        del self.registar
+        del current_work_registar
+
+
+def only_forward_ops():
+    return current_work_registar.get_forward_ops()
+
+
+def has_backward_ops():
+    return current_work_registar.get_backward_ops()
+
+
+def all_ops():
+    return current_work_registar.get_all_ops()
 
 
 __all__ = [
--- a/src/flag_gems/__init__.py
+++ b/src/flag_gems/__init__.py
@@ -10,13 +10,13 @@ from .runtime.register import Register
 __version__ = "2.1"
 device = runtime.device.name
 aten_lib = torch.library.Library("aten", "IMPL")
-registar = Register
-current_work_registar = None
+registrar = Register
+current_work_registrar = None
 
 
-def enable(lib=aten_lib, unused=None, registar=registar):
-    global current_work_registar
-    current_work_registar = registar(
+def enable(lib=aten_lib, unused=None, registrar=registrar):
+    global current_work_registrar
+    current_work_registrar = registrar(
         (
             ("abs", abs, Autograd.disable),
             ("add.Tensor", add, Autograd.disable),
@@ -211,29 +211,21 @@ class use_gems:
     def __init__(self, unused=None):
         self.lib = torch.library.Library("aten", "IMPL")
         self.unused = [] if unused is None else unused
-        self.registar = Register
+        self.registrar = Register
 
     def __enter__(self):
-        enable(lib=self.lib, unused=self.unused, registar=self.registar)
+        enable(lib=self.lib, unused=self.unused, registrar=self.registrar)
 
     def __exit__(self, exc_type, exc_val, exc_tb):
-        global current_work_registar
+        global current_work_registrar
         del self.lib
         del self.unused
-        del self.registar
-        del current_work_registar
-
-
-def only_forward_ops():
-    return current_work_registar.get_forward_ops()
-
-
-def has_backward_ops():
-    return current_work_registar.get_backward_ops()
+        del self.registrar
+        del current_work_registrar
 
 
 def all_ops():
-    return current_work_registar.get_all_ops()
+    return current_work_registrar.get_all_ops()
 
 
 __all__ = [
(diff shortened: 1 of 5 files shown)


src/flag_gems/__init__.py
```diff
@@ -10,12 +10,13 @@
 __version__ = "2.1"
 device = runtime.device.name
 aten_lib = torch.library.Library("aten", "IMPL")
+registar = Register
```
StrongSpoon: register

src/flag_gems/__init__.py
```diff
@@ -201,20 +202,38 @@ def enable(lib=aten_lib, unused=None):
...
 
     def __exit__(self, exc_type, exc_val, exc_tb):
+        global current_work_registar
         del self.lib
+        del self.unused
+        del self.registar
+        del current_work_registar
+
+
+def only_forward_ops():
```
StrongSpoon: I think users don't need these functions in deployment.

src/flag_gems/runtime/backend/backend_utils.py
```diff
@@ -13,10 +13,20 @@ class VendorInfoBase:
...
 
 def get_tune_config(vendor_name, file_mode="r"):
-    vendor_name = "_" + vendor_name
-    script_path = os.path.abspath(__file__)
-    file_path = os.path.dirname(script_path) + "/" + vendor_name + "/tune_configs.yaml"
-    with open(file_path, file_mode) as file:
-        config = yaml.safe_load(file)
+    try:
+        if not vendor_name.isalnum():
+            raise ValueError("Invalid vendor name")
```
StrongSpoon: function is called inside runtime code, so vendor_name can hardly be invalid.

src/flag_gems/runtime/backend/backend_utils.py
```diff
@@ -13,10 +13,20 @@ class VendorInfoBase:
...
+        vendor_name = "_" + vendor_name
```
StrongSpoon, line 27: why prefix an underline?

src/flag_gems/runtime/register.py
```diff
@@ -60,20 +62,25 @@ def register_impl(self, key, fn, has_backward):
...
-    def _set_info(self, config):
-        for _, fn, hasbackward in config:
+    def _set_info(self):
+        for _, fn, hasbackward in self.config:
             fn_name = fn.__name__
             self.backward_ops.append(
                 fn_name
-            ) if hasbackward else self.forward_ops.append(fn_name)
+            ) if hasbackward is commom_utils.Autograd.enable else self.forward_ops.append(
+                fn_name
```
StrongSpoon: change this expression to a common conditional one.

src/flag_gems/runtime/register.py
```diff
@@ -60,20 +62,25 @@ def register_impl(self, key, fn, has_backward):
...
+            )
+
+    def get_all_ops(self):
+        return self.forward_ops + self.backward_ops if self.debug else []
```
StrongSpoon: just return them all.

src/flag_gems/runtime/register.py
```diff
@@ -60,20 +62,25 @@ def register_impl(self, key, fn, has_backward):
...
 
     def get_forward_ops(self):
         return self.forward_ops if self.debug else []
```
StrongSpoon: ditto

src/flag_gems/runtime/register.py
```diff
@@ -1,37 +1,39 @@
...
         self.debug = debug
         self.forward_ops = []
         self.backward_ops = []
-        self.config = config
         self.vendor_extend_configs = self.get_vendor_extend_op()
         self.vendor_unused_ops_list = self.get_vendor_unused_op()
         self.unused_ops = user_unused_ops_list + self.vendor_unused_ops_list
-        self.check_backend()
-        self.for_each(config)
+        self.config = config + self.vendor_extend_configs
```
StrongSpoon, line 23: do we allow vendors to have their own operators?

src/flag_gems/runtime/register.py
```diff
@@ -1,37 +1,39 @@
...
         user_unused_ops_list=None,
         lib=None,
-        debug=False,
+        debug=True,
     ):
         # lib is a instance of torch.library.Library
-        self.lib = lib
         self.device = DeviceDetector()
+        self.vendor_list = list(commom_utils.vendors_map)
+        self.check_backend()
```
StrongSpoon: I don't think it's necessary. as long as Register successfully initializes self.device, backend won't be wrong.

src/flag_gems/__init__.py
```diff
@@ -10,12 +10,13 @@
...
+registar = Register
+current_work_registar = None
 
 
-def enable(lib=aten_lib, unused=None):
-    if unused is None:
-        unused = []
-    Register(
+def enable(lib=aten_lib, unused=None, registar=registar):
+    global current_work_registar
```
StrongSpoon: why do we need a global variable to keep the register?

StrongSpoon (commented):
done

src/flag_gems/__init__.py
Galaxy1458: Actually want to use registrar， and registar is actually wrong

src/flag_gems/__init__.py
Galaxy1458: 'current_work_registrar' is used to indicate the current Register context and to provide the all_ops() method using the 'current_work_registrar' object. Otherwise we don't know whether the Register object is from enable() or use_gems()

src/flag_gems/__init__.py
Galaxy1458: removed

src/flag_gems/runtime/register.py
Galaxy1458: Yes，this is for 
1. if the implementation of large sections of its code is different from that of nv
2. An operator with good performance for a vendor but low performance on nv

src/flag_gems/runtime/register.py
Galaxy1458: removed

src/flag_gems/runtime/register.py
Galaxy1458: double check, Prevents user-transmitted vendor errors

src/flag_gems/runtime/register.py
Galaxy1458: Done

src/flag_gems/runtime/backend/backend_utils.py
Galaxy1458: removed

src/flag_gems/runtime/backend/backend_utils.py
Galaxy1458: Distinguish from 'nvidia library'

src/flag_gems/runtime/register.py
StrongSpoon: I got it. it actually means the specific implementation by vendors, instead of extension. I suggest rename 'op's and 'config's carefully.

src/flag_gems/__init__.py
StrongSpoon: theoretically there can't be enable() and use_gems() at the same time. we don't have to know where Register is from.

src/flag_gems/runtime/backend/backend_utils.py
StrongSpoon: what do you want to prevent the misunderstanding of?

src/flag_gems/runtime/register.py
StrongSpoon: isn't vendor detected from backend? why permit user to transmit it?

StrongSpoon (commented):
done

StrongSpoon (approved):
lgtm

PR #373, opened 2024-12-20
[Polish code]

Galaxy1458:
### Progress


src/flag_gems/utils/triton_lang_helper.py
```diff
@@ -0,0 +1,17 @@
+import triton
```
iclementine, line 1: I think it is better to import `tl_extra_shim` from this module into `flag_gems.utils`, so we can use it with a shallower path.

`from triton_lang_helper import tl_extra_shim`

iclementine (approved):
LGTM

PR #380, opened 2024-12-30
add a cpp extension

iclementine:
### PR Category

Other

### Type of Change

Other

### Description

Add a cpp extension(flag_gems._C) to flag_gems.

It finds torch include dirs & libraries and also cuda include dirs & libraries(if necessary).

### Progress

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -9,7 +9,7 @@ from torch.utils.cpp_extension import BuildExtension, CppExtension
 # ----------------------------- check triton -----------------------------
 # NOTE: this is used to check whether pytorch-triton or triton is installed. Since
 # the name for the package to be import is the name, but the names in package manager
-# are different. So we check it in this way
+# are different. So we check it in this way:
 # 1. If the triton that is installed via pytorch-triton, then it is the version that is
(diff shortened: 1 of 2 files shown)


setup.py
```diff
@@ -62,6 +71,58 @@ def _is_package_installed(package_name: str) -> bool:
...
+library_name = "flag_gems"
+
+
+def get_extensions():
+    debug_mode = os.getenv("DEBUG", "0") == "1"
+    use_cuda = os.getenv("USE_CUDA", "1") == "1"
+    if debug_mode:
+        print("Compiling in debug mode")
+
+    use_cuda = use_cuda and torch.cuda.is_available() and CUDA_HOME is not None
```
Galaxy1458: Expect using torch_device_fn.is_available() instead of  torch.cuda.is_available() 
and torch_device_fn from（from ..runtime import torch_device_fn）

iclementine: Since it is a part of `flag_gems`, it may not use the utilities provided by `flag_gems`. So I removed cuda extensions since it is planned to use c++-only extensions here.

iclementine:
New PR to add operators with triton jit c++ runtime has been reworked in PR 531

PR #400, opened 2025-01-06
Add details in tests

tianxiao-baai:
### Description
add pytest detail output when run tests by pytest hook


tests/conftest.py
```diff
@@ -96,3 +101,31 @@ def pytest_runtest_teardown(item, nextitem):
...
+
+    test_results[item.nodeid]['params'] = param_values
+
+@pytest.hookimpl(tryfirst=True)
+def pytest_runtest_logreport(report):
+    if report.when == "call":
+        test_results[report.nodeid]['result'] = report.outcome
+
+def pytest_terminal_summary(terminalreporter):
+    if os.path.exists('result.json'):
```
tongxin: Shall we use a less general file name? And is the file supposed to be found in the root of repos?

tianxiao-baai: How about correctness.result? The content of the file is the result of execution. It is recommended to place it under the tests directory.

tongxin: I would suggest the output files be able to set by commandline args or be configurable so that different output files can be created for different runs.

tianxiao-baai: Okay, I will consider using command line methods in my future implementations.

Bowen12992 (approved):

PR #406, opened 2025-01-08
[Operator]update metax custom op

Alvin-YCHEN:
update metax custom op

### DIFF
--- a/src/flag_gems/runtime/backend/_metax/__init__.py
+++ b/src/flag_gems/runtime/backend/_metax/__init__.py
@@ -1,0 +1,20 @@
+from backend_utils import VendorInfoBase  # noqa: E402
+
+from .heuristics_config_utils import HEURISTICS_CONFIGS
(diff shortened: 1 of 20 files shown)


src/flag_gems/utils/codegen_config_utils.py
```diff
@@ -38,3 +45,20 @@ def get_codegen_config():
     if device.vendor not in CODEGEN_COFIGS:
         return CODEGEN_COFIGS.get(vendors.NVIDIA)
     return CODEGEN_COFIGS.get(device.vendor)
+
+
+def get_heuristics_for_num_warps(tile_size):
+    if device.vendor == vendors.METAX:
```
Galaxy1458: It is suggested that write it as a function. And i happen to have a PR here, I changed it easily, and then pull the latest code.

Galaxy1458 (approved):

PR #411, opened 2025-01-12
[fix] rewrite do_bench with fixed warmup/rep runs mode

ConvolutedDog:
### PR Category

Benchmark

### Type of Change

Bug Fix

### Description

In the original implementation of triton's `do_bench` function, when users specify the `warmup` and `rep` parameters, the behavior may not align with their expectations. This is because:

- In `triton.testing.do_bench`, the `warmup` and `rep` parameters are in milliseconds instead of reps, so that triton dynamically adjusts the number of warmup and repetition runs. And this is inconsistent with torch's measurement.

Therefore, this PR rewrites the `triton.testing.do_bench` function to support a new `fixed_warmup_rep_runs` parameter, which allows users to choose between two modes:

1. **Dynamic Mode** (the original implementation of `triton.testing.do_bench`): Estimates the runtime of the kernel and dynamically adjusts the number of warmup and repetition runs based on the provided `warmup` and `rep` times (in milliseconds).
2. **Fixed Mode** (default in this rewritten version, and consistent with torch's testing): Uses the provided `warmup` and `rep` values directly as the number of warmup and repetition runs.

### Issue

None

### Progress

### Performance

No need

### DIFF
--- a/benchmark/performance_utils.py
+++ b/benchmark/performance_utils.py
@@ -55,6 +55,92 @@ def SkipVersion(module_name, skip_pattern):
         return (major, minor) > (M, N)
 
 
+def triton_testing_do_bench_rewritting(fn, warmup=25, rep=100, grad_to_none=None,
+                                       quantiles=None, fast_flush=True, return_mode="mean",
+                                       device_type="cuda", fixed_warmup_rep_runs=True):
+    """
+    This is a rewritten version of the original `triton.testing.do_bench` function.
+
+    Benchmark the runtime of the provided function. By default, return the median runtime
+    of :code:`fn` along with the 20-th and 80-th performance percentile.
+
+    This function supports two modes for determining the number of warmup and repetition
+    runs, by appending a parameter called `fixed_warmup_rep_runs`:
+    1. Dynamic Mode (the original implementation of `triton.testing.do_bench`):
+       Estimates the runtime of the kernel and dynamically adjusts the number of warmup and
+       repetition runs based on the provided `warmup` and `rep` times (in milliseconds).
+    2. Fixed Mode (default in this rewritten version, and consistent with torch's testing):
+       Uses the provided `warmup` and `rep` values directly as the number of warmup and
+       repetition runs.
+
+    Please refer to the original implementation of `triton.testing.do_bench` function for
+    more details:
+    https://github.com/triton-lang/triton/blob/199fd8a239068318e94d39843c4c676f44883bd3/python/triton/testing.py#L162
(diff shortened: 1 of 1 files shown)


StrongSpoon:
thanks for your idea and contribution! I recommend reformat the code according to CONTRIBUTING.md to pass the checks. we will review it soon.

ConvolutedDog:
@StrongSpoon Thank you for your feedback! I've reformatted the code according to CONTRIBUTING.md. Please let me know if there's anything else that needs to be adjusted.

kiddyjinjin:
It's great that you spotted the difference in the warmup and rep parameters between our understanding and the actual implementation—and fixed it!
However, I have a small concern: since we are in the process of rewriting the function, we might need to continuously track changes in the original implementation and adjust our version accordingly. (I noticed that the latest `do_bench` in Triton has been modified to reduce runtime dependencies on Torch.)

benchmark/performance_utils.py
```diff
@@ -55,6 +54,102 @@ def SkipVersion(module_name, skip_pattern):
...
+def triton_do_bench_rewritten(
+    fn,
+    warmup=25,
+    rep=100,
+    grad_to_none=None,
+    quantiles=None,
+    fast_flush=True,
+    return_mode="mean",
+    device_type="cuda",
+    fixed_warmup_rep_runs=True,
```
kiddyjinjin, line 66: Set `fixed_warmup_rep_runs` default to `False` for backward compatibility.

benchmark/performance_utils.py
```diff
@@ -55,6 +54,102 @@ def SkipVersion(module_name, skip_pattern):
...
+    2. Fixed Mode (default in this rewritten version, and consistent with torch's testing):
+       Uses the provided `warmup` and `rep` values directly as the number of warmup and
+       repetition runs.
+
+    Please refer to the original implementation of `triton.testing.do_bench` function for
+    more details:
+    https://github.com/triton-lang/triton/blob/199fd8a239068318e94d39843c4c676f44883bd3/python/triton/testing.py#L162
+    """
+
+    assert return_mode in ["min", "max", "mean", "median"]
```
kiddyjinjin, line 88: Sorry for reaching out after so long. Could you please update this PR so that when `fixed_warmup_rep_runs=False`, it directly calls the original `triton.testing.do_bench` function (instead of duplicating its core logic)? When it's True, keep the current logic in your PR. This avoids the need to manually sync updates from Triton's implementation in the future.

PR #413, opened 2025-01-12
[Fix InstanceNorm] Fix a memory out-of-bound access

mark14wu:
### PR Category

Operator

### Type of Change

Bug Fix

### Description

This change fixes potential out-of-bound access in the following two triton kernels by adding a missing `row_mask` to the `tl.load` operations for `Mean` and `Rstd`. Default values are provided (mean=0.0, rstd=1.0) to handle masked-out elements.
#### `instance_norm_backward_kernel`
```diff
- mean = tl.load(Mean).to(tl.float32)
- rstd = tl.load(Rstd).to(tl.float32)
+ mean = tl.load(Mean, mask=row_mask, other=0.0).to(tl.float32)
+ rstd = tl.load(Rstd, mask=row_mask, other=1.0).to(tl.float32)
```
#### `weight_bias_backward_kernel`
```diff
- mean = tl.load(Mean + mid).to(tl.float32)
- rstd = tl.load(Rstd + mid).to(tl.float32)
+ mean = tl.load(Mean + mid, mask=row_mask).to(tl.float32)
+ rstd = tl.load(Rstd + mid, mask=row_mask).to(tl.float32)
```
p.s. A similar issue is found in #414. 
### Issue
N/A.

### Progress

### Performance

N/A

### DIFF
--- a/src/flag_gems/ops/instancenorm.py
+++ b/src/flag_gems/ops/instancenorm.py
@@ -332,6 +332,8 @@ def instance_norm_backward_kernel(
 ):
     pid = tl.program_id(0) * BLOCK_ROW_SIZE + tl.arange(0, BLOCK_ROW_SIZE)[:, None]
     c_offsets = pid % C
+    if pid >= M:
+        return
     row_mask = pid < M
     dY += pid * N
     X += pid * N
@@ -339,8 +341,8 @@ def instance_norm_backward_kernel(
     Mean += pid
     Rstd += pid
 
-    mean = tl.load(Mean, mask=row_mask, other=0.0).to(tl.float32)
-    rstd = tl.load(Rstd, mask=row_mask, other=1.0).to(tl.float32)
+    mean = tl.load(Mean).to(tl.float32)
+    rstd = tl.load(Rstd).to(tl.float32)
     if HAS_WEIGHT_BIAS:
         w = tl.load(W + c_offsets, mask=row_mask).to(tl.float32)
     else:
--- a/src/flag_gems/ops/instancenorm.py
+++ b/src/flag_gems/ops/instancenorm.py
@@ -339,8 +339,8 @@ def instance_norm_backward_kernel(
     Mean += pid
     Rstd += pid
 
-    mean = tl.load(Mean).to(tl.float32)
-    rstd = tl.load(Rstd).to(tl.float32)
+    mean = tl.load(Mean, mask=row_mask, other=0.0).to(tl.float32)
+    rstd = tl.load(Rstd, mask=row_mask, other=1.0).to(tl.float32)
     if HAS_WEIGHT_BIAS:
         w = tl.load(W + c_offsets, mask=row_mask).to(tl.float32)
     else:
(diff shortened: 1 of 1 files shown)


src/flag_gems/ops/instancenorm.py
```diff
@@ -339,8 +339,8 @@ def instance_norm_backward_kernel(
     Mean += pid
     Rstd += pid
 
-    mean = tl.load(Mean).to(tl.float32)
-    rstd = tl.load(Rstd).to(tl.float32)
+    mean = tl.load(Mean, mask=row_mask, other=0.0).to(tl.float32)
```
Jokeren, line 342: Same as the other PR, I think you can skip computation entirely without using the mask

src/flag_gems/ops/instancenorm.py
```diff
@@ -409,8 +409,8 @@ def weight_bias_backward_kernel(
         bid = b_off + tl.arange(0, BLOCK_BATCH_SIZE)[:, None]
         mid = bid * C + cid
         row_mask = mid < M
-        mean = tl.load(Mean + mid).to(tl.float32)
-        rstd = tl.load(Rstd + mid).to(tl.float32)
+        mean = tl.load(Mean + mid, mask=row_mask).to(tl.float32)
```
Jokeren, line 412: This is fine to me because `row_mask` is changing every iteration

src/flag_gems/ops/instancenorm.py
mark14wu: @Jokeren Done.

mark14wu: @Jokeren If I skip the computation, errors occur:
```
python3: /source/llvm-project/llvm/include/llvm/Support/Casting.h:566: decltype(auto) llvm::cast(const From &) [To = mlir::detail::TypedValue<mlir::IntegerType>, From = mlir::Value]: Assertion `isa<To>(Val) && "cast<Ty>() argument of incompatible type!"' failed.
```

src/flag_gems/ops/instancenorm.py
```diff
@@ -332,6 +332,8 @@ def instance_norm_backward_kernel(
 ):
     pid = tl.program_id(0) * BLOCK_ROW_SIZE + tl.arange(0, BLOCK_ROW_SIZE)[:, None]
     c_offsets = pid % C
+    if pid >= M:
```
Jokeren: Similar to the other PR, we should compute a scalar condition to skip

StrongSpoon: pid is actually an offset here. it might not work with skip condition.

src/flag_gems/ops/instancenorm.py
StrongSpoon: correct. I prefer setting row_mask as `bid < B`, which is more clear.

StrongSpoon (commented):
thanks for your correction!

src/flag_gems/ops/instancenorm.py
tongxin: Early skipping does not apply here since the rows are not aligned..

mark14wu: Now reverted to the masked load.

src/flag_gems/ops/instancenorm.py
mark14wu: @StrongSpoon Now changed to `row_mask = bid < B`.

StrongSpoon (approved):
lgtm

tongxin (approved):
LGTM

Issue #402, opened 2025-01-06
test_accuracy_weightnorm_interface failed when run correctness test 


tianxiao-baai:
### Describe the bug

env:
nv
torch: 2.3.0a0+40ec155e58.nv24.3
pytorch-triton： 2.2.0+e28a256d7

![image](https://github.com/user-attachments/assets/4a6c7bfb-7a42-41a9-8eaf-aa92ed482269)
![image](https://github.com/user-attachments/assets/6a4db7da-a310-4c4b-a38c-796f942132b9)

the method to reproduce the issue： pytest test_norm_ops.py::test_accuracy_weightnorm_interface

tongxin:
@TZWX-0, could you please take a look? Thanks.

tongxin:
I think this issue is solved. Could you please verify at your convenience?

tianxiao-baai:
> I think this issue is solved. Could you please verify at your convenience?

Sure，I will recheck it soon

bin913:
Is there a conclusion on this issue? Can we close it now?

PR #414, opened 2025-01-12
[Fix LayerNorm] Fix a memory out-of-bound access

mark14wu:
### PR Category

Operator

### Type of Change

Bug Fix

### Description

Similar issue to #413. 
This change adds a missing `row_mask` to the `tl.load` operations for `Mean` and `Rstd` to avoid potential out-of-bound access.
```diff
- mean = tl.load(Mean).to(tl.float32)
- rstd = tl.load(Rstd).to(tl.float32)
+ mean = tl.load(Mean, mask=row_mask).to(tl.float32)
+ rstd = tl.load(Rstd, mask=row_mask).to(tl.float32)
```
### Issue

N/A

### Progress

### Performance

N/A

### DIFF
--- a/src/flag_gems/ops/layernorm.py
+++ b/src/flag_gems/ops/layernorm.py
@@ -244,8 +244,8 @@ def layer_norm_backward_kernel(
     Mean += pid
     Rstd += pid
 
-    mean = tl.load(Mean).to(tl.float32)
-    rstd = tl.load(Rstd).to(tl.float32)
+    mean = tl.load(Mean, mask=row_mask).to(tl.float32)
+    rstd = tl.load(Rstd, mask=row_mask).to(tl.float32)
 
     dx_part2 = tl.zeros([BLOCK_ROW_SIZE, BLOCK_COL_SIZE], dtype=tl.float32)
     dx_part3 = tl.zeros([BLOCK_ROW_SIZE, BLOCK_COL_SIZE], dtype=tl.float32)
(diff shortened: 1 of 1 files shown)


src/flag_gems/ops/layernorm.py
```diff
@@ -244,8 +244,8 @@ def layer_norm_backward_kernel(
     Mean += pid
     Rstd += pid
 
-    mean = tl.load(Mean).to(tl.float32)
-    rstd = tl.load(Rstd).to(tl.float32)
+    mean = tl.load(Mean, mask=row_mask).to(tl.float32)
```
Jokeren, line 247: It might be better to just skip the body with the following code:

```
if pid >= M:
  return
```

Jokeren:
@iclementine

src/flag_gems/ops/layernorm.py
mark14wu: @Jokeren Done.

src/flag_gems/ops/layernorm.py
```diff
@@ -237,6 +237,8 @@ def layer_norm_backward_kernel(
     BLOCK_COL_SIZE: tl.constexpr,
 ):
     pid = tle.program_id(0) * BLOCK_ROW_SIZE + tl.arange(0, BLOCK_ROW_SIZE)[:, None]
+    if pid >= M:
```
Jokeren: `pid` here is a tensor but not a scalar

Jokeren: We can only return when the condition is a scalar comparison

StrongSpoon: maybe our naming mislead you. we should rename it as offset instead of pid here.

src/flag_gems/ops/layernorm.py
tongxin: This new change won't compile, please use the masked load instead.

mark14wu: @tongxin Reverted to masked load now.

mark14wu:
CI failed probably because of this issue: #402.

tongxin (approved):
LGTM. Just ignore the CI failures and we'll fix those later.

PR #431, opened 2025-01-21
[TEST] fix error in argmin UT when dtype=int16

awayzjj:
### Issue

1. The tl.min function will upcast an int16 input tensor to int32 on iluvatar. Therefore, the min_values should be initialized with dtype=int32; otherwise, the Triton IR cannot be generated.
2. add tune config of batchnorm

### Progress


iclementine (approved):
LGTM.

It is a temporary workaround to avoid promoting `min_values` to float32 when `value` is of type `tl.int16`.

PR #437, opened 2025-02-07
Fix a Setup error

sgjzfzzf:
### PR Category
User Experience

### Type of Change
Bug Fix

### Description
When I tried to install it with pip from Github with `pip install git+https://github.com/FlagOpen/FlagGems.git`, and then use it in my project, it reports an error `FileNotFoundError: [Errno 2] No such file or directory: '**/.venv/lib/python3.10/site-packages/flag_gems/runtime/backend/_nvidia/tune_configs.yaml'`. It's due to the Yaml files are not packaged correctly during building. Now I reset the package searching patterns, and it works. You can try it with `pip install git+https://github.com/sgjzfzzf/FlagGems.git`.

### Progress

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -102,7 +102,7 @@ setup(
     package_dir={"": "src"},
     include_package_data=True,  # To include non-Python files, e.g., README
     package_data={
-        "flag_gems.runtime": ["**/*.yaml"],
+        "flag_gems.runtime": ["*/**/*.yaml"],
     },
     setup_requires=["setuptools"],
 )


StrongSpoon:
@Galaxy1458 please take a look.

Galaxy1458:
Thanks for your contribution! At the same time, could you please provide us with your environment and the versions of torch, triton and python or any other useful information you think, so that we can better test, because this problem does not occur within the scope we have tested.

sgjzfzzf:
> Thanks for your contribution! At the same time, could you please provide us with your environment and the versions of torch, triton and python or any other useful information you think, so that we can better test, because this problem does not occur within the scope we have tested.

```
python
Python 3.10.12 (main, Jan 17 2025, 14:35:34) [GCC 11.4.0] on linux
Type "help", "copyright", "credits" or "license" for more information.
>>> import torch
>>> import triton
>>> torch.__version__
'2.6.0+cu124'
>>> triton.__version__
'3.2.0'
```

This problem occurred when I tried to install `flag_gems` in a new virtual environment. I found that if it's installed with `pip install -e .` or from a pre-built environment, this problem won't happen. You can try to start with a purely new environment and install it with `pip install .` to replay it.

Galaxy1458 (commented):
lgtm

Galaxy1458 (approved):

PR #438, opened 2025-02-08
fix bugs on database search

sgjzfzzf:
### PR Category
Model Test

### Type of Change
Bug Fix

### Description
When reading the cache of the compiled kernel, the read index sometimes could be an error. You can replay this bug with the following script.
```
import flag_gems
import torch


x = torch.randn((256, 256), device="cuda")
y = torch.randn((256, 256), device="cuda")
with flag_gems.use_gems():
    z0 = torch.matmul(x, y)
    z1 = torch.matmul(x, y)
```
In the PR, it reads by `filter`, not `index`, so it has no relationship with the parameter storage position.

### Progress

### DIFF
--- a/src/flag_gems/utils/libentry.py
+++ b/src/flag_gems/utils/libentry.py
@@ -82,9 +82,12 @@ class LibTuner(triton.runtime.Autotuner):
             attrs = -5 if major_version == 2 else -4
             for k, v in cfg_ls[:attrs]:
                 config.kwargs[k] = eval(v)
-            config.num_warps = eval(cfg_ls[attrs][1])
-            config.num_ctas = eval(cfg_ls[attrs + 1][1])
-            config.num_stages = eval(cfg_ls[attrs + 2][1])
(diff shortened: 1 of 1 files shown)


StrongSpoon:
could you please provide the environment information?

StrongSpoon:
I reproduced it with Triton v3.2. This error is caused by the difference between definitions of class Config. In version 3.2, developers added arguments like num_buffers_warp_spec, num_consumer_groups and so on. To adapt to the update, I suggest reconstructing the function `preload`. These new arguments should be loaded into config as well.

StrongSpoon:
PR #446 targeting at the same issue is already merged. thanks for your suggestion! and if you don't mind, we will close this PR ;)

sgjzfzzf:
>  PR #446 targeting at the same issue is already merged. thanks for your suggestion! and if you don't mind, we will close this PR ;)

SGTM! Thank you for your work!

PR #449, opened 2025-02-18
[Metax] Add metax backend conv tune config

mx-flaggems-user:
### PR Category

Operator

### Type of Change

Bug fix

### Description

Add conv op configs in metax backend (copied from nvida backend configs)


StrongSpoon (approved):
lg

PR #462, opened 2025-02-26
[bugfix] delete parameter not initialized for triton2.2

StrongSpoon:
### PR Category

Other
### Type of Change

Bug Fix
### Description

In the definition of triton.Config, attribute enable_persistent is not acclaimed as initialization parameter. While it's stored in FlagGems' config cache, resulting the error in preload.
Since in most cases, enable_persistent is set as default value, I ignore it when reconstructing triton.Config.
### Progress

### DIFF
--- a/src/flag_gems/utils/libentry.py
+++ b/src/flag_gems/utils/libentry.py
@@ -93,6 +93,7 @@ class LibTuner(triton.runtime.Autotuner):
                 kwargs[k] = eval(v)
             for k, v in cfg_ls[-attrs:]:
                 numargs[k] = eval(v)
+            numargs.pop("enable_persistent", None)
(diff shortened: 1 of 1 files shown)


src/flag_gems/utils/libentry.py
```diff
@@ -93,6 +93,7 @@ def preload(self):
                 kwargs[k] = eval(v)
             for k, v in cfg_ls[-attrs:]:
                 numargs[k] = eval(v)
+            numargs.pop("enable_persistent", None)
```
tongxin, line 98: Could you add a comment line here describing this config option for reference?

StrongSpoon: done

tongxin (approved):
LG

PR #500, opened 2025-03-18
[Iluvatar] fix kron scatter randperm quantile

awayzjj:
### Description

For randerm, I think using 'cg' is sufficient instead of 'wt' and it improves readability. I also benchmarked it and found that it does not impact performance, although I initially expected it might provide some benefits.

### Progress


StrongSpoon (approved):
lgtm

PR #553, opened 2025-04-21
Fix addmm op bug

Salv1a:
### PR Category

[ Operator ]

### Type of Change

[ Bug Fix ]

### Description

The first argument of `torch.addmm` op is the `bias` matrix, the shape of which should be broadcastable with the output shape. But the implementation of `flag_gems.addmm` only takes the row of `bias` into consideration. e.g. When the shape of `bias` is `(m x n)`, the result is incorrect.

### Progress

### DIFF
--- a/src/flag_gems/ops/addmm.py
+++ b/src/flag_gems/ops/addmm.py
@@ -6,7 +6,7 @@ import triton.language as tl
 
 from .. import runtime
 from ..runtime import torch_device_fn
-from ..utils import libentry
+from ..utils import broadcastable_to, libentry
 from ..utils import triton_lang_extension as tle
 
 
@@ -44,7 +44,6 @@ def addmm_kernel(
     offs_k = tl.arange(0, BLOCK_SIZE_K)
     a_ptrs = a_ptr + (offs_am[:, None] * stride_am + offs_k[None, :] * stride_ak)
     b_ptrs = b_ptr + (offs_k[:, None] * stride_bk + offs_bn[None, :] * stride_bn)
-    bias_ptrs = bias_ptr + offs_bn
 
     accumulator = tl.zeros((BLOCK_SIZE_M, BLOCK_SIZE_N), dtype=tl.float32)
     for k in range(0, tl.cdiv(K, BLOCK_SIZE_K)):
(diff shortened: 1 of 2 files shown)


tests/test_blas_ops.py
```diff
@@ -28,7 +28,7 @@
 def test_accuracy_addmm(M, N, K, scalar, dtype):
     mat1 = torch.randn((M, K), dtype=dtype, device=flag_gems.device)
     mat2 = torch.randn((K, N), dtype=dtype, device=flag_gems.device)
-    bias = torch.randn((N,), dtype=dtype, device=flag_gems.device)
+    bias = torch.randn((M, N), dtype=dtype, device=flag_gems.device)
```
StrongSpoon: I think both shapes make sense and suggest reserving the old one.

src/flag_gems/ops/addmm.py
```diff
@@ -61,26 +60,32 @@ def addmm_kernel(
...
         b_ptrs += BLOCK_SIZE_K * stride_bk
-    bias = tl.load(bias_ptrs, mask=offs_bn < N, other=0.0)
-    accumulator = accumulator * alpha + bias * beta
-    c = accumulator.to(bias.dtype)
 
     offs_cm = pid_m * BLOCK_SIZE_M + tl.arange(0, BLOCK_SIZE_M)
     offs_cn = pid_n * BLOCK_SIZE_N + tl.arange(0, BLOCK_SIZE_N)
     c_ptrs = c_ptr + stride_cm * offs_cm[:, None] + stride_cn * offs_cn[None, :]
     c_mask = (offs_cm[:, None] < M) & (offs_cn[None, :] < N)
+    bias_ptrs = bias_ptr + offs_cm[:, None] * stride_cm + offs_cn[None, :] * stride_cn
```
StrongSpoon: perhaps delivering the strides of bias tensor is helpful.

StrongSpoon (commented):
thanks for your contribution! it's a valuable fix for addmm and we're glad to accept this PR.

tests/test_blas_ops.py
Salv1a: Agree.

src/flag_gems/ops/addmm.py
Salv1a: You are right, the strides of bias are not equivalent to that of c.

StrongSpoon (approved):
lgtm

PR #587, opened 2025-04-28
[LibTuner] simplify autotuner for faster inference

StrongSpoon:
### PR Category

Other
### Type of Change

New Feature
### Description

1. allow autotuner to choose configs according to log of key arguments
2. set shared config cache for similar operators
### Progress

### DIFF
--- a/src/flag_gems/ops/all.py
+++ b/src/flag_gems/ops/all.py
@@ -7,7 +7,7 @@ import triton.language as tl
 
 from .. import runtime
 from ..runtime import torch_device_fn
-from ..utils import dim_compress, libentry
+from ..utils import dim_compress, libentry, libtuner
 from ..utils import triton_lang_extension as tle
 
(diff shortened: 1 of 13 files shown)


src/flag_gems/ops/all.py
```diff
@@ -21,7 +21,11 @@ def reduce_all(a, b):
 
 
 @libentry()
-@triton.autotune(configs=runtime.get_tuned_config("all"), key=["M", "N"])
+@libtuner(
+    configs=runtime.get_tuned_config("naive_reduction"),
+    key=["M", "N"],
+    share="naive_reduction",
```
kiddyjinjin, line 27: is there any case where the key in `get_tuned_config` is  "native_reduction", but the corresponding key in the DB is something else?
If not, then perhaps we don’t need to add the "share" parameter?

kiddyjinjin (approved):

PR #636, opened 2025-05-23
add a function to locate python files in src/flag_gems

iclementine:
### PR Category

Other

### Type of Change

Other

### Description

1. Add a function to get source of ops defined in src/flag_gems(via environment);
2. fix CMakeList.txt errors in the use of findin_package;
3. clean cmake files.

TODO: most source files in flag_gems/ops can not be imported alone, since it has a lot of relative imports, we need to fix it.

### Progress


CMakeLists.txt
```diff
@@ -1,20 +1,99 @@
...
+# Automatically add all linked folders that are NOT in the build directory to
+# the rpath (per library?)
+set(CMAKE_INSTALL_RPATH_USE_LINK_PATH TRUE)
+
+
+# --------------------------- project options ---------------------------
+option(FLAGGEMS_USE_EXTERNAL_TRITON_JIT "whether to use external triton jit library" OFF)
+option(FLAGGEMS_USE_EXTERNAL_PYBIND11 "whether to use external pybind11 library" ON)
+option(FLAGGEMS_BUILD_C_EXTENSIONS "whether to build c extension" OFF)
+option(FLAGGEMS_BUILD_CTESTS "Whether to build CPP unit tests" ${FLAGGEMS_BUILD_C_EXTENSIONS})
```
Bowen12992, line 34: I think that when building a C++ extension, ctest should not be a mandatory build option—especially as the number of ctest cases increases in the future, which would unnecessarily increase the build time.

Bowen12992 (approved):
LGTM, but why not use `Options`, `Dependencies` and `Settings` cmakefiles any more?

iclementine:
> LGTM, but why not use `Options`, `Dependencies` and `Settings` cmakefiles any more?

Jumping back and forth between these files makes me crazy.

CMakeLists.txt
iclementine: Yes, it is the default value. But it can be configured via arguments to cmake.

PR #649, opened 2025-05-27
[AdvancedCompiler]add contiguous-backward and tests

AdvancedCompiler:
### PR Category

Operator
### Type of Change

New Feature
### Description

Implement the function of contiguous-backward op
### Progress

### Performance

image


src/flag_gems/ops/contiguous.py
```diff
@@ -1,16 +1,48 @@
...
+@pointwise_dynamic(promotion_methods=[(0, "DEFAULT")])
+@triton.jit
+def contiguous_backward(out, grad):
+    return grad
+
+
+ALL_FLOAT_DTYPES = (torch.bfloat16, torch.float16, torch.float32, torch.float64)
+
+
+class Contiguous(torch.autograd.Function):
```
iclementine: We recommend not to use autograd.Function since it may not be compatible with torch.compile.

iclementine: since contiguous is a CompositeAutogradImplicit op, adding a Backend-specific implementation make it loss its backward, so we may need to add a backward formula for it. We can try torch.register_backward for it.
But I am not sure if it would work for torch.compile, since it go back to the problem that: how to add backward formula(only for a backend? or in a backend-agnostic way)

w1120029931-bit (changes requested):
The code architecture has been adjusted,
test_special_ops.py ...
does not exist currently, Each operator requires separate functional test files and performance test files. Please refer to the relevant code modifications in the repository

w1120029931-bit:
Is the PR still being followed up

PR #656, opened 2025-05-28
[Runtime] libtuner fix for triton2

StrongSpoon:
### PR Category

Other
### Type of Change

Bug Fix
### Description

triton.Config in triton2 doesn't have function named all_kwargs. this pr sets attribute for it.

ERROR
```
FAILED test_reduction_perf.py::test_general_reduction_perf[all-all-dtypes0] - Failed: 'Config' object has no attribute 'all_kwargs'
```
### Progress

### DIFF
--- a/src/flag_gems/utils/libentry.py
+++ b/src/flag_gems/utils/libentry.py
@@ -26,6 +26,29 @@ ATTRS = {
 version = triton.__version__.split(".")
 major_version, minor_version = eval(version[0]), eval(version[1])
 
+if major_version == 2:
+
+    def all_kwargs(self):
+        return {
+            **self.kwargs,
+            **{
+                k: getattr(self, k)
+                for k in (
+                    "num_warps",
+                    "num_ctas",
+                    "num_stages",
+                    "num_buffers_warp_spec",
+                    "num_consumer_groups",
+                    "reg_dec_producer",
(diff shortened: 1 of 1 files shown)


src/flag_gems/utils/libentry.py
```diff
@@ -26,6 +26,29 @@
...
+                    "num_warps",
+                    "num_ctas",
+                    "num_stages",
+                    "num_buffers_warp_spec",
+                    "num_consumer_groups",
+                    "reg_dec_producer",
+                    "reg_inc_consumer",
+                    "maxnreg",
+                )
+                if hasattr(self, k)
```
tongxin, line 46: I think it would be more clear and easier to verify these runtime params if we include a permlink to the triton codebase location where these params are actually used.

StrongSpoon: yes, but vendors are waiting for the stable version. I suggest preserving it as a todo. orz

iclementine (approved):

tongxin (approved):
LGTM

Rayfxl:
I noticed that this fix only applies when `major_version == 2`.  
However, Triton 3.0.0 also does not provide `Config.all_kwargs`, so the same issue still exists there.

PR #687, opened 2025-06-17
[UTILS] Change get_device_properties api

zhzhcookie:
### PR Category
UTILS

### Type of Change
New Feature

### Description
If ```torch_device_fn.get_device_properties``` is not supported, use ```triton.runtime.driver.active.utils.get_device_properties```

### Progress

### DIFF
--- a/src/flag_gems/utils/triton_driver_helper.py
+++ b/src/flag_gems/utils/triton_driver_helper.py
@@ -1,0 +1,8 @@
+try:
+    from ..runtime import torch_device_fn
+
+    get_device_properties = torch_device_fn.get_device_properties
+except ImportError:
+    import triton
+
+    get_device_properties = triton.runtime.driver.active.utils.get_device_properties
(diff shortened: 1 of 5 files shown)


src/flag_gems/utils/triton_driver_helper.py
```diff
@@ -0,0 +1,8 @@
+try:
+    from ..runtime import torch_device_fn
+
+    get_device_properties = torch_device_fn.get_device_properties
```
tongxin, line 4: As backend/runtime utils getting chaotic, the code is getting harder to maintain. Can you define an interface class where individual backends can register their own instances?

zhzhcookie: > As backend/runtime utils getting chaotic, the code is getting harder to maintain. Can you define an interface class where individual backends can register their own instances?

In fact, all Triton backends have this interface, but FlagGems previously obtained it from the torch interface (however, not all hardware deeply will extend torch).

PR #699, opened 2025-06-20
Add mm.out and sum_dim_out op

0x45f:
### PR Category

Operator
### Type of Change

New Feature
### Description

Add mm.out op
### Progress

### DIFF
--- a/src/flag_gems/ops/sum.py
+++ b/src/flag_gems/ops/sum.py
@@ -118,6 +118,25 @@ def sum(inp, *, dtype=None):
     return out
 
 
+def sum_out(inp, *, dtype=None, out):
+    logger.debug("GEMS SUM_OUT")
+    M = inp.numel()
+    if dtype is None:
+        dtype = inp.dtype
+        if dtype is torch.bool:
+            inp = inp.to(torch.int64)
+            dtype = torch.int64
+    block_size = triton.next_power_of_2(math.ceil(math.sqrt(M)))
+    mid_size = triton.cdiv(M, block_size)
+    block_mid = triton.next_power_of_2(mid_size)
+
+    mid = torch.empty((mid_size,), dtype=dtype, device=inp.device)
+    with torch_device_fn.device(inp.device):
+        sum_kernel_1[(mid_size, 1, 1)](inp, mid, M, block_size)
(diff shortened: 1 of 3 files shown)


src/flag_gems/ops/mm.py
```diff
@@ -146,3 +146,42 @@ def mm(a, b):
...
     return c
+
+
+def mm_out(a, b, *, out):
+    logger.debug("GEMS MM_OUT")
+    # handle non-contiguous inputs if necessary
+    if a.stride(0) > 1 and a.stride(1) > 1:
+        a = a.contiguous()
+    if b.stride(0) > 1 and b.stride(1) > 1:
+        b = b.contiguous()
```
tongxin, line 157: It's slightly faster to trim down method calls and operators 

```suggestion
    if 1 not in a.stride():
        a = a.contiguous()
    if 1 not in b.stride():
        b = b.contiguous()
```

src/flag_gems/ops/sum.py
```diff
@@ -118,6 +118,25 @@ def sum(inp, *, dtype=None):
...
+    if dtype is None:
+        dtype = inp.dtype
+        if dtype is torch.bool:
+            inp = inp.to(torch.int64)
+            dtype = torch.int64
+    block_size = triton.next_power_of_2(math.ceil(math.sqrt(M)))
+    mid_size = triton.cdiv(M, block_size)
+    block_mid = triton.next_power_of_2(mid_size)
+
+    mid = torch.empty((mid_size,), dtype=dtype, device=inp.device)
```
tongxin, line 133: I think the intermediate and final sums are both prone to overflow for ints.  It would be safe to use int64 as accumulate dtype for all int typed inputs.

src/flag_gems/ops/mm.py
```diff
@@ -146,3 +146,42 @@ def mm(a, b):
...
+            a.stride(0),
+            a.stride(1),
+            b.stride(0),
+            b.stride(1),
+            c.stride(0),
+            c.stride(1),
+            dot_out_dtype=dot_out_dtype,
+            GROUP_M=8,
+        )
+    return c
```
Bowen12992, line 187: Do we still need `return c`？

Issue #718, opened 2025-06-28
Autotuning introduces non-trivial overhead and should be disabled in some operators.


tongxin:
### Describe the bug

#712 reveals that enabling online autotuning for certain ops can harm user experience because they can be quite frequently invoked in data inspection, debugging and other non-compute cases. Their peak performance is not really as relevant as those like gemms and activation functions. We can safely turn off autotuning  for them.

Among these ops are masked_select, nonzero.

PR #719, opened 2025-06-29
Update masked_select and nonzero

tongxin:
### PR Category

Operator

### Type of Change

Refactor & optimization

### Description

This PR curbs autotuning for `masked_select` and `nonzero`, to avoid confusion during testing and debugging. `masked_select` is optimized in the following way:

- work requires at most two kernel launches
- compute partial sums of mask and cumsum over the partial sums in a single kernel
- write back the selected input data in the second kernel.
- no extra total sum kernel involved

Performance gains are clear for larger input sizes.

image

### Issue
#718 #712 

### Progress

### DIFF
--- a/src/flag_gems/ops/masked_select.py
+++ b/src/flag_gems/ops/masked_select.py
@@ -8,12 +8,13 @@ from .. import runtime
 from ..runtime import torch_device_fn
 from ..utils import broadcastable, libentry
 from ..utils import triton_lang_extension as tle
+from ..utils.shape_utils import bracket_next_power_of_2
 
 logger = logging.getLogger(__name__)
 
 
 @libentry()
-@triton.autotune(configs=runtime.get_tuned_config("masked_select"), key=["n_elements"])
+@triton.heuristics(runtime.get_heuristic_config("elementwise_generic"))
 @triton.jit
 def masked_select_kernel(
     inp_ptr,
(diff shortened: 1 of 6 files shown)


src/flag_gems/ops/masked_select.py
```diff
@@ -46,15 +144,59 @@ def masked_select(inp, mask):
     inp, mask = torch.broadcast_tensors(inp, mask)
 
     inp = inp.contiguous()
-    mask = mask.contiguous()
+    mask = mask.ravel()
```
iclementine: As torch's documentaion says

The shapes of the mask tensor and the input tensor don’t need to match, but they must be broadcastable.

So make them contiguous is not enough. If the kernel does not handle broadcasting itself, it must be broadcasted and contiguous and ravel beforehand.

src/flag_gems/ops/masked_select.py
```diff
@@ -4,34 +4,132 @@
...
+            acc += select_ints
+            offset += BLOCK_SIZE
+    else:
+        for block_id in range(
+            start_block, min(num_blocks, start_block + num_blocks_per_row)
+        ):
+            select = tl.load(mask_ptr + offset, mask=offset < N, other=0)
+            select_ints = select.to(tl.constexpr(part_sums_ptr.dtype.element_ty))
+            acc += select_ints
+            offset += BLOCK_SIZE
```
iclementine: These 2 blocks looks mostly the same. The only difference is the end of the iteration. What about computing the iteration end beforehand.

src/flag_gems/ops/masked_select.py
```diff
@@ -4,34 +4,132 @@
...
+    np = tl.num_programs(0)
+    if count == np - 1:
+        mask = tl.arange(0, NP_BLOCK) < np
+        part_sums = tl.load(part_sums_ptr + tl.arange(0, NP_BLOCK), mask=mask)
+        final_sum = tl.sum(part_sums, axis=0)
+        pre_sums = tl.cumsum(part_sums, axis=0)
+        tl.store(
+            part_sums_ptr + tl.arange(0, NP_BLOCK), pre_sums - part_sums, mask=mask
+        )
+        tl.store(part_sums_ptr + np, final_sum)
```
iclementine, line 86: So this kernel computes the CTA-level exclusive-prefix-scan.

src/flag_gems/ops/masked_select.py
```diff
@@ -46,15 +144,59 @@ def masked_select(inp, mask):
...
+        mask,
+        part_sums,
+        out,
+        N,
+        n_blocks,
+        n_blocks_per_row,
+        NP_BLOCK=triton.next_power_of_2(np),
+        BLOCK_SIZE=BLOCK_SIZE,
+        num_warps=num_warps,
+    )
```
iclementine: A reduce-then-scan prefix sum implementation, LGTM.

iclementine (changes requested):

src/flag_gems/ops/masked_select.py
tongxin: You are right.

src/flag_gems/ops/masked_select.py
tongxin: Thanks for pointing out the error!

iclementine (approved):

src/flag_gems/ops/masked_select.py
iclementine: OH, no. Sorry there was broadcast_tensors before. No problem here.

src/flag_gems/ops/masked_select.py
```diff
@@ -4,34 +4,125 @@
...
+    last_block_id = min(num_blocks - 1, start_block + num_blocks_per_row - 1)
+
+    for block_id in range(start_block, last_block_id):
+        select = tl.load(mask_ptr + offset)
+        select_ints = select.to(tl.constexpr(part_sums_ptr.dtype.element_ty))
+        acc += select_ints
+        offset += BLOCK_SIZE
+    # Peeled last block
+    select = tl.load(mask_ptr + offset, mask=offset < N, other=0)
+    select_ints = select.to(tl.constexpr(part_sums_ptr.dtype.element_ty))
```
iclementine: `tl.constexpr(part_sums_ptr.dtype.element_ty)` can be simply `part_sums_ptr.dtype.element_ty` as dtypes are always known at compile time.

tongxin: I'm always confused by lining tl.constexpr in kernel. Is there a rule of thumb on how to use it?

iclementine: I only use annotated assignment in triton kernels.

`variable: tl.constexpr = rhs`

In this casem the rhs can call non triton jit functions like `math.exp` etc.

tongxin: updated.

iclementine (dismissed):
LGTM

meinie0826 (dismissed):

src/flag_gems/ops/masked_select.py
```diff
@@ -46,15 +165,59 @@ def masked_select(inp, mask):
...
+        n_blocks,
+        n_blocks_per_row,
+        NP_BLOCK=NP_BLOCK,
+        BLOCK_SIZE=BLOCK_SIZE,
+        num_warps=num_warps,
+    )
 
-    n_elements = inp.numel()
-    grid = lambda meta: (triton.cdiv(n_elements, meta["BLOCK_SIZE"]),)
-    with torch_device_fn.device(inp.device):
```
iclementine, line 58: Removing device guard may cause "inaccessible address"

iclementine (changes requested):
Removing device guard may cause inaccessible address.

src/flag_gems/ops/masked_select.py
tongxin: Done.

PR #747, opened 2025-07-08
modify: use jitFuntion cache_key instead of source code as hash

meinie0826:
### PR Category

libtuner hash value
### Type of Change

Bug Fix
### Description

Now, the method of hash generation is use source code of kernel function. This will result in changes to the subfunction not being reflected in the hash value. 
So, This PR use jitFuntion cache_key instead of source code as hash.
### Progress

### DIFF
--- a/src/flag_gems/utils/libentry.py
+++ b/src/flag_gems/utils/libentry.py
@@ -200,15 +200,18 @@ class LibTuner(triton.runtime.Autotuner):
         self.keys = key
         self.strategy = strategy
         # Use table name with hash instead of hash in key
-        self.kernel_hash = self.get_kernel_hash
-        self.table_name = f"{self.__name__}_{self.kernel_hash}"
+        self.kernel_hash = None
+        self.table_name = f"{self.__name__}_{self.get_kernel_hash()}"
         self.cache = libcache[self.table_name]
         if strategy:
             assert len(self.strategy) == len(self.keys), "Invalid number of strategies"
 
     def get_kernel_hash(self):
         if self.kernel_hash is None:
-            func_hash = self.fn.cache_key
+            jit_fn = self.fn
+            while not isinstance(jit_fn, triton.runtime.JITFunction):
+                jit_fn = jit_fn.fn
+            func_hash = jit_fn.cache_key
             config_strs = [str(config) for config in self.configs]
             combined_content = f"{func_hash}{config_strs}"
             self.kernel_hash = hashlib.md5(
--- a/src/flag_gems/utils/libentry.py
+++ b/src/flag_gems/utils/libentry.py
@@ -214,9 +214,7 @@ class LibTuner(triton.runtime.Autotuner):
             func_hash = jit_fn.cache_key
             config_strs = [str(config) for config in self.configs]
             combined_content = f"{func_hash}{config_strs}"
-            self.kernel_hash = hashlib.md5(
-                combined_content.encode("utf-8")
-            ).hexdigest()[:8]
+            self.kernel_hash = hashlib.md5(combined_content.encode("utf-8")).hexdigest()
         return self.kernel_hash
 
     def get_key(self, args):
--- a/src/flag_gems/utils/libentry.py
+++ b/src/flag_gems/utils/libentry.py
@@ -33,18 +33,6 @@ version = triton.__version__.split(".")
 major_version, minor_version = eval(version[0]), eval(version[1])
 
 
-def get_kernel_hash(func, configs):
-    if hasattr(func, "fn"):
-        original_func = func.fn
-    else:
-        original_func = func
-
-    source_code = inspect.getsource(original_func)
-    config_strs = [str(config) for config in configs]
-    combined_content = f"{source_code}{config_strs}"
-    return hashlib.md5(combined_content.encode("utf-8")).hexdigest()[:8]
-
-
 if major_version == 2:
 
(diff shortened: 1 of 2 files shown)


meinie0826:
please give some advice @StrongSpoon  @iclementine  @tongxin

Bowen12992 (approved):
LGTM

src/flag_gems/utils/libentry.py
```diff
@@ -211,13 +199,23 @@ def __init__(
         self.__name__ = self.base_fn.__name__
         self.keys = key
         self.strategy = strategy
-        self.kernel_hash = get_kernel_hash(self.base_fn, self.configs)
         # Use table name with hash instead of hash in key
+        self.kernel_hash = self.get_kernel_hash
```
tongxin: This seems assigning function to the hash variable? I don't quite get it.

meinie0826: I'm sorry about that, I forget to modify the call method, Thanks!

src/flag_gems/utils/libentry.py
```diff
@@ -211,13 +199,26 @@ def __init__(
         self.__name__ = self.base_fn.__name__
         self.keys = key
         self.strategy = strategy
-        self.kernel_hash = get_kernel_hash(self.base_fn, self.configs)
         # Use table name with hash instead of hash in key
-        self.table_name = f"{self.__name__}_{self.kernel_hash}"
+        self.kernel_hash = None
+        self.table_name = f"{self.__name__}_{self.get_kernel_hash()}"
```
tongxin, line 204: This is a bit convoluted. If self.kernel_hash is set to None, then next self.get_kernel_hash() will definitely take the true branch inside. For sake of simplicity, it's OK to run the hashing eagerly.
```suggestion
        self.kernel_hash = self.get_kernel_hash()
        self.table_name = f"{self.__name__}_{self.kernel_hash}"
```

src/flag_gems/utils/libentry.py
```diff
@@ -211,13 +199,26 @@ def __init__(
...
         self.cache = libcache[self.table_name]
         if strategy:
             assert len(self.strategy) == len(self.keys), "Invalid number of strategies"
 
+    def get_kernel_hash(self):
+        if self.kernel_hash is None:
```
tongxin, line 210: Then remove this.

src/flag_gems/utils/libentry.py
```diff
@@ -211,13 +199,26 @@ def __init__(
...
+            jit_fn = self.fn
+            while not isinstance(jit_fn, triton.runtime.JITFunction):
+                jit_fn = jit_fn.fn
```
tongxin, line 213: Maybe self.base_fn is good enough.

src/flag_gems/utils/libentry.py
meinie0826: I think `if self.kernel_hash is None:` is better. By this way, the method of obtaining hash can be unified into the get_cerner_hash function. If not, when you retrieve self.hash, it is possible to get None. which is in preparation for lazy load.

src/flag_gems/utils/libentry.py
meinie0826, line 213: I have retest, and I find `self.base_fn != jit_fn`. So, if we use base_fn, then base_fn will be write, other function will get dirty value.Do I understand correctly?

src/flag_gems/utils/libentry.py
```diff
@@ -211,13 +199,26 @@ def __init__(
...
+            func_hash = jit_fn.cache_key
+            config_strs = [str(config) for config in self.configs]
+            combined_content = f"{func_hash}{config_strs}"
```
tongxin, line 216: hash should also be updated with self.keys, self.strategy and so on

src/flag_gems/utils/libentry.py
```diff
@@ -211,13 +199,26 @@ def __init__(
...
+            self.kernel_hash = hashlib.md5(
+                combined_content.encode("utf-8")
+            ).hexdigest()[:8]
```
tongxin: Is it safe to take a slice of the hash?

src/flag_gems/utils/libentry.py
meinie0826: Actually, keys and strategy was get by get_key function, which used as key of cache. And, hash value was used as table_name, It was in different level. I'm not sure if they should be put together, should I modify them according to your suggestion？

tongxin: OK. I didn't notice that. I think it's fine now.

src/flag_gems/utils/libentry.py
meinie0826: Normally, I think it's sufficient, but I can also make some changes

iclementine (approved):

tests/test_libentry.py
```diff
@@ -289,3 +285,66 @@ def kernel_a_copy(x, y):
...
+
+    @libtuner(
+        configs=[
+            triton.Config({"TILE_N": 32}),
+            triton.Config({"TILE_N": 64}),
+        ],
+        key=["x"],
+    )
+    @triton.jit
+    def main_kernel(x, y):
```
iclementine, line 320: Can we avoid defining `main_kernel` and use the one defined before?

StrongSpoon (approved):
lgtm

PR #751, opened 2025-07-08
[METAX] fix mha_varlan_fwd and fill, skip var_mean

HuangLianghong:
### PR Category

Operator and OP Test
### Type of Change

Bug Fix
### Description

1. shrink block size in mha_varlan_fwd due to shared memory out of source

2. replace metax fill op with default fill op

3. skip var_mean test, reason: core dumped

### Progress


tongxin (changes requested):
Please refrain from including vendor specific checks in the main flag_gems/ops source files. You may change the code to be backend neutral using gems runtime apis.  I suggest consulting with @Galaxy1458

HuangLianghong:
@Galaxy1458 Hi, could you please show me which runtime api can be used to modify the code? I need to shrink the block size in mha_varlana_fwd, and maybe I need to perform this change in other op in the future tests, since the size of shared memory in C500 is smaller than A100.

HuangLianghong:
> Hi, Lianghong. I removed the hardcoded kernel params. #753 Please make further changes based in that.

That's great, I will make changes after #753 is merged.

mx-flaggems-user (approved):

tongxin:
Do we also need to change fill?

HuangLianghong:
> Do we also need to change fill?

Yes, because it cannot pass the accuracy test while  'fill' in flag_gems/ops pass the accuracy test and do not have any performance degradation.
I think use ops from  flag_gems/ops source files whenever possible is better, because they might be updated in the future. Currently we only implement ops in flag_gems/runtime/backend/_metax when we need to improve performance or need to modify kernel code to adapt it to C500.

PR #755, opened 2025-07-09
[libtuner]Expand libtuner search strategy interface

meinie0826:
### PR Category

libtuner
### Type of Change

New Feature
### Description

The current libtuner uses brute force search strategy for automatic tuning, which is very inconvenient for extending other strategies. 
Therefore, this PR abstracts the extension strategy into an interface, allowing users to freely define their own extension strategy.
### Progress


src/flag_gems/utils/libentry.py
```diff
@@ -205,6 +222,10 @@ def __init__(
         self.cache = libcache[self.table_name]
         if strategy:
             assert len(self.strategy) == len(self.keys), "Invalid number of strategies"
+        if isinstance(search_strategy, str) and search_strategy in SEARCH_STRATEGIES:
+            self.search_strategy = SEARCH_STRATEGIES[search_strategy]
+        else:
+            self.search_strategy = default_search_strategy
```
sgjzfzzf: Why not set the default value of `search_strategy` as `brute` directly? The `else` branch looks not so necessary.

meinie0826: Actually, if-else is unavoidable, and self. search_stategy could be none, brute, or any other field, so I think this writing style is more concise.

sgjzfzzf: I really don't get your meaning. You can delete the `else` branch and provide default value for `search_strategy`, and I think that could be a more readable way for developers to find the default strategy. Anyway it's just a coding style issue and other parts look good to me.

meinie0826: Imagine if the search_stategy passed in by the user is a string that is not defined in SEARCH-STATEGIES, then search_stategy will become a string and subsequent calls will report an error

sgjzfzzf: Print a warning if it's not found.

StrongSpoon (approved):
lg

PR #756, opened 2025-07-09
[kunlunxin] update attention & add mmout,sumout,sumdimout

### LINKED ISSUE #748: Can't parse flash_fwd_splitkv_kernel name in Python 3.8
### Describe the bug

In Python 3.8, `inspect.getsource(fn)` maybe cannot correctly get the source code for Triton kernels,

```
triton/runtime/jit.py:717: in __init__
    self.src = self.src[re.search(r"^def\s+\w+\s*\(", self.src, re.MULTILINE).start():]
E   AttributeError: 'NoneType' object has no attribute 'start'
```

and this issue appears to affect only the `flash_fwd_splitkv_kernel`

```
src/flag_gems/ops/flash_kernel.py:695: in <module>
    def flash_fwd_splitkv_kernel(
```

Python 3.8.20
Torch: 2.4.1
Triton: 3.1.0

DuanYaQi:
### Issue

Resolves #748 

### Progress

### DIFF
--- a/src/flag_gems/ops/flash_kernel.py
+++ b/src/flag_gems/ops/flash_kernel.py
@@ -53,7 +53,7 @@ def apply_dropout_mask(
     if encode_dropout_in_sign_bit:
         P = tl.where(mask, -P, P)
     else:
-        P = tl.where(mask, (P * 0).to(P.dtype), P)
+        P = tl.where(mask, P * 0, P)
     return P
 
 


src/flag_gems/ops/flash_kernel.py
```diff
@@ -674,8 +682,8 @@ def flash_fwd_bh_parallel_kernel():
 @libentry()
 @triton.heuristics(
     values={
-        "BLOCK_M": lambda args: block_m_splitkv_heuristic(args["d"]),
-        "BLOCK_N": lambda args: block_n_splitkv_heuristic(args["d"]),
+        "BLOCK_M": lambda args: block_m_splitkv_heuristic_spec_argd,
```
duanyaqi: @tongxin it looks like a bug with python's builtin module inspect, which is used in triton's frontend to parse the source of triton jit function. In short, inspect fails to parse function correctly when some form of lambda is in the function decorator.

tongxin: Oh, I didn't notice this. I guess you want to replace lambdas in the heuristics with function like the following right?
```
"BLOCK_M": block_m_splitkv_heuristic_spec_args,
"BLOCK_N": block_n_splitkv_heuristic_spec_args
```

tongxin: Let's apply the change.

duanyaqi: ok

nianqi-tian (approved):

PR #717, opened 2025-06-27
add hash key and early stop mechanism for libtuner

meinie0826:
### PR Category

Other
### Type of Change

New Feature
### Issue
The current libtuner lacks a field representing the code, which may result in enabling outdated cache after modifying the code. To address this issue, this PR has added the md5 value of the source code as the hash key.
There is a phenomenon of excessive tuning in the way libtuner searches for the optimal config, which requires excessive computation to improve performance by 1%. Therefore, this PR has added an early stop mechanism.

### Progress

- [ ] Change is properly reviewed (1 reviewer required, 2 recommended).
- [ ] Change is responded to an issue.
- [ ] Change is fully covered by a UT.

### Performance

Test by the following code on H800, Through the following code test, on H800, the first end-to-end tuning time decreased from 11.51 seconds to 8.53 seconds, while the second end-to-end inference time increased from 1.16 seconds to 1.67 seconds.
```
# SPDX-License-Identifier: Apache-2.0
import random
import time

import numpy as np
import torch
from vllm import LLM, SamplingParams

import flag_gems


flag_gems.enable()  # Enable gems for PyTorch (aten) operators
flag_gems.apply_gems_patches_to_vllm(verbose=True)  # Patch vLLM custom ops


def set_seed(seed: int = 42):
    random.seed(seed)
    np.random.seed(seed)
    torch.manual_seed(seed)
    torch.cuda.manual_seed(seed)
    torch.cuda.manual_seed_all(seed)
    torch.backends.cudnn.deterministic = True
    torch.backends.cudnn.benchmark = False


set_seed(42)


# Sample prompts.
prompts = [
    "Hello, my name is",
    "The president of the United States is",
    "The capital of France is",
    "The future of AI is",
]

# Create a sampling params object.
sampling_params = SamplingParams(temperature=0.8, top_p=0.95, max_tokens=120)


def main():
    # Create an LLM.
    llm = LLM(
        model="Qwen/Qwen2.5-7B-Instruct",
        tensor_parallel_size=1,
        max_model_len=1024,
        gpu_memory_utilization=0.5,
        # enforce_eager=True,
    )

    # test time 2 times
    for _ in range(2):
        start_time = time.time()
        # Generate texts from the prompts. The output is a list of RequestOutput
        # objects that contain the prompt, generated text, and other information.
        outputs = llm.generate(prompts, sampling_params)
        end_time = time.time()
        print(f"Time taken for generation: {end_time - start_time:.2f} seconds")



if __name__ == "__main__":
    main()
```

PR #762, opened 2025-07-10
MultiArmedBandit Autotuner

sgjzfzzf:
### PR Category
Autotune

### Type of Change
Performance Optimization

### Description

This PR provides a new `LibTuner` implementation based on multi-armed-bandit policy. It allows autotuning step by step and accelerates this process wisely.

### Issue

PR #717

### Progress

### DIFF
--- a/src/flag_gems/utils/libentry.py
+++ b/src/flag_gems/utils/libentry.py
@@ -7,10 +7,11 @@ import os
 import sqlite3
 import threading
 import time
+import sys
 import weakref
 from collections import OrderedDict
-from itertools import starmap
-from typing import Dict, Optional
+from itertools import chain, starmap
+from typing import Any, Dict, List, Optional, Tuple, Union
 
 import triton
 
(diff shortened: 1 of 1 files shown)


sgjzfzzf:
I design a [benchmark](https://gist.github.com/sgjzfzzf/77bd0c1deeb145ca162879c367b6747c) based on @meinie0826's PR #717. The results follow
```
End to End Performance
Default: 154386.97923959792ms
Multi-Armed-Bandit: 107352.17661622167ms
```

Here is the step by step plot
e2e_timecost
This is the graph delete the first step(due to it's too high)
e2e_timecost
The Multi-Armed-Bandit Autotuner oscillates quickly at first, then gradually converges to the performance of the default Autotuner, while also having an advantage in the final performance.

sgjzfzzf:
Known problems include
- It doesn't work with the default `LibEntry` implementation, because the cache mechanism only triggers the autotune once. I need more detailed discussion with @StrongSpoon.
- It depends on the `di` interface to obtain the runtime event information, so it's risky to add this feature to non NVIDIA and AMD platforms, based on the feedback from @zhzhcookie, @Galaxy1458.
- It cannot run on vllm, because [`di.synchronize`](https://github.com/FlagOpen/FlagGems/blob/ce925295c5c8c1b69fe93d93ab3d8a7d88549d73/src/flag_gems/utils/libentry.py#L447) will lead to a runtime exception during the CUDA graph capturing.

sgjzfzzf:
Look forward to your opinions!

StrongSpoon:
Is there a comparison between the AmoTuner and early-stop mechanism mentioned in #717?

sgjzfzzf:
> Is there a comparison between the AmoTuner and early-stop mechanism mentioned in #717?

I run the same benchmark on @meinie0826's PR #717. The result follows

```
End to End Performance
Default: 154386.97923959792ms
Multi-Armed-Bandit: 107352.17661622167ms
Early Stopping: 309926.12229275703.
```

e2e_timecost

e2e_timecost

The first step of early stopping takes too much time. I'm unsure whether I run it in a wrong way. I'll do it again to double check the result, and please let @meinie0826 take a look at this.

meinie0826:
I advice use register_search_strategy and self.search_strategy in https://github.com/FlagOpen/FlagGems/pull/755 , then we can use different strategy on different operators, and we will be conveniented  to compare different strategies

meinie0826:
> > Is there a comparison between the AmoTuner and early-stop mechanism mentioned in #717?
> 
> I run the same benchmark on @meinie0826's PR #717. The result follows
> 
> ```
> End to End Performance
> Default: 154386.97923959792ms
> Multi-Armed-Bandit: 107352.17661622167ms
> Early Stopping: 309926.12229275703.
> ```
> 
> e2e_timecost
> 
> e2e_timecost
> 
> The first step of early stopping takes too much time. I'm unsure whether I run it in a wrong way. I'll do it again to double check the result, and please let @meinie0826 take a look at this.

I noticed that early stop is slower than the default setting in the first round, and I suspect this is because # 717 used align32 and hash key, so the first round is not informative. You can observe the results after the first round to draw a conclusion.

PR #763, opened 2025-07-10
[libtuner]Expand libtuner search and stop interface, add early stop a…

meinie0826:
…nd bayesian strategy

### PR Category

libtuner
### Type of Change

New Feature
### Description

In order to extend the search and stop strategies of libtuner and support different search strategies, I have extended two interfaces using a decorator in this PR. These interfaces are named `register_dearch_strategy` and `register_should_search_hook`. This approach is well compatible with the original libtuner framework and has backward scalability.
This PR provides two examples at the same time, namely Bayesian and early stop, which provide good examples for future developers. These two methods contain some hyperparameters, and I have not conducted extensive testing, so the selection of hyperparameters may not be optimal. @sgjzfzzf  has more experience in choosing hyperparameters, please provide some suggestions.
@sgjzfzzf  I'm not sure if you agree with this extension method and if it's compatible with AmoTuner. Please provide more suggestions.
### Issue
https://github.com/FlagOpen/FlagGems/pull/762

### Progress

### DIFF
--- a/src/flag_gems/utils/libentry.py
+++ b/src/flag_gems/utils/libentry.py
@@ -164,7 +164,7 @@ def register_should_search_hook(name):
     return decorator
 
 
-@register_should_search_hook("default")
+@register_should_search_hook("brute")
 def default_should_search_hook(tuner, key):
     return key not in tuner.cache
 
@@ -347,7 +347,7 @@ class LibTuner(triton.runtime.Autotuner):
         self.search_strategy_name = search_strategy
         self.search_fn = SEARCH_STRATEGIES[search_strategy]
         self.should_search_fn = SHOULD_SEARCH_HOOKS.get(
-            search_strategy, SHOULD_SEARCH_HOOKS["default"]
+            search_strategy, SHOULD_SEARCH_HOOKS["brute"]
         )
         self.call_counts = {}
         self.timings_cache = {}
--- a/tests/test_libentry.py
+++ b/tests/test_libentry.py
@@ -9,6 +9,7 @@ from triton import language as tl
 import flag_gems
 from flag_gems.runtime import torch_device_fn
 from flag_gems.utils import libentry, libtuner
+from flag_gems.utils.libentry import SEARCH_STRATEGIES, SHOULD_SEARCH_HOOKS
 
 
 # not_raises is copied from https://gist.github.com/oisinmulvihill/45c14271fad7794a4a52516ecb784e69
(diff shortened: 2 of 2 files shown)


meinie0826:
The test results for Bayesian, early stop, and critical search are as follows:
The test script can be found at this link:https://gist.github.com/sgjzfzzf/77bd0c1deeb145ca162879c367b6747c
mm_configs
mm_configs2
mm_configs3

sgjzfzzf:
Could you please remove the first step and draw the plot again? I couldn't see any difference from them. And I think it would be more helpful if you could put them in the same plot.

src/flag_gems/utils/libentry.py
```diff
@@ -12,6 +13,9 @@
 from typing import Dict, Optional
 
 import triton
+from skopt import gp_minimize
+from skopt.space import Categorical
+from skopt.utils import use_named_args
```
sgjzfzzf, line 18: Should we consider to add `skopt` to `pyproject.toml`?

src/flag_gems/utils/libentry.py
```diff
@@ -151,13 +156,127 @@ def decorator(fn):
...
+    return key not in tuner.cache
+
+
+@register_should_search_hook("bayesian")
+def bayesian_should_search_hook(tuner, key):
+    if key not in tuner.cache:
+        return True
+    tuner.call_counts[key] = tuner.call_counts.get(key, 0) + 1
+    # Starts with a 20% chance of exploration, decays over calls
+    explore_prob = 0.2 * (0.95 ** (tuner.call_counts[key] - 1))
```
sgjzfzzf, line 178: Do you think we should define these hyperparameters by function arguments so the user could adjust them easily if they want? I know it's a little hard to design an interface to pass them. Just a suggestion.

src/flag_gems/utils/libentry.py
```diff
@@ -151,13 +156,127 @@ def decorator(fn):
...
+    timings = tuner.timings_cache.get(key, {})
+
+    # Define search space
+    search_space = []
+    if not configs:
+        return None, {}
+
+    param_names = list(configs[0].kwargs.keys())
+    for name in param_names:
+        values = sorted(list(set(c.kwargs[name] for c in configs)))
```
sgjzfzzf: Looks like you don't need to convert it into `list`.
```suggestion
        values = sorted(set(c.kwargs[name] for c in configs))
```

src/flag_gems/utils/libentry.py
meinie0826: I have considered this idea and found that if all hyperparameters are placed in the function signature, it is not good. Each strategy requires a hyperparameter, and the function signature is too long. What do you think?

src/flag_gems/utils/libentry.py
meinie0826: Yes, you are right! Thanks!

src/flag_gems/utils/libentry.py
meinie0826: Yes! you are right !

sgjzfzzf:
> > Could you please remove the first step and draw the plot again? I couldn't see any difference from them. And I think it would be more helpful if you could put them in the same plot.
> 
> As I mentioned earlier, the focus of this PR is on expanding the framework rather than search methods. I think the current hyperparameter settings are not good enough and may need your assistance in considering:)

If so I'll recommend you converting this PR into draft. It should be very careful to merge a PR into the main stream. Let's talk about more details on this new feature before the merge. We can continue to work on this branch together and make it ready for the final review.

src/flag_gems/utils/libentry.py
sgjzfzzf: If you could define it as a class, maybe you can pass them to the constructor.

src/flag_gems/utils/libentry.py
```diff
@@ -225,7 +344,13 @@ def __init__(
         assert (
             isinstance(search_strategy, str) and search_strategy in SEARCH_STRATEGIES
         ), "Invalid search strategy"
-        self.search_strategy = SEARCH_STRATEGIES[search_strategy]
+        self.search_strategy_name = search_strategy
```
StrongSpoon, line 347: I suggest setting the default choice of search strategy as "brute"

meinie0826: I wrote this before, and at @sgjzfzzf  strong suggestion, I moved 'bruce' to the function signature:)
```
def libtuner(
    configs,
    key,
    prune_configs_by=None,
    reset_to_zero=None,
    restore_value=None,
    pre_hook=None,
    post_hook=None,
    warmup=25,
    rep=100,
    use_cuda_graph=False,
    do_bench=None,
    strategy=None,
    search_strategy="brute",
):
```

PR #771, opened 2025-07-11
New Interfaces for `LibTuner`

sgjzfzzf:
### PR Category

libtuner

### Type of Change

New Feature

### Description

This is a proposal to provide a new interface to extend the current `LibTuner`. It redefined `LibTuner` as a base class and allows extending it by overriding the `policy` method. For easily usage, we also provide a subclass called `OfflineLibTuner` so our current implementation, including the default one and new strategies, could be added to the new `LibTuner` system by `make` static method. For the future extension including online autotuner, it could be integrated by providing new subclasses. Totally our design principle is to allow different granularity to control the autotuner details.

Except that, we also provide a dispatch mechanism so new policies and `LibTuner`s could be registered easily into the global dispatcher easily and use them with their name.

If this work is approved, future work includes migrate #755, #763, and implement online autotuner #762 on it.

### Issue

This is a competitive proposal against  #763. Please have a detailed discussion on which design we should work on, or anything we can make better.

### Progress


src/flag_gems/utils/libentry.py
```diff
@@ -244,39 +253,133 @@ def get_kernel_hash(self):
...
+            def policy(
+                self,
+                fn: triton.runtime.KernelInterface,
+                configs: Iterator[triton.Config],
+                args: Tuple[Any],
+                kwargs: Dict[str, Any],
+            ) -> Tuple[triton.Config, Dict[str, float]]:
+                return policy(fn, configs, args, kwargs)
+
+        return AnonymousLibTuner
```
tongxin: Hi, Jinjie. Can the local class be accessible outside the static method scope?

sgjzfzzf: I don't think it should be available for the outside scape, but if it's needed, they can still be accessed by their keys through `LibTuner.get`.

sgjzfzzf: I rewrite this part as a decorator.

src/flag_gems/utils/libentry.py
```diff
@@ -244,39 +253,133 @@ def get_kernel_hash(self):
...
+            if self.strategy is None
+            else tuple(
+                strategy(arg)
+                for strategy, arg in starmap(
+                    lambda idx0, idx1: (STRATEGY[self.strategy[idx0]], args[idx1]),
+                    enumerate(self.keys),
+                )
+            )
+            + tuple(str(arg.dtype) for arg in args.values() if hasattr(arg, "dtype"))
+        )
```
tongxin: I think ternary expression only fits for oneliners. This looks daunting.. 
```
if self.strategy is None:
    key = tuple(args[k] for k in self.keys if k in args)
else:
    key = tuple(startmap(lambda idx0, idx1: (STRATEGY[self.strategy[idx0]])(args[idx1]), enumerate(self.keys))
```

sgjzfzzf: Fix.

src/flag_gems/utils/libentry.py
tongxin: Yeah, the decorator version is more straightforward.

src/flag_gems/utils/libentry.py
tongxin: We missed the other part  `+ tuple(str(arg.dtype) for arg in args.values() if hasattr(arg, "dtype"))`.  Can you add it back?

sgjzfzzf: My bad. Sorry for the carelessness. Fix.

tongxin (dismissed):
LGTM

src/flag_gems/utils/libentry.py
```diff
@@ -245,38 +238,120 @@ def get_kernel_hash(self):
...
-            v = s(args[k])
-            key.append(v)
+            key = tuple(args[k] for k in self.keys if k in args)
+        else:
+            key = tuple(
+                starmap(
+                    lambda idx0, idx1: self.strategy[idx0](args[idx1]),
+                    enumerate(self.keys),
+                )
+            ) + tuple(str(arg.dtype) for arg in args.values() if hasattr(arg, "dtype"))
```
StrongSpoon: append the data types for both status

sgjzfzzf: Fix.

StrongSpoon (dismissed):
lgtm

src/flag_gems/utils/libentry.py
```diff
@@ -245,38 +238,121 @@ def get_kernel_hash(self):
...
-            key.append(v)
+            key = tuple(args[k] for k in self.keys if k in args)
+        else:
+            key = tuple(
+                starmap(
+                    lambda idx0, idx1: self.strategy[idx0](args[idx1]),
+                    enumerate(self.keys),
+                )
+            )
+        key += tuple(str(arg.dtype) for arg in args.values() if hasattr(arg, "dtype"))
```
tongxin, line 249: Is it OK that `self.keys` is not filtered against `args` in the else branch?

src/flag_gems/utils/libentry.py
```diff
@@ -245,38 +238,121 @@ def get_kernel_hash(self):
...
+    ) -> Type["LibTuner"]:
+        """A decorator to register a policy for `LibTuner`.
+
+        This decorator allows you to create a new `LibTuner` subclass without defining a new class explicitly.
+        The new subclass will have the `policy` method set to the provided policy function and will be registered under
+        the specified name in the `LibTuner` dispatch table.
+        """
+
+        def decorator(
+            policy: Callable[
```
tongxin: What about renaming it as `policy_impl` to distinguish from `AnonymousLibTunerImpl`'s `policy` method. It's bit confusing.

src/flag_gems/utils/libentry.py
```diff
@@ -245,38 +238,121 @@ def get_kernel_hash(self):
...
+                    super().__init__(*args, **kwargs)
+
+                def policy(
+                    self,
+                    fn: triton.runtime.KernelInterface,
+                    configs: Iterator[triton.Config],
+                    args: Tuple[Any],
+                    kwargs: Dict[str, Any],
+                ) -> Tuple[triton.Config, Dict[str, float]]:
+                    return policy(fn, configs, args, kwargs)
```
tongxin: Calling `policy_impl` instead.

src/flag_gems/utils/libentry.py
```diff
@@ -245,38 +238,121 @@ def get_kernel_hash(self):
...
+        def decorator(
+            strategy: Union[Callable[[Any], Any], List[Callable[[Any], Any]]],
+        ):
+            LibTuner._strategy_table[name] = strategy
+            return strategy
+
+        return decorator
+
     def run(self, *args, **kwargs):
         self.nargs = dict(zip(self.arg_names, args))
```
tongxin, line 342: Please add a comment for clarification, noting that arg_names correspond to a JITFunction's parameter names in its signature.

src/flag_gems/utils/libentry.py
```diff
@@ -220,17 +209,21 @@ def __init__(
...
+        assert len(strategy) == len(
+            self.keys
+        ), f"the length of strategy {len(strategy)} must match the length of keys {len(self.keys)}"
+        strategy: List[Callable[[Any], Any]] = [
+            LibTuner.get_strategy(s) if isinstance(s, str) else s for s in strategy
+        ]
         self.strategy = strategy
         # Use table name with hash instead of hash in key
         self.kernel_hash = None
         self.table_name = f"{self.__name__}_{self.get_kernel_hash()}"
```
tongxin, line 225: To deal with re-entrance is similar to what jit's _do_compile does in the async_mode, where the cache_key is voluntarily calculated with `get_cache_key `.  I think it's more appropriate than just reusing cache key?
```
cache_key = get_cache_key(src, backend, options, env_vars)
```

src/flag_gems/utils/libentry.py
sgjzfzzf: I think so. The previous implementation doesn't check `args` and works well. Do you think I should add it here?

tongxin: Not a big issue assuming `self.key` not including kwargs.

src/flag_gems/utils/libentry.py
sgjzfzzf: Take a look plz. I'm unsure whether it meets your requirements. Thanks!

src/flag_gems/utils/libentry.py
sgjzfzzf: Fix.

src/flag_gems/utils/libentry.py
sgjzfzzf: Fix.

tongxin (approved):

PR #792, opened 2025-07-16
【c++ wrapper】Add compatibility control for C extension usage

kiddyjinjin:
### PR Category

User Experience

### Type of Change

Added global flags:  
  - `has_c_extension`: whether the C extension is importable  
  - `use_c_extension`: whether to actually use the C extension based on import success and env var

### Progress

### DIFF
--- a/src/flag_gems/__init__.py
+++ b/src/flag_gems/__init__.py
@@ -1,14 +1,7 @@
 import logging
+import os
 
 import torch
-
-# C extensions
-try:
-    from flag_gems import ext_ops  # noqa: F401
-
-    has_c_extension = True
-except ImportError:
-    has_c_extension = False
 
 from . import testing  # noqa: F401
 from . import runtime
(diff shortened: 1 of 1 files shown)


src/flag_gems/__init__.py
```diff
@@ -19,6 +12,20 @@
...
+
+try:
+    from flag_gems import c_operators
+
+    has_c_extension = True
+except ImportError:
+    c_operators = None  # avoid import error if c_operators is not available
+    has_c_extension = False
+
+if has_c_extension and os.environ.get("USE_C_EXTENSION", "0") == "1":
```
sgjzfzzf, line 26: Should we give the users a warning if they enable the C extension through the environment variable but no extensions found?

src/flag_gems/__init__.py
```diff
@@ -19,6 +12,20 @@
...
 
+has_c_extension = False
+use_c_extension = False
+
+try:
+    from flag_gems import c_operators
+
+    has_c_extension = True
+except ImportError:
+    c_operators = None  # avoid import error if c_operators is not available
```
sgjzfzzf, line 23: Is this line necessary? If `has_c_extension` is set as `False`, it looks like `c_operators` won't be accessed.

src/flag_gems/__init__.py
tongxin: I would vote for bail out if users require `use_c_extension` and import fails.

tongxin: Another question I want to raise is should we consider enabling C operators and registering C operators as seprate use cases?

tongxin: I would propose adding one more argument in `flag_gems.enable` to flag C operator registration.

src/flag_gems/__init__.py
kiddyjinjin: I agree with your point. Unfortunately, I had to include this line to make the pre-commit check pass.

src/flag_gems/__init__.py
kiddyjinjin: > Should we give the users a warning if they enable the C extension through the environment variable but no extensions found?

This has been fixed in PR  #824    — you can check it out there.

PR #816, opened 2025-07-22
[Operator] Add max in C++

scatyf3:
### PR Category
Operator and OP Test 

### Type of Change
New Feature

### Description
- Add max in C++ and it's unit test
- Fix Cmake flag error in document

### Progress

### DIFF
--- a/ctests/test_triton_reduction.cpp
+++ b/ctests/test_triton_reduction.cpp
@@ -30,3 +30,29 @@ TEST(reduction_op_test, nonzero) {
 
   EXPECT_TRUE(torch::allclose(out_torch, out_triton, 1e-5, 1e-8));
 }
+
+TEST(reduction_op_test, max_keepdim) {
+    const torch::Device device(torch::kCUDA, 0);
+    for (bool keepdim : {false, true}) {
+        torch::Tensor a = torch::randn({32, 1024}, device);
+        auto out_torch = at::max(a, 1, keepdim); 
+        torch::Tensor max_torch = std::get<0>(out_torch);
+        torch::Tensor index_torch = std::get<1>(out_torch);
+        auto out_triton = flag_gems::max_dim(a, 1, keepdim, std::nullopt); 
+        torch::Tensor max_triton = std::get<0>(out_triton);
+        torch::Tensor index_triton = std::get<1>(out_triton);
+        LOG(INFO) << "Input shape: " << a.sizes();
+        LOG(INFO) << "Max Torch shape: " << max_torch.sizes();
+        LOG(INFO) << "Index Torch shape: " << index_torch.sizes();
+        LOG(INFO) << "Max Triton shape (keepdim=" << keepdim << "): " << max_triton.sizes();
+        LOG(INFO) << "Index Triton shape (keepdim=" << keepdim << "): " << index_triton.sizes();
+        if (!torch::allclose(max_torch, max_triton, 1e-5, 1e-8)) {
+            LOG(INFO) << "Max value difference (keepdim=" << keepdim << "):\n" << max_torch - max_triton;
+        }
+        EXPECT_TRUE(torch::allclose(max_torch, max_triton, 1e-5, 1e-8));
+        if (!torch::allclose(index_torch, index_triton, 1e-5, 1e-8)) {
+            LOG(INFO) << "Index difference (keepdim=" << keepdim << "):\n" << index_torch - index_triton;
+        }
+        EXPECT_TRUE(torch::allclose(index_torch, index_triton, 1e-5, 1e-8));
+    }
+}
--- a/lib/max.cpp
+++ b/lib/max.cpp
@@ -1,0 +1,93 @@
+#include "flag_gems/operators.h"
+#include "flag_gems/utils.h"
+
+#include <iostream>
+#include "c10/cuda/CUDAStream.h"
+#include "torch/torch.h"
+#include "triton_jit/triton_jit_function.h"
+
+#include <filesystem>
+#include "ATen/WrapDimUtils.h"
+#include "ATen/native/ReduceOpsUtils.h"
+#include "c10/util/DimVector.h"
+
+
+
+namespace {
+std::tuple<at::Tensor, int64_t, int64_t> permute_reduction_axes_right(
+    const at::Tensor &tensor, at::OptionalIntArrayRef reduction_axes_opt) {
+  int64_t dim = tensor.dim();
+  c10::DimVector reduction_axes;
+
+  if (reduction_axes_opt.has_value()) {
+    reduction_axes = reduction_axes_opt.value().vec();
+  }
+
+  std::unordered_set<int64_t> reduction_set(reduction_axes.begin(), reduction_axes.end());
+
+  c10::DimVector left_axes, right_axes;
+  int64_t non_reduction_size = 1, reduction_size = 1;
+
+  for (int64_t i = 0; i < dim; ++i) {
+    if (reduction_set.count(i)) {
+      right_axes.push_back(i);
+      reduction_size *= tensor.size(i);
+    } else {
+      left_axes.push_back(i);
+      non_reduction_size *= tensor.size(i);
+    }
+  }
+
+  // Concatenate left and right axes to form the new permutation order
+  c10::DimVector permute_order = left_axes;
+  permute_order.insert(permute_order.end(), right_axes.begin(), right_axes.end());
+
+  return {tensor.permute(permute_order), non_reduction_size, reduction_size};
+}
+}  // anonymous namespace
+
+
+namespace flag_gems {
+    using namespace triton_jit;
+    ::std::tuple<at::Tensor, at::Tensor> max_dim(const at::Tensor &self,
+                        at::OptionalIntArrayRef dim,
+                        bool keepdim,
+                        ::std::optional<at::ScalarType> dtype){
+        at::DimVector dims_ = at::native::make_dim_vector(dim, self.dim());
+        at::maybe_wrap_dims(dims_, self.dim());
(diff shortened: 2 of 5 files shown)


ctests/test_triton_reduction.cpp
```diff
@@ -30,3 +30,29 @@ TEST(reduction_op_test, nonzero) {
 
   EXPECT_TRUE(torch::allclose(out_torch, out_triton, 1e-5, 1e-8));
 }
+
+TEST(reduction_op_test, max_keepdim) {
+    const torch::Device device(torch::kCUDA, 0);
+    for (bool keepdim : {false, true}) {
```
iclementine: For multiple tests, it is recommended to use parametrized tests.

https://google.github.io/googletest/reference/testing.html

Also, there are some examples 
https://github.com/FlagOpen/FlagGems/blob/e6ec851e9b4da5abaf9117384cc28e3fa938df39/ctests/test_triton_blas.cpp#L25C1-L43C1

triton_src/max.py
```diff
@@ -0,0 +1,91 @@
...
+from flag_gems.utils import dim_compress, libentry, libtuner
+from flag_gems.utils import triton_lang_extension as tle
+from flag_gems.utils.limits import get_dtype_min
+
+logger = logging.getLogger(__name__)
+
+
+@libentry()
+@triton.jit
+def max_kernel_1(
```
iclementine: Thanks for being fast-pacing, but we can now use triton jit function from src/flag_gems/ops now. See examples in 

https://github.com/FlagOpen/FlagGems/pull/817

lib/max.cpp
```diff
@@ -0,0 +1,93 @@
...
+  // Concatenate left and right axes to form the new permutation order
+  c10::DimVector permute_order = left_axes;
+  permute_order.insert(permute_order.end(), right_axes.begin(), right_axes.end());
+
+  return {tensor.permute(permute_order), non_reduction_size, reduction_size};
+}
+}  // anonymous namespace
+
+
+namespace flag_gems {
```
iclementine, line 43: Please install pre-commit in your environment to apply hooks before committing code. It enforces some code style checking and applies reformatting.

lib/max.cpp
```diff
@@ -0,0 +1,93 @@
...
+    using namespace triton_jit;
+    ::std::tuple<at::Tensor, at::Tensor> max_dim(const at::Tensor &self,
+                        at::OptionalIntArrayRef dim,
+                        bool keepdim,
+                        ::std::optional<at::ScalarType> dtype){
+        at::DimVector dims_ = at::native::make_dim_vector(dim, self.dim());
```
iclementine: The example `sum.cpp` may not be a good example, as I have not implemented the cases where dim has no value. In this case, torch would perform a full reduction here.

We can discuss it in detail.

lib/max.cpp
scatyf3: If I understand correctly, `permute_reduction_axes_right` is a function for all reduce kernel. We either need to refactor this part of the code, similar to the Python side in `src/flag_gems/utils`, or since we are still in the early stages of development, there's no rush to introduce this complexity. Any suggestion?

lib/max.cpp
scatyf3: Refer to `src/flag_gems/ops/max.py`, max has 2 types:
- reduce to a specified dimension
- `(1)` tensor.
I will implement both of them in the following commit, and also add  sum to a `(1)` tensor warpper in sum.cpp

iclementine: Yes. But to be clear, the second case is a reduction to scalar tensor, a.k.a. tensor with shape `()`. Since shape `(1,)` and `(1, 1)` or `(1, 1, ....)` all have a single element, but we only refer to the tensor with shape `()` as a scalar tensor.

lib/max.cpp
iclementine: I would prefer putting the reduction operators in a single source file, and have a reduction_utils.h/cpp.

But yes, we may change the way we implement reduction operators, so let's keep it for now.

lib/max.cpp
```diff
@@ -44,50 +42,93 @@ std::tuple<at::Tensor, int64_t, int64_t> permute_reduction_axes_right(
...
+                                             ::std::optional<at::ScalarType> dtype) {
+  at::DimVector dims_ = at::native::make_dim_vector(dim, self.dim());
+  at::maybe_wrap_dims(dims_, self.dim());
+  at::DimVector shape = at::meta::get_reduction_shape(self, dims_, keepdim, false);
+  at::Tensor out_value = at::empty(shape, self.options());
+  at::Tensor out_index = at::empty(shape, self.options().dtype(at::kLong));
+  auto [permuted_self, non_reduction_size, reduction_size] = permute_reduction_axes_right(self, dims_);
+  permuted_self = permuted_self.contiguous();
+  const TritonJITFunction &f =
+      TritonJITFunction::getInstance(std::string(utils::get_triton_src_path() / "max.py"), "max_kernel");
```
tongxin: I guess it's OK to use gems kernels with `get_flag_gems_src_path() / "ops"` by default.

triton_src/max.py
iclementine: If this file is no longer needed, we  can delete it.

scatyf3: OK, I removed that

lib/max.cpp
```diff
@@ -0,0 +1,137 @@
...
+#include "triton_jit/triton_jit_function.h"
+
+#include <filesystem>
+#include "ATen/WrapDimUtils.h"
+#include "ATen/native/ReduceOpsUtils.h"
+#include "c10/util/DimVector.h"
+
+namespace {
+std::tuple<at::Tensor, int64_t, int64_t> permute_reduction_axes_right(
+    const at::Tensor &tensor, at::OptionalIntArrayRef reduction_axes_opt) {
```
iclementine: You can simplify 的 reduction_axes to an integer, since we are implementing these 2 overloads. (Although there are more overloads for `max`).

Check `aten/src/ATen/native/native_functions.yaml` for more details.

- func: max(Tensor self) -> Tensor
- func: max.dim(Tensor self, int dim, bool keepdim=False) -> (Tensor values, Tensor indices)

iclementine: Actually torch has a `structured delegate` mechanism to delegate `max.dim` to `max.dim_max` where the output values and indices are passed in as output parameters. We may also implement that.

include/flag_gems/operators.h
```diff
@@ -14,7 +14,11 @@ at::Tensor sum_dim(const at::Tensor &self,
                    at::OptionalIntArrayRef dim,
                    bool keepdim = false,
                    ::std::optional<at::ScalarType> dtype = ::std::nullopt);
-
+std::tuple<at::Tensor, at::Tensor> max_dim(const at::Tensor &self,
+                                           at::OptionalIntArrayRef dim,
+                                           bool keepdim,
+                                           ::std::optional<at::ScalarType> dtype);
+at::Tensor max(const at::Tensor &self);
```
iclementine, line 18: We also need to register those operators to torch libraries `flag_gems` and `aten`.

See an example in `https://github.com/FlagOpen/FlagGems/blob/master/src/flag_gems/csrc/cstub.cpp` for registering to `flag_gems`

```c++
TORCH_LIBRARY(flag_gems, m) {
  // blas
  m.def("addmm(Tensor self, Tensor mat1, Tensor mat2, *, Scalar beta=1, Scalar alpha=1) -> Tensor");
}
```

and
```c++
TORCH_LIBRARY_IMPL(flag_gems, CUDA, m) {
  // blas
  m.impl("addmm", TORCH_FN(addmm));
}
```

and also registering to `aten`

https://github.com/FlagOpen/FlagGems/blob/master/src/flag_gems/csrc/aten_patch.cpp

```c++
namespace flag_gems {
TORCH_LIBRARY_IMPL(aten, CUDA, m) {
  REGISTER_AND_LOG("addmm", addmm);
  REGISTER_AND_LOG("bmm", bmm);
  REGISTER_AND_LOG("mm", mm_tensor);
}
```

ctests/test_triton_reduction.cpp
```diff
@@ -8,7 +8,7 @@ TEST(reduction_op_test, sum) {
 
   torch::Tensor out_torch = at::sum(a, {1});
   torch::Tensor out_triton = flag_gems::sum_dim(a, {1});
-
+  torch::cuda::synchronize();
```
iclementine: I think explicit synchronization is not necessary.

src/flag_gems/csrc/cstub.cpp
```diff
@@ -28,6 +29,7 @@ TORCH_LIBRARY(flag_gems, m) {
       "zeros(SymInt[] size, ScalarType? dtype=None,Layout? layout=None, Device? device=None, bool? "
       "pin_memory=None) -> Tensor");
   m.def("sum.dim_IntList(Tensor self, int[1]? dim, bool keepdim=False, *, ScalarType? dtype=None) -> Tensor");
+  m.def("max.dim(Tensor self, int dim, bool keepdim=False) -> (Tensor values, Tensor indices)");
```
iclementine, line 33: We can define both operator overloads and implement them.

max and max.dim

src/flag_gems/csrc/aten_patch.cpp
```diff
@@ -29,5 +29,6 @@ TORCH_LIBRARY_IMPL(aten, CUDA, m) {
   REGISTER_AND_LOG("addmm", addmm);
   REGISTER_AND_LOG("bmm", bmm);
   REGISTER_AND_LOG("mm", mm_tensor);
+  REGISTER_AND_LOG("max.dim", max_dim)
```
iclementine, line 32: Also, here. We can register "max", too.

iclementine (approved):
LGTM

PR #817, opened 2025-07-22
[Cpp wrappers]Refactor cpp operator impl and registration 

iclementine:
### PR Category

Operator

### Type of Change

Refactor

### Description

1. Make cpp wrappers use the same triton jit functions as their python counterparts(use environment variable "FLAGGEMS_SOURCE_DIR" to specify the src dir of flag_gems when using from cpp instead of python package, so the ctest now needs an environment variable to be set);
2. split the C extension into 2 parts; aten_patch and c_operators; aten_patch includes patch to aten library and an empty python module, while c_operators includes a torch library flag_gems and a python module. The split is intended for better control on whether to use a specific library.

### Progress

### DIFF
--- a/lib/mm.cpp
+++ b/lib/mm.cpp
@@ -35,7 +35,7 @@ at::Tensor mm_tensor(const at::Tensor &mat1, const at::Tensor &mat2) {
     /* grid_y = */ 1,
     /* grid_z = */ 1,
     /* num_warps = */ 4,
-    /* num_stages = */ 4,
+    /* num_stages = */ 2,
     mat1,
     mat2,
     out,


lib/addmm.cpp
```diff
@@ -26,19 +26,22 @@ at::Tensor addmm(const at::Tensor &self,
...
+  int BLOCK_N = 64;
+  unsigned int grid_x = ((mat1_sizes[0] + BLOCK_M - 1) / BLOCK_M);
+  unsigned int grid_y = ((mat2_sizes[1] + BLOCK_N - 1) / BLOCK_N);
   f(/* CUstream = */ raw_stream,
     /* grid_x = */ grid_x,
     /* grid_y = */ grid_y,
     /* grid_z = */ 1,
-    /* num_warps = */ 4,
-    /* num_stages = */ 1,
+    /* num_warps = */ 2,
```
tongxin, line 43: Is 2 warps good?

src/flag_gems/ops/mm.py
```diff
@@ -12,13 +12,18 @@
 logger = logging.getLogger(__name__)
 
 
+@triton.jit
+def prev_multiple_of(a, b):
```
tongxin, line 16: What about `round_down_to_multiple_of`?

src/flag_gems/ops/mm.py
```diff
@@ -57,37 +58,40 @@ def mm_kernel(
...
-            b = tl.load(B)
-        else:
-            k_remaining = K - k * (BLOCK_K * SPLIT_K)
-            _0 = tl.zeros((1, 1), dtype=C.dtype.element_ty)
-            a = tl.load(A, mask=rk[None, :] < k_remaining, other=_0)
-            b = tl.load(B, mask=rk[:, None] < k_remaining, other=_0)
+    prev_multiple = prev_multiple_of(K, BLOCK_K)
+
+    acc = tl.zeros((BLOCK_M, BLOCK_N), dtype=tl.float32)
+    for start_k in range(0, prev_multiple, BLOCK_K):
```
tongxin, line 64: I think it's more clear to just step range over K and then test if a trailing iteration is needed.
```
    start_k = 0
    for _ in range(0, K, BLOCK_K):
        ...
        start_k += BLOCK_K
  
    if start_k < K:
        ...
```

lib/addmm.cpp
iclementine: Before we have implemented the functionality to partition the L1 data cache and shared memory to prefer shared memory, and the required shared memory is within the increased limit, we can only use num-warps=2 in this case for RTX-3090.

src/flag_gems/ops/mm.py
iclementine: There are multiple copy of this utility function now. We may considering moving it to a shared utility module.

PR #824, opened 2025-07-24
[new feature] enable c extension from python

kiddyjinjin:
### PR Category

User Experience

### Type of Change

New Feature 

### Description

This PR introduces support for optionally enabling c_extension functionalities in FlagGems. The key updates include:

This PR introduces support for optionally enabling `c_extension` functionalities in FlagGems. The key updates include:

- **Python-side access to registered C++ ops**  
  Added `get_registered_ops()` in `aten_patch` to expose the list of ops overridden by `aten_patch`.  
  Python can now query this list to determine which ops have c_extension  implementations.

- **Avoid redundant registration**  
  Python op registration is skipped if a C++ implementation already exists in `aten_patch`, reducing overhead and avoiding conflicts.

- **Environment variable for unified control**  
  Introduced the `USE_C_EXTENSION` environment variable. Set to `"1"` to enable loading of `c_extension`; otherwise, it will be disabled by default.

- **Updated documentation**  
  The `how_to_use_flaggems.md` document is updated to explain how to enable or disable `c_extension` and the behavior change in op registration.

This change improves flexibility and allows runtime control over whether to use native C++ operator overrides.

### Progress


meinie0826:
LGTM

meinie0826 (dismissed):

src/flag_gems/config.py
```diff
@@ -0,0 +1,23 @@
...
+
+try:
+    from flag_gems import c_operators
+
+    has_c_extension = True
+except ImportError:
+    c_operators = None  # avoid import error if c_operators is not available
+    has_c_extension = False
+
+if has_c_extension and os.environ.get("USE_C_EXTENSION", "0") == "1":
```
iclementine: I think it is better to be explicit, for example "FLAGGEMS_PATCH_ATEN".

Also, maybe I shall rename `c_operators` to `flag_gems_c`.

kiddyjinjin: As we just discussed, I think it's better not to introduce a separate `FLAGGEMS_PATCH_ATEN` environment variable. Having too many environment variables may confuse users who are not familiar with the internal implementation. Instead, we use the existing `USE_C_EXTENSION` to uniformly control the switch between the Python and C++ implementations.

kiddyjinjin: Regarding renaming `c_operators` to `flag_gems_c`, I think that's a good idea

src/flag_gems/__init__.py
```diff
@@ -301,7 +294,9 @@ def enable(
             ("zeros", zeros),
             ("zeros_like", zeros_like),
         ),
-        user_unused_ops_list=[] if unused is None else unused,
+        user_unused_ops_list=list(
+            set([] if unused is None else unused) & set(aten_patch_list)
+        ),
```
iclementine, line 299: Does it mean that operators in unused or aten_patch_list should be excluded from `enable`?

If so, then I think it should be the union of the two sets, rather than the intersection of them.

kiddyjinjin: you are right!

src/flag_gems/config.py
iclementine: Yes. Here is the rationale.

For user of flag_gems library, there are several ways to use the operators:

- directly use: `flag_gems.op_name`, `torch.ops.flag_gems.op_name`, flag_gems nn modules;
- apply patch and use it implicitly: patch aten, patch vllm, etc.

Whether to apply the patches is implemented as a user setting, for example `flag_gems.enable()` (we may better rename it to patch_aten), and `flag_gems.apply_gems_patches_to_vllm` (we may better rename it to patch_vllm). Those settings express patching explicitly.

Whether to use operator implementation in C is a configuration intended for finer control of how flag_gems is used in both cases(directly use and via patch). If USE_C_EXTENSION is True, `flag_gems.op_name`, `torch.ops.flag_gems.op_name` and flag_gems nn modules may switch to the cpp implementation if available. Also, if the patches are applied, `torch.op_name` and `torch.ops.aten.op_name` and vllm modules may also switch to the cpp implementation of flag_gems.

kiddyjinjin: Totally agree

iclementine (approved):
LGTM

meinie0826 (approved):

PR #830, opened 2025-07-25
[Benchmark] support profiling wrapper cpu time cost

StrongSpoon:
### PR Category

Benchmark
### Type of Change

New Feature
### Description

set new choices 'kernel', 'operator' and 'wrapper' for option mode.
'kernel' mode: get the latency of triton kernel on accelerator
'operator' mode: get the latency of gems operator from end to end on cpu
'wrapper' mode: get the latency of gems operator implemented by python or cpp
### Progress

### DIFF
--- a/benchmark/attri_util.py
+++ b/benchmark/attri_util.py
@@ -138,6 +138,12 @@ def get_recommended_shapes(
     return _shapes_sort(DEFAULT_SHAPES)
 
 
+class BenchMode(Enum):
+    KERNEL = "kernel"
+    OPERATOR = "operator"
+    WRAPPER = "wrapper"
+
+
 class BenchLevel(Enum):
     COMPREHENSIVE = "comprehensive"
     CORE = "core"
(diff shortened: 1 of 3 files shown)


benchmark/attri_util.py
```diff
@@ -138,6 +138,12 @@ def _shapes_sort(shapes):
     return _shapes_sort(DEFAULT_SHAPES)
 
 
+class BenchMode(Enum):
+    KERNEL = "kernel"
+    OPERATOR = "operator"
+    WRAPPER = "wrapper"
+
+
```
iclementine, line 146: Good to have an enum.

iclementine: We may better rename it to 

- device timer: time that records timestamp from device;
- synchronized timer: time that records timestamp (from host) after synchronization(actually the same timestamp from host or device, since the synchronization is done);
- cpu timer: time that records timestamp from cpu(host);

StrongSpoon: device, e2e, host?

iclementine (approved):
LGTM

PR #832, opened 2025-07-27
[mthreads] musa backend adaptation

Kylin1207:
### PR Category

OP Test 
### Description

musa backend adaptation
1. enable concat_and_cache_mla op
2. enable diagonal op
3. disable conv1d op temporarily
### Progress


tests/test_attention_ops.py
```diff
@@ -901,7 +901,10 @@ def test_flash_attn_varlen_func_swap_qg(
 BLOCK_SIZES_MLA = [16]
 NUM_BLOCKS_MLA = [8]
 SEEDS = [0]
-CUDA_DEVICES = [f"cuda:{i}" for i in range(1 if torch.cuda.device_count() == 1 else 2)]
+DEVICES = [
+    flag_gems.device if flag_gems.device == "musa" else f"cuda:{i}"
+    for i in range(1 if torch.cuda.device_count() == 1 else 2)
+]
```
tongxin: Just fyi, the orignal device ids were picked for pinning ci jobs on certain ranks of our test machine. This rule may not apply to other backends.

meinie0826: I have noticed that other tests only use flaggems.device like device=flag_gems.device directly. Is it more concise?

tongxin: Not if you want to run tests on multiple devices. But I'm not sure why we need to replicate the same tests on two devices.

tongxin: I guess the intention was to be able to use either one from [rank0, rank2] if multiple devices are available.  We should change the code.

Kylin1207: Thx, I'll revert the changes to this file and just keep the musa backend custom op.

tongxin: It;s OK to change the test file. I think there;s an issue with our original version. Feel free to make corrections.

meinie0826: Yes, it makes sense.

PR #838, opened 2025-07-29
Fix: logger conflict by import vllm

meinie0826:
### PR Category

Benchmark 
### Type of Change

Bug Fix 
### Description

The bug was import by flash_attn_varlen_func benchmark. 
flash_attn_varlen_func test will try to import vllm, and then all pytest logger config will be changed by unknown reason.
### Progress

### DIFF
--- a/benchmark/performance_utils.py
+++ b/benchmark/performance_utils.py
@@ -52,8 +52,7 @@ def SkipVersion(module_name, skip_pattern):
         raise ValueError("Cannot parse version number from skip_pattern.")
 
     try:
-        module = importlib.import_module(module_name)
-        version = module.__version__
+        version = importlib.metadata.version(module_name)
         major, minor = map(int, version.split(".")[:2])
     except Exception:
         raise ImportError(f"Cannot determine version of module: {module_name}")
--- a/benchmark/test_attention_perf.py
+++ b/benchmark/test_attention_perf.py
@@ -294,6 +294,9 @@ class FlashAttnVarlenBenchmark(Benchmark):
 @pytest.mark.skipif(flag_gems.vendor_name == "cambricon", reason="TypeError")
 @pytest.mark.flash_attn_varlen_func
 def test_perf_flash_attn_varlen_func():
+    import os
+
+    os.environ["VLLM_CONFIGURE_LOGGING"] = "0"
     from vllm.vllm_flash_attn.flash_attn_interface import flash_attn_varlen_func
 
     bench = FlashAttnVarlenBenchmark(


benchmark/test_attention_perf.py
```diff
@@ -294,8 +294,20 @@ def flash_attn_varlen_input_fn(self, config, dtype, device):
...
+    from benchmark.conftest import Config
+
+    logging.basicConfig(
+        filename="result_{}.log".format("_".join(Config.cmd_args)).replace("_-", "-"),
+        filemode="a",
+        level=logging.INFO,
+        format="[%(levelname)s] %(message)s",
+        force=True,
+    )
+
```
awayzjj: Hello, I was wondering since we can retrieve the version of vllm using `version = importlib.metadata.version(module_name)` without importing the vllm module, why do we still need to call logging.basicConfig() here?

meinie0826: Actually, I have found that `from vllm.vllm_flash_attn.flash_attn_interface import flash_attn_varlen_func`  also import vllm, and then the gems logger will lose efficacy.

awayzjj: Got it! Thanks.

benchmark/performance_utils.py
```diff
@@ -52,8 +52,7 @@ def SkipVersion(module_name, skip_pattern):
         raise ValueError("Cannot parse version number from skip_pattern.")
 
     try:
-        module = importlib.import_module(module_name)
-        version = module.__version__
+        version = importlib.metadata.version(module_name)
```
iclementine, line 55: Actually, a package's __version__ is arbitrarily defined. It may not agree with the version number of the package.

benchmark/test_attention_perf.py
```diff
@@ -294,6 +294,9 @@ def flash_attn_varlen_input_fn(self, config, dtype, device):
 @pytest.mark.skipif(flag_gems.vendor_name == "cambricon", reason="TypeError")
 @pytest.mark.flash_attn_varlen_func
 def test_perf_flash_attn_varlen_func():
+    import os
+
+    os.environ["VLLM_CONFIGURE_LOGGING"] = "0"
```
iclementine, line 299: It sounds like `vllm` has made some modifications to the root logger so that the logger used in the benchmark is affected. Maybe we can use a dedicated logger for the benchmark to avoid such effects.

iclementine (approved):

benchmark/test_attention_perf.py
tongxin: vllm controls the root logger by default. I agree that we should use a dedicated logger for gems.

PR #839, opened 2025-07-30
Support triton 3.4

meinie0826:
### PR Category

Other
### Type of Change

Bug Fix
### Description

Support triton 3.4
### Progress

### DIFF
--- a/src/flag_gems/utils/libentry.py
+++ b/src/flag_gems/utils/libentry.py
@@ -30,6 +30,7 @@ ATTRS = {
     (3, 1): 4,
     (3, 2): 4,
     (3, 3): 8,
+    (3, 4): 4,
 }
 # Set (3, 2) to 9 for cambricon (special Autotune config)
 if vendor_module.vendor_info.vendor_name == "cambricon":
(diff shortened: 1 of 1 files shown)


src/flag_gems/utils/libentry.py
```diff
@@ -588,7 +589,7 @@ def run(self, *args, **kwargs):
                 k_args[param_names[i]] = arg
                 dns_args.append(arg)
             else:
-                if major_version == 3 and minor_version == 3:
+                if major_version == 3 and minor_version >= 3:
```
sgjzfzzf: I'll suggest replace it with
```Python
if major_version == 3 and 3 <= minor_version <=4:
```
we don't know whether the configs will be updated in version 3.5.0. We'd better check them when it's published instead of assuming they won't be changed. Or at least you can provide a warning here if it's not checked by us manually. (It's very painful to be confused by the update of triton `Config`😮‍💨

meinie0826: Yes, you are right!

sgjzfzzf (approved):

PR #840, opened 2025-07-31
[AdvancedCompiler]Div&div (cpp wrapper)

AdvancedCompiler:
### PR Category

Operator
### Type of Change

New Feature
### Description

CPP wrapper packaging for div&div_ op.
### Progress


lib/div.cpp
```diff
@@ -0,0 +1,457 @@
...
+
+at::Tensor div_mode(const at::Tensor& a_,
+                    const at::Tensor& b_,
+                    const c10::optional<std::string>& rounding_mode) {
+  auto res = torch::broadcast_tensors({a_, b_});
+  at::Tensor a = res[0].contiguous();
+  at::Tensor b = res[1].contiguous();
+
+  at::ScalarType out_dtype = at::promote_types(a.scalar_type(), b.scalar_type());
+  at::Tensor out = at::empty(a.sizes(), at::TensorOptions().dtype(out_dtype).device(a.device()));
```
tongxin, line 321: The output shape is `res.sizes()`, isn't it?

tongxin: Nevermind, `a` is already not the original input.

lib/div.cpp
```diff
@@ -0,0 +1,457 @@
...
+    kernel_name = "true_div_kernel";
+  } else {
+    TORCH_CHECK(false, "div_mode rounding_mode must be 'floor', 'trunc', or empty.");
+  }
+  const TritonJITFunction& f =
+      TritonJITFunction::getInstance((utils::get_triton_src_path() / "div.py").string(), kernel_name);
+
+  int64_t tile_size = 1024;
+  int64_t n = out.numel();
+  const unsigned int num_blocks = (n + tile_size - 1) / tile_size;
```
tongxin: On NVidia GPU grid.x's limit is 2^31 - 1. Though it's a super narrow chance to overflow, but using signed int is better here.

lib/div.cpp
```diff
@@ -0,0 +1,457 @@
...
+
+#include <c10/cuda/CUDAStream.h>
+#include <c10/util/TypeTraits.h>
+#include <iostream>
+#include "triton_jit/triton_jit_function.h"
+
+namespace flag_gems {
+using namespace triton_jit;
+
+at::Tensor true_div(const at::Tensor& a_, const at::Tensor& b_) {
```
tongxin: The trailing underscore convention is reserved for inplace modification only. I think that argument names should be `a` and `b`.

lib/div.cpp
```diff
@@ -0,0 +1,457 @@
...
+  if (a_.dim() == 0 && b_.dim() > 0) {
+    at::Tensor a = at::full_like(b_, a_.item<double>());
+    return true_div(a, b_);
+  }
+
+  // Case 3: Tensor / Scalar
+  if (a_.dim() > 0 && b_.dim() == 0) {
+    at::Tensor b = at::full_like(a_, b_.item<double>());
+    return true_div(a_, b);
+  }
```
tongxin, line 65: In the worst case the data movement could be doubled for materializing the broadcast tensor and then loading it. We should consider adding triton kernels for mixed scalar and tensor operands cases.

lib/div.cpp
```diff
@@ -0,0 +1,457 @@
...
+    return true_div_(a_, b_expand);
+  }
+
+  // Tensor // Tensor
+  auto res = torch::broadcast_tensors({a_, b_});
+  at::Tensor a_broadcast = res[0].contiguous();
+  at::Tensor b_broadcast = res[1].contiguous();
+  TORCH_CHECK(a_.numel() == a_broadcast.numel(), "true_div_: broadcasted shape mismatch");
+
+  TORCH_CHECK(a_broadcast.dim() >= 2, "true_div_: input must be at least 2D");
```
tongxin: Why do we have this constraint?

triton_src/div.py
```diff
@@ -0,0 +1,143 @@
...
+    a = tl.load(a_ptr + offsets, mask=mask)
+    b = tl.load(b_ptr + offsets, mask=mask)
+    r = a % b
+    c = (r != 0.0) & ((r < 0.0) != (b < 0.0))
+    result = tl.where(c, r + b, r)
+    tl.store(c_ptr + offsets, result, mask=mask)
+
+
+@triton.jit
+def true_div_kernel_(a_ptr, b_ptr, c_ptr, M, N, BLOCK_SIZE: tl.constexpr):
```
tongxin: Why are we breaking down input shapes into separate 2d sizes rather than simply numel?  I don't see any 2d specific indexings in this kernel.

lib/div.cpp
```diff
@@ -0,0 +1,457 @@
...
+  int64_t N = a_broadcast.numel() / M;
+
+  at::Tensor out = at::empty_like(a_broadcast);
+
+  const TritonJITFunction& f =
+      TritonJITFunction::getInstance((utils::get_triton_src_path() / "div.py").string(), "true_div_kernel_");
+
+  int64_t tile_size = 1024;
+  int64_t n = M * N;
+  int num_blocks = (n + tile_size - 1) / tile_size;
```
tongxin: I'm not sure why the input is viewed as a 2d tensor. Once you've made the inputs contiguous, its simply a 1d elementwise operation.

lib/div.cpp
```diff
@@ -0,0 +1,457 @@
...
+  c10::cuda::CUDAStream stream = c10::cuda::getCurrentCUDAStream();
+  CUstream raw_stream = static_cast<CUstream>(stream.stream());
+  c10::DeviceGuard guard(a_.device());
+
+  f(raw_stream, num_blocks, 1, 4, 2, 0, a_broadcast, b_broadcast, out, M, N, tile_size);
+
+  return a_.resize_as_(out).copy_(out);
+}
+
+at::Tensor trunc_div(const at::Tensor& a_, const at::Tensor& b_) {
```
tongxin: Again, naming conventions.

lib/div.cpp
```diff
@@ -0,0 +1,457 @@
...
+  int64_t N = a.numel() / M;
+
+  const TritonJITFunction& f =
+      TritonJITFunction::getInstance((utils::get_triton_src_path() / "div.py").string(),
+                                     "trunc_divide_kernel_");
+
+  int64_t tile_size = 1024;
+  int64_t n = M * N;
+  int num_blocks = (n + tile_size - 1) / tile_size;
+
```
tongxin: See previous comments.

lib/div.cpp
```diff
@@ -0,0 +1,457 @@
...
+    at::Tensor b_expand = at::full_like(a_, b_.item<double>());
+    bool is_int = c10::isIntegralType(a_.scalar_type(), true) && c10::isIntegralType(b_.scalar_type(), true);
+    return launch_floor_div_kernel(a_, b_expand, is_int);
+  }
+  // ========== Scalar // Tensor ==========
+  if (a_.dim() == 0 && b_.dim() > 0) {
+    at::Tensor a_expand = at::full_like(b_, a_.item<double>());
+    bool is_int = c10::isIntegralType(a_.scalar_type(), true) && c10::isIntegralType(b_.scalar_type(), true);
+    return launch_floor_div_kernel(a_expand, b_, is_int);
+  }
```
tongxin, line 259: Need dedicated kernels for mixed scalar/tensor operands cases.

lib/div.cpp
```diff
@@ -0,0 +1,457 @@
...
+  TORCH_CHECK(a.device().is_cuda(), "Only CUDA supported for floor_div");
+
+  at::ScalarType out_dtype = at::promote_types(a.scalar_type(), b.scalar_type());
+
+  if (a_.dtype() != out_dtype) {
+    a_ = a_.to(out_dtype);
+  }
+  if (!a_.sizes().equals(a.sizes())) {
+    a_ = at::empty(a.sizes(), a_.options().dtype(out_dtype));
+  }
```
tongxin, line 281: This is incorrect. In the case of inplace operation, the first operand is also the output and it should match sizes of the broadcast shape. If they don't match then fail.

lib/div.cpp
```diff
@@ -0,0 +1,457 @@
...
+  TORCH_CHECK(a_.sizes() == a.sizes(), "a_ size mismatch after broadcast.");
```
tongxin: OK. I think this line should be moved up.

lib/div.cpp
```diff
@@ -0,0 +1,457 @@
...
+  std::string kernel_name = use_integer_kernel ? "int_floordiv_kernel_" : "float_floordiv_kernel_";
+  const TritonJITFunction& f =
+      TritonJITFunction::getInstance((utils::get_triton_src_path() / "div.py").string(), kernel_name);
+
+  TORCH_CHECK(a.dim() >= 2, "Input tensor must be at least 2D for broadcasting");
+  int64_t M = a.size(0);
+  int64_t N = a.numel() / M;
+  int64_t tile_size = 1024;
+  int64_t total = M * N;
+  int num_blocks = (total + tile_size - 1) / tile_size;
```
tongxin: See previous comments on similar issues.

triton_src/div.py
```diff
@@ -0,0 +1,143 @@
...
+
+    q = tl.where(b == 0.0, div, q)
+    tl.store(c_ptr + offsets, q, mask=mask)
+
+
+@triton.jit
+def int_floordiv_kernel_(a_ptr, b_ptr, c_ptr, M, N, BLOCK_SIZE: tl.constexpr):
+    pid = tl.program_id(0)
+    offsets = pid * BLOCK_SIZE + tl.arange(0, BLOCK_SIZE)
+    mask = offsets < M * N
```
tongxin: replace M, N with numel.

triton_src/div.py
```diff
@@ -0,0 +1,143 @@
...
+
+    a = tl.load(a_ptr + offsets, mask=mask)
+    b = tl.load(b_ptr + offsets, mask=mask)
+    div = a / b
+    c = tl.where(div >= 0, tl.math.floor(div), tl.math.ceil(div))
+    tl.store(c_ptr + offsets, c, mask=mask)
+
+
+@triton.jit
+def float_floordiv_kernel_(a_ptr, b_ptr, c_ptr, M, N, BLOCK_SIZE: tl.constexpr):
```
tongxin: replace M, N with numel.

triton_src/div.py
```diff
@@ -0,0 +1,143 @@
...
+    mask = offsets < M * N
+
+    a = tl.load(a_ptr + offsets, mask=mask)
+    b = tl.load(b_ptr + offsets, mask=mask)
+    c = a / b
+    tl.store(c_ptr + offsets, c, mask=mask)
+
+
+@triton.jit
+def trunc_divide_kernel_(a_ptr, b_ptr, c_ptr, M, N, BLOCK_SIZE: tl.constexpr):
```
tongxin: replace M, N with numel.

tongxin:
Hi, can you let know how things were going?

lib/div.cpp
you-and-you: In FlagGems, there is a Triton operator for full_like (located at src/flag_gems/ops/full_like.py). Can we call it directly, or do we need to rewrite a Triton kernel for full_like?

ctests/test_triton_div.cpp
```diff
@@ -0,0 +1,125 @@
...
+
+  EXPECT_TRUE(torch::allclose(out_torch, out_inplace, 1e-4, 1e-6));
+}
+
+TEST(blas_op_test, trunc_div) {
+  const torch::Device device(torch::kCUDA, 0);
+  torch::Tensor a = torch::randn({64, 64}, device);
+  torch::Tensor b = torch::randn({1, 64}, device).clamp_min(1e-3);
+
+  auto out_torch = torch::trunc(a / b);
```
zhangpeiyang1, line 33: maybe torch.trunc_divide(a, b) is better？

ctests/test_triton_div.cpp
```diff
@@ -0,0 +1,125 @@
...
+}
+
+TEST(blas_op_test, floor_div_) {
+  const torch::Device device(torch::kCUDA, 0);
+
+  torch::Tensor a = torch::randn({4, 8}, device) * 10;
+  torch::Tensor b = torch::randn({1, 8}, device).clamp_min(1e-3);
+
+  auto out_torch = torch::floor_divide(a, b);
+  auto out_triton = flag_gems::floor_div_(a, b);
```
zhangpeiyang1, line 70: you used to use a.clone as the inplace op input . why not so here?

ctests/test_triton_div.cpp
```diff
@@ -0,0 +1,125 @@
...
+  flag_gems::div_mode_(triton_out, b, c10::make_optional<std::string>("floor"));
+
+  EXPECT_TRUE(torch::allclose(torch_out, triton_out, 1e-4, 1e-6));
+}
+
+TEST(blas_op_test, remainder) {
+  const torch::Device device(torch::kCUDA, 0);
+
+  torch::Tensor a = torch::randn({32, 32}, device) * 10;
+  torch::Tensor b = torch::randn({32, 32}, device).clamp_min(0.5);
```
zhangpeiyang1, line 103: can you update some int case？

triton_src/div.py
```diff
@@ -0,0 +1,143 @@
...
+    pid = tl.program_id(0)
+    offsets = pid * BLOCK_SIZE + tl.arange(0, BLOCK_SIZE)
+    mask = offsets < size
+
+    a = tl.load(a_ptr + offsets, mask=mask)
+    b = tl.load(b_ptr + offsets, mask=mask)
+
+    div = a / b
+    result = tl.math.floor(div)
+    c = tl.where(b == 0.0, div, result)
```
zhangpeiyang1, line 46: why ？ for nan inf ？

zhangpeiyang1: tl.math.floor(inf) → inf，tl.math.floor(-inf) → -inf；
tl.math.floor(NaN) → NaN；

triton_src/div.py
```diff
@@ -0,0 +1,143 @@
...
+    a = tl.load(a_ptr + offsets, mask=mask)
+    b = tl.load(b_ptr + offsets, mask=mask)
+    r = a % b
+    c = (r != 0.0) & ((r < 0.0) != (b < 0.0))
+    result = tl.where(c, r + b, r)
+    tl.store(c_ptr + offsets, result, mask=mask)
+
+
+@triton.jit
+def true_div_kernel_(a_ptr, b_ptr, c_ptr, size, BLOCK_SIZE: tl.constexpr):
```
zhangpeiyang1, line 82: looks all the same with true_div_kernel ， do we really need to write  this kernel again？

triton_src/div.py
```diff
@@ -0,0 +1,143 @@
...
+
+    a = tl.load(a_ptr + offsets, mask=mask)
+    b = tl.load(b_ptr + col, mask=mask)
+
+    div = a / b
+    q = tl.math.floor(div)
+
+    diff = div - q
+    threshold = 1e-6
+    fix = (diff < -threshold) & ((a < 0) != (b < 0))
```
zhangpeiyang1, line 121: where‘s this algorithm from？

triton_src/div.py
```diff
@@ -0,0 +1,143 @@
...
+def int_floordiv_kernel(a_ptr, b_ptr, c_ptr, size, BLOCK_SIZE: tl.constexpr):
+    pid = tl.program_id(0)
+    offsets = pid * BLOCK_SIZE + tl.arange(0, BLOCK_SIZE)
+    mask = offsets < size
+    a = tl.load(a_ptr + offsets, mask=mask).to(tl.int32)
+    b = tl.load(b_ptr + offsets, mask=mask).to(tl.int32)
+    r = a % b
+    c1 = r != 0
+    c2 = (a < 0) ^ (b < 0)
+    q = a // b
```
zhangpeiyang1, line 61: where‘s this algorithm from？

ctests/test_triton_div.cpp
you-and-you: This is an API in Python, but we need to use an API in C++

ctests/test_triton_div.cpp
you-and-you: Because this line of code `auto out_torch = torch::floor_divide(a, b);` is not an in-place operation and does not change the value of `a`, there is no need to use `clone` afterwards

triton_src/div.py
you-and-you: The Triton kernel has existed before, and this is just a project involving a C++ wrapper

ctests/test_triton_div.cpp
you-and-you: ok

triton_src/div.py
you-and-you: The line of code `result = tl.math.floor(div)` uses the floor function to round down. For example, if `div[i] = 3.7`, then `result[i] = 3.0`

triton_src/div.py
you-and-you: They are identical, but the call in div.cpp differs. true_div_kernel_ is used for in-place operations

triton_src/div.py
you-and-you: float_floordiv_kernel_is different from float_floordiv_kernel, and it is a kernel used for in-place operations

triton_src/div.py
you-and-you: int_floordiv_kernel is different from int_floordiv_kernel_ , as it is a kernel used for non-in-place operations

PR #843, opened 2025-07-31
[Ascend] Fixed the error in the mul operator and resolved the logger naming issue for the ascend operator.

Yanguan619:


src/flag_gems/runtime/backend/_ascend/ops/addmm.py
```diff
@@ -9,7 +9,7 @@
 from flag_gems.utils import broadcastable_to, libentry
 from flag_gems.utils import triton_lang_extension as tle
 
-logger = logging.getLogger(__name__)
+logger = logging.getLogger(f'flag_gems.runtime._ascend.ops.{__name__.split(".")[-1]}')
```
kiddyjinjin, line 12: why this works?

Yanguan619: When I used `__name__` as the log name in `runtime/backend/_ascend/ops`, I found that operator execution was not recorded in the `oplist.log` because their logger prefixes are different.

PR #849, opened 2025-08-01
prune configs for mm autotune

sgjzfzzf:
### PR Category

Operator

### Type of Change

New Feature

### Description

We found that when the shared memory usage is too large, Triton will be blocked on the compilation before throwing an exception, which could be tens of minutes. This problem comes from Triton compiler, and we cannot avoid it in FlagGems. The solution we want to introduce now is to prune the configs which are very possible to exceed the maximum of shared memory usage before tuning. We give a formula to compute the usage and drop it if so.

### Progress

### DIFF
--- a/src/flag_gems/ops/mm.py
+++ b/src/flag_gems/ops/mm.py
@@ -1,8 +1,10 @@
 import logging
+from typing import List
 
 import torch
 import triton
 import triton.language as tl
+import triton.runtime
 
 from flag_gems import runtime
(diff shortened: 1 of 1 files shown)


StrongSpoon (commented):
I believe we should gather information on whether vendors support interfaces for triton.runtime.driver.

StrongSpoon (approved):

bin913:
@sgjzfzzf Do you still keep track of this PR?

PR #866, opened 2025-08-08
Add ops diff

Liguanlong123:
### PR Category
Operator

### Type of Change
New Feature

### Description
This operator is a diff operation based on Triton, supporting 1D, 2D, 3D data. It dynamically selects the optimal kernel based on the input dimension and shape (e.g., flattening, vectorization, 2D tiling, 3D optimization) and enhances memory access and parallel efficiency through vectorized loading, cache-friendly layout, and dynamic BLOCK/TILE parameters, thereby efficiently performing functions similar to torch.diff.

### Progress
- [ ] Initial implementation of the diff operator.

### Performance
Accuracy rate: n taken from 1-50 is accurate. 

准确率测试

Performance testing: on par with torch implementation in a series of test cases

性能测试1
性能测试2

### DIFF
--- a/src/flag_gems/ops/diff.py
+++ b/src/flag_gems/ops/diff.py
@@ -1,0 +1,299 @@
+import torch
+import triton
+import triton.language as tl
+from torch import Tensor
+from typing import Optional
+
+@triton.jit
+def diff_kernel_flat(
+    input_ptr, output_ptr,
+    B,
+    L,
+    out_L,
+    input_stride,
+    output_stride,
+    N: tl.constexpr,
+    BLOCK: tl.constexpr
+):
+    pid = tl.program_id(0)
+    start = pid * BLOCK
+    offs  = start + tl.arange(0, BLOCK)
+    total = B * out_L
+
+    mask = offs < total
+    batch = offs // out_L
+    pos   = offs % out_L
+
+    idx0 = batch * input_stride + pos
+    idx1 = idx0 + N
+
+    x0 = tl.load(input_ptr + idx0, mask=mask)
+    x1 = tl.load(input_ptr + idx1, mask=mask)
+    res = x1 - x0
+
+    out_idx = batch * output_stride + pos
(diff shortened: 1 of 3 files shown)


meinie0826:
Does this implementation only support 1D, 2D and 3D?

src/flag_gems/ops/diff.py
```diff
@@ -0,0 +1,299 @@
...
+
+
+@triton.jit
+def diff_kernel_1d_vectorized(
+    input_ptr, output_ptr,
+    N, out_L,
+    BLOCK_SIZE: tl.constexpr,
+    VECTOR_WIDTH: tl.constexpr
+):
+    """1D向量化内核"""
```
meinie0826, line 45: It would be best to provide the annotations in English.

src/flag_gems/ops/diff.py
```diff
@@ -0,0 +1,299 @@
...
+
+    if prepend is not None:
+        input = torch.cat([prepend, input], dim=dim)
+    if append is not None:
+        input = torch.cat([input, append], dim=dim)
+    
+    out = input
+
+    for _ in range(n):
+        out = _diff_once(out, dim)
```
StrongSpoon, line 298: I suggest implementing "diff" within a single kernel. For n greater than 1, the shape of the output tensor can be readily calculated, and the values of higher-order differences adhere to the pattern of Pascal's triangle.

StrongSpoon:
I'm concerned that the performance may rapidly decline once n exceeds one.

src/flag_gems/ops/diff.py
Liguanlong123: Okay, I will rewrite my code as you mentioned.

Liguanlong123:
> I'm concerned that the performance may rapidly decline once n exceeds one.

I will test cases when n is greater than 1 after fixing the issues, aiming to maintain performance.

bin913:
@Liguanlong123 Do you still keep track of this PR?

Issue #312, opened 2024-11-22
Code Contribution: 【Lv3】【Operator Development】 conv3d


StrongSpoon:
##  Description 任务介绍
Develop forward function for conv3d operator.
开发conv3d算子的前向功能。
## Requirements 任务要求
__Interface 接口__
`conv3d(Tensor input, Tensor weight, Tensor? bias=None, SymInt[3] stride=1, SymInt[3] padding=0, SymInt[3] dilation=1, SymInt groups=1) -> Tensor`
__Function reference 功能参考__
https://pytorch.org/docs/stable/generated/torch.nn.Conv3d.html
__Implementation reference 实现参考__
https://github.com/FlagOpen/FlagGems/blob/7016596a5071a3f2a969e19034e7dae58b699b36/src/flag_gems/ops/conv2d.py

The operator should support all optional arguments defined in the interface. 
算子应支持接口中定义的所有参数选项。
## DDL 提交时间
Please submit a Pull Request within 3 weeks after accepting the assignment.
请于接取任务后三周内提交PR。

Gxiandy:
Gxian认领

Miner-debug:
don't submit a Pull Request within 3 weeks

IdleIdiot:
Idiot认领

IdleIdiot:
Completed. https://github.com/FlagOpen/FlagGems/pull/876

PR #876, opened 2025-08-12
[Operator] Add conv3d forward

IdleIdiot:
### Type of Change
New Feature

### Description
Implement the 3D convolution operator

### Issue
#312 

### Progress

### Performance

image

accuracy
image


benchmark/core_shapes.yaml
```diff
@@ -184,6 +184,17 @@ ConvBenchmark:
...
+Conv3DBenchmark:
+  shapes:
+    - [32, 64, 64, 64, 64, 32, 3, 3, 3, 1, 1, 1]
+    - [32, 64, 64, 64, 64, 32, 3, 3, 3, 1, 2, 1]
+    - [104, 16, 32, 32, 32, 32, 4, 4, 4, 1, 0, 1]
+    - [64, 32, 18, 18, 18, 32, 5, 5, 5, 2, 1, 1]
+    - [16, 32, 12, 12, 12, 24, 3, 3, 3, 2, 1, 1]
+    - [16, 32, 24, 24, 24, 24, 3, 3, 3, 1, 1, 2]
+    - [16, 32, 24, 24, 24, 24, 3, 3, 3, 2, 2, 2]
+    - [16, 32, 24, 24, 24, 24, 3, 3, 3, 1, 2, 2]
```
StrongSpoon: generally we set 5 shapes for core mode ;)

benchmark/test_special_perf.py
```diff
@@ -343,6 +343,56 @@ def conv2d_input_fn(shape, dtype, device):
...
+            "padding": padding,
+        },
+
+    torch.backends.cudnn.allow_tf32 = False
+    bench = Conv3DBenchmark(
+        input_fn=conv3d_input_fn,
+        op_name="conv3d",
+        torch_op=torch.nn.functional.conv3d,
+        dtypes=FLOAT_DTYPES,
+    )
```
StrongSpoon, line 392: Since `conv3d` is not yet registered in the ATEN library, we should explicitly set `gems_op` using `bench.set_gems`.

IdleIdiot: Thank you for your suggestion. After explicitly specifying the operator call, the performance issue was exposed. I am currently working hard to identify the root cause of the performance problem.
image

IdleIdiot:
image
I have made some optimizations to the auto-tuning configuration, which has improved the operator's performance metrics to some extent. However, after trying various parameter combinations, the performance of the fp16 and bf16 cases still cannot achieve satisfactory results. Currently, I suspect that this might be related to the grid dividing the input data all at once for computation, where excessive thread block allocation leads to performance degradation. Next, I will try to add the upper limit of thread block allocation and simultaneously modify the operator code for verification.
Like this: 

```
grid = lambda META: (
    min(8192, triton.cdiv(in_n * out_depth * out_height * out_width, META["BLOCK_NI_DO_HO_WO"])),
    triton.cdiv(out_c // groups, META["BLOCK_CO"]),
    groups,
)

```

Issue #894, opened 2025-08-12
Code Contribution: 【Hard】【Operator Development】avg_pool2d


0x45f:
## Description 任务介绍
Develop forward function for avg_pool2d operator.
开发avg_pool2d算子的前向功能。
## Requirements 任务要求
__Interface 接口__
`avg_pool2d(Tensor self, int[2] kernel_size, int[2] stride=[], int[2] padding=0, bool ceil_mode=False, bool count_include_pad=True, int? divisor_override=None) -> Tensor`
__Function reference 功能参考__
https://docs.pytorch.org/docs/stable/generated/torch.nn.functional.avg_pool2d.html
__Implementation reference 实现参考__
https://github.com/FlagOpen/FlagGems/blob/master/src/flag_gems/ops/upsample_nearest2d.py

The operator should support all optional arguments defined in the interface. 
算子应支持接口中定义的所有参数选项。
## DDL 提交时间
Please submit a Pull Request within 2 weeks after accepting the assignment.
请于接取任务后两周内提交PR。
Please provide both accuracy test and performance test code.
请同时提供实现正确性测试与性能测试代码。

StrongSpoon:
@Blury233

IdleIdiot:
idiot 认领

w1120029931-bit:
Has this operator been developed? Can it be closed?

PR #959, opened 2025-09-12
Add avg_pool2d forward and backward implementation

Blury233:
### PR Category

Operator 
### Type of Change

New Feature
### Description

This PR adds:
- avg_pool2d forward
- avg_pool2d backward
### Issue
Associated with Feature #894

### Progress

### Performance

Accuracy Test
acc
Performance Test
perf1
perf2
perf3

### DIFF
--- a/tests/test_reduction_ops.py
+++ b/tests/test_reduction_ops.py
@@ -1214,7 +1214,7 @@ def test_accuracy_avg_pool2d_forward(
     gems_assert_close(res_out, ref_out, dtype)
 
 
-@pytest.mark.avg_pool2d
+@pytest.mark.avg_pool2d_bwd
 @pytest.mark.parametrize(
     "shape, kernel_size, stride, padding, ceil_mode, count_include_pad, divisor_override",
     AVGPOOL2D_CONFIGS,
@@ -1242,7 +1242,7 @@ def test_accuracy_avg_pool2d_backward(
         count_include_pad=count_include_pad,
         divisor_override=divisor_override,
     )
-    out_grad = torch.randn_like(ref_out, dtype=inp.dtype)
+    out_grad = torch.randn_like(ref_out, dtype=inp.dtype, device=flag_gems.device)
     ref_out_grad = to_reference(out_grad, True)
     ref_inp_grad = torch.ops.aten.avg_pool2d_backward(
         ref_out_grad,


src/flag_gems/ops/max_pool2d.py
```diff
@@ -0,0 +1,285 @@
...
+    padding: int,
+    dilation: int,
+    ceil_mode: bool = False,
+) -> int:
+    effective_kernel_size = (kernel_size - 1) * dilation + 1
+    numerator = in_size + 2 * padding - effective_kernel_size
+    if ceil_mode:
+        return (numerator + stride - 1) // stride + 1
+    else:
+        return numerator // stride + 1
```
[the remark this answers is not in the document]
iclementine: I agree

src/flag_gems/ops/max_pool2d.py
```diff
@@ -0,0 +1,285 @@
...
+                stride_w,
+                padding_h,
+                padding_w,
+                dilation_h,
+                dilation_w,
+            )
+
+        ctx.save_for_backward(indices)
+        ctx.in_shape = input.shape
+        ctx.in_strides = (in_c * in_h * in_w, in_h * in_w, in_w, 1)
```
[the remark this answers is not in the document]
iclementine: I agree.

src/flag_gems/ops/max_pool2d.py
```diff
@@ -0,0 +1,285 @@
...
+
+        if output.numel() > 0:
+            # 2D Grid for the optimized kernel
+            grid = lambda meta: (
+                in_n * in_c,
+                triton.cdiv(out_h, meta["BLOCK_H"])
+                * triton.cdiv(out_w, meta["BLOCK_W"]),
+            )
+
+            max_pool2d_forward_kernel_optimized[grid](
```
iclementine: add device guard to ensure that the kernel is launched to the correct device.

src/flag_gems/ops/max_pool2d.py
```diff
@@ -0,0 +1,285 @@
...
+        out_h, out_w = grad_output.shape[2], grad_output.shape[3]
+
+        original_dtype = grad_output.dtype
+        grad_input = torch.zeros(
+            in_shape, device=grad_output.device, dtype=torch.float32
+        )
+
+        grid = (grad_output.numel(),)
+        if grad_output.numel() > 0:
+            max_pool2d_backward_kernel[grid](
```
iclementine: In this case, a CTA computes an element in the output. Maybe this is too-small.

src/flag_gems/ops/max_pool2d.py
```diff
@@ -0,0 +1,285 @@
...
+    grad_out_val = tl.load(grad_output_ptr + pid)
+    max_idx_flat = tl.load(indices_ptr + pid)
+
+    # Pointer to the start of the current batch and channel in the gradient input tensor
+    grad_input_base_ptr = grad_input_ptr + n_idx * in_stride_n + c_idx * in_stride_c
+
+    if max_idx_flat != -1:
+        grad_input_offset = max_idx_flat
+        # Atomically add the gradient to the correct location in grad_input
+        tl.atomic_add(
```
iclementine: Maybe consider rewrite it with a max-pool-like kernel, which partitions tasks w.r.t. to grad_input to avoid atomic add to accumulate results.

src/flag_gems/ops/max_pool2d.py
```diff
@@ -0,0 +1,285 @@
...
+            grad_input_base_ptr + grad_input_offset,
+            grad_out_val,
+        )
+
+
+class MaxPool2d(torch.autograd.Function):
```
iclementine: Avoid using `autograd.Function` since it is not the encouraged way now.

Overwrite `max_pool2d_with_indices_backward.grad_input` or `max_pool2d_with_indices_backward`.

src/flag_gems/ops/avg_pool2d.py
```diff
@@ -0,0 +1,426 @@
...
+    padding: int,
+    dilation: int,
+    ceil_mode: bool = False,
+) -> int:
+    effective_kernel_size = (kernel_size - 1) * dilation + 1
+    numerator = in_size + 2 * padding - effective_kernel_size
+    if ceil_mode:
+        return (numerator + stride - 1) // stride + 1
+    else:
+        return numerator // stride + 1
```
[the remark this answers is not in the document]
iclementine: We accept it now. Maybe somtime we would refactor the code the extract some utility functions for convolution, convolution_transpose, and pooling.

src/flag_gems/ops/avg_pool2d.py
```diff
@@ -0,0 +1,426 @@
...
+    )
+    in_write_mask = (h_in_offsets[:, None] < in_h) & (w_in_offsets[None, :] < in_w)
+    tl.store(
+        grad_input_store_ptr,
+        grad_acc.to(grad_input_ptr.type.element_ty),
+        mask=in_write_mask,
+    )
+
+
+class AvgPool2d(torch.autograd.Function):
```
iclementine: I suggest not to subclass `torch.autograd.Function` and add implementions for the forward and backward operator, separately.

- `avg_pool2d` or `avg_pool2d.out`
- `avg_pool2d_backward` and `avg_pool2d_backward.grad_input`

To do that, we need to check their schema in `pytorch/aten/src/ATen/native/native_functions.yaml` and their signature in the implementation files to be sure that our implementation uses the same signature.

src/flag_gems/ops/avg_pool2d.py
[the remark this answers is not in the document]
iclementine: There may be some risk that the output size is different from the result of Aten counterpart.

The ceil_mode has 2 effects:
1. It allow the window's right edge to extend bybind the padded input;
2. but It does not allow the window to start from right padding region.

So follow ATen's implementation to be consistent or else in some cases the output size is larger the Aten's output by 1.

```cpp
template<typename T>
inline T pooling_output_shape_pad_lr(
        T inputSize, T kernelSize, T pad_l, T pad_r, T stride, T dilation,
        bool ceil_mode) {
    T outputSize = div_rtn<T>(
        inputSize + pad_l + pad_r - dilation * (kernelSize - 1) - 1 +
        (ceil_mode ? stride - 1 : 0), stride) + 1;
    if (ceil_mode) {
        // ensure that the last pooling starts inside the image
        // needed to avoid problems in ceil mode
        if ((outputSize - 1) * stride >= inputSize + pad_l) {
          --outputSize;
        }
    }
    return outputSize;
}
```

src/flag_gems/ops/avg_pool2d.py
```diff
@@ -0,0 +1,426 @@
...
+        input,
+        kernel_size,
+        stride,
+        padding,
+        ceil_mode,
+        count_include_pad,
+        divisor_override,
+    ):
+        logger.debug("GEMS AVG_POOL2D FORWARD")
+        input = input.contiguous()
```
iclementine: Add argument checking that 
- stride should not be zero
- pad should not be negative
- pad is less than half of the window size

iclementine: Please Keep in mind that Aten's implementation would try to preserve & propagate channel_last layout when possible, but this implementation may introduce extra transposition.

```cpp
  const auto memory_format = input_.suggest_memory_format();

  Tensor input = input_.contiguous(memory_format);
```

Maybe we can leave it for further optimizations.

src/flag_gems/ops/avg_pool2d.py
```diff
@@ -0,0 +1,426 @@
...
+                kernel_h,
+                kernel_w,
+                stride_h,
+                stride_w,
+                padding_h,
+                padding_w,
+                dilation_h,
+                dilation_w,
+                COUNT_INCLUDE_PAD=count_include_pad,
+                divisor_override=divisor_override or 0,
```
iclementine: I advice to check that divisor is not zero if it is not None.
If it is None, just pass it to the jit function. Do not use 0 as a magic number.

src/flag_gems/ops/avg_pool2d.py
```diff
@@ -0,0 +1,426 @@
...
+            w_valid_map = (w_out_num >= 0) & ((w_out_num % stride_w) == 0)
+
+            h_out = h_out_num // stride_h
+            w_out = w_out_num // stride_w
+
+            h_out_mask = h_valid_map & (h_out < out_h)
+            w_out_mask = w_valid_map & (w_out < out_w)
+            out_mask = h_out_mask & w_out_mask
+
+            if divisor_override > 0:
```
iclementine: divisor may also be negative.

src/flag_gems/ops/avg_pool2d.py
```diff
@@ -0,0 +1,426 @@
...
+    kernel_w: tl.constexpr,
+    stride_h: tl.constexpr,
+    stride_w: tl.constexpr,
+    padding_h: tl.constexpr,
+    padding_w: tl.constexpr,
+    dilation_h: tl.constexpr,
+    dilation_w: tl.constexpr,
+    # AvgPool specific parameters
+    COUNT_INCLUDE_PAD: tl.constexpr,
+    divisor_override: tl.constexpr,
```
iclementine: `divisor_override` should not be a constexpr since it may be a float, which may not be critical to performance.

src/flag_gems/ops/avg_pool2d.py
```diff
@@ -0,0 +1,426 @@
...
+                        is_valid = (
+                            (h_in_for_count >= 0)
+                            & (h_in_for_count < in_h)
+                            & (w_in_for_count >= 0)
+                            & (w_in_for_count < in_w)
+                        )
+                        count += is_valid.to(tl.int32)
+                divisor = count.to(tl.float32)
+
+            divisor = tl.where(divisor == 0, 1.0, divisor)
```
iclementine, line 235: In what case may divisor be zero?

iclementine (approved):

Blury233:
> @Blury233 Thanks for you contribution. Please give me your email address. I'll send you an E-certificate of honor.

Thank you for your message. My email address is blurylee@outlook.com

PR #965, opened 2025-09-15
optimize some operators on ascend backend

yanminghui123:
### Progress

### DIFF
--- a/src/flag_gems/runtime/backend/_ascend/tune_configs.yaml
+++ b/src/flag_gems/runtime/backend/_ascend/tune_configs.yaml
@@ -80,22 +80,18 @@ mm:
     BLOCK_M: 128
     BLOCK_N: 128
     BLOCK_K: 128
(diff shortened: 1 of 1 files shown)


src/flag_gems/runtime/backend/_ascend/ops/argmax.py
```diff
@@ -140,17 +146,30 @@ def argmax(inp, dim=None, keepdim=False, *, dtype=None):
...
+            with torch_device_fn.device(inp.device):
+                argmax_kernel[grid](
+                    inp,
+                    out_index,
+                    M,
+                    N,
+                    K,
+                )
+        else:
+            inp_reshaped = inp.view(M, N, K).permute(2, 0, 1).contiguous()
```
iclementine, line 164: Maybe in this case, you can use a larger block size on dimension K and avoid permuting the input. It would save extra memory operation.

PR #969, opened 2025-09-16
[Bug Fix] Explicitly zero out masked elements in mm_kernel for the CPU backend

ziliangzl:
### PR Category

Operator
### Type of Change

Bug Fix
### Description

The test case test_accuracy_mm fails when running on the CPU backend:
```
pytest -v tests/test_blas_ops.py::test_accuracy_mm[dtype1-495-5333-71]
```
output: 
```
E       AssertionError: Tensor-likes are not close!
E       
E       Mismatched elements: 62715 / 2639835 (2.4%)
E       Greatest absolute difference: nan at index (0, 16) (up to 0.01 allowed)
E       Greatest relative difference: nan at index (0, 16) (up to 0 allowed)
```
This issue can also be reproduced for other shape when K is not a power of two (e.g., M, N, K = (32, 32, 34)).

The error seems to be caused by different mask-handling strategies across Triton, Triton-Shared, and Flagtree.
As a fix, this PR explicitly sets the masked elements in mm_kernel to zero for the CPU backend, which provides consistent behavior and resolves the test failure.
### Progress


ziliangzl:
> Are you saying that running the results on Triton Shared and Flagtree is correct? Will the modification have any impact on both?

Hi @meinie0826, no — if other is not specified, the results on Triton-Shared and Flagtree will be incorrect, because masked elements are not set to zero by default. With this PR, Triton-Shared and Flagtree can now pass this test and produce results consistent with Triton.

ziliangzl:
Hi @meinie0826 , Could you please help merge it? Thanks a lot!

PR #1019, opened 2025-10-28
Support Multi-DB-Backends for LibCache

sgjzfzzf:
### PR Category

Other

### Type of Change

New Feature

### Description

This PR supports multi-sql-backends for `libcache`. It's powered by `sqlalchemy`. The previous SQL backend is sqlite3, which doesn't support multi-writers at the same time, and may lead to some writing conflicts in the multi-workers cases. By introducing PostgreSQL as an available option for `libcache`, which supports several writings at the same time, it could solve the problem.

The details are recored in the document `docs/libcache_db_backend.md`. Please refer to it for more details.

### Progress

### DIFF
--- a/src/flag_gems/utils/libentry.py
+++ b/src/flag_gems/utils/libentry.py
@@ -6,14 +6,24 @@ import logging
 import math
 import multiprocessing
 import os
-import sqlite3
-import threading
 import time
 from abc import abstractmethod
 from collections import OrderedDict
 from functools import cached_property
 from itertools import starmap
-from typing import Any, Callable, Dict, Iterator, List, Optional, Tuple, Type, Union
+from pathlib import Path
+from typing import (
+    Any,
+    Callable,
+    Dict,
+    Final,
+    Iterator,
+    List,
+    Optional,
+    Tuple,
+    Type,
+    Union,
+)
 
 import torch
 import triton
@@ -22,6 +32,7 @@ from flag_gems import runtime
 from flag_gems.runtime import torch_device_fn
 from flag_gems.runtime.backend import vendor_module
 from flag_gems.utils.code_cache import config_cache_dir
+from flag_gems.utils.models import PersistantModel, SQLPersistantModel
 
 logger = logging.getLogger(__name__)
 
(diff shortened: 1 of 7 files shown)


sgjzfzzf:
> > Hi, could you please provide us with the test script you're using here? It looks like a database problem with duplicated primary keys. @ArashiFeng
> 
> Sorry that I can't provide the test script because it's for internal testing. In short, this problem is caused by this line
> 
> https://github.com/FlagOpen/FlagGems/blob/11efecfcfa04c8c6d509adf04a4a067f1e6da524/src/flag_gems/utils/models/sql.py#L176
> 
> .
> `INSERT OR REPLACES`(https://github.com/FlagOpen/FlagGems/blob/master/src/flag_gems/utils/libentry.py#L258) can replace data when primary keys are detected duplicated. But `session.add()` cannot handle this.
> Change to following code can deal with `postgresql` on conflict.
> ```python
>   from sqlalchemy.dialects.postgresql import insert
>   obj: Base = ConfigCls(**key_dict, **config)
>   cmd = insert(ConfigCls).values(**key_dict, **config)
>   primary_keys = [col for col in ConfigCls.__table__.primary_key.columns]
>   update_fields = {k: getattr(cmd.excluded, k) for k in key_dict.keys() if k not in [c.name for c in primary_keys]}
>   update_fields.update({k: getattr(cmd.excluded, k) for k in config.keys() if k not in [c.name for c in primary_keys]})
>   cmd = cmd.on_conflict_do_update(index_elements=primary_keys, set_=update_fields)
>   session.execute(cmd)
>   session.commit()
> ```
> 
> The following code may also work, but not validated.
> 
> ```python
>   with Session(engine) as session:
>     obj = session.get(ConfigCls, key_dict["xxx"])
>     if obj is None:
>         obj = ConfigCls(**key_dict, **config)
>         session.add(obj)
>     else:
>         for k, v in config.items():
>             setattr(obj, k, v)
>     session.commit()
> ```

Hi thank you for your reply and comment! I'll try this solution later. My machine is broken so I need to wait for the repair😭. I'll let you know when there is any new update.

src/flag_gems/utils/libentry.py
```diff
@@ -427,23 +287,23 @@ def __init__(
...
             jit_fn = jit_fn.fn
         return jit_fn.cache_key
 
     @cached_property
-    def kernel_hash(self):
+    def kernel_hash(self) -> str:
         return hashlib.md5(
             f"{self.cache_key}{self.configs_hash}".encode("utf-8")
-        ).hexdigest()
+        ).hexdigest()[:32]
```
0x45f, line 300: Why only use first 32 bits?

sgjzfzzf: Because the length of the table name is limited, we truncate it here.

PR #1109, opened 2025-11-27
[Iluvatar] enable test_perf_flash_attn_varlen_func, fix bmm and conv3d

awayzjj:
### Description

1. Enable `test_perf_flash_attn_varlen_func` using the legacy FA interface for reference, as the latest interface is not supported yet.
3. Since FlagGems may not support **auto** calling ops from the backend when other ops are used internally (e.g., attempting `from flag_gems import bmm` `from flag_gems.ops.bmm import bmm`fails), we directly call `bmm` in the Iluvatar backend.
5. Add tuning configurations for `conv3d`.

### Progress


Galaxy1458 (approved):
LGTM

PR #1126, opened 2025-11-28
cambricon: update cambricon code based on master

chenmiao1919:
### PR Category

Operator

### Type of Change

Performance Optimization

### Description

update cambricon to master

### Progress

### DIFF
--- a/benchmark/test_unary_pointwise_perf.py
+++ b/benchmark/test_unary_pointwise_perf.py
@@ -266,8 +266,8 @@ class BinaryPointwiseBenchmark(Benchmark):
     def get_input_iter(self, cur_dtype) -> Generator:
         for shape in self.shapes:
             inp1 = generate_tensor_input(shape, cur_dtype, self.device)
-            shift_amount = torch.randint(
-                0, 8, shape, dtype=cur_dtype, device=self.device
+            shift_amount = torch.randint(0, 8, shape, dtype=cur_dtype, device="cpu").to(
+                self.device
             )
             yield inp1, shift_amount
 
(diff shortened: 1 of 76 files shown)


benchmark/test_unary_pointwise_perf.py
```diff
@@ -265,8 +265,8 @@ def set_more_shapes(self):
     def get_input_iter(self, cur_dtype) -> Generator:
         for shape in self.shapes:
             inp1 = generate_tensor_input(shape, cur_dtype, self.device)
-            shift_amount = torch.randint(
-                0, 8, shape, dtype=cur_dtype, device=self.device
+            shift_amount = torch.randint(0, 8, shape, dtype=cur_dtype, device="cpu").to(
+                self.device
```
kiddyjinjin, line 270: Is there any difference between these two expressions？

pytest.ini
```diff
@@ -2,3 +2,4 @@
 testpaths = tests
 pythonpath = src
 filterwarnings = ignore::pytest.PytestUnknownMarkWarning
+addopts = -p no:faulthandler
```
kiddyjinjin: Why do we disable the faulthandler plugin

benchmark/test_unary_pointwise_perf.py
chenmiao1919: Since the randint function in our torch-mlu does not support specific data types (e.g., i16), we’ve made this temporary modification for now. Support for these types will be added in future.

pytest.ini
chenmiao1919: The change has been deleted

PR #1152, opened 2025-12-03
【cpp  wrapper】extract and cache device_info in cpp wrapper

kiddyjinjin:
### Type of Change

Performance Optimization

### Description

extract and cache device_info in cpp wrapper

### DIFF
--- a/lib/device_info.cpp
+++ b/lib/device_info.cpp
@@ -1,0 +1,83 @@
+#include "flag_gems/device_info.h"
+
+#include <ATen/cuda/CUDAContext.h>
+#include <cuda_runtime_api.h>
+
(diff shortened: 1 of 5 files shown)


lib/device_info.cpp
```diff
@@ -0,0 +1,83 @@
...
+  }
+
+  std::mutex &cache_mutex() {
+    static std::mutex mutex;
+    return mutex;
+  }
+}  // namespace
+
+const DeviceInfo &get_device_info(int device_id) {
+  {
```
meinie0826, line 43: should device_id == 0 be treated as an exception ?

meinie0826 (approved):
lgtm

PR #1190, opened 2025-12-09
【bug fix】Fix large-tensor overflow for exp operator, and mm contiguity issues

kiddyjinjin:
### PR Category

Operator

### Type of Change

Bug Fix

### Description

- Use the 1D fast-path only when all tensors share the same shape and numel <= int32_max. Otherwise fall back to the original N-dimensional scheduling to avoid 32-bit overflow. (`pointwise_dynamic.py`)
- Avoid GPU OOM in large-tensor comparisons (`testing/__init__.py`)
- adds 64-bit wrappers for program IDs to avoid overflow in kernels using large strides （used in `mm` operator）
- Using torch.empty(..., memory_format=torch.contiguous_format) instead of allocating then calling .to().contiguous().

### DIFF
--- a/tests/accuracy_utils.py
+++ b/tests/accuracy_utils.py
@@ -61,9 +61,6 @@ POINTWISE_SHAPES = (
     if QUICK_MODE
     else [(), (1,), (1024, 1024), (20, 320, 15), (16, 128, 64, 60), (16, 7, 57, 32, 29)]
 )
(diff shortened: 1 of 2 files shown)


zhangpeiyang1:
LGTM

zhangpeiyang1 (approved):

PR #1241, opened 2025-12-19
[KUNLUNXIN][BUF-FIX]fix that transformer engine disable benchmark logger

dongjibin1996:
the transformer_engine may disable flagGems benchMark logging, which lead to "pytest --record log" failure

### PR Category
OP Test 

### Type of Change
Bug Fix

### Progress

repair issue https://github.com/flagos-ai/FlagGems/issues/1242


Ason93 (dismissed):

benchmark/performance_utils.py
```diff
@@ -372,7 +371,7 @@ def run(self):
                 shape_desc=self.shape_desc,
             )
             print(attri)
```
tengqm, line 373: Do we still want the result printed on stdout?

benchmark/performance_utils.py
```diff
@@ -425,7 +424,7 @@ def run(self):
                 result=metrics,
             )
             print(result)
```
tengqm, line 426: Same question here.

benchmark/test_transformer_engine_perf.py
```diff
@@ -8,6 +8,8 @@
 try:
     from transformer_engine.pytorch import cpp_extensions as tex
 
+    # note: the transformer_engine has its' own logging which may have conflict with flag_gems logging
```
tengqm: so ... what do you mean by conflict?
both of them are logging to stdout/stderr?

what's your intention by adding this note here?

benchmark/conftest.py
```diff
@@ -165,12 +167,21 @@ def pytest_configure(config):
             for arg in config.invocation_params.args
         ]
 
-        logging.basicConfig(
-            filename="result_{}.log".format("_".join(cmd_args)).replace("_-", "-"),
-            filemode="w",
-            level=logging.INFO,
-            format="[%(levelname)s] %(message)s",
-        )
+        log_file = "result_{}.log".format("_".join(cmd_args)).replace("_-", "-")
```
tengqm, line 170: where to find this log file then?
shall we specify a directory that can be easily guessed/found?

benchmark/test_transformer_engine_perf.py
dongjibin1996: The "conflict" mentioned here refers to a Root Logger configuration conflict , rather than a simple stdout/stderr output issue.

The FlagGems benchmark relies on logging.basicConfig(filename=...) to write performance results to result_*.log . However, logging.basicConfig is designed to be a no-op (ignored) if the Root Logger already has handlers configured.

We observed that in certain versions of transformer_engine (specifically on Python 3.10 in our environment), importing the module registers a StreamHandler to the Root Logger during initialization. This causes our subsequent basicConfig call to be ignored, resulting in missing result files.

This behavior can be verified with the following one-liner:
`python -c "import logging; print('before', logging.root.handlers); import transformer_engine.pytorch.cpp_extensions as tex; print('after', logging.root.handlers)"`
Output (Python 3.10):
```
before []
after [<StreamHandler <stderr> (NOTSET)>]
```
As shown, the Root Logger acquires a StreamHandler after the import. This issue has also been discussed in [NVIDIA/TransformerEngine#1065](https://github.com/NVIDIA/TransformerEngine/issues/1065) .

Fix: To resolve this robustly, we have switched to using a dedicated RecordLogger with an explicit FileHandler for writing benchmark results, which operates independently of the Root Logger's state.

benchmark/conftest.py
dongjibin1996: This PR focuses on fixing the missing log file issue caused by the logger conflict, rather than changing the existing file structure.

To answer your question: the log file is generated in the current working directory where the pytest command is executed (e.g., the project_root/benchmark/). This behavior follows the existing convention set by previous contributors, and I intend to keep it consistent in this PR to minimize side effects.

benchmark/performance_utils.py
dongjibin1996: Yes, print(attri) is the existing behavior that outputs the result to stdout for immediate visibility, which is independent of the logging system.

Since this PR focuses on fixing the missing log file issue (by using a dedicated file logger), I prefer to keep the stdout printing behavior unchanged to maintain backward compatibility for users who rely on console output.

benchmark/test_transformer_engine_perf.py
tengqm: Great. Thanks for the elaborated feedback. We'll track this issue.

benchmark/conftest.py
tengqm: Thanks. We'll try solve this problem globally.

tengqm (dismissed):

kiddyjinjin (approved):

benchmark/conftest.py
```diff
@@ -165,12 +167,21 @@ def pytest_configure(config):
...
+
+        for h in list(recordLogger.handlers):
+            recordLogger.removeHandler(h)
+            try:
+                h.close()
+            except Exception:
+                pass
```
SamYuan1990: pass?
should we expose the reason why exception?

benchmark/conftest.py
```diff
@@ -165,12 +167,21 @@ def pytest_configure(config):
...
+
+        handler = logging.FileHandler(log_file, mode="w", encoding="utf-8")
+        handler.setLevel(logging.INFO)
```
SamYuan1990, line 182: wait a min, the default logger in global level is debug, should we keep the same log level?

benchmark/conftest.py
```diff
@@ -165,12 +167,21 @@ def pytest_configure(config):
...
+        handler.setFormatter(logging.Formatter("[%(levelname)s] %(message)s"))
+        recordLogger.addHandler(handler)
+        recordLogger.setLevel(logging.INFO)
```
SamYuan1990, line 185: may I know why setLevel twice?

benchmark/performance_utils.py
```diff
@@ -372,7 +371,7 @@ def run(self):
                 shape_desc=self.shape_desc,
             )
             print(attri)
-            logging.info(attri.to_dict())
+            recordLogger.info(attri.to_dict())
```
SamYuan1990, line 374: Sorry, I don't get time to go through all 374 lines above. 
@kiddyjinjin , do we tested `logger = logging.getLogger("flag_gems.benchmark.record")` before?

My question: as default logger(`logging.getLogger`) is able to log file by `enable_gems(....)`, and in this file, which using logging without getLogger seems wrong.

Back to `Since this PR focuses on fixing the missing log file issue...`. Which means, we confirmed and tested the log file will missing in benchmark testing, when both `enable_gems(....)` and `logging.getLogger` configured by design?

dongjibin1996: Let me clarify the relationship between enable_gems and this fix, and provide evidence for the issue.

1. Relationship with enable_gems(...) : They are independent.

- enable_gems(...) configures the runtime debug logger (e.g., logging.getLogger("flag_gems") ) for operator dispatch info.
- This PR addresses the benchmark result logger (previously using Root Logger, now flag_gems.benchmark.record ), which stores the final performance metrics (JSON).
  Using logging.getLogger in this PR is the correct practice to isolate these two purposes and avoid conflicts.
2. The Verified Issue: We confirmed that transformer_engine (verified on version 1.13.0+7a6085c with Python 3.10) automatically configures the Root Logger during import. This causes the subsequent logging.basicConfig in our benchmark suite to be ignored (no-op), resulting in missing log files .

3. Verification Steps: You can verify this conflict with this one-liner:

```
python -c "import logging; print('Before:', logging.root.handlers); import transformer_engine.
pytorch.cpp_extensions as tex; print('After:', logging.root.handlers)"
```
    Output:

```
Before: []
After: [<StreamHandler <stderr> (NOTSET)>]  <-- This handler blocks our basicConfig
```
Conclusion: We have tested that without this fix, pytest ... --record log fails to generate the result file in this environment. This PR fixes it by using a dedicated logger that bypasses the polluted Root Logger.

benchmark/conftest.py
dongjibin1996: This logger is dedicated to recording benchmark results (metrics & JSON summaries) rather than debug traces.

Using logging.INFO is intentional here because we only want to capture the final performance data. Setting it to DEBUG isn't necessary for this purpose and might introduce unwanted noise if we ever add debug logs to this specific logger in the future.

We are strictly maintaining the original behavior, as the previous code also used level=logging.INFO.
bfb5aa19f46f80b5ba8b00d43e3da0a1

benchmark/performance_utils.py
dongjibin1996: Give the answer in another comment

benchmark/conftest.py
dongjibin1996: This is necessary because we are configuring a named non-root logger to isolate our benchmark results, which follows the standard Python logging behavior (as seen in the [Logging Cookbook ](https://docs.python.org/3/howto/logging-cookbook.html#using-logging-in-multiple-modules)).

Why set level twice?

1. recordLogger.setLevel(logging.INFO) :
    Purpose : Overrides the default level inheritance.
    Reason : By default, a named logger inherits the Root Logger's level (usually WARNING ). If we don't explicitly set this logger to INFO , our messages will be filtered out before they even reach any handlers.

2. handler.setLevel(logging.INFO) :
    Purpose : Ensures this specific FileHandler records INFO messages.
    Reason : While strictly setting the Logger level is often sufficient if the Handler is NOTSET , explicitly setting the Handler level is a best practice recommended by the Python docs to guarantee that this specific sink receives the intended logs, regardless of future changes to the logger hierarchy.
  
So, setting both ensures our log isolation works correctly and robustly."

benchmark/conftest.py
dongjibin1996: This logic is for cleanup (removing stale handlers before adding a new one).

If h.close() fails (e.g., file already closed or permission issue on an old handle), we generally want to ignore it because:

1. It shouldn't block the new test run from starting.
2. We are removing this handler from the logger anyway ( removeHandler is called right before).
However, you are right that a silent pass might hide underlying issues. I have updated the code to raise a warning ( warnings.warn ) instead of silently passing, so we can be aware if something goes wrong while still keeping the cleanup process robust.

benchmark/performance_utils.py
SamYuan1990: @tengqm , I will skip review this PR for now.

https://github.com/flagos-ai/FlagGems/blob/4eda19ff41955ec929188366a6874f8ffaa4a5e0/benchmark/performance_utils.py#L427-L428

I am confusing....it seems we are not using standard pytest or pytest-benchmark plugin....
and this PR just to save metrics result....
seems we have print and logging... so... just pytest xxx > some_file and change result.to_json() from logging.info to print will work?

@kiddyjinjin , @dongjibin1996  please move/go ahead.

tengqm (approved):

PR #1267, opened 2025-12-24
flag_gems only_enable api and unit tests

littlekid-zxt:
### PR Category

User Experience | OP Test

### Type of Change

New Feature

### Description

#### Changes:

- Encapsulate the op config in `src/flag_gems/__init__.py` into the `_get_full_config()` api for easier access and modification
- Add the `only_enable(...)` api in `src/flag_gems/__init__.py` to support specifying only_used ops for replacing the FlagGems implementation
- Improve the implementation of the `only_enable(...)` api and `only_used` param in the `__init__`, `__enter__`, and `__exit__` functions of the `class use_gems`
- Add unit test case for three scenarios: `flag_gems.enable(...)`, `flag_gems.enable(unused=unused_op,...)`, `flag_gems.only_enable(only_used=only_used_op,...)`
- Based on the `class use_gems`, add unit test case for three scenarios: `flag_gems.use_gems(... | unused=unused_op,... | only_used=only_used_op,...)`

#### Usage:

```
   only_used_op = ["sum", "mul", "add"]
   flag_gems.only_enable(
      only_used=only_used_op,
      record=True,
      path=log_file_name
   )
```
```
   only_used_op = ["mul", "add"]
   with flag_gems.use_gems(only_used=only_used_op, record=True, path=log_file_name):
      ...
```
```
   cd FlagGems/tests/
   pytest test_enable_api.py -s
```
### Issue

N/A

### Progress

### Performance

N/A

### DIFF
--- a/benchmark/test_special_perf.py
+++ b/benchmark/test_special_perf.py
@@ -561,3 +561,72 @@ def test_perf_moe_sum():
     )
     bench.set_gems(gems_op)
     bench.run()
+
+
+try:
+    import os
+
+    os.environ["VLLM_CONFIGURE_LOGGING"] = "0"
+    import vllm._custom_ops as vllm_ops
+
+    HAS_VLLM = True
+except ImportError:
+    HAS_VLLM = False
+
+
+@pytest.mark.moe_align_block_size
+@pytest.mark.skipif(not HAS_VLLM, reason="vllm not installed")
+def test_perf_moe_align_block_size():
+    def moe_align_block_size_input_fn(shape, dtype, device):
+        # ------------ parameters ------------
+        num_experts = shape[0]
+        block_size = shape[1]
+        dtype = torch.int32
+        topk_ids = torch.randint(0, num_experts, (3, 4), dtype=dtype, device=device)
+        max_num_tokens_padded = topk_ids.numel() + num_experts * (block_size - 1)
+
+        # padded_num_experts in vllm._custom_ops.moe_align_block_size
+        # must be less than 1024
+        if max_num_tokens_padded >= 1024:
+            return
+
+        sorted_ids = torch.empty((max_num_tokens_padded,), dtype=dtype, device=device)
+        max_num_m_blocks = max_num_tokens_padded // block_size
+        expert_ids = torch.empty((max_num_m_blocks,), dtype=dtype, device=device)
+        num_tokens_post_pad = torch.empty(1, dtype=dtype, device=device)
+
+        yield (
+            topk_ids,
+            num_experts,
+            block_size,
+            sorted_ids,
+            expert_ids,
+            num_tokens_post_pad,
+        )
+
+    class MoeAlignBlockSizeBenchmark(GenericBenchmark2DOnly):
+        def set_more_shapes(self):
+            return [
+                (16, 8),
+                (16, 16),
+                (16, 32),
+                (32, 8),
+                (32, 16),
+                (32, 32),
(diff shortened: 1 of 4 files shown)


src/flag_gems/__init__.py
```diff
@@ -37,358 +358,98 @@ def enable(
...
+        tuple(used_config),
+        user_unused_ops_list=[],
+        cpp_patched_ops_list=list(set(aten_patch_list)),
+        lib=lib,
+    )
+    setup_flaggems_logging(path=path, record=record, once=once)
 
 class use_gems:
-    def __init__(self, unused=None, record=False, once=False, path=None):
+    def __init__(self, unused=None, only_used=None, record=False, once=False, path=None):
```
tengqm: I'd suggest we rename `unused` and `only_used` to `exclude` and `include` respectively. "only_used" is not a good name for a parameter, "unused" is not a good name either. These words are okay to describe a status, but not very accurate for describing your intent.

Consider these:

```python
use_gems(exclude=["foo"])
use_gems(include=["bar", "baz"])
```

These read more natural.

We may also want to add some comment clarifying that what happens if a user specifies both `include` and `exclude`. Is it an invalid setting?
Or we will ignore `exclude` if `include` is specified?

src/flag_gems/__init__.py
```diff
@@ -37,358 +358,98 @@ def enable(
...
         self.lib = torch.library.Library("aten", "IMPL")
         self.unused = [] if unused is None else unused
+        self.only_used = [] if only_used is None else only_used
```
tengqm: I'd write

```python
self.include = include if isinstance(include, list) else []
```

tests/test_enable_api_register_error.py
```diff
@@ -0,0 +1,95 @@
+# RuntimeError: This is not allowed since there's already a kernel registered from python overriding add.Tensor's behavior for CUDA dispatch key and aten namespace.
+import flag_gems
+import pytest
+import torch
+import time
+import os
+import re
+
+# @pytest.mark.enable
```
tengqm: Please try `@pytest.mark.skip` (https://docs.pytest.org/en/stable/how-to/skipping.html#skipping-test-functions) for tests that are skipped temporarily instead of commenting the code out.

tests/test_enable_api.py
```diff
@@ -0,0 +1,100 @@
...
+# it seems to be related to FileHandler.
+import flag_gems
+import pytest
+import torch
+import time
+import os
+import re
+
+@pytest.mark.enable
+def test_enable():
```
tengqm: What's the purpose of this test case?

tests/test_enable_api.py
```diff
@@ -0,0 +1,100 @@
...
+      b = torch.tensor([4.0, 5.0, 6.0], device=flag_gems.device)
+      c = a + b
+      d = a * b
+      e = flag_gems.sum(a)
+      f = torch.sum(a)
+
+      print(f"add result: [{c}]")
+      print(f"mul result: [{d}]")
+      print(f"sum result flaggems: [{e}]")
+      print(f"sum result torch: [{f}]")
```
tengqm: This test case is not asserting anything, why?
Test cases are supposed to be executed silently without "polluting" the console ... shall we remove these `print` calls?

src/flag_gems/__init__.py
kiddyjinjin: nice suggestion~

PR #1293, opened 2025-12-28
Remove star imports for the top-level module

tengqm:
### PR Category

Other

### Type of Change

Other

### Description

Star-imports (`from foo import *`) is a bad programming practices in Python, though valid.
It hurts the readability of the source code. Sometimes it introduces ambiguity and subtle
bugs that are very difficult to identify.

The current `src/flag_gems/__init__.py` fails the ruff lint check for this reason and
ruff has a good reason to complain for these code. This PR is an attempt to fix it for the
top-level module.

This PR revises the top-level module so that func calls are prefixed with a package name.
This helps clarify where each func comes from and it helps eliminate the ambiguous func names
that are keyword or built-in functions such as:

- `sum`
- `any`
- `all`
- `max`
- `min`
- `abs`

This PR also cleans the `src/flag_gems/__init__.py` for unused imports such as
`from flag_gems import testing` which is not only unused but also an anti-pattern for module
dependency.

### Issue

N/A

### Progress

### Performance

N/A


tengqm:
This doesn't sound like a viable solution. CLosing for now.
Need to revisit this later.

PR #1296, opened 2025-12-29
update flag_gems only_enable api and unit tests

littlekid-zxt:
### PR Category

User Experience | OP Test

### Type of Change

New Feature

### Description

#### Changes:

- Encapsulate the op config in `src/flag_gems/__init__.py` into the `_get_full_config()` api for easier access and modification;
- Rename `unused` to `exclude`;
- Add the `only_enable(...)` api in `src/flag_gems/__init__.py` to support specifying include ops for replacing the FlagGems implementation;
- Improve the implementation of the `only_enable(...)` api and `include` param in the `__init__`, `__enter__`, and `__exit__` functions of the `class use_gems`;
- Based on the `class use_gems`, add unit test cases for three scenarios: `flag_gems.use_gems(... | exclude=exclude_op,... | include=include_op,...)`.

#### Usage:

```
   exclude_op = ["mul", "add"]

   flag_gems.enable(
      exclude=exclude_op,
      record=True,
      path=log_file_name
   )
   # or
   with flag_gems.use_gems(exclude=exclude_op, record=True, path=log_file_name):
      ...
```
```
   include_op = ["sum", "mul", "add"]

   flag_gems.only_enable(
      include=include_op,
      record=True,
      path=log_file_name
   )
   # or
   with flag_gems.use_gems(include=include_op, record=True, path=log_file_name):
      ...
```
```
   cd FlagGems/tests/
   pytest test_enable_api.py
```

### Issue

N/A

### Progress

### Performance

N/A

### DIFF
--- a/.github/CODEOWNERS
+++ b/.github/CODEOWNERS
@@ -3,14 +3,14 @@
 # @global-owner1 and @global-owner2 will be requested for
 # review when someone opens a pull request.
 
-/src/flag_gems/*      @kiddyjinjin   @0x45f
+/src/flag_gems/*      @kiddyjinjin   @0x45f   @zhangpeiyang1  @huangyiqun
 /src/flag_gems/runtime/backend/*
 
-/tests/*             @kiddyjinjin   @0x45f
-/modules_tests/*     @kiddyjinjin   @0x45f
-/benchmark/*         @kiddyjinjin   @0x45f
+/tests/*             @kiddyjinjin   @0x45f   @zhangpeiyang1  @huangyiqun
+/modules_tests/*     @kiddyjinjin   @0x45f   @zhangpeiyang1  @huangyiqun
+/benchmark/*         @kiddyjinjin   @0x45f   @zhangpeiyang1  @huangyiqun
 
 
-/lib/*                 @zhangpeiyang1  @huangyiqun
-/ctests/*              @zhangpeiyang1  @huangyiqun
-/include/*             @zhangpeiyang1  @huangyiqun
+/lib/*                @kiddyjinjin   @0x45f   @zhangpeiyang1  @huangyiqun
+/ctests/*             @kiddyjinjin   @0x45f   @zhangpeiyang1  @huangyiqun
+/include/*            @kiddyjinjin   @0x45f   @zhangpeiyang1  @huangyiqun


src/flag_gems/__init__.py
```diff
@@ -26,369 +26,429 @@
...
-        ),
-        user_unused_ops_list=list(set(unused or [])),
+        _get_full_config(),
+        user_unused_ops_list=list(set(exclude or [])),
         cpp_patched_ops_list=list(set(aten_patch_list)),
         lib=lib,
     )
     setup_flaggems_logging(path=path, record=record, once=once)
 
+def only_enable(
```
tengqm: From semantics' perspective, it seems that there is no need to add the `only_enable` function. We can use the `enable()` method/function to do the job. Users can either specify `include` for operators that are explicitly included, or use `exclude` to specify operators that are explicitly excluded.

src/flag_gems/__init__.py
```diff
@@ -26,369 +26,429 @@
 def torch_ge(v):
     return version.parse(torch.__version__) >= version.parse(v)
 
+def _get_full_config():
```
tengqm: There is no moving parts in this function. The only purpose, as far as I can tell, is to construct a variable. We are then passing that variable around.
A more straightforward way could be just define this as a "global" variable and reference that variable wherever needed. Wrapping this into a helper function doesn't buy us anything.
Thoughts?

src/flag_gems/__init__.py
```diff
@@ -26,369 +26,429 @@
...
+        ("vstack", vstack),
+        ("where.ScalarOther", where_scalar_other),
+        ("where.ScalarSelf", where_scalar_self),
+        ("where.self", where_self),
+        ("where.self_out", where_self_out),
+        ("zeros", zeros),
+        ("zeros_like", zeros_like),
+        ("scatter_add_", scatter_add_),
+        ("dreglu", dreglu),
+        ("reglu", reglu),
```
tengqm: This tuple is out of order, thus difficult to maintain.
Please consider sort it in alphabetic order.

tests/test_enable_api.py
```diff
@@ -0,0 +1,111 @@
...
+   path_file = get_path_log_name()
+   with flag_gems.use_gems(exclude=exclude_op, record=True, path=path_file):
+      a = torch.tensor([1.0, 2.0, 3.0], device=flag_gems.device)
+      b = torch.tensor([4.0, 5.0, 6.0], device=flag_gems.device)
+      _ = a + b
+      _ = a * b
+      _ = flag_gems.sum(a)
+      _ = torch.sum(a)
+  
+   op_names_str = "_".join(exclude_op).replace(".", "_").replace("-", "_")
```
tengqm: This is a useful function.
We may want to consider making it a util rather than copying it everywhere.

tests/test_enable_api.py
```diff
@@ -0,0 +1,111 @@
...
+   path_file = get_path_log_name()
+   with flag_gems.use_gems(record=True, path=path_file):
+      a = torch.tensor([1.0, 2.0, 3.0], device=flag_gems.device)
+      b = torch.tensor([4.0, 5.0, 6.0], device=flag_gems.device)
+      _ = a + b
+      _ = a * b
+      _ = flag_gems.sum(a)
+      _ = torch.sum(a)
+
+   log_file = f"./gems_enable_all_ops.log"
```
tengqm: The `f` prefix is not needed

tests/test_enable_api.py
```diff
@@ -0,0 +1,111 @@
+import flag_gems
+import pytest
+import torch
+import os
+import re
+import atexit
+
+
+def get_path_log_name():
+   return "./gems_enable_test.log"
```
tengqm: Please clarify the reason behind creating a new file for verification.

src/flag_gems/__init__.py
kiddyjinjin: From a semantics point of view, I agree that enable() could theoretically cover this use case.
The current enable() signature is designed as:

```
def enable(
    lib=aten_lib,
    unused=None,
    registrar=registrar,
    record=False,
    once=False,
    path=None,
):
```

the enable() API is already used by both internal users (flagscale) and external projects (e.g. RWKV). Changing its parameters or behavior would break backward compatibility.

To avoid breaking existing users, we propose introducing a new `only_enable` API to handle this more specific use case, where only the explicitly listed operators should be enabled.

tengqm: What I was suggesting is that adding a new keyword argument `exclude` to the `enable` method. Let `exclude` default to None`. You are not breaking existing users by adding an optional argument.

tengqm:
why close this PR? @littlekid-zxt

littlekid-zxt:
> why close this PR? @littlekid-zxt

I need to revise and resubmit based on your suggestions.

PR #1304, opened 2025-12-29
[Feature] Enable experimental operators through enable tag

Dongxu-H:
### PR Category
Operator Implementation
Operator Test

### Type of Change
New tag support

### Description
Experimental ops are a lot but we can't import them one by one.
We'd better to add a tag to support to enable them together.

In this design, we import a new tag called "experimental_ops" in enable key.
All experimental_ops can be called through `flag_gems.enable(experimental_ops=["logit", "rms_norm"])`

what will happen if we use `flag_gems.enable(experimental_ops=["logit", "rms_norm"])` instead of `flag_gems.enable()`

1. if the experimental op name is not a valid aten op, e.g., `flag_gems.enable(experimental_ops=["blablabla"])`, just ignore it;
2. if the experimental op name is valid and it is not in flag_gems standard list, just add it and use it;
3. if the experimental op name is valid and there is another the same op in flag_gems standard list, use experimental op. Because we treat uses want to use experimental op first so as to add its name here by default.

### Issue
#1305

### Progress

- [√] Change is responded to an issue.
- [√] Change is fully covered by a UT.


src/flag_gems/__init__.py
```diff
@@ -30,333 +31,354 @@ def torch_ge(v):
...
+        ("conv2d.padding", conv2d),
+        ("conv3d.padding", conv3d),
+    ]
+
+    experimental_keys = {item[0] for item in experimental_ops_mappings}
+    filtered_standard_mappings = [
+        item for item in standard_mappings if item[0] not in experimental_keys
+    ]
+
+    final_mappings = filtered_standard_mappings + experimental_ops_mappings
```
tengqm, line 381: say operator 'foo' is experimental, we can exclude 'foo' from the `standard_mapping` list, why bother adding it and then remove it?

tengqm:
@Dongxu-H There is a merge conflict to be resolved, fyi.

tengqm:
@Dongxu-H Can/shall we split this PR into two? One for operator renaming, the other for adding the experimental_ops parameter? These two changes are independent from each other, IIUC.

src/flag_gems/__init__.py
Dongxu-H: > say operator 'foo' is experimental, we can exclude 'foo' from the `standard_mapping` list, why bother adding it and then remove it?

in line "filtered_standard_mappings = [item for item in standard_mappings if item[0] not in experimental_keys]"
we exclude 'foo' from the standard_mapping to build a new mapping called "filtered_standard_mappings"

then "filtered_standard_mappings " + "experimental_ops_mappings ", there is only one 'foo' in the final list (from experimental_ops)

tengqm: At the end of the day, you are using `final_mappings` for registration.
- If all experimental operators are supposed to be included in the "standard" mapping, the standard mapping is a superset of the experimental ones. We can simply use the "standard" mapping for registration.
- If no experimental operators are supposed to appear in the "standard" mapping, and we want to register the union of the two mapping, we don't need to do filtering.
- If some experimental operators are to appear in the "standard" mapping, while others won't show up, we are creating a huge problem for future maintenance.

I'm not challenging what you are doing here. I'm wondering whether we can make the logic straightforward and easy to maintain.

Dongxu-H: Got it.
I think this is a design issue, maybe we need to make this design clearer.

If a user adds `flag_gems.enable(experimental_ops=["logit"])` and the function `logit` is already implemented in both standard and experimental, which should be called.
Here I set to call experimental op because user add it explicitly

tengqm: Is it possible we have the same operator implemented as both a standard one and an experimental one? I can get the idea that the maturity level of an operator may change over time (alpha->beta->stable), but I don't quite get the idea how the same operator shows up in these two sets.

tengqm:
@Dongxu-H Do we still need this PR? There are conflicts to be solved.

bin913:
@Dongxu-H Do you still keep track of this PR?

PR #1316, opened 2025-12-31
Enhance register

lxd-cumt:
### PR Category
Other

### Type of Change
Bug Fix

### Description
Enhance registry: for the same Op, allow multiple enable calls without actually registering it and without raising an error.

### DIFF
--- a/src/flag_gems/runtime/register.py
+++ b/src/flag_gems/runtime/register.py
@@ -44,7 +44,6 @@ class Register:
         return []
 
     def register_impl(self, key, fn):
-        device_key = self.reg_key
         # Check if this operation has already been registered
         if key in _registered_ops:
             # Skip registration - already registered
@@ -52,7 +51,7 @@ class Register:
         # Register the operation
         self.all_ops.append(key)
         _registered_ops.add(key)
-        self.lib.impl(key, fn, device_key)
+        self.lib.impl(key, fn, self.reg_key)
 
     def for_each(self):
         try:


kiddyjinjin (dismissed):
lgtm

src/flag_gems/runtime/register.py
```diff
@@ -42,7 +45,14 @@ def get_vendor_unused_op(self):
 
     def register_impl(self, key, fn):
         device_key = self.reg_key
+        # Check if this operation has already been registered
+        op_key = (key, device_key)
+        if op_key in _registered_ops:
```
Galaxy1458: It is impossible for different `device_key` to appear within the same process, so you can directly use the `self.all_ops` attribute to perform the check.

Galaxy1458:
I’d like to ask in what scenarios this issue would occur？ Since the registered operators are all built into `Flaggems`

lxd-cumt:
> I’d like to ask in what scenarios this issue would occur？ Since the registered operators are all built into `Flaggems`

The current training flow is: Megatron-LM → TransformerEngine-FL.

The ops inside TransformerEngine-FL are split into three backends: gems, cuda, and torch. When gems is required, we activate it with the context manager with flag_gems.use_gems().

The problem is: if I also want the ops used by Megatron-LM itself to run through gems, I have to invoke with flag_gems.use_gems() (or flag_gems.enable()) again, which triggers a “duplicate registration” error.

src/flag_gems/runtime/register.py
lxd-cumt: done

src/flag_gems/runtime/register.py
```diff
@@ -42,7 +45,13 @@ def get_vendor_unused_op(self):
 
     def register_impl(self, key, fn):
         device_key = self.reg_key
```
tengqm: The `device_key` variable is not needed?

lxd-cumt: The device_key is used for lib register, and not used for duplicate ops check.
ref: https://github.com/flagos-ai/FlagGems/pull/1316#discussion_r2654681912

tengqm: I mean, how about remove line 47 and use `self.reg_key` directly in line 55?

lxd-cumt: done, thx

src/flag_gems/runtime/register.py
```diff
@@ -41,9 +44,14 @@ def get_vendor_unused_op(self):
         return []
 
     def register_impl(self, key, fn):
-        device_key = self.reg_key
+        # Check if this operation has already been registered
+        if key in _registered_ops:
+            # Skip registration - already registered
+            return
```
tengqm, line 50: Okay, now back to the logic itself.
We now have two places for the whole set/collection of ops: a) `self.all_ops`, b) global var `_registered_ops`.
The logic from line 52 and 53 means we are duplicating the registration to two places. Why is that necessary? Can we simply do `if key in self.all_ops:` and remove the new `_registered_ops` variable?

lxd-cumt: _registered_ops is global to prevent duplicate registrations into the same library across instances.

If each Register instance uses a different lib, checking with self.all_ops is sufficient.
However, if multiple Register instances share the same library (e.g., the default aten_lib), global deduplication is still necessary to avoid duplicate registrations.

Therefore, I recommend keeping the global _registered_ops.

tengqm: Please confirm that `Register` (which should be renamed to something like `Registry` or `OpsRegistry`) is not a singleton.
I'm not sure if having multiple `Register` instances is needed or if it is a good design.

lxd-cumt: I’ll confirm it with the FlagGems design team, thx

tengqm: @lxd-cumt Where are we now?

lxd-cumt: We have bypassed the duplicate registration on the framework side and will no longer use this PR.

PR #1338, opened 2026-01-03
[LibCache] fix warnings on "AutoBase.prepare()"

### LINKED ISSUE #1323: Eliminate the SAWarning at the sqlalchemy layer
The current DB table automap mechanism is triggering an annoying warning when running tests. This warning is not hurting the correctness of the software but it may have some slight impact over the performance results. It is pretty annoying when there are real defects discovered from unit tests.

One of the candidate solutions is to add the following logic to the `src/flag_gems/utils/models/sql.py` file:

```python
import warnings
from sqlalchemy.exc import SAWarning
warnings.filterwarnings("ignore", category=SAWarning)
```

There could be other alternatives though.

sgjzfzzf:
### PR Category

This PR is used to resolve the warning that occurs when multiple workers attempt to create tables simultaneously. Although the original code provides a cache to avoid preparing the same `AutoBase` multiple times, it lacks a thread-level lock, so in multi-threaded cases, a race condition could occur, allowing some threads to attempt to prepare the same `AutoBase`. A thread-level lock is necessary to resolve this warning.

### Issue

Resolves #1323 .

### Progress

### DIFF
--- a/src/flag_gems/utils/models/sql.py
+++ b/src/flag_gems/utils/models/sql.py
@@ -94,17 +94,13 @@ class SQLPersistantModel(PersistantModel):
             ModelCls: Optional[Type[Base]] = self.sql_model_pool.get(name)
             if ModelCls is not None:
                 return ModelCls
-            ModelCls: Optional[Type[Base]] = SQLPersistantModel.build_sql_model_by_db(
-                name, self.engine
-            )
+            ModelCls = SQLPersistantModel.build_sql_model_by_db(name, self.engine)
             if ModelCls is not None:
                 self.sql_model_pool[name] = ModelCls
                 return ModelCls
             if not keys or not values:
                 return None
-            ModelCls: Type[Base] = SQLPersistantModel.build_sql_model_by_py(
-                name, keys, values
-            )
+            ModelCls = SQLPersistantModel.build_sql_model_by_py(name, keys, values)
             with self.engine.begin() as conn:
                 conn.execute(
                     sqlalchemy.schema.CreateTable(


src/flag_gems/utils/models/sql.py
```diff
@@ -90,26 +90,29 @@ def get_sql_model(
...
-            return ModelCls
-        ModelCls: Optional[Type[Base]] = SQLPersistantModel.build_sql_model_by_db(
-            name, self.engine
-        )
-        if ModelCls is not None:
+        with self.lock:
+            ModelCls: Optional[Type[Base]] = self.sql_model_pool.get(name)
+            if ModelCls is not None:
+                return ModelCls
+            ModelCls: Optional[Type[Base]] = SQLPersistantModel.build_sql_model_by_db(
```
tengqm:
```suggestion
            ModelCls = SQLPersistantModel.build_sql_model_by_db(name, self.engine)
```

src/flag_gems/utils/models/sql.py
```diff
@@ -90,26 +90,29 @@ def get_sql_model(
...
+                name, self.engine
+            )
+            if ModelCls is not None:
+                self.sql_model_pool[name] = ModelCls
+                return ModelCls
+            if not keys or not values:
+                return None
+            ModelCls: Type[Base] = SQLPersistantModel.build_sql_model_by_py(
```
tengqm:
```suggestion
            ModelCls = SQLPersistantModel.build_sql_model_by_py(name, keys, values)
```

tengqm (approved):
Nice.

kiddyjinjin (approved):

src/flag_gems/utils/models/model.py
```diff
@@ -10,6 +11,7 @@ class PersistantModel(object):
 
     def __init__(self, *args, **kwargs) -> None:
         super().__init__(*args, **kwargs)
+        self.lock: Final[threading.Lock] = threading.Lock()
```
SamYuan1990, line 14: @tengqm, @kiddyjinjin we should consider Rlock in the feature.

SamYuan1990 (commented):
From performance considering, we should consider Lock and RLock.

src/flag_gems/utils/models/sql.py
```diff
@@ -90,26 +90,25 @@ def get_sql_model(
...
         values: Mapping[str, Union[Any, Type]] = {},
     ) -> Callable[[str, Optional[Mapping[str, Type]]], Optional[Type[Base]]]:
-        ModelCls: Optional[Type[Base]] = self.sql_model_pool.get(name)
-        if ModelCls is not None:
-            return ModelCls
-        ModelCls: Optional[Type[Base]] = SQLPersistantModel.build_sql_model_by_db(
-            name, self.engine
-        )
-        if ModelCls is not None:
+        with self.lock:
```
SamYuan1990, line 93: consider with lock scope.

Issue #1320, opened 2025-12-31
Explicitly declare Python packages requirements for the project.


tengqm:
### The feature, motivation and pitch

We cannot assume that users have each and every Python package we use in our development environment. For a smooth installation, the project has to explicitly declare its package dependencies, for example, using the `pyproject.toml` file.

This is an umbrella issue. We need to collect requirements for the following scenarios:

- runtime requirements: this can be broken down by platforms, models, etc.
- development requirements: packages that are needed only during development.
- testing requirements: including linting, unit tests, ... etc.
- tooling requirements: packages required by different tools.

With a clear hierarchy of requirements, the dependent packages can be installed with one command.

A related issue is to pin down the specific version of packages we use. This is for:

- SBOM inspection, ensuring that we can quickly respond to security incidents;
- Bugs triage, making sure that a problem reported can be reproduced and diagnosed quickly.

bin913:
Shall we follow up on this further?

PR #1346, opened 2026-01-04
Update project package dependency

tengqm:
### PR Category

Other

### Type of Change

Other

### Description

This PR updates the project package dependency. 

### Issue

- Relates #1320

### Progress

### Performance

N/A

### DIFF
--- a/docs/installation.md
+++ b/docs/installation.md
@@ -9,23 +9,26 @@ Since FlagGems has not yet been released on PyPI, it can only be installed from 
 ### Clone the source
 
 ```shell
-git clone https://github.com/FlagOpen/FlagGems
+git clone https://github.com/flagos-ai/FlagGems
 cd FlagGems/
 ```
 
 ### Install FlagTree
 
 [FlagTree](https://github.com/flagos-ai/flagtree/) is an open source, unified compiler for multiple AI chips project.
-If you want to use the native Triton instead of FlagTree, please skip this step.
 
 For details on environment requirements for other backends, please check the
 [build](https://github.com/flagos-ai/FlagTree/blob/main/documents/build.md)
 documentation in the FlagTree project.
 
 ```shell
-# Other backends: replace with requirements_backendxxx.txt
-pip install -r flag_tree_requirements/requirements_nvidia.txt
+uv pip install .[flagtree_nvidia]
 ```
+
+> [!Tip]
+> - If you want to use the native Triton instead of FlagTree, you can skip this step.
+> - The currently supported variants are `flagtree_ascend`, `flagtree_enflame`, `flagtree_hcu`,
+>   `flagtree_iluvatar`, `flagtree_nvidia`, `flagtree_xpu`.
 
 ### Build system
 
@@ -55,7 +58,7 @@ install the required packages.
 Example command:
 
 ```shell
-pip install -U scikit-build-core>=0.11 pybind11 ninja cmake
+pip install -U scikit-build-core>=0.11 pybind11>3.0.1 ninja>=1.13.0 cmake>=4.2.1
 ```
 
 FlagGems can be installed as either a pure python package or a package with C extensions.
@@ -123,10 +126,10 @@ Commonly used pip options are:
 
 Example recipes are provided for easy copy-and-paste.
 
-Editable installation with external TritonJIT
+Editable installation with external TritonJIT:
 
 ```shell
-pip install -U scikit-build-core ninja cmake pybind11
+pip install -U scikit-build-core>=0.11 pybind11>3.0.1 ninja>=1.13.0 cmake>=4.2.1
 
 CMAKE_ARGS="-DFLAGGEMS_BUILD_C_EXTENSIONS=ON -DFLAGGEMS_USE_EXTERNAL_TRITON_JIT=ON -DTritonJIT_ROOT=<install path of triton-jit>" \
 pip install --no-build-isolation -v -e .
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -1,14 +1,13 @@
 [build-system]
-requires = ["setuptools>=64.0", "scikit-build-core>=0.11", "pybind11"]
+requires = ["setuptools>=64.0", "scikit-build-core>=0.11", "pybind11>=3.0.1", "uv>=0.9.26"]
 build-backend = "scikit_build_core.build"
-
 
 [project]
 name = "flag_gems"
 version = "4.2.1.rc.0"
 description = "FlagGems is a function library written in Triton."
 readme = { file = "README.md", content-type = "text/markdown" }
-requires-python = ">=3.8.0"
+requires-python = ">=3.10.0"
 license = { text = "Apache Software License" }
 authors = [
     { name = "Zhixin Li", email = "strongspoon@outlook.com" },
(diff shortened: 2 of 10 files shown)


pyproject.toml
```diff
@@ -22,27 +21,55 @@ classifiers = [
     "License :: OSI Approved :: Apache Software License",
 ]
 dependencies = [
-    "torch>=2.2.0",
-    "packaging",
-    "pybind11",
-    "PyYAML",
-    "packaging",
-    "sqlalchemy",
+    "acl>=0.2",  # used by Ascend cumsum operator
```
SamYuan1990: Ascend is one of deps for this repo? not conditional one?

pyproject.toml
```diff
@@ -22,27 +21,55 @@ classifiers = [
...
+    "typing_extensions>=4.15.0",  # core model dependency
+    "vllm>=0.13.0",  # core dependency
 ]
 
+[project.urls]
+Homepage = "https://github.com/flagos-ai/FlagGems"
+
 [project.optional-dependencies]
+benchmark = [
+    "aiohttp>=3.13.3",  # models benchmark, Apache-2.0 and MIT
```
SamYuan1990: license can leave to SBOM, there lots of python project SBOM scan tool. I suppose no need duplicate in pyproject.toml

SamYuan1990 (commented):
LGTM, but it's hard to generate SBOM without requirements.txt or xxx.lock file.
Considering with vendor case....

Can we lock dependency version by this PR?
Anyway, a update is a good start.

pyproject.toml
tengqm: good question. I'm not sure if we have an option to opt-out certain backends. My assumption is that all backends are built/delivered in the same package.

pyproject.toml
tengqm: This is a comment for ourselves. If we know the license of a dependent package, we don't need a fancy tool to detect it.
We can remove these comments later if 1) SBOM is in place; 2) these notes are seemed unnecessary.

pyproject.toml
```diff
@@ -22,27 +21,55 @@ classifiers = [
...
-    "packaging",
-    "sqlalchemy",
+    "acl>=0.2",  # used by Ascend cumsum operator
+    "numpy>=2.4.0",  # used by backends, tests and utils
+    "packaging>=25.0",  # used for packaging and version determination
+    "pybind11>=3.0.1",  # TODO(Qiming): check if this is a runtime dependency
+    "PyYAML>=6.0.3",  # aka. 'yaml', used for benchmark, backend and runtime.
+    "SQLAlchemy>=2.0.45",  # aka. 'sqlalchemy', used in core
+    "torch>=2.2.0",  # core dependency
+    "torch_npu>=2.8.0",  # core dependency
```
Galaxy1458: `torch_npu` is only applied for the `ascend` backend.

tengqm: it is imported in `src/flag_gems/runtime/backend/device.py` which is a generic module.

Galaxy1458: Yes, but `src/flag_gems/runtime/backend/device.py ` is a generic module for obtaining multiple backend devices. It's normal for there to be specialized backend modules in this module. And the `torch_npu`  is in the `try-except` block, specially designed for detecting `HuaweiAscend` backend [device.py#L86](https://github.com/flagos-ai/FlagGems/blob/master/src/flag_gems/runtime/backend/device.py#L86)

pyproject.toml
```diff
@@ -22,27 +21,54 @@ classifiers = [
...
-    "packaging",
-    "sqlalchemy",
+    "numpy>=2.4.0",  # used by backends, tests and utils
+    "packaging>=25.0",  # used for packaging and version determination
+    "pybind11>=3.0.1",  # TODO(Qiming): check if this is a runtime dependency
+    "PyYAML>=6.0.3",  # aka. 'yaml', used for benchmark, backend and runtime.
+    "SQLAlchemy>=2.0.45",  # aka. 'sqlalchemy', used in core
+    "torch>=2.2.0",  # core dependency
+    "torch_npu>=2.8.0",  # core dependency
+    "triton>=3.5.1",  # core dependency
```
Galaxy1458: `"triton>=3.5.1"` ： The restrictions on the `triton` version is too high, recommend `"triton>=3.1"`

pyproject.toml
```diff
@@ -22,27 +21,54 @@ classifiers = [
...
+    "typing_extensions>=4.15.0",  # core model dependency
+    "vllm>=0.13.0",  # core dependency
```
Galaxy1458: `vllm`  should not be a dependency of `flag_gems`.  @kiddyjinjin Please confirm.

pyproject.toml
tengqm: This dependency is complicated. We are not using the "acl" PyPi package, we are actually depending on a Huawei specific Ascend Communications Library.

pyproject.toml
tengqm: In that case, we should refactor the source to push the dependency down to HuaweiAscend backend.

pyproject.toml
tengqm: Okay. Makes sense.

pyproject.toml
tengqm: Double checked the source code. `vllm` is referenced in the `src/flag_gems/patches` module for the purpose of optimizing vLLM performance (as mentioned in the `docs/how_to_use_flaggems.md`). The `flags_gems.apply_gems_patches_to_vllm` API is an "additional" API, not an "optional" API.

So there are two possibilities here:

- The `apply_gems_patches_to_vllm` API is an optional API, we can choose to built it in the deliverable or not. When we do NOT always ship this API, the dependency on `vllm` becomes *optional*.
- The `apply_gems_patches_to_vllm` API is an additional API, always shipped with the deliverable. Users can opt in/out this feature. The dependency is there. When users install `flag_gems`, the installer automatically installs `vllm` as a dependency, no matter the users use it or not.

pyproject.toml
tengqm: Moved `torch_npu` to optional dependency. By semantics, it IS an optional dependency.
I'm filing an issue to kick this import to `ascend` backend.

tengqm:
@Galaxy1458 file revised, please help review again, thanks.

tengqm:
@zhangpeiyang1 Can you help review?

zhangpeiyang1 (dismissed):

pyproject.toml
kiddyjinjin: I don’t think `vllm` should be a dependency of `flag_gems`. 
The vLLM patch only makes sense when there is already a vLLM service running, and anything that touches vLLM is not on the essential path of our main API (`flag_gems.enable()`). 
Also, `flag_gems` is not inference-only — it’s used in training setups as well (e.g. `Megatron`). In those environments, installing `vLLM` is unnecessary and unreasonable. 
So even if we always ship the vLLM patch API, the dependency itself should stay optional and opt-in, not something every `flag_gems` user has to install.

docs/installation.md
```diff
@@ -126,7 +129,7 @@ Example recipes are provided for easy copy-and-paste.
 Editable installation with external TritonJIT
 
 ```shell
-pip install -U scikit-build-core ninja cmake pybind11
+pip install .
```
kiddyjinjin: I’m not sure this can work properly. 
`pip install . `installs both the dependencies and `flag_gems` itself at the same time,
but in this scenario we actually need a specific build of `flag_gems` that can run with `libtriton_jit_runtime`, controlled by those CMake flags. 
Installing a default build first may lead to unexpected behavior, and it’s not obvious that the later editable install will fully replace it in all cases. 
I think we should test this flow carefully before changing the install instructions.

tengqm: Thanks for the review.
Are you referring to `-U scikit-build-core ninja cmake pybind11` part or the revised `pyproject.toml` file as a whole?
The `scikit-build-core ninja cmake pybind11` part is now declared in the `[build-system]` section in the `pyproject.toml` file, which means that these tools are prerequisites for the package to be installed, as shown below:

```toml
[build-system]
requires = ["setuptools>=64.0", "scikit-build-core>=0.11", "pybind11>=3.0.1", "uv>=0.9.26"]
build-backend = "scikit_build_core.build"
```

As for `cmake`, I don't think it is a PyPi package to be installed. I checked with you before regarding this dependency. It smells to me that we need a `apt-get install cmake` before installing FlagGems (I could be wrong though). In that case, it is out of the scope of `pyproject.toml`, just like we are not supposed to do `pip install cmake`.

tengqm: So ... my guess is that we need to remove `pip install .` completely, or we do a `apt-get install ...` as a preparation.

kiddyjinjin: You are right, thanks for the clarification.  it seems we can remove the `pip install .` line.

Let me test this in a fresh environment to be sure.

pyproject.toml
tengqm: Okay. I'm gonna move it into `project.optional-dependencies` section, with a group named `inference`. Users will need to do an extra `uv pip install .[inference]` step to get this installed.

PR #1378, opened 2026-01-06
[METAX] add metax op log

Alvin-YCHEN:
### Progress


Alvin-YCHEN:
> @Alvin-YCHEN Please check if your email setting in `git config user.email` is the same one you used in your Github account. Also please sign the CLA using the correct email account in order for your PR to be reviewed.

Thanks for your review. It's ok now.

tengqm:
@Alvin-YCHEN Thanks for your response. It smells like a requirement for the logging mechanism. The fix works but it is a little bit dirty. Can you help file an issue on this? My concern is that other backends may have similar requirements, and we don't want you all to solve it again and again.

PR #1392, opened 2026-01-07
【LibCache】 fix warnings on "AutoBase.prepare()" again

kiddyjinjin:
### PR Category

Other

### Type of Change

Bug Fix

### Description

This PR is used to resolve the warning that occurs when multiple workers attempt to create tables simultaneously. 
<img width="1874" height="1124" alt="img_v3_02tj_4c557b0f-e91d-4ac5-808a-864ef0f5403g" 
img_v3_02ti_aa2ac904-0f1a-4c5c-8f0d-885871f9d1dg

### Issue
#1323 

### Progress

### DIFF
--- a/src/flag_gems/utils/models/sql.py
+++ b/src/flag_gems/utils/models/sql.py
@@ -62,7 +62,7 @@ class SQLPersistantModel(PersistantModel):
         engine: sqlalchemy.engine.Engine,
     ) -> Optional[Type[Base]]:
         AutoBase: sqlalchemy.ext.automap.AutomapBase = (
-            sqlalchemy.ext.automap.automap_base(Base)
+            sqlalchemy.ext.automap.automap_base()
         )
         AutoBase.prepare(engine)
         ModelCls: Optional[Type[Base]] = AutoBase.classes.get(name)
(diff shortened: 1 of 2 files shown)


src/flag_gems/utils/models/sql.py
```diff
@@ -90,25 +90,26 @@ def get_sql_model(
...
-            if not keys or not values:
-                return None
-            ModelCls = SQLPersistantModel.build_sql_model_by_py(name, keys, values)
-            with self.engine.begin() as conn:
-                conn.execute(
-                    sqlalchemy.schema.CreateTable(
-                        ModelCls.__table__, if_not_exists=True
-                    )
-                )
+        ModelCls: Optional[Type[Base]] = self.sql_model_pool.get(name)
```
tengqm, line 93: my previous experience with sqlalchemy wasn't good. One of the reason is that when A writes something, I can never ensure that B sees that new value.
Let's hope this revision solves the problem. If the problem persists, we may have to revisit the design, replace sqlalchemy with something else, maybe.

src/flag_gems/utils/models/sql.py
```diff
@@ -90,25 +90,26 @@ def get_sql_model(
...
+        )
+        if ModelCls is not None:
             self.sql_model_pool[name] = ModelCls
             return ModelCls
+        if not keys or not values:
+            return None
+        ModelCls: Type[Base] = SQLPersistantModel.build_sql_model_by_py(
+            name, keys, values
+        )
+        with self.engine.begin() as conn:
```
tengqm, line 107: Emm ... this sounds promising...

src/flag_gems/utils/models/sql.py
kiddyjinjin: I understand the concern — I’m not a  fan of SQLAlchemy either.
In our current libcache design, we don’t strictly require read-after-write consistency across devices. This is more of an eventual consistency use case, and short delays in visibility are acceptable.
I’ve done some basic validation in a 4-device setup with a real model, and the results look reasonable so far. If issues show up, I agree we should revisit the design and possibly replace SQLAlchemy.

PR #1400, opened 2026-01-08
【refactor】extract device info 

kiddyjinjin:
### PR Category

Other

### Type of Change

Refactor

### Description

- Extracted device-related helper functions from `mm.py` into a new module
`utils/device_info.py`.  All functions are cached via `lru_cache` to avoid repeated device queries.

### Progress

### DIFF
--- a/src/flag_gems/utils/device_info.py
+++ b/src/flag_gems/utils/device_info.py
@@ -51,9 +51,11 @@ def get_device_info() -> DeviceInfo:
     l2_cache_size = None
     sm_count = None
     if props is not None:
-        l2_cache_size = getattr(props, "L2_cache_size", None) or getattr(
-            props, "l2_cache_size", None
-        )
+        l2_cache_size = None
+        if hasattr(props, "L2_cache_size"):
+            l2_cache_size = props.L2_cache_size
+        elif hasattr(props, "l2_cache_size"):
+            l2_cache_size = props.l2_cache_size
         sm_count = getattr(props, "multi_processor_count", None) or getattr(
             props, "multiProcessorCount", None
         )


src/flag_gems/utils/device_info.py
```diff
@@ -0,0 +1,66 @@
...
+    try:
+        if torch.cuda.is_available():
+            return torch.cuda.get_device_capability(device_id)
+    except Exception:
+        pass
+    return (0, 0)
+
+
+@lru_cache(maxsize=1)
+def get_device_info() -> tuple[int, int, int]:
```
tengqm: I'd suggest we return a key-value struct rather than a tuple.
Next time when we need/want to add a new field, we don't need to change the function prototype.

kiddyjinjin: I’ve updated the return type to a `@dataclass` instead of a raw tuple. This improves readability via named fields, keeps backward compatibility reasonably well, and avoids the weaker type guarantees of a plain key-value/dict-style return.

huangyiqun (dismissed):

src/flag_gems/utils/device_info.py
```diff
@@ -0,0 +1,78 @@
...
+        return torch_device_fn.current_device()
+    except Exception:
+        return 0
+
+
+@lru_cache(maxsize=1)
+def get_device_properties():
+    device_id = get_device_id()
+    try:
+        return torch_device_fn.get_device_properties(device_id)
```
tengqm, line 28: Is this `get_device_properties` function a new interface for backends to implement?

src/flag_gems/utils/device_info.py
```diff
@@ -0,0 +1,78 @@
...
+
+@lru_cache(maxsize=1)
+def get_device_info() -> DeviceInfo:
+    props = get_device_properties()
+    l2_cache_size = None
+    sm_count = None
+    if props is not None:
+        l2_cache_size = getattr(props, "L2_cache_size", None) or getattr(
+            props, "l2_cache_size", None
+        )
```
tengqm: I'd write simple and straightforward code like


```suggestion
        l2_cache_size = None
        if hasattr(props, "L2_cache_size"):
            l2_cache_size = props.L2_cache_size
        elif hasattr(props, "l2_cache_size"):
            l2_cache_size = props.l2_cache_size
```

src/flag_gems/utils/device_info.py
```diff
@@ -0,0 +1,78 @@
...
+        sm_count = getattr(props, "multi_processor_count", None) or getattr(
+            props, "multiProcessorCount", None
+        )
+    if l2_cache_size is None:
+        # default L2 cache size to 40MB for A100
+        l2_cache_size = 40 * 1024 * 1024
+    if sm_count is None:
+        # default sm_count to 108 for A100
+        sm_count = 108
```
tengqm, line 67: This smells pretty dangerous to me...

src/flag_gems/utils/device_info.py
```diff
@@ -0,0 +1,78 @@
...
+    return DeviceInfo(
+        device_id=get_device_id(),
+        l2_cache_size=l2_cache_size,
+        sm_count=sm_count,
+    )
```
tengqm, line 72: Why not just a simple dict like this? I'm not quite familiar with dataclasses.


```suggestion
    return {
        "device_id": get_device_id(),
        "l2_cache_size": l2_cache_size,
        "sm_count": sm_count,
    }
```

src/flag_gems/utils/device_info.py
kiddyjinjin: No, it’s a very fundamental interface that every backend/vendor should already have implemented.

src/flag_gems/utils/device_info.py
kiddyjinjin: Mostly personal preference. I prefer using a dataclass here because it gives us a fixed, explicit structure (typed fields), and it helps avoid typos/mistakes in string keys compared to a plain dict.

kiddyjinjin: maybe a habit from my C++ coding style.

src/flag_gems/utils/device_info.py
kiddyjinjin: well, in a properly working backend/device, these fields should always be available, so we should never hit the default values.
I added this fallback mainly for robustness — in case `props` is incomplete for some reason, we can still return a usable DeviceInfo instead of crashing downstream.

PR #1418, opened 2026-01-12
[BugFix] fix bugs when model primary keys are not aligned

sgjzfzzf:
### PR Category

[SQL]

### Type of Change

Bug Fix

### Description

For `LayerNorm` operator, the kernel signature would change based on the parameter types, which could lead to a SQL unmatched error on the table definition. This PR adds the type of request as a suffix to the model name to differentiate these and fix the known bugs.

### Progress

- [ ] Change is properly reviewed (1 reviewer required, 2 recommended).
- [ ] Change is responded to an issue.
- [ ] Change is fully covered by a UT.

PR #1427, opened 2026-01-13
[BugFix] fix bug on test_libentry

sgjzfzzf:
### PR Category

OP Test

### Type of Change

Bug Fix

### Description

This PR resolves the bug introduced by #1418 which will lead to the failure on `test_libentry`. This bug occur because the previous PR renames the table to avoid definition conflicts on the SQL model based on different kernel signatures, but in the previous implementation of the unit test, it still tries to access this table directly.

This PR disables `test_libcache_concurrent_write_on_signal` temporarily and plan to enable it again when `get_configs` is ready.

### Progress

### DIFF
--- a/tests/test_libentry.py
+++ b/tests/test_libentry.py
@@ -10,6 +10,7 @@ from triton import language as tl
 import flag_gems
 from flag_gems.runtime import torch_device_fn
 from flag_gems.utils import libentry, libtuner
+from flag_gems.utils.libentry import libcache
 
 
 # not_raises is copied from https://gist.github.com/oisinmulvihill/45c14271fad7794a4a52516ecb784e69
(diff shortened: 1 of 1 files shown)


tests/test_libentry.py
```diff
@@ -397,18 +396,15 @@ def child_process():
...
-            cursor.execute("SELECT COUNT(*) FROM test_vllm_operator")
-            count = cursor.fetchone()[0]
-            if count > 0:
-                cache_saved = True
-            cursor.execute("DELETE FROM test_vllm_operator")
-            conn.commit()
-            conn.close()
-        except sqlite3.OperationalError:
-            pass
+        from flag_gems.utils.libentry import libcache
```
tengqm: please avoid import in the middle of a file, i.e. move this import to the head of the file if possible.

PR #1360, opened 2026-01-05
Update flag_gems only_enable api and unit tests

littlekid-zxt:
### PR Category

User Experience | OP Test

### Type of Change

New Feature

### Description

Add the `flag_gems.only_enable(...)` function, allowing it to specify supported operators to be replaced with the optimized FlagGems implementation, and also add unit tests for the `flag_gems.enable(...)` and `flag_gems.only_enable(...)`.

#### Changes:

- Encapsulate the op config in `src/flag_gems/__init__.py` into the `_get_full_config()` api for easier access and modification;
- Rename `unused` to `exclude`;
- Add the `only_enable(...)` api in `src/flag_gems/__init__.py` to support specifying include ops for replacing the FlagGems implementation;
- Improve the implementation of the `only_enable(...)` api and `include` param in the `__init__`, `__enter__`, and `__exit__` functions of the `class use_gems`;
- Based on the `class use_gems`, add unit test cases for three scenarios: `flag_gems.use_gems(... | exclude=exclude_op,... | include=include_op,...)`.

#### Usage:

```
   exclude_op = ["mul", "add"]

   flag_gems.enable(
      exclude=exclude_op,
      record=True,
      path=log_file_name
   )
   # or
   with flag_gems.use_gems(exclude=exclude_op, record=True, path=log_file_name):
      ...
```
```
   include_op = ["sum", "mul", "add"]

   flag_gems.only_enable(
      include=include_op,
      record=True,
      path=log_file_name
   )
   # or
   with flag_gems.use_gems(include=include_op, record=True, path=log_file_name):
      ...
```
```
   cd FlagGems/tests/
   pytest test_enable_api.py
```

### Issue

N/A

### Progress

- [ ] Change is properly reviewed (1 reviewer required, 2 recommended).
- [ ] Change is responded to an issue.
- [ ] Change is fully covered by a UT.

### Performance

N/A

PR #1438, opened 2026-01-14
【KMPL】 Added the FlagGems only_enable() API and unit tests

littlekid-zxt:
### PR Category

User Experience | OP Test

### Type of Change

New Feature

### Description

#### Changes:

- Encapsulate the op config in `src/flag_gems/__init__.py` into the `_FULL_CONFIG` for easier access and modification;
- Add the `only_enable(...)` api in `src/flag_gems/__init__.py` to support specifying include ops for replacing the FlagGems implementation;
- Improve the implementation of the `only_enable(...)` api and `include` param in the `__init__`, `__enter__`, and `__exit__` functions of the `class use_gems`;
- Based on the `class use_gems`, add unit test cases for three scenarios: `flag_gems.use_gems(... | unused=unused_op,... | include=include_op,...)`.

#### Usage:

```
   unused_op = ["mul", "add"]

   flag_gems.enable(
      unused=unused_op,
      record=True,
      path=log_file_name
   )
   # or
   with flag_gems.use_gems(unused=unused_op, record=True, path=log_file_name):
      ...
```
```
   include_op = ["sum", "mul", "add"]

   flag_gems.only_enable(
      include=include_op,
      record=True,
      path=log_file_name
   )
   # or
   with flag_gems.use_gems(include=include_op, record=True, path=log_file_name):
      ...
```
```
   cd FlagGems/tests/
   pytest test_enable_api.py
```

### Issue
#1267 
#1296 
#1360 

N/A

### Progress

### Performance

N/A

### DIFF
--- a/tests/test_enable_api.py
+++ b/tests/test_enable_api.py
@@ -29,6 +29,9 @@ def save_log_file(ori_log, sv_log):
         with open(sv_log, "wb") as f_out:
             f_out.write(content)
 
+        with open(ori_log, "wb") as f_ori:
+            f_ori.truncate(0)
+
 
 def cleanup_log_file():
     path_file = get_path_log_name()


src/flag_gems/__init__.py
```diff
@@ -37,359 +359,94 @@ def enable(
...
+        user_unused_ops_list=[],
+        cpp_patched_ops_list=list(set(aten_patch_list)),
+        lib=lib,
+    )
+    setup_flaggems_logging(path=path, record=record, once=once)
+
 
 class use_gems:
-    def __init__(self, unused=None, record=False, once=False, path=None):
+    def __init__(self, unused=None, include=None, record=False, once=False, path=None):
```
kiddyjinjin: change `unused` to `exclude` here, and add a comment explaining that `include` takes precedence over `exclude`.

tests/test_enable_api.py
```diff
@@ -0,0 +1,120 @@
...
+   with flag_gems.use_gems(record=True, path=path_file):
+      a = torch.tensor([1.0, 2.0, 3.0], device=flag_gems.device)
+      b = torch.tensor([4.0, 5.0, 6.0], device=flag_gems.device)
+      _ = a + b
+      _ = a * b
+      _ = flag_gems.sum(a)
+      _ = torch.sum(a)
+
+   log_file = "./gems_enable_all_ops.log"
+   save_log_file(path_file, log_file)
```
kiddyjinjin: Why are we writing the log file twice — once to `path_file` and once to `log_file`?

littlekid-zxt: I explained this issue in the `get_path_log_name()` api of `test_enable_api.py`.
The main reason is that when repeatedly using `flag_gems.use_gems(record=True, path=path_file)`, the logging system initializes a logger named `flag_gems`. However, during `use_gems.__exit__`, `logging.root.handlers` is cleared, which causes path_file to not update `logging.FileHandler`. Therefore, the test cases currently use this method to implement verification.

kiddyjinjin: Got it, thanks. This seems like a bug/behavior in` use_gems()` (handler cleanup in __exit__) that we should fix later. Your current test workaround makes sense for now.

tengqm: In that case, please file an issue in case the ball gets dropped.
We are not supposed to manipulate the logging mechanism during `__enter__` or `__exit__`. There must be an elegant solution.

src/flag_gems/__init__.py
```diff
@@ -37,359 +359,98 @@ def enable(
...
+            )
+        else:
+            enable(
+                lib=self.lib,
+                unused=self.exclude,
+                registrar=self.registrar,
+                record=self.record,
+                once=self.once,
+                path=self.path,
+            )
```
tengqm, line 451: Here are my two cents to simplify/unify the logic.

I'd suggest we combine `only_enable` and `enable` to a single helper function `prefilter(include, exclude)`. In that function we centralize the processing of `include` and `exclude` lists. With proper function docstrings, we can ensure that the overall logic is clear and maintainable.

tests/test_enable_api.py
tengqm: I believe we don't want to make the logging infrastructure more and more complex/fragile. @littlekid-zxt Can you please help spend some time on summarizing our requirements on logging? For example:

- We cannot use the builtin logging from Triton, or Torch, because ...
- We want logs to be dumped into a different file because ...
- Our runtime backends are not using `logger = logging.getLogger(__name__)` because they have some unique requirements. The requirements are ...
- We know that logging can have notable impact on gems benchmarks, but we also want to know how frequent each operator is invoked. We know that `logger.debug` has performance impacts but we are not aware of any alternatives ...
- We have different requirements for logs from test cases and logs from production code, such as ...
- We are opening log files in `"w+"` mode rather than `"a+"` mode because ...
- ...

littlekid-zxt: Okay, no problem. I will submit an issue later and provide a detailed explanation based on the issue's symptoms and the relevant code.

PR #1440, opened 2026-01-14
[KUNLUNXIN] fix baddbmm kernel

dongjibin1996:
### Progress


huangyiqun:
Could clarify what issue this fixes? Is this addressing an operator error on KUNLUNXIN?

src/flag_gems/ops/baddbmm.py
```diff
@@ -235,4 +235,10 @@ def compute_B_grad(A, d_output, alpha):
...
 
 def baddbmm(bias, A, B, beta=1.0, alpha=1.0):
-    return BaddbmmFunction.apply(bias, A.contiguous(), B.contiguous(), beta, alpha)
+    return BaddbmmFunction.apply(
+        bias.contiguous(),
+        A.contiguous(),
+        B.contiguous(),
+        beta,
+        alpha,
+    )
```
tengqm, line 244: Can we add a simple test case for this change?

dongjibin1996: Thanks for the suggestion — that makes sense. I’ll add a small regression test in a follow-up PR.

- The test will explicitly construct bias as a broadcasted view (via broadcast_to/expand ) so it carries zero strides on the expanded dimensions, since that is the layout that exposed the KUNLUNXIN issue.
- It will use a minimal dot case (dot/accumulation path on XPU) and compare against the PyTorch reference to validate correctness.
- This should be clearer and more targeted than a generic accuracy test, which may not reliably hit the broadcasted/zero‑stride bias path.
I’ll keep it concise and easy to understand.

PR #1441, opened 2026-01-14
【KernelGen】update experimental test workflow

Dongxu-H:
### PR Category
CI workflow

### Type of Change
workflow

### Description
1. fix experimental test workflow issue in master (non PR mode), base SHA is none.
2. fix experimental test list containing space
3. add more tests file check, including accuracy test file and speed up test file.

### Progress

- [ √] Change is fully covered by a UT.

### DIFF
--- a/.github/workflows/gems-experimental-test.yaml
+++ b/.github/workflows/gems-experimental-test.yaml
@@ -31,90 +31,11 @@ jobs:
 
       - name: FlagGems experimental tests on hopper
         shell: bash
+        env:
+          BASE_SHA: ${{ github.event_name == 'pull_request' && github.event.pull_request.base.sha || github.event.before }}
+          HEAD_SHA: ${{ github.sha }}
         run: |
           source "/home/zhangzhihui/miniconda3/etc/profile.d/conda.sh"
           conda activate flag_gems
           source tools/run_command.sh
-          HEAD_SHA=${{ github.sha }}
-          if [[ "${{ github.event_name }}" == "pull_request" ]]; then
-            BASE_SHA=${{ github.event.pull_request.base.sha }}
-          else
-            BASE_SHA=${{ github.event.before }}
-          fi
-          echo "Diffing $BASE_SHA...$HEAD_SHA"
-
-          changed_files=$(git diff --name-only --diff-filter=AM $BASE_SHA...$HEAD_SHA || true)
-          if [[ -z "$changed_files" ]]; then
-            echo "No changed files detected, skipping."
-            exit 0
-          fi
-
-          functional_tests_to_run=""
-          performance_tests_to_run=""
-          functional_missing_tests=""
-          performance_missing_tests=""
-
-          for f in $changed_files; do
-              if [[ $f == src/flag_gems/experimental_ops/*.py ]]; then
-                  if [[ $(basename "$f") == __*__* ]]; then continue; fi
-
-                  base=$(basename "$f" .py)
-                  functional_test_file="experimental_tests/functional/${base}_test.py"
-                  if [[ -f "$functional_test_file" ]]; then
-                      functional_tests_to_run="$functional_tests_to_run $functional_test_file"
-                  else
-                      functional_missing_tests="$functional_missing_tests $functional_test_file"
-                  fi
-
-                  performance_test_file="experimental_tests/performance/${base}_test.py"
-                  if [[ -f "$performance_test_file" ]]; then
-                      performance_tests_to_run="$performance_tests_to_run $performance_test_file"
-                  else
-                      performance_missing_tests="$performance_missing_tests $performance_test_file"
-                  fi
-
-              elif [[ $f == experimental_tests/functional/*_test.py ]]; then
-                  if [[ -f "$f" ]]; then
-                      functional_tests_to_run="$functional_tests_to_run $f"
-                  fi
-              elif [[ $f == experimental_tests/performance/*_test.py ]]; then
-                  if [[ -f "$f" ]]; then
-                      performance_tests_to_run="$performance_tests_to_run $f"
-                  fi
-              fi
-
-          done
-
-          if [[ -n "$functional_missing_tests" ]]; then
-              echo "::error:: The following operator files were modified, but their corresponding functional test files are missing:"
-              for m in $functional_missing_tests; do
-                  echo "  - Expected: $m"
-              done
-              exit 1
-          fi
-
-          if [[ -n "$performance_missing_tests" ]]; then
-              echo "::error:: The following operator files were modified, but their corresponding performance test files are missing:"
-              for m in $performance_missing_tests; do
-                  echo "  - Expected: $m"
-              done
-              exit 1
-          fi
-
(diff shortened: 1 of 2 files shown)


.github/workflows/gems-experimental-test.yaml
```diff
@@ -50,30 +53,36 @@ jobs:
...
+          if [[ -n "$missing_tests" ]]; then
               echo "Error: The following operator files were modified, but their corresponding test files are missing:"
               for m in $missing_tests; do
                   echo "  - Expected: $m"
               done
               exit 1
           fi
 
-          unique_tests=$(echo "$tests_to_run" | tr ' ' '\n' | sort -u | tr '\n' ' ')
+          unique_tests=$(echo "$tests_to_run" | tr ' ' '\n' | sort -u | xargs)
```
tengqm: Not sure if you meant this:

```suggestion
          unique_tests=$(echo "$tests_to_run" | tr ' ' '\n' | uniq)
```

.github/workflows/gems-experimental-test.yaml
```diff
@@ -50,30 +53,36 @@ jobs:
...
 
-          if [ -n "$unique_tests" ]; then
+          if [[ -n "$unique_tests" ]]; then
```
tengqm: This change is not necessary?

.github/workflows/gems-experimental-test.yaml
```diff
@@ -50,30 +53,36 @@ jobs:
...
+                      tests_to_run="$tests_to_run $accuracy_test_file"
+                  elif [[ -f "$speedup_test_file" ]]; then
+                      tests_to_run="$tests_to_run $speedup_test_file"
                   else
                       missing_tests="$missing_tests $test_file"
                   fi
 
               elif [[ $f == experimental_tests/*_test.py ]]; then
-                  if [ -f "$f" ]; then
+                  if [[ -f "$f" ]]; then
```
tengqm: This change is not necessary?

.github/workflows/gems-experimental-test.yaml
```diff
@@ -50,30 +53,36 @@ jobs:
...
                       tests_to_run="$tests_to_run $f"
                   fi
               fi
           done
 
-          if [ -n "$missing_tests" ]; then
+          if [[ -n "$missing_tests" ]]; then
```
tengqm: `[[]]` is a bash dialect, maybe stick to `[]` is a better choice?

.github/workflows/gems-experimental-test.yaml
```diff
@@ -50,30 +53,36 @@ jobs:
 
                   base=$(basename "$f" .py)
                   test_file="experimental_tests/${base}_test.py"
-                  if [ -f "$test_file" ]; then
+                  accuracy_test_file="experimental_tests/${base}_accuracy_test.py"
+                  speedup_test_file="experimental_tests/${base}_speedup_test.py"
```
tengqm: We may want to separate test cases into different subdirectories rather than enforcing file naming rules.

.github/workflows/gems-experimental-test.yaml
Dongxu-H: no, I want to sort all tests file in order (sort) and remove space (xargs)
uniq can't remove spaces and this would cause testing error

.github/workflows/gems-experimental-test.yaml
Dongxu-H: GitHub Actions uses bash. using [[ ... ]] is much better and much safer

.github/workflows/gems-experimental-test.yaml
Dongxu-H: GitHub Actions uses bash. using [[ ... ]] is much better and much safer

.github/workflows/gems-experimental-test.yaml
Dongxu-H: GitHub Actions uses bash. using [[ ... ]] is much better and much safer

.github/workflows/gems-experimental-test.yaml
Dongxu-H: it depends on your following design, we could modify it when your design ready.

tengqm: It is much more simpler to group the admission unit tests (focusing on correctness/accuracy) into `experimental_tests/unit` and move performance tests to `experimental_tests/perf`. I don't see a file following this naming pattern. This means the naming pattern is a newly introduced convention. Please stop here before going further down this road.

Dongxu-H: fixed

.github/workflows/gems-experimental-test.yaml
```diff
@@ -35,56 +35,86 @@ jobs:
...
 
-          if [[ -n "$unique_tests" ]]; then
-              echo "Running tests: $unique_tests"
-              run_command pytest -s $unique_tests
+          if [[ -n "$performance_unique_tests" ]]; then
+              echo "Running tests: $performance_unique_tests"
+              run_command pytest -s $performance_unique_tests
           else
               echo "No relevant changes in ops or tests, skipping."
           fi
```
tengqm: Okay, this logic is getting complicated as I originally imagined ...
Do you want to extract the bash script into something like `/tools/experimental-op-test.sh`?

Dongxu-H: fixed

tools/experimental-op-test.sh
```diff
@@ -0,0 +1,72 @@
...
+
+# 3. Categorize Tests
+functional_tests_to_run=""
+performance_tests_to_run=""
+functional_missing_tests=""
+performance_missing_tests=""
+
+for f in $changed_files; do
+    # Logic for files in experimental_ops
+    if [[ $f == src/flag_gems/experimental_ops/*.py ]]; then
```
tengqm: consider use `source_dir="src/flag_gems/experimental_ops"`,
and `test_dir="experimental_tests"`,
you won't have to scan the script for directory name to fix later.

Note that you don't have to define vars for unit tests or performance tests.

tools/experimental-op-test.sh
```diff
@@ -0,0 +1,72 @@
...
+        if [[ $(basename "$f") == __*__* ]]; then continue; fi
+
+        base=$(basename "$f" .py)
+
+        # Check Functional
```
tengqm: Please note that unit tests are different from functional tests.
Unit tests are mainly for testing the control flow of a module, i.e. the module behaves as designed given some mocked inputs.
Functional tests are mainly for a group of modules (i.e. a subsystem). For example, given a specific problem, the subsystem can handle it properly, including valid/invalid inputs.

Besides these, there are integration tests (SVT), smoke tests ...
So I don't think we want to use the word "functional" here.

tools/experimental-op-test.sh
Dongxu-H: fixed

tools/experimental-op-test.sh
Dongxu-H: fixed

tengqm (approved):

PR #1444, opened 2026-01-15
Test: add accuracy_utils for cpp wrapper

wlxjhyf:
### PR Category

Test

### Type of Change

New Feature

### Description

Add accuracy_utils to implement unified precision testing control in ctest aligned with Python code

### Progress

### DIFF
--- a/include/flag_gems/accuracy_utils.h
+++ b/include/flag_gems/accuracy_utils.h
@@ -1,0 +1,28 @@
+#pragma once
+
+#include <c10/core/ScalarType.h>
+#include <torch/torch.h>
+
+namespace flag_gems::accuracy_utils {
+
+extern bool TO_CPU;
+
+struct CheckCloseResult {
+  bool ok;
+  std::string message;
+};
+
+torch::Tensor to_reference(torch::Tensor inp, bool upcast = false);
+
+torch::Tensor to_cpu(torch::Tensor res, const torch::Tensor& ref);
+
+CheckCloseResult gems_assert_close(torch::Tensor res,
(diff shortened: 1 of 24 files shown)


include/flag_gems/accuracy_utils.h
```diff
@@ -0,0 +1,23 @@
...
+
+void gems_assert_close(torch::Tensor res,
+                       torch::Tensor ref,
+                       c10::ScalarType dtype = c10::ScalarType::Undefined,
+                       bool equal_nan = false,
+                       int64_t reduce_dim = 1,
+                       double atol = 1e-4);
+
+void gems_assert_equal(torch::Tensor res, torch::Tensor ref, bool equal_nan = false);
+
```
tengqm, line 27: Wondering if the above two functions should return `bool` as `torch.allclose()` does.
Or else how would we know if the "assertion" was a success or not?

wlxjhyf: > Wondering if the above two functions should return `bool` as `torch.allclose()` does. Or else how would we know if the "assertion" was a success or not?

These functions are assertion helpers, not ordinary boolean checks. They do not return a value because their success is implicit. If the comparison fails, an exception is raised, which the test framework (here is gtest in cpp wrapper) interprets as a test failure. Unlike torch.allclose(), which returns True/False, assertion helpers rely on exception-based control flow rather than boolean results. So you “know” the assertion succeeded if no exception is raised.

tengqm: Thanks for the confirmation. That answered my question.

PR #1451, opened 2026-01-16
cambricon: fix the sum and triu operators

tututuski:
### PR Category
Operator

### Type of Change
Bug Fix

#### triu op
There is a problem with obtaining the number of bytes for element types in the triu operator; it can only retrieve the number of bytes for floating-point types.

#### sum op

Under certain input shapes, the original grid exceeds hardware limitations.  Therefore, the original grid is broken down within the kernel, and loops are used to replace the original program.

### Issue
Null

### Progress

### Performance
Null


tututuski:
I'll fork the repository later and resubmit the PR.

PR #1455, opened 2026-01-16
To decouple from torch_npu.

zhangpeiyang1:
### Progress


src/flag_gems/runtime/backend/device.py
Galaxy1458, line 88: This is a good robustness repair; Meanwhile, it is recommended to retain the logic of `import torch_npu` to provide users with a better experience when importing `flag_gems`

PR #1456, opened 2026-01-16
Refactor cumsum.py with NPU properties and optimizations

fengrui886:
### Progress

### DIFF
--- a/src/flag_gems/runtime/backend/_ascend/ops/cumsum.py
+++ b/src/flag_gems/runtime/backend/_ascend/ops/cumsum.py
@@ -4,12 +4,18 @@ import math
 import torch
 import triton
 import triton.language as tl
+import triton.runtime.driver as driver
 
 from flag_gems.runtime import device, torch_device_fn
 from flag_gems.utils import libentry
 from flag_gems.utils import triton_lang_extension as tle
 
 logger = logging.getLogger(f'flag_gems.runtime._ascend.ops.{__name__.split(".")[-1]}')
+
+
+def get_npu_properties():
+    device = torch.npu.current_device()
+    return driver.active.utils.get_device_properties(device)
 
 
 device = device.name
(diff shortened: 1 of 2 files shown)


src/flag_gems/runtime/backend/device.py
```diff
@@ -83,14 +83,6 @@ def _get_vendor_from_quick_cmd(self):
...
             if hasattr(torch, flag):
                 return vendor_name
-        try:
-            import torch_npu
-
-            for vendor_name, flag in cmd.items():
-                if hasattr(torch_npu, flag):
-                    return vendor_name
-        except:  # noqa: E722
-            pass
```
tengqm, line 96: We are not ready to drop these lines IN THIS PR, right?

src/flag_gems/runtime/backend/_ascend/ops/cumsum.py
```diff
@@ -382,9 +388,7 @@ def normed_cumsum(inp, dim=-1):
     out = torch.empty_like(inp)
     with torch_device_fn.device(inp.device.index):
         # Pass one, scan a (batch, n_tiles * TILE) sized block within each cta
-        import acl
-
-        num_sms = acl.rt.get_device_info(acl.rt.get_device()[0], 201)[0]
+        num_sms = get_npu_properties()["num_vectorcore"]
```
tengqm, line 391: This is NICE!!!

src/flag_gems/runtime/backend/device.py
zhangpeiyang1: nonono。 since prhttps://github.com/flagos-ai/FlagGems/pull/1455/files  merged。 so we can。

Galaxy1458: It is recommended to retain the logic of import `torch_npu`to provide users with a better experience when importing `flag_gems`.

zhangpeiyang1: That shouldn't be in this public module.

Galaxy1458: This is a faster way to detect the backend. Since `import triton` and `import torch` have already consumed too much time when `import flag_gems` the backend detection should be as fast as possible.

src/flag_gems/runtime/backend/_ascend/ops/cumsum.py
tengqm, line 391: Are we 100% sure that the key `num_vectorcore` can always be found in the dict returned?

zhangpeiyang1 (approved):

PR #1052, opened 2025-11-12
[Advanced Compiler]Add TE.scaled_softmax forward&backward

AdvancedCompiler:
### PR Category

Operator
### Type of Change

New Feature
### Description

Add transformer_engine_torch.scaled_softmax_forward&transformer_engine_torch.scaled_softmax_backward
### Progress

- [ ] Change is properly reviewed (1 reviewer required, 2 recommended).
- [ ] Change is responded to an issue.
- [ ] Change is fully covered by a UT.

### Performance

7c8c27d18ed7ef1e81ee70d107d38f81

PR #1464, opened 2026-01-19
[Iluvatar] fix test_accuracy_moe_align_block_size when using cpu as r…

awayzjj:
1. Fix `test_accuracy_moe_align_block_size` failing when using `--ref cpu`.

2. Remove `_load_library` call in TE and directly `import transformer_engine_torch as tex`; this avoids issues for backends that do not implement `_load_library` which seems unnecessary.

3. Set random seed for `test_accuracy_normal` and `test_accuracy_randn` like other backends.

4. Add `kwargs` support for `flash_attn_varlen_legacy` to resolve `unexpected keyword` errors.

### Progress

### DIFF
--- a/benchmark/test_reduction_perf.py
+++ b/benchmark/test_reduction_perf.py
@@ -516,10 +516,7 @@ class ScaledSoftmaxBenchmark(GenericBenchmark):
 @pytest.mark.scaled_softmax
 def test_perf_scaled_softmax_forward():
     try:
-        from transformer_engine.common import _load_library
-
-        _load_library()
-        import transformer_engine_torch as tex  # type: ignore
+        from transformer_engine.pytorch import cpp_extensions as tex
     except ImportError:
         pytest.skip("TransformerEngine is not available, skipping performance test")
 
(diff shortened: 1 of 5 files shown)


benchmark/test_reduction_perf.py
```diff
@@ -516,10 +516,7 @@ def get_input_iter(self, cur_dtype) -> Generator:
 @pytest.mark.scaled_softmax
 def test_perf_scaled_softmax_forward():
     try:
-        from transformer_engine.common import _load_library
-
-        _load_library()
-        import transformer_engine_torch as tex  # type: ignore
+        from transformer_engine.pytorch import cpp_extensions as tex
```
huangyiqun, line 519: Why was this import changed? Are both versions equivalent in functionality?

tengqm: Same question here ...

But I'm leaning toward this substitution because this change (and the following one) kills an additional dependency from the FlagGems project, and THAT IS GOOD!

awayzjj: @huangyiqun 
1. Because we do not implement `_load_library` in our Transformer Engine (TE).
2. I believe these two approaches are equivalent. I already raised this question in the PR https://github.com/flagos-ai/FlagGems/pull/1052 discussion, but haven’t received a response yet:
Moreover, this was how it had been used previously (for example, https://github.com/flagos-ai/FlagGems/blob/master/tests/test_unary_pointwise_ops.py#L8). I’m not sure why it was changed to the current implementation.

huangyiqun: Thanks for the explanation. I’ll look into the question about the import method change at a later time.
The import patterns shown in the referenced PR also need to be addressed, we should aim to keep the import style consistent.

PR #1465, opened 2026-01-19
【KernelGen】Split experimental tests into unit and performance directories

factnn:
- Created experimental_tests/unit/ for correctness tests
- Created experimental_tests/performance/ for performance/benchmark tests
- Removed original test files from experimental_tests/ root directory
- Updated sys.path in all test files to use "../.." for correct imports
- Removed unused imports to pass flake8 checks
- All pre-commit checks now pass


tengqm:
Thanks for this huge effort.

tengqm (dismissed):

tengqm (approved):

PR #1408, opened 2026-01-09
【KernelGen】 delete rmsnorm from experimental_ops

factnn:
- rmsnorm Triton implementation underperforms PyTorch official API
- Performance test shows only 1/15 cases faster, 14/15 cases slower (0.28-0.38x speedup)
- PyTorch torch.nn.functional.rms_norm provides better performance
- Removed files: rmsnorm.py, rmsnorm_test.py
- Updated __init__.py to remove rmsnorm import
- Updated README.md to remove rmsnorm examples

PR #1472, opened 2026-01-21
Revert "【KernelGen】 delete rmsnorm from experimental_ops"

Dongxu-H:
Reverts flagos-ai/FlagGems#1408


src/flag_gems/experimental_ops/rmsnorm.py
```diff
@@ -0,0 +1,92 @@
+# Generated by KernelGen v1.0
```
tengqm, line 1: This file has been relocated ...

PR #1484, opened 2026-01-23
[DNM] Test publishing coverage data

tengqm:
### PR Category

CI/CD

### Type of Change

Testing

### Description

This PR is an attempt expose the test coverage data via artifacts.

### Issue

N/A

### Progress

### Performance

N/A

### DIFF
--- a/.github/workflows/op-test-quick-cpu.yaml
+++ b/.github/workflows/op-test-quick-cpu.yaml
@@ -1,20 +1,32 @@
 # This workflow will install Python dependencies, run tests and lint with a single version of Python
 # For more information see: https://docs.github.com/en/actions/automating-builds-and-tests/building-and-testing-python
 
+# - id: upload_pages
+#   uses: peaceiris/actions-gh-pages@v4
+#   with:
+#     github_token: ${{ secrets.GITHUB_TOKEN }}
+#     publish_dir: htmlcov
+#     destination_dir: 1484
+#     publish_branch: gh-pages
+#     user_name: 'github-actions[bot]'
+#     user_email: '41898282+github-actions[bot]@users.noreply.github.com'
 name: op-test-quick-cpu
 
 on:
   workflow_call:
+    secrets:
+      PAT_TOKEN:
+        required: true
 
 jobs:
   container-unit-test:
     runs-on: jiuding
     steps:
-      - name: Checkout code
-        uses: actions/checkout@v4
+      - id: checkout_code
+        uses: actions/checkout@v6
 
-      - uses: dorny/paths-filter@v3
-        id: experimental_ops
+      - id: filter
+        uses: dorny/paths-filter@v3
         with:
           filters: |
             run_ci:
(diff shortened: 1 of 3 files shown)


tools/op-test-quick-cpu.sh
```diff
@@ -18,5 +18,5 @@ run_command coverage run ${COVERAGE_ARGS} -m pytest -s tests/test_pointwise_type
 run_command coverage run ${COVERAGE_ARGS} -m pytest -s tests/test_tensor_constructor_ops.py --ref=cpu --mode=quick && \
 run_command coverage run ${COVERAGE_ARGS} -m pytest -s tests/test_attention_ops.py --ref=cpu --mode=quick
 
-mkdir -p /home/zhangbo/PR_Coverage/PR${PR_ID}/${ID_SHA}
-mv ${ID_SHA}* /home/zhangbo/PR_Coverage/PR${PR_ID}/${ID_SHA}
+# mkdir -p /home/zhangbo/PR_Coverage/PR${PR_ID}/${ID_SHA}
```
SamYuan1990: Will you remove all this mkdir in different shells?

.github/workflows/op-test-quick-cpu.yaml
```diff
@@ -41,4 +41,14 @@ jobs:
             echo "This is a push event. The relate PR number is ${PR_NUMBER}"
             PR_ID=${PR_NUMBER}
           fi
+          rm -f .coverage
           bash tools/op-test-quick-cpu.sh ${PR_ID}
+          coverage combine coverage-*
+          mkdir -p cov_html
+          coverage html -d cov_html
+
+      - id: upload_artifacts
```
SamYuan1990: I am worry about report size. if the report size too large, over the artifact size limit.

.github/workflows/op-test-quick-cpu.yaml
```diff
@@ -41,4 +41,14 @@ jobs:
             echo "This is a push event. The relate PR number is ${PR_NUMBER}"
             PR_ID=${PR_NUMBER}
           fi
+          rm -f .coverage
           bash tools/op-test-quick-cpu.sh ${PR_ID}
+          coverage combine coverage-*
+          mkdir -p cov_html
+          coverage html -d cov_html
```
SamYuan1990: Current job seems collect all result from a hard code folder, after all job passed.
if so... we need move this page generate logic from here to .github/workflows/python-coverage.yaml after a download artifacts.

tools/op-test-quick-cpu.sh
tengqm: Yes. That is the plan. But this has to be done one by one.

.github/workflows/op-test-quick-cpu.yaml
tengqm: the current report is 280 KB.
I'm not expecting it to be huge.

.github/workflows/op-test-quick-cpu.yaml
tengqm: That is too big a step forward.
I have to gain a little bit confidence before revising other scripts and workflows.
The eventual goal is to cleanup the many workflows.

.github/workflows/op-test-quick-cpu.yaml
tengqm: Okay. I'm converting the sqllite db directly to HTML pages for human examination. The size is about 5 MB now.

PR #1502, opened 2026-01-26
Fix get_sm_version_num function to check for attribute

Galaxy1458:
### Progress


src/flag_gems/fused/cutlass_scaled_mm.py
```diff
@@ -10,8 +10,9 @@
 
 
 def get_sm_version_num():
-    major, minor = torch_device_fn.get_device_capability()
-    return major * 10 + minor
+    if hasattr(torch_device_fn, "get_device_capability"):
```
tengqm, line 13: This smells to me like a compromise ... we are hoping that vendors provide this function but sometimes it is not provided. Am I understanding this correctly?

PR #1504, opened 2026-01-26
【Operator】add FLA（DeltaNet） operator

kiddyjinjin:
### PR Category

Operator

### Type of Change

New Feature

### Progress


benchmark/test_fused_recurrent_gated_delta_rule_perf.py
```diff
@@ -0,0 +1,176 @@
...
+    if contiguous:
+        return query.contiguous(), key.contiguous(), value.contiguous()
+    else:
+        return query, key, value
+
+
+class FusedRecurrentGatedDeltaRuleBenchmark(Benchmark):
+    DEFAULT_DTYPES = [torch.bfloat16]
+
+    def __init__(self, *args, qkv_contiguous: bool, **kwargs):
```
tengqm: There are two acceptable ways for declaring this function:

- treat `qkv_contiguous` as a positional argument

  ```python
  def __init__(self, qkv_contiguous: bool, *args, **kwargs):
  ```

  `qkv_contiguous` is now a positional argument, followed by other positional argument(s) and/or keyword arguments.

- treat `qkv_contiguous` as a keyword argument (my preference)

  ```python
  def __init__(self, *args, **kwargs):
      self.qkv_contiguous = kwargs.pop("qkv_contiguous", False)
      super().__init__(*args, **kwargs)
  ```

  This is less intrusive to the existing design/implementation, with `qkv_contiguous` added as an optional keyword argument.

kiddyjinjin: nice suggestion

PR #1576, opened 2026-02-10
[KUNLUNXIN] fix conv2d

geishinen:
### PR Category
Operator

### Type of Change
Bug Fix

### Description
Problem: Conv2d backward pass produced severe accuracy errors when computing input gradients under specific parameter combinations (43.7% element mismatch, max error up to 49.9).

Root Cause: When using transposed convolution to compute input gradients in the backward pass, the calculation formula for new_out_height/width lacked output_padding compensation. When (input_size + 2*padding - dilation*(kernel-1) - 1) % stride != 0, the original formula produced undersized dilated gradient dimensions, causing subsequent convolution index misalignment.

Fix: Changed the new_out_height/width calculation from an indirect formula based on output size to a direct formula based on input size, which automatically incorporates output_padding compensation


huangyiqun:
Just a gentle reminder: Please sign the CLA.

PR #1578, opened 2026-02-10
improve corner case for topk

physics31415926:
### PR Category
Operator

### Type of Change
Refactor

### Description
improve corner case for topk

### DIFF
--- a/src/flag_gems/ops/topk.py
+++ b/src/flag_gems/ops/topk.py
@@ -278,7 +278,7 @@ def topk(x, k, dim=-1, largest=True, sorted=True):
     # Early return for k=0 to avoid Triton kernel compilation error.
     # Triton's tl.arange(0, BLOCK_SIZE) requires BLOCK_SIZE > 0.
     # When k=0, stage2_elem_cnt becomes 0, leading to BLOCK_SIZE=0.
-    if k <= 0:
+    if k == 0:
         out_shape = list(x.shape[:-1]) + [0]
         return (
             torch.empty(out_shape, device=x.device, dtype=x.dtype),


src/flag_gems/ops/topk.py
```diff
@@ -275,6 +275,11 @@ def topk(x, k, dim=-1, largest=True, sorted=True):
     assert dim == x.ndim - 1, "Currently only support topk in last dimension"
     # assert sorted, "Currently only support sorted == True"
 
+    if k <= 0:
```
tengqm: Better add a comment before this line so that other contributors know why `k<=0` warrants a special treatment.

src/flag_gems/ops/topk.py
```diff
@@ -275,6 +275,16 @@ def topk(x, k, dim=-1, largest=True, sorted=True):
...
 
+    # Early return for k=0 to avoid Triton kernel compilation error.
+    # Triton's tl.arange(0, BLOCK_SIZE) requires BLOCK_SIZE > 0.
+    # When k=0, stage2_elem_cnt becomes 0, leading to BLOCK_SIZE=0.
+    if k <= 0:
+        out_shape = list(x.shape[:-1]) + [0]
+        return (
+            torch.empty(out_shape, device=x.device, dtype=x.dtype),
+            torch.empty(out_shape, device=x.device, dtype=torch.int64),
+        )
```
tengqm, line 286: Change looks good to me ...
wondering if the existing test case can cover this branch?

physics31415926: test cases now may not cover these ops

tengqm: How about augmenting the existing test case (no necessarily a new one) for the case `k = -1`?

physics31415926: added -1 as test case in test_topk

PR #1582, opened 2026-02-11
add Moore Threads CI support for FlagGems

douxetpur:
### Progress

### DIFF
--- a/.github/workflows/unittest.yaml
+++ b/.github/workflows/unittest.yaml
@@ -39,12 +39,10 @@ jobs:
   wait-triage:
     runs-on: ubuntu-latest
     steps:
-      - run: |
-          echo "done"
-      # - name: Wait for PR triage
-      #   uses: ArcticLampyrid/action-wait-for-workflow@v1
-      #   with:
-      #     workflow: .github/workflows/triage.yaml
+      - name: Wait for PR triage
+        uses: ArcticLampyrid/action-wait-for-workflow@v1
+        with:
+          workflow: .github/workflows/triage.yaml
 
   build-doc:
     needs: wait-triage


tengqm:
@douxetpur Please sign the CLA in order for your PR to be reviewed.

.github/workflows/gems-test-on-moore.yaml
```diff
@@ -0,0 +1,114 @@
...
+          source tools/run_command.sh
+          GEMS_VENDOR=moore run_command python3 -m pytest -s \
+            tests/test_pointwise_dynamic.py
+
+      - name: FlagGems distribution ops on moore
+        shell: bash
+        run: |
+          source tools/run_command.sh
+          GEMS_VENDOR=moore run_command python3 -m pytest -s \
+            tests/test_distribution_ops.py
```
tengqm: It is okay to add this workflow as is at the moment, however, when you look at the global picture, adding new workflow for each and every backend is not a sustainable approach. The overall direction for the team to pursue is to simply and unify the unit test workflows.

douxetpur: I completely agree that implementing a unified matrix-based workflow would be the ideal architectural approach long-term. However, given our current development timeline and the urgent need to validate Moore Threads functionality, I propose we treat this initial CI implementation as a temporary but necessary stepping stone - it will serve the immediate purpose of unblocking Moore Threads testing while providing valuable implementation insights that will directly inform the future unified workflow design.

tengqm: 💯

douxetpur:
@tengqm Thanks for revisiting this PR and for the suggestions !
I agree with the direction. This workflow is intended as a pragmatic first step to get Moore Threads CI running and unblock validation. I’ll follow up with separate PRs to 
(1) refactor this into a reusable backend-specific sub-workflow, and 
(2) consolidate the multiple test steps into a single step that invokes a shell script under `tools/`, so GitHub-specific logic stays minimal and the test orchestration lives in repo scripts.

PR #1602, opened 2026-02-13
[KUNLUNXIN] fix max_pool2d and avg_pool2d

geishinen:
### PR Category
Benchmark | Operator

### Type of Change
Bug Fix

### Description
Description
Fix test_perf_max_pool2d_backward benchmark test failure caused by incompatible indices format between PyTorch and FlagGems.

Root Cause:

PyTorch's max_pool2d_with_indices and FlagGems' max_pool2d_with_indices produce indices in different formats

The original test used PyTorch forward to generate indices, then passed them to FlagGems backward

This format mismatch caused illegal memory access errors

Changes:

Benchmark test fix: Modified test_perf_max_pool2d_backward to use flag_gems.max_pool2d_with_indices for forward pass, ensuring indices are compatible with flag_gems.max_pool2d_backward

Indices dtype optimization: Changed indices tensor dtype from int64 to int32 in:

FlagGems/src/flag_gems/ops/max_pool2d_with_indices.py
FlagGems/src/flag_gems/runtime/backend/_kunlunxin/ops/max_pool2d_with_indices.py
C++ layer robustness: Added int64→int32 conversion handling in launch_max_pool2d_backward for cases where indices come as int64 (e.g., from PyTorch native forward), with proper synchronization to prevent use-after-free of temporary buffers

### DIFF
--- a/benchmark/test_reduction_perf.py
+++ b/benchmark/test_reduction_perf.py
@@ -399,11 +399,14 @@ def test_perf_max_pool2d_backward():
         for forward_args in max_pool2d_input_fn(shape, dtype, device):
             inp, params = forward_args
             inp.requires_grad_(True)
-            output, indices = torch.nn.functional.max_pool2d_with_indices(inp, **params)
+            # Use FlagGems forward to produce indices compatible with FlagGems backward
+            # Note: FlagGems indices format differs from PyTorch's format
+            output, indices = flag_gems.max_pool2d_with_indices(inp, **params)
             grad_output = torch.randn_like(output)
             yield grad_output, inp, indices, params
 
     def torch_max_pool2d_backward_wrapper(grad_output, input, indices, **kwargs):
+        # For torch baseline, we use torch forward to get compatible indices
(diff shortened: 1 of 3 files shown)


src/flag_gems/ops/max_pool2d_with_indices.py
```diff
@@ -292,7 +293,7 @@ def max_pool2d_with_indices(
         (in_n, in_c, out_h, out_w), device=input.device, dtype=input.dtype
     )
     indices = torch.empty(
-        (in_n, in_c, out_h, out_w), device=input.device, dtype=torch.int64
+        (in_n, in_c, out_h, out_w), device=input.device, dtype=torch.int32
```
kiddyjinjin: Could you explain why the type was changed from int64 to int32?

kiddyjinjin: What I meant is: why do we need to change the general implementation in `src/flag_gems/ops/max_pool2d_with_indices.py` instead of only updating the backend-specific logic in `src/flag_gems/runtime/backend/_kunlunxin/ops/max_pool2d_with_indices.py`?

PR #1626, opened 2026-02-15
[Iluvatar] use torch_moe_align_block_size as reference.

awayzjj:
### Description

`vllm_op.moe_align_block_size` on Iluvatar has accuracy  issues, use torch implementation as reference instead. 

### Progress

### DIFF
--- a/tests/test_special_ops.py
+++ b/tests/test_special_ops.py
@@ -1665,12 +1665,101 @@ def test_moe_sum(shape, dtype):
     gems_assert_close(res_out, ref_out, dtype)
 
 
-try:
-    import vllm._custom_ops as vllm_ops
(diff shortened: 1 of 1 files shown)


tests/test_special_ops.py
```diff
@@ -1666,13 +1666,108 @@ def test_moe_sum(shape, dtype):
 
 
 try:
-    import vllm._custom_ops as vllm_ops
-
```
tengqm, line 1670: Why this change?

awayzjj: `vllm_op.moe_align_block_size` on Iluvatar has accuracy issues, use torch implementation as reference instead.

tengqm: Then we don't need the `try ... except` block, right?

awayzjj: yes, I removed the `try ... except` block. Thanks!

PR #1668, opened 2026-02-22
Fix script of GPU check on iluvatar runners

tengqm:
### PR Category

CI/CD

### Type of Change

Bug Fix

### Description

The tool chain for Iluvatar accelerator lives in a special directory.
This has to be configured before running the GPU check and test cases so that the latter can find GPUs to use.
This PR also attempts to turn all scripts on for Iluvatar platforms, while treating MooreThreads as a special beast.

### Progress


tengqm:
This one is no longer needed.

PR #1673, opened 2026-02-24
[TSINGMICRO] skip mm and addmm with float32 input since low precision of tx81 hw

tsingmicro-public-e:
### Progress


tests/test_blas_ops.py
```diff
@@ -39,6 +39,9 @@
 @pytest.mark.parametrize("dtype", FLOAT_DTYPES)
 @pytest.mark.parametrize("b_column_major", [True, False])
```
huangyiqun, line 40: Please use
"""
@pytest.mark.skipif(vendor_name == "tsingmicro", reason="Skipping fp32 addmm test on Tsinghua Micro platform")
"""
to conditionally skip the test, rather than modifying the test logic inside the test function itself.

tsingmicro-public-e: I only want to skip dtype==float32

PR #1686, opened 2026-02-25
fix(ci): fix workflow bugs and update runner labels

douxetpur:
### Bug Fixes

1. **Remove invalid `types` field from `push` event**
   - `push` event does not support `types` (only valid for `pull_request`)

2. **Fix typo in `build-doc` job condition**
   - `outpus` → `outputs` (this bug prevented doc builds from ever triggering)

3. **Fix unsafe object reference in `quick-cpu-op`**
   - Added event type guard for `github.event.pull_request` access

4. **Use `fromJson()` for label matching**
   - Changed from substring match to proper JSON array lookup

### Runner Updates(use validated environment)
 
- `ascend` → `ascend_CI`
- `iluvatar` → `iluvatar_CI`

---

## 💬 Discussion Points

While fixing these issues, I found some architectural topics worth discussing:

### 1. Label System vs Path Filtering

**Current State:**
- The workflow uses labels to trigger different tests
- Adding new tests requires modifying multiple configuration files

**Issues Found:**
- Too many labels increase maintenance burden
- Some vendor labels have no corresponding CI tests
- Rules are scattered across multiple files

**Trade-off:**

| Aspect | Label-driven | Path Filtering |
|--------|-------------|----------------|
| Manual override (skip tests) | ✅ Remove label | ❌ Not possible |
| PR visualization | ✅ Labels visible | ❌ Check logs |
| Files to maintain | 2 | 1 |

**Are there scenarios where we need to manually add/remove labels? If not, should we consider simplifying?**

### 2. Environment Isolation for Shared Clusters

**Current State:** CI runs directly on self-hosted runners without container isolation.

**Challenge:** If migrating to organization-shared clusters, we may face:
- Dependency conflicts between projects
- Environment pollution
- Difficulty reproducing issues locally

**Current Progress:**
- Some base images exist, but CI-specific images need to be rebuilt
- Workflow files need to be modified to add `container` configuration or use `docker run`

**Proposed Solution:** Use Docker containers for test isolation.

### 3. Additional Shared Cluster Considerations

- **Resource contention**: GPU scheduling, memory limits
- **Environment consistency**: CUDA/Python version management
- **Security**: Secret isolation between projects
- **Observability**: Centralized logging and monitoring

### DIFF
--- a/.github/workflows/unittest.yaml
+++ b/.github/workflows/unittest.yaml
@@ -5,7 +5,6 @@ on:
 
   push:
     branches: [ "master" ]
-    types: [opened, synchronize, reopened, labeled, unlabeled]
 
   pull_request:
     branches: [ "master" ]
@@ -59,17 +58,17 @@ jobs:
 
   build-doc:
     needs: preprocess
-    if: contains(needs.preprocess.outpus.labels, 'documentation')
+    if: contains(fromJson(needs.preprocess.outputs.labels), 'documentation')
     uses: ./.github/workflows/mkdocs.yaml
 
   cpp-op:
     needs: preprocess
-    if: contains(needs.preprocess.outputs.labels, 'ops/cpp')
+    if: contains(fromJson(needs.preprocess.outputs.labels), 'ops/cpp')
     uses: ./.github/workflows/cpp-op-test.yaml
 
   blas-op:
     needs: preprocess
-    if: contains(needs.preprocess.outputs.labels, 'tests')
+    if: contains(fromJson(needs.preprocess.outputs.labels), 'tests')
     concurrency:
       group: blas-op-test-${{ github.event.pull_request.number || github.ref }}
       cancel-in-progress: true
@@ -82,8 +81,8 @@ jobs:
     needs: preprocess
     # NOTE: This contains a temporary hack to skip this job for PRs from FODC contest.
     if: |
-      contains(needs.preprocess.outputs.labels, 'tests') &&
-      ! contains(github.event.pull_request.labels.*.name, 'competition')
+      contains(fromJson(needs.preprocess.outputs.labels), 'tests') &&
+      (github.event_name != 'pull_request' || ! contains(github.event.pull_request.labels.*.name, 'competition'))
     concurrency:
       group: op-test-quick-cpu-${{ github.event.pull_request.number || github.ref }}
       cancel-in-progress: true
@@ -94,7 +93,7 @@ jobs:
 
   pointwise-op:
     needs: preprocess
-    if: contains(needs.preprocess.outputs.labels, 'tests')
+    if: contains(fromJson(needs.preprocess.outputs.labels), 'tests')
     concurrency:
       group: pointwise-op-test-${{ github.event.pull_request.number || github.ref }}
       cancel-in-progress: true
@@ -105,7 +104,7 @@ jobs:
 
   reduction-op:
     needs: preprocess
-    if: contains(needs.preprocess.outputs.labels, 'tests')
+    if: contains(fromJson(needs.preprocess.outputs.labels), 'tests')
     concurrency:
       group: reduction-op-test-${{ github.event.pull_request.number || github.ref }}
       cancel-in-progress: true
@@ -116,7 +115,7 @@ jobs:
 
   special-op:
     needs: preprocess
-    if: contains(needs.preprocess.outputs.labels, 'tests')
+    if: contains(fromJson(needs.preprocess.outputs.labels), 'tests')
     concurrency:
       group: special-op-test-${{ github.event.pull_request.number || github.ref }}
       cancel-in-progress: true
@@ -127,7 +126,7 @@ jobs:
 
   utils:
     needs: preprocess
-    if: contains(needs.preprocess.outputs.labels, 'tests')
+    if: contains(fromJson(needs.preprocess.outputs.labels), 'tests')
     concurrency:
       group: op-utils-test-${{ github.event.pull_request.number || github.ref }}
       cancel-in-progress: true
@@ -138,7 +137,7 @@ jobs:
 
   examples:
     needs: preprocess
-    if: contains(needs.preprocess.outputs.labels, 'examples')
+    if: contains(fromJson(needs.preprocess.outputs.labels), 'examples')
     concurrency:
       group: examples-test-${{ github.event.pull_request.number || github.ref }}
       cancel-in-progress: true
@@ -149,7 +148,7 @@ jobs:
 
   backend-nvidia:
     needs: preprocess
-    if: contains(needs.preprocess.outputs.labels, 'vendor/NVIDIA')
+    if: contains(fromJson(needs.preprocess.outputs.labels), 'vendor/NVIDIA')
     uses: ./.github/workflows/backend-test.yaml
     with:
       vendor: nvidia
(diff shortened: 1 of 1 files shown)


.github/workflows/unittest.yaml
```diff
@@ -59,17 +58,17 @@ jobs:
 
   build-doc:
     needs: preprocess
-    if: contains(needs.preprocess.outpus.labels, 'documentation')
+    if: contains(fromJson(needs.preprocess.outputs.labels), 'documentation')
```
tengqm, line 61: The existing code actually works because we didn't bother to do a `toJson` when generating the `labels` array.

.github/workflows/unittest.yaml
```diff
@@ -82,8 +81,8 @@ jobs:
     needs: preprocess
     # NOTE: This contains a temporary hack to skip this job for PRs from FODC contest.
     if: |
-      contains(needs.preprocess.outputs.labels, 'tests') &&
-      ! contains(github.event.pull_request.labels.*.name, 'competition')
+      contains(fromJson(needs.preprocess.outputs.labels), 'tests') &&
+      (github.event_name != 'pull_request' || ! contains(github.event.pull_request.labels.*.name, 'competition'))
```
tengqm, line 85: Thanks for this fix.
The `competition` label is currently manually applied.
And, when the labels for a PR is changed, the unittest.yaml workflow is retriggered.
The problem with `github.event.pull_request.labels.*.name` is that the labels is cached.
The new labels we automatically attached won't be effective if we use the cached value.

.github/workflows/unittest.yaml
```diff
@@ -159,27 +158,27 @@ jobs:
 
   backend-ascend:
     needs: preprocess
-    if: contains(needs.preprocess.outputs.labels, 'vendor/Ascend')
+    if: contains(fromJson(needs.preprocess.outputs.labels), 'vendor/Ascend')
     uses: ./.github/workflows/backend-test.yaml
     with:
       vendor: ascend
-      runner_label: ascend
+      runner_label: ascend_CI
```
tengqm, line 165: No. We use the shared CI runner pool.
We don't want each and every flagos-ai project to host their own runner pool.
It is a huge wasted of resource and it creates silos between projects and extra burden for managing the runners.
All `<backend>_CI` runners are supposed to be migrated to the organization level for sharing, or to be retired.

tengqm:
Reply to your questions/comments, @douxetpur 

> Current State:
> The workflow uses labels to trigger different tests

Yes. That is one of the intentions behind the refactoring. We don't want to trigger all unit-tests for a simple PR that modifies a `.md` file. We don't want to run all unit-tests when a PR adds an operator to one backend.

> Adding new tests requires modifying multiple configuration files.

This has been a huge pain for the project. We used to have multiple workflows (each bound to a different shell script) to drive a subset of unit tests. The unit test file names were sometimes hardcoded in the workflow and sometimes hardcoded in the shell scripts.

The next steps of this refactorization are:

- converge all shell scripts into one or two
- make the shell scripts generic enough for all test case files
- converge benchmark test cases into the same framework
- split the huge test case files into smaller ones (ideally, each operator has its own test case file)

The eventual goal of the refactor are:

- each PR only triggers the unit test that is relevant to the PR
- shorten the unit test time from 80+ minutes to less than 5 minutes
- improve the CI runners' utilization (No PR will lock a runner for more than 5 minutes,
  compared to the current situation where each PR will occupy a runner for 80+ minutes,
  you can imagine the throughput improvement)
- with proper engineering of the workflow/scripts, we can drive scheduled workflows to
  trigger a thorough unit-test run (including accuracy and performance) during night time,
  so we won't miss the whole picture when we drastically reduce PR turn-around time.

> Issues Found:
> - Too many labels increase maintenance burden

This is not a problem. Most (if not all) labels are attached to a PR automatically. No one enjoys the burden of manually triaging a PR and/or labelling them. Human burden is hugely reduced.

> Some vendor labels have no corresponding CI tests

This has been another pain point in the project. We are working on preparing runners for most (if not all) vendors. The goal is to have all vendor-specific changes tested/gated instead of merged blindly. Feel free to join this effort.

> Rules are scattered across multiple files

Feel free to propose improvements in this space. No one likes duplicated code or rules.

.github/workflows/unittest.yaml
douxetpur: Thanks for the clarification!

I added fromJson() mainly to make the label handling explicit,
but I see that the current implementation works without it.
Happy to adjust if you prefer the original style.

.github/workflows/unittest.yaml
douxetpur: Thanks for the explanation.

That makes sense — using preprocess outputs avoids the cached
github.event.pull_request.labels issue.

.github/workflows/unittest.yaml
douxetpur: Thanks for the clarification.

We'll switch to the unified runner labels after the stability testing.
I'm still updating the workflow and will push the changes later today.

.github/workflows/unittest.yaml
tengqm: We can watch the CI job log closely later and see how `contains()` actually work.
If `fromJson` is a better choice, we can do it when generating the `labels` output in the `preprocess` job. What do you think?

.github/workflows/unittest.yaml
tengqm: Inspired by your question, I have added an extra step to label a pull request with `competition` automatically, so this dirty extra check can be safely removed now.

.github/workflows/unittest.yaml
tengqm: That is a reasonable compromise. I agree we can test the backend CI jobs on repository runners and migrate them to organization runners when things become stable and manageable.
Looking forward to the results from backend specific tests!

tengqm:
> @tengqm Thanks for the detailed explanation — I really appreciate the clarification on the CI design and the refactoring direction. I agree with the approach you described. I spent some time thinking through the points you raised, so my reply came a bit later.
> 

Never mind. An open discussion helps everyone to better understand the overall problem and to evaluate the various options.

> Right now I'm focusing on getting CI set up correctly for several backends first. Once the stability tests are verified, we can move them to the shared runner pool.

Sounds a reasonable plan. I might be too eager to get things done in a consistent, global manner. Rome was not built in one day. We have to do this step by step.

> I’m also thinking a bit about the security side of CI/CD. Since CI jobs essentially provide an RCE surface, and there have been CI-related attacks in the past (including in GitHub’s ecosystem), do you have any recommendations or best practices for improving CI security in this setup?

Right. I have heard similar stories as well. There are several things we have to live with:

- The whole open-source ecosystem is based on trust, and the belief shared by people is that being open is more secure than being close.
- When a security issue is identified, the whole community will propose a solution quickly, but we have to actively watch this.
- All software have bugs, it is simply a matter of severity.

As for what we can do ... several thoughts from my side:

- To make sure we can respond to security problem quickly, we may need to explicitly declare the versions/tags of the libraries/packages we depend on. Then we can enable the `dependabot` from GitHub. It will help tell us that `torch==2.2.0` has CVEs to fix.
- When we are using GitHub actions (which are themselves JS programs), we pin the actions to specific versions. That is, we treat them as third-party software.
- We have to actively review each and every PRs to the repo.
- We need to protect our `master` branch.
- When writing CI jobs, pay close attention to security related information such as tokens, passwords, keys.
- ...

> I’ll continue working on the CI improvements on my side as well.

Cool.

.github/workflows/unittest.yaml
douxetpur: That makes sense. If we eventually move to `fromJson`, generating the labels as a proper JSON array in the `preprocess` job would probably be cleaner.

For now I'm fine with keeping the current behavior and observing how `contains()` behaves in the CI logs.

.github/workflows/unittest.yaml
douxetpur: Nice, that makes the workflow cleaner. Thanks for adding the automatic `competition` labeling — removing that extra check definitely simplifies the logic.

.github/workflows/unittest.yaml
douxetpur: Thanks for confirming! I'm still updating the workflow and testing the backend-specific CI jobs. I'll push the changes once everything is ready.

douxetpur:
> As for what we can do ... several thoughts from my side:
> 
> * To make sure we can respond to security problem quickly, we may need to explicitly declare the versions/tags of the libraries/packages we depend on. Then we can enable the `dependabot` from GitHub. It will help tell us that `torch==2.2.0` has CVEs to fix.
> * When we are using GitHub actions (which are themselves JS programs), we pin the actions to specific versions. That is, we treat them as third-party software.
> * We have to actively review each and every PRs to the repo.
> * We need to protect our `master` branch.
> * When writing CI jobs, pay close attention to security related information such as tokens, passwords, keys.
> * ...
> 
Thanks for the insights and suggestions on CI security. These are very helpful. I’ll keep them in mind while working on the CI updates.

tengqm:
@douxetpur I'm gonna close this.
Please raise a new PR if there are still some defects to be fixed. Thanks.

PR #1775, opened 2026-02-27
[FlagGems Operator Development Competition] Add ctc_loss operator

BuaaAlban:
## Summary
- Implement `torch.nn.functional.ctc_loss` (forward + backward) with Triton
- Register at `aten::ctc_loss.Tensor` and `aten::ctc_loss.IntList` so `F.ctc_loss` reaches the Triton path directly
- Supports padded (2D) and concatenated (1D) targets, batched (3D) and unbatched (2D) `log_probs`, fp32 / fp16 / bf16
- All reduction modes (`"none"` / `"mean"` / `"sum"`) and `zero_infinity` flag

## Implementation
- **Forward kernel** (`_ctc_loss_forward_kernel`): one program per batch element, vectorized across S' label states. Computes alpha DP in log-space with on-the-fly can-skip mask; stores the full alpha table for backward; NLL via final logaddexp.
- **`_ctc_loss_init_grad_kernel`**: single vectorized launch over `(T, N, C)` initializes `grad = exp(log_probs) * scale` with reduction-aware scaling and `zero_infinity` handling.
- **Backward kernel** (`_ctc_loss_backward_kernel`): reverse-time beta DP with double-buffered scratch and `atomic_add` scatter to the gradient tensor; reduction (none/mean/sum) and `zero_infinity` handled inside the kernel — no extra launches required.
- **Python wrapper** (`_CtcLossFunction`, `torch.autograd.Function`): handles 2D/3D `log_probs`, list/tuple lengths, concatenated 1D targets, float targets; promotes fp16/bf16 → fp32 for compute and casts the result back.

## Test Coverage

50 tests passing. Coverage matrix:

| Dimension | Coverage |
|---|---|
| Reduction | `none` / `mean` / `sum` |
| Target layout | padded `(N, S)` / concatenated 1D |
| Length form | `Tensor` / Python `list` (IntList overload) |
| dtype | `float32` / `float16` / `bfloat16` (when supported) |
| Blank label | `0` / non-zero |
| Input lengths | uniform `T` / per-sample varying |
| Edge cases | unbatched 2D `log_probs` / repeated labels / empty target row / `zero_infinity` / non-contiguous `log_probs` |
| Error paths | invalid `blank` / invalid `reduction` / mismatched concatenated targets |

### Test Functions

| # | Test | Description |
|---|---|---|
| 1 | `test_accuracy_ctc_loss` | Forward + backward over the (dtype × reduction) matrix |
| 2 | `test_accuracy_ctc_loss_concatenated_targets` | 1D concatenated targets |
| 3 | `test_accuracy_ctc_loss_intlist_lengths` | `IntList` overload (Python list lengths) |
| 4 | `test_accuracy_ctc_loss_varying_input_lengths` | Per-sample varying `input_lengths` |
| 5 | `test_accuracy_ctc_loss_nonzero_blank` | `blank=0` and `blank=3` |
| 6 | `test_accuracy_ctc_loss_zero_infinity` | `zero_infinity=True` (impossible alignment) |
| 7 | `test_accuracy_ctc_loss_unbatched` | Unbatched 2D `log_probs` |
| 8 | `test_accuracy_ctc_loss_repeated_labels` | Repeated adjacent labels (skip-disallow) |
| 9 | `test_accuracy_ctc_loss_empty_target` | Sample with `target_length = 0` |
| 10 | `test_accuracy_ctc_loss_noncontiguous` | Non-contiguous `log_probs` |
| 11 | `test_accuracy_ctc_loss_invalid_blank` | Negative / out-of-range `blank` |
| 12 | `test_accuracy_ctc_loss_invalid_reduction` | Unknown `reduction` string |
| 13 | `test_accuracy_ctc_loss_invalid_concatenated_targets` | `sum(target_lengths) != targets.numel()` |

### Forward + Backward (fp32, `reduction='mean'`)

| Shape (T, N, C, S) | PyTorch (ms) | FlagGems (ms) | Speedup |
|---|---|---|---|
| (64, 4, 32, 16) | 3.629 | 0.557 | **6.52x** |
| (256, 16, 64, 48) | 4.188 | 0.944 | **4.44x** |
| (512, 32, 64, 48) | 4.965 | 1.375 | **3.61x** |
| (1024, 32, 128, 96) | 6.201 | 4.599 | **1.35x** |

### Forward only (fp32, `reduction='mean'`)

| Shape (T, N, C, S) | PyTorch (ms) | FlagGems (ms) | Speedup |
|---|---|---|---|
| (64, 4, 32, 16) | 2.024 | 0.501 | **4.04x** |
| (256, 16, 64, 48) | 2.407 | 0.677 | **3.55x** |
| (512, 32, 64, 48) | 2.704 | 0.892 | **3.03x** |
| (1024, 32, 128, 96) | 3.078 | 2.071 | **1.49x** |

All shapes meet the official ≥ 0.9x speedup baseline; both forward and forward+backward exceed 1.0x across the entire benchmark sweep.


src/flag_gems/ops/__init__.py
```diff
@@ -313,6 +314,9 @@
     "cos",
     "cos_",
     "count_nonzero",
+    "ctc_loss",
+    "_ctc_loss_impl",
+    "_ctc_loss_backward_impl",
```
tengqm: Pls help keep the list strictly sorted.

src/flag_gems/__init__.py
```diff
@@ -117,6 +117,10 @@ def torch_ge(v):
     ("cos", cos),
     ("cos_", cos_),
     ("count_nonzero", count_nonzero),
+    ("_ctc_loss", _ctc_loss_impl),
+    ("_ctc_loss.Tensor", _ctc_loss_impl),
+    ("_ctc_loss_backward", _ctc_loss_backward_impl),
+    ("_ctc_loss_backward.Tensor", _ctc_loss_backward_impl),
```
tengqm: pls keep the list strictly sorted, '_' matters as well.

benchmark/test_ctc_loss_perf.py
```diff
@@ -3,64 +3,105 @@
 import pytest
 import torch
 
+import flag_gems
 from benchmark.attri_util import BenchLevel
 from benchmark.performance_utils import Config, GenericBenchmark
```
tengqm: These lines are deprecated, please check the new code base.

benchmark/test_ctc_loss_perf.py
```diff
@@ -3,64 +3,105 @@
...
+        yield log_probs, concat_targets, input_lengths, target_lengths, {
+            "blank": blank,
+            "reduction": "mean",
+            "zero_infinity": False,
+        }
+
+
+CTC_BENCH_DTYPES = [torch.float32, torch.float16]
+if flag_gems.runtime.device.support_bf16:
+    CTC_BENCH_DTYPES.append(torch.bfloat16)
```
tengqm: Not sure if we want to compromise on this ...

benchmark/test_ctc_loss_perf.py
```diff
@@ -3,64 +3,105 @@
...
+
 
 @pytest.mark.ctc_loss
 def test_perf_ctc_loss():
     bench = CTCLossBenchmark(
         input_fn=ctc_loss_input_fn,
         op_name="ctc_loss",
-        torch_op=torch.nn.functional.ctc_loss,
-        dtypes=[torch.float32],
+        torch_op=_reference_ctc_loss,
```
tengqm: I don't like this wrapping logic ...
It smells like that we are not comparing to the *real* torch function.
How about remove the `_reference_ctc_loss` function and move the data type manipulation logic into this function?

benchmark/test_ctc_loss_perf.py
```diff
@@ -3,64 +3,105 @@
...
-        for shape in ctc_shapes:
+        for shape in CTC_DEFAULT_SHAPES:
             yield from self.input_fn(shape, cur_dtype, self.device)
 
+    def set_more_shapes(self):
+        return []
+
 
 @pytest.mark.ctc_loss
 def test_perf_ctc_loss():
```
tengqm: Please drop the `_perf` part.
It is meaningless and redundant.

src/flag_gems/ops/__init__.py
```diff
@@ -339,8 +339,6 @@
 __all__ = [
     "_assert_async",
     "_conv_depthwise2d",
-    "_ctc_loss_backward_impl",
-    "_ctc_loss_impl",
```
tengqm: why drop these entries?

benchmark/test_ctc_loss_perf.py
BuaaAlban: ok torch only supports for fp32

src/flag_gems/ops/__init__.py
BuaaAlban: Good catch — these were dropped during the rewrite when I switched the user-facing path                                                                                                                                                                                                               
  to a torch.autograd.Function attached at ctc_loss.{Tensor,IntList}. Functionally that                                                                                                                                                                                                                 
  makes F.ctc_loss work end-to-end (forward goes through ctc_loss(...) and backward goes                                                                                                                                                                                                                
  through _CtcLossFunction.backward), so I had assumed the legacy decomposition entries                                                                                                                                                                                                                 
  were unreachable. But you're right that we shouldn't silently shrink the dispatch coverage                                                                                                                                                                                                            
  on an op that has explicit aten::_ctc_loss / aten::_ctc_loss_backward overloads.                                                                                                                                                                                                                      
                                                                                                                                                                                                                                                                                                        
  Restored in 21b73e74:                                                                                                                                                                                                                                                                                 
                                                                                                                                                                                                                                                                                                        
  - Added _ctc_loss_impl and _ctc_loss_backward_impl as thin wrappers that share the                                                                                                                                                                                                                    
    same Triton kernels (_ctc_loss_forward_kernel, _ctc_loss_init_grad_kernel,
    _ctc_loss_backward_kernel) used by _CtcLossFunction. They expose the legacy                                                                                                                                                                                                                         
    signatures: _ctc_loss returns (neg_log_likelihood, log_alpha), and                                                                                                                                                                                                                                  
    _ctc_loss_backward consumes that pair plus grad and returns grad_log_probs.                                                                                                                                                                                                                         
  - Re-registered them on aten::_ctc_loss[.Tensor] and aten::_ctc_loss_backward[.Tensor]                                                                                                                                                                                                                
    in _FULL_CONFIG, and re-added _ctc_loss_impl / _ctc_loss_backward_impl to                                                                                                                                                                                                                           
    flag_gems.ops.__all__.                                                                                                                                                                                                                                                                              
  - The high-level ctc_loss.{Tensor,IntList} path is unchanged.                                                                                                                                                                                                                                         
                                                                                                                                                                                                                                                                                                        
  I cross-checked both dispatch paths against torch (50 unit tests + a manual diff                                                                                                                                                                                                                      
  against torch._VariableFunctions): both produce matching results.

douxetpur:

PR #1778, opened 2026-02-27
[Fix] addcmul out tensor shape mismatch when broadcasting

fangfangssj:
## Summary
- Fix `addcmul` raising `RuntimeError` when `out` tensor shape doesn't match the broadcasted result shape
- PyTorch natively resizes `out` to the correct broadcast shape, but FlagGems' `pointwise_dynamic` strictly rejects shape mismatches
- Also fix the `out=None` path to let `pointwise_dynamic` handle output allocation instead of using `torch.empty_like(inp)`, which could be incorrect for broadcast cases

## Root Cause
When `torch.addcmul(inp, t1, t2, value=v, out=out)` is called with inputs that broadcast to a larger shape (e.g., `(1,256,56,56)`) but `out` has a smaller shape (e.g., `(1,256,1,1)`), the original code passed `out` directly to `pointwise_dynamic` which validates `out.shape == broadcast_shape` and raises:
```
RuntimeError: out tensor at index 0 shape is invalid, should be (1, 256, 56, 56) but is torch.Size([1, 256, 1, 1])!
```

## Fix
- Compute `broadcast_shape = torch.broadcast_shapes(inp.shape, tensor1.shape, tensor2.shape)`
- Call `out.resize_(broadcast_shape)` when shapes differ (matching PyTorch native behavior)
- When `out is None`, delegate output allocation to `pointwise_dynamic` framework

## Test Plan
- [x] Added `test_accuracy_addcmul_out_broadcast` with 4 broadcast shape combinations x 2 dtypes (float16, float32)
- [x] Verified numerical correctness against PyTorch native output
- [x] Verified `out` tensor is correctly resized to broadcast shape
- [x] Ran GraphNet model `convnextv2_pico.fcmae` end-to-end (Speedup e2e: 1.165, gpu: 1.165)

### DIFF
--- a/tests/test_binary_pointwise_ops.py
+++ b/tests/test_binary_pointwise_ops.py
@@ -2158,7 +2158,9 @@ def test_accuracy_addcmul(shape, dtype):
     ],
 )
 @pytest.mark.parametrize("dtype", FLOAT_DTYPES)
-def test_accuracy_addcmul_out_broadcast(inp_shape, t1_shape, t2_shape, out_shape, dtype):
+def test_accuracy_addcmul_out_broadcast(
+    inp_shape, t1_shape, t2_shape, out_shape, dtype
+):
     res_inp = torch.randn(inp_shape, dtype=dtype, device=flag_gems.device)
     t1 = torch.randn(t1_shape, dtype=dtype, device=flag_gems.device)
     t2 = torch.randn(t2_shape, dtype=dtype, device=flag_gems.device)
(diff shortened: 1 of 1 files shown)


tests/test_binary_pointwise_ops.py
```diff
@@ -2147,6 +2147,42 @@ def test_accuracy_addcmul(shape, dtype):
...
+    t1 = torch.randn(t1_shape, dtype=dtype, device=flag_gems.device)
+    t2 = torch.randn(t2_shape, dtype=dtype, device=flag_gems.device)
+
+    ref_inp = to_reference(res_inp, True)
+    ref_t1 = to_reference(t1, True)
+    ref_t2 = to_reference(t2, True)
+
+    v = float(np.float32(random.random()))
+
+    ref_out_tensor = torch.randn(out_shape, dtype=dtype, device="cpu")
```
tengqm: you cannot hard code the device to "cpu" here ... that is the reason behind unit test failure.

fangfangssj: Fixed

huangyiqun (approved):

PR #1799, opened 2026-03-02
[Fix] arange: cast size to int to handle SymInt inputs

fangfangssj:
### PR Category
Operator

### Type of Change
Bug Fix

### Description
Fix `arange_start` TypeError when `size` is a `torch.SymInt` instead of plain Python `int`.

When `arange_start` receives SymInt-typed `start`/`end`/`step` arguments (from PyTorch 2.x dynamic shapes / `__torch_function__` dispatch), the computed `size` is also a SymInt. `torch.empty((size,), ...)` requires plain Python int elements in the size tuple, so passing a SymInt causes:
```
TypeError: empty() received an invalid combination of arguments - got (tuple, pin_memory=bool, dtype=torch.dtype, device=torch.device)
```

The fix adds an explicit `size = int(size)` cast after computing the output length.

Reproduced on `xlnet-base-cased` model via GraphNet test framework.

### Issue
- N/A (discovered during model testing)

### Progress

### Performance
Tested on xlnet-base-cased:
- Speedup[e2e]: 1.02028
- Speedup[gpu]: 1.02014

No performance regression (single `int()` cast on scalar value).

### DIFF
--- a/src/flag_gems/ops/arange.py
+++ b/src/flag_gems/ops/arange.py
@@ -29,19 +29,23 @@ def arange_start(
     start, end, step=1, *, dtype=None, layout=None, device=None, pin_memory=None
 ):
     logger.debug("GEMS ARANGE")
-    # Check for float step with int64 dtype - must raise error to match PyTorch behavior
-    # When step is float (e.g., 0.5) and dtype is int64, the float converts to int as 0
-    if dtype is torch.int64 and (
-        isinstance(step, float) or isinstance(start, float) or isinstance(end, float)
-    ):
-        # Convert to check what the effective integer step would be
-        int_step = int(step)
-        if int_step == 0:
-            raise RuntimeError("step must be nonzero")
     if dtype is torch.int64:
+        start = int(start)
+        end = int(end)
+        step = int(step)
+        if step == 0:
+            raise RuntimeError("step must be nonzero")
         sgn = (step > 0) - (step < 0)
         size = (end - start + step - sgn) // step
     else:
+        if dtype is torch.int64 and (
+            isinstance(step, float)
+            or isinstance(start, float)
+            or isinstance(end, float)
+        ):
+            int_step = int(step)
+            if int_step == 0:
+                raise RuntimeError("step must be nonzero")
         size = math.ceil((end - start) / step)
     size = int(size)
 


tests/test_tensor_constructor_ops.py
```diff
@@ -326,3 +326,38 @@ def test_accuracy_one_hot():
...
+        (5, 50, 3),
+        (0, 10, 2),
+        (0.0, 5.0, 0.5),
+        (1.0, 10.0, 1.5),
+    ],
+)
+@pytest.mark.parametrize("dtype", [torch.float32, torch.float16, torch.int64])
+def test_accuracy_arange_start(start, end, step, dtype):
+    if dtype == torch.int64 and isinstance(start, float):
+        pytest.skip("int64 with float start/end/step not applicable")
```
tengqm: is skipping the case when `start` or `end` is float a good idea?
What is the native behavior?
Should we assert pytest exception here?

tests/test_tensor_constructor_ops.py
```diff
@@ -326,3 +326,38 @@ def test_accuracy_one_hot():
...
+    ref_out = torch.arange(start, end, step, dtype=dtype, device="cpu")
+    gems_assert_equal(res_out.cpu(), ref_out)
+
+
+@pytest.mark.arange
+@pytest.mark.parametrize("end", [10, 100, 1000, 5.0])
+@pytest.mark.parametrize("dtype", [torch.float32, torch.float16, torch.int64])
+def test_accuracy_arange(end, dtype):
+    if dtype == torch.int64 and isinstance(end, float):
+        pytest.skip("int64 with float end not applicable")
```
tengqm: Same as above.

tests/test_tensor_constructor_ops.py
fangfangssj: 1. Is skipping the case when start or end is float a good idea?
Not the best practice. Skipping hides the real behavior - we should explicitly verify the exception.

2. What is the native behavior?
torch.arange(0.0, 5.0, 0.5, dtype=torch.int64) → Throws RuntimeError: step must be nonzero
Reason: float step (0.5) converts to int as 0, making step=0

3. Should we assert pytest exception here?
Yes, we should use pytest.raises(RuntimeError) to verify FlagGems has the same error behavior as PyTorch.

PR #1800, opened 2026-03-02
[Advanced Compiler]Add-Upsample_nearest3d

AdvancedCompiler:
### PR Category

Operator
### Type of Change

New Feature
### Description

add Upsample_nearest3d
### Progress

### Performance

tests:
3b19acf263deede57f22a1cb3462dbe0
benchmark:
ed5cf3b88838a4ab4d70da52924d2f60


src/flag_gems/ops/upsample_nearest3d.py
```diff
@@ -0,0 +1,147 @@
...
+            C,
+            OD,
+            OH,
+            OW,
+            ID,
+            IH,
+            IW,
+            reciprocal_scale_d,
+            reciprocal_scale_h,
+            reciprocal_scale_w,
```
huangyiqun, line 145: Isn't ‘USE_INT32_IDX’ here?

tests/test_special_ops.py
```diff
@@ -847,6 +848,22 @@ def test_upsample_nearest2d(dtype, shape, scale):
...
 
+@pytest.mark.upsample_nearest3d
+@pytest.mark.parametrize(
+    "scale", [(2, 2, 2), (1.5, 2.1, 3.7), (0.5, 0.5, 0.5), (0.3, 1.3, 0.7)]
+)
+@pytest.mark.parametrize("shape", UPSAMPLE_SHAPES_3D)
+@pytest.mark.parametrize("dtype", FLOAT_DTYPES)
+def test_upsample_nearest3d(dtype, shape, scale):
+    input = torch.randn(shape, dtype=dtype, device=flag_gems.device)
+    ref_i = to_reference(input).to(torch.float32)
```
huangyiqun, line 859: Why ref_i's dtype = float32?

src/flag_gems/ops/upsample_nearest3d.py
henghengxiedaima: Thanks for your comment! USE_INT32_IDX is not passed manually here because it's automatically derived by the heuristic function upsample_nearest3d_USE_INT32_IDX (defined in the runtime heuristic config for upsample_nearest3d)

tests/test_special_ops.py
henghengxiedaima: Hi, the reason I added .to(torch.float32) for the 3D test is that I followed the same pattern as our 1D and 2D upsampling test cases. I have tested the scenario without converting to float32, and the tests still pass— it seems that different data types are also supported natively. Could we modify this part to remove the float32 conversion?

huangyiqun: Sounds good. Since the tests pass without it, please go ahead and remove it.

PR #1802, opened 2026-03-02
[Fix] pad reflect/circular mode assertion uses wrong dimension indices

fangfangssj:
### PR Category
Operator

### Type of Change
Bug Fix

### Description
Fix pad reflect/circular mode assertion using wrong dimension indices.

The `pad()` function in `flag_gems/ops/pad.py` incorrectly halved `ndim` (`ndim //= 2`) and used wrong dimension indices (`self.shape[ndim - (2*i+1) - 1]`) when validating padding sizes for reflect and circular modes. For a 4D tensor `[1, 64, 112, 112]` with `pad=(1,1,1,1)`, this caused it to check `self.shape[0]=1` instead of `self.shape[3]=112`, rejecting valid padding with an AssertionError.

The fix uses `pad_pairs = len(pad) // 2` for iteration count and `self.shape[ndim - 1 - i]` for dimension indexing, matching PyTorch native last-dimension-first padding semantics.

### Issue
- N/A (discovered during model testing with resnetblur50)

### Progress

### Performance
N/A (bug fix in input validation logic only, no kernel changes)

### DIFF
--- a/src/flag_gems/ops/pad.py
+++ b/src/flag_gems/ops/pad.py
@@ -158,6 +158,11 @@ def generate_destination_passing_padding_wrapper(
 
         code.newline()
 
+        code.writeline("# Check which dimensions have padding")
+        for i in range(rank):
+            code.writeline(
+                f"dim{i}_has_pad = pad_before[{i}] > 0 or pad_after[{i}] > 0"
+            )
         code.writeline("IS_CONSTANT = mode == 'constant'")
         code.writeline("IS_REFLECT = mode == 'reflect'")
         code.writeline("IS_REPLICATE = mode == 'replicate'")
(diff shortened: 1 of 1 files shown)


tests/test_special_ops.py
```diff
@@ -781,6 +781,41 @@ def test_pad(shape, dtype, pad_mode, contiguous):
...
+        ([1, 64, 112, 112], [1, 1, 1, 1]),
+        # 4D tensor with larger padding
+        ([2, 32, 56, 56], [2, 2, 2, 2]),
+        # 3D tensor with reflect pad on last dim
+        ([4, 64, 128], [3, 3]),
+    ],
+)
+@pytest.mark.parametrize("dtype", [torch.float32, torch.float16])
+@pytest.mark.parametrize("pad_mode", ["reflect", "replicate", "circular"])
+def test_pad_high_dim_non_constant(shape, pad_params, dtype, pad_mode):
```
huangyiqun: Could merge this test case into the existing tests?

fangfangssj: done

PR #2068, opened 2026-03-23
[CI] Add competition CI pipeline with trusted harness and automated score upload

douxetpur:
### PR Category
CI/CD

### Type of Change
New Feature

### Description

This PR implements an automated CI/CD scoring system for the **FlagGems Operator Development Competition** with the following key features:

**1. Authoritative Trusted Harness Mechanism**
- CI exports `competition/`, `tests/`, and `benchmark/` directories from the base branch as authoritative test code
- Prevents participants from tampering with test/scoring logic through PR modifications
- Ensures fairness and security in evaluation

**2. Multi-Task Support**
- Single PR can submit multiple competition tasks simultaneously
- Task resolution supports: explicit specification, PR title matching, and changed files matching
- Execution flow: correctness tests → performance tests → scoring → upload

**3. Automated Scoring and Upload**
- Performance score calculated using `latency_base` baseline (no external baseline file dependency)
- Secure upload to FlagOS platform using signature mechanism
- API credentials managed via GitHub Actions Secrets

**Pending Validation**:
- [ ] End-to-end CI execution verification
- [ ] Trusted harness effectiveness confirmation (check "Authoritative competition ref" in logs)
- [ ] Score upload success verification (requires GitHub Secrets configuration)
- [ ] Anti-tampering mechanism testing (observe CI behavior after modifying tasks.yaml)
 
**Security**:
- Trusted harness exported from base branch, PRs cannot tamper with evaluation logic
- API credentials injected via GitHub Actions Secrets
- Upload requests use signature + timestamp to prevent forgery/replay attacks

### DIFF
--- a/tools/setup_vendor.sh
+++ b/tools/setup_vendor.sh
@@ -118,14 +118,18 @@ case $VENDOR in
         "torch==2.7.1+musa.4.0.0" \
         "torch_musa==2.7.1" \
         "numpy==1.26.4" \
-        "flagtree==0.5.0+mthreads3.1" \
         "mkl==2024.0.0"
 
     # Replace flagtree with Triton if requested
     if [ -n "${USE_TRITON}" ]; then
       uv pip uninstall flagtree
+      uv pip uninstall triton
       uv pip install --index $FLAGOS_PYPI \
         "triton==3.1.0+musa1.4.6"
+    else
+      uv pip uninstall triton
+      uv pip install --index $FLAGOS_PYPI \
+        "flagtree==0.5.0+mthreads3.1"
     fi
     ;;
 


benchmark/conftest.py
```diff
@@ -175,7 +175,11 @@ def pytest_configure(config):
     Config.record_log = config.getoption("--record") == "log"
     if Config.record_log:
         cmd_args = [
-            arg.replace(".py", "").replace("=", "_").replace("/", "_")
+            arg.replace(".py", "")
+            .replace("=", "_")
+            .replace("/", "_")
```
tengqm: This is a little bit dangerous

benchmark/performance_utils.py
```diff
@@ -71,7 +71,7 @@ class Benchmark:
     DEFAULT_DTYPES = FLOAT_DTYPES
     DEFAULT_SHAPES = DEFAULT_SHAPES
     DEFAULT_SHAPE_DESC = "M, N"
-    DEFAULT_SHAPE_FILES = "core_shapes.yaml"
+    DEFAULT_SHAPE_FILES = os.path.join(os.path.dirname(__file__), "core_shapes.yaml")
```
tengqm: This doesn't smell like belonging to this PR?

benchmark/test_competition_perf.py
```diff
@@ -0,0 +1,729 @@
+"""
+Competition Performance Benchmark Tests
+
+为 tasks.yaml 中定义的 20 个竞赛算子提供性能测试。
```
tengqm: please write comments in English.

competition/calculate_competition_score.py
```diff
@@ -0,0 +1,169 @@
+#!/usr/bin/env python3
```
tengqm: I'd suggest moving the whole `competition` directory into `tools/`.

competition/calculate_competition_score.py
```diff
@@ -0,0 +1,169 @@
...
+        geometric_mean = 0.0
+        normalized_score = 0.0
+
+    return {
+        "total_score": round(normalized_score, 2),
+        "geometric_mean_speedup": round(geometric_mean, 4),
+        "num_tests": len(speedups),
+        "num_failed": len(benchmark_results) - len(speedups),
+        "details": details,
+    }
```
tengqm: We may want to generalize these utils for benchmarking other operators as well.
In other words, these utils are applicable to non-competition operators.

competition/calculate_competition_score.py
douxetpur: ok

benchmark/performance_utils.py
douxetpur: Good point — this change is not really in scope for this PR and it adds unnecessary noise / potential behavior changes. I’ll revert this part from the current PR, and if it’s still needed I’ll follow up with a separate PR. Thanks for the catch.

benchmark/conftest.py
douxetpur: Thanks for the callout. I agree it’s risky to derive log filenames directly from raw pytest invocation args. To avoid touching the shared benchmark/test infra too much, I addressed this in the competition runner script instead.

competition/calculate_competition_score.py
douxetpur: Agree — the parsing/scoring utilities in competition/calculate_competition_score.py are generally useful beyond the competition pipeline. For this PR I’d like to keep the scope limited to the competition CI integration, but I’m happy to follow up with a refactor to extract the reusable pieces into a more generic benchmarking utility module (e.g. a shared log parser + score/summary calculator), and keep competition/* as a thin wrapper around it. I’ll track this as a separate follow-up PR to avoid mixing concerns here.

.github/workflows/unittest.yaml
```diff
@@ -5,7 +5,6 @@ on:
 
   push:
     branches: [ "master" ]
-    types: [opened, synchronize, reopened, labeled, unlabeled]
```
tengqm, line 10: Don't touch this types line unless there is a good reason.

tests/test_competition_ops.py
```diff
@@ -0,0 +1,540 @@
...
+SHAPES_2D = [(64, 64), (256, 256), (1024, 1024)]
+SHAPES_GENERAL = [(1024, 1024), (20, 320, 15), (16, 128, 64, 60)]
+
+
+# ============================================================
+# 1. log10
+# ============================================================
+
+
+@pytest.mark.log10
```
tengqm: Please avoid using the operator marker if possible.
There is a high possibility that this conflicts with the existing operator tests for formal ones.

.github/workflows/unittest.yaml
douxetpur: I removed the on: push trigger because this workflow is label-driven, and **a push event does not have PR label context.**
According to GitHub Actions docs, label-based triggering should use the labeled/unlabeled activity types on pull_request (or pull_request_target) events, not push (see “Events that trigger workflows”: https://docs.github.com/en/actions/using-workflows/events-that-trigger-workflows).
So, removing push is intentional to avoid a trigger that cannot support the label-based control flow this workflow relies on.

tests/test_competition_ops.py
douxetpur: Thanks for the suggestion! You’re right that using operator-name pytest markers (e.g. `@pytest.mark.log10`) can easily conflict with existing/formal operator tests and make `-m <op>` selection ambiguous. I’ll update the tests accordingly to avoid operator markers (switch to a dedicated non-conflicting marker / or remove such markers) to keep test selection and CI behavior deterministic.

PR #2072, opened 2026-03-24
[KernelGen] Add fused moe_align_block_size on Ascend.

ftgreat:
### PR Category
Operator | OP Test

### Type of Change
New Feature

### Description
Add fused moe_align_block_size

### Issue
None

### Progress

### Performance
img_v3_02103_385c4b01-7cfc-4f9f-a599-32791e09742g

### DIFF
--- a/benchmark/test_special_perf.py
+++ b/benchmark/test_special_perf.py
@@ -755,8 +755,81 @@ except ImportError:
     HAS_VLLM = False
 
 
+def torch_moe_align_block_size(
+    topk_ids,
+    num_experts,
+    block_size,
+    sorted_token_ids,
+    experts_ids,
+    num_tokens_post_pad,
+):
+    max_num_tokens_padded = topk_ids.numel() + num_experts * (block_size - 1)
+    flattened_token_indices = torch.arange(
+        topk_ids.numel(), device=topk_ids.device, dtype=torch.int32
+    )
+    flattened_expert_ids = topk_ids.flatten()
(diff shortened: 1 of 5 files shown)


benchmark/test_special_perf.py
```diff
@@ -809,11 +877,14 @@ def set_shapes(self, shape_file_path: None):
         def set_more_shapes(self):
             return None
 
+    torch_op = (
+        torch_moe_align_block_size if vendor_name == "ascend" else vllm_ops.moe_align_block_size
```
tengqm, line 881: the `torch_moe_align_block_size` logic is implemented in the test case rather than in the operator implmentation?

ftgreat: yes, used in testing cases as reference impl.

benchmark/test_special_perf.py
```diff
@@ -755,8 +755,81 @@ def torch_op(input_tensor, output_tensor):
     HAS_VLLM = False
 
 
+def torch_moe_align_block_size(
```
0x45f, line 758: can we import `torch_moe_align_block_size` from tests/test_special_ops.py?

0x45f, line 758: and why not use vllm `moe_align_block_size`

ftgreat: no vllm moe_align_block_size available on ascend.

ftgreat: > can we import `torch_moe_align_block_size` from tests/test_special_ops.py?

If we reuse the tests, there are issues with the import chain, which cause failures.

PR #2105, opened 2026-03-26
fix: fix sum_dim_out and trunc_div type issues

physics31415926:
## Summary

- **`ops/sum.py`**: Fix `sum_dim_out` / `aten.sum.out` — resize externally-provided `out` tensor to the correct output shape before kernel dispatch, and skip the post-kernel `squeeze()` call (which incorrectly operated on an already-reduced tensor). This fixes both the `IndexError: Dimension out of range` and the `ValueError: Unbroadcastable` errors triggered by `torch.logsumexp`.
- **`ops/div.py`**: Fix `trunc_divide` for integer inputs — use explicit `if/else` instead of early-return style so Triton eliminates the `div_rz` branch at compile time for integer types. Previously Triton compiled both branches, causing `KeyError` when `libdevice.div_rz` received integer type arguments.

## Root Causes

- `aten.sum.out` (called internally by `torch.logsumexp`) passes a zero-size placeholder `out = torch.empty({0})`; FlagGems did not resize it before use.
- `libdevice.div_rz` only supports `float32`/`float64`; Triton's early-return `if` does not eliminate dead branches at compile time — explicit `if/else` is required.

## Test plan

- [ ] `torch.logsumexp` on multi-dimensional tensors no longer raises `IndexError` or `ValueError`
- [ ] `torch.div(int_tensor, int_scalar, rounding_mode='trunc')` no longer raises `KeyError` at kernel compile time


src/flag_gems/ops/div.py
```diff
@@ -65,19 +65,28 @@ def true_divide_(A, B):
 @pointwise_dynamic(promotion_methods=[(0, 1, "DEFAULT")])
 @triton.jit
 def trunc_div_func(x, y):
-    return trunc(div_rz(x, y))
+    if x.type.scalar.is_int():
+        return x // y
```
tengqm: Does it make sense to move this `is_int` testing to GPU for execution?

src/flag_gems/ops/div.py
```diff
@@ -65,19 +65,28 @@ def true_divide_(A, B):
...
+    else:
+        return trunc(div_rz(x, y))
 
 
 @pointwise_dynamic(is_tensor=[True, False], promotion_methods=[(0, 1, "DEFAULT")])
 @triton.jit
 def trunc_div_func_tensor_scalar(x, y):
-    return trunc(div_rz(x, y))
+    if x.type.scalar.is_int():
+        return x // tl.cast(y, x.dtype)
```
tengqm: Does it make sense to do the above operations on GPU?

src/flag_gems/ops/div.py
```diff
@@ -65,19 +65,28 @@ def true_divide_(A, B):
...
+    else:
+        return trunc(div_rz(x, tl.cast(y, x.dtype)))
 
 
 @pointwise_dynamic(is_tensor=[False, True], promotion_methods=[(0, 1, "DEFAULT")])
 @triton.jit
 def trunc_div_func_scalar_tensor(x, y):
-    return trunc(div_rz(x, y))
+    if y.type.scalar.is_int():
+        return tl.cast(x, y.dtype) // y
```
tengqm: Same question here.

tests/test_binary_pointwise_ops.py
```diff
@@ -686,6 +686,56 @@ def test_accuracy_trunc_divide_scalar_scalar(dtype):
         gems_assert_close(res_out, ref_out, dtype)
 
 
+@pytest.mark.div
+@pytest.mark.parametrize("shape", POINTWISE_SHAPES)
+@pytest.mark.parametrize("dtype", INT_DTYPES + [torch.int64])
+def test_accuracy_trunc_div_int(shape, dtype):
```
tengqm: Rename this function (and the following ones)?
We don't have to stress that this is an accuracy test in function names.
When we are collecting function coverage or unit coverage, lengthy function names are annoying.

```suggestion
def test_trunc_div_int(shape, dtype):
```

src/flag_gems/ops/div.py
physics31415926: x.type.scalar.is_int() is a Triton JIT compile-time predicate, not a GPU runtime check. The if/else ensures dead-branch elimination works correctly — with early-return style, Triton compiles both branches and div_rz fails on integer types.

src/flag_gems/ops/div.py
physics31415926: Same as above

src/flag_gems/ops/div.py
physics31415926: Same as above

src/flag_gems/ops/div.py
physics31415926: Same as above

src/flag_gems/ops/div.py
physics31415926: Same as above

src/flag_gems/ops/div.py
physics31415926: Same as above

src/flag_gems/ops/div.py
tengqm: I mean do we lose anything if we move this check to somewhere before line 95?

physics31415926: Good point. We can move the integer check to the Python dispatch layer in trunc_divide() using A.is_floating_point(), and dispatch integer inputs directly to torch.div(..., rounding_mode='trunc') without going through a Triton kernel. This would simplify the kernel and avoid the compile-time branching entirely. I'll update the PR.

tengqm (approved):

PR #2187, opened 2026-03-30
fix: clamp start/end in slice_backward to prevent out-of-bounds kernel write

physics31415926:
## Problem
When `end > dim_size`, `slice_len` was computed without clamping, causing
`dim_index = start + slice_idx * step` to exceed `dim_size`. The kernel then
wrote to an out-of-bounds address, triggering `cudaErrorIllegalAddress`.

PyTorch allows `end` to exceed the dimension size (it clamps silently), but
the FlagGems kernel did not handle this case.

## Fix
Clamp `start` and `end` to `[0, dim_size`] before computing `slice_len`,
matching PyTorch's behavior.

## Test
Added `test_slice_backward_oob_end` to reproduce the out-of-bounds case.

### DIFF
--- a/src/flag_gems/ops/slice_backward.py
+++ b/src/flag_gems/ops/slice_backward.py
@@ -52,6 +52,9 @@ def slice_backward(
 
     shape = list(input_sizes)
 
+    if dim < 0:
+        dim += len(shape)
+
     outer = 1
     for i in range(dim):
         outer *= shape[i]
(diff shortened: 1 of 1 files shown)


src/flag_gems/ops/slice_backward.py
```diff
@@ -64,6 +62,10 @@ def slice_backward(
 
     dim_size = shape[dim]
 
+    actual_start = max(0, min(start, dim_size))
+    slice_len = grad_output.shape[dim]
+    start = actual_start
+
```
tengqm:
```suggestion
    slice_len = grad_output.shape[dim]
    start = max(0, min(start, dim_size))

```

tengqm (commented):
picky nits.
otherwise lgtm.

heavyrain-lzy:
@0x45f Please help to review~

PR #2197, opened 2026-03-31
[kunlunxin] fix index put

up-to-star:
## Problem
`index_put` on KunlunXin backend crashes with `RuntimeError` when indices
contain a mix of `None` and tensor values (e.g. `indices=[None, tensor]`
or `[tensor, None, None, tensor]`).

Root cause:
1. `target_shape` computation ignored dimensions corresponding to `None`
   positions, producing a shape too short for `broadcast_to`.
2. Missing input permutation logic — the Triton kernel assumes tensor-indexed
   dimensions come first, but no transpose was performed when tensor indices
   were non-contiguous or preceded by `None`.

## Fix
Ported the transpose/permutation logic from the default `index_put`
implementation:
- Pad indices with `None` to match input dimensions
- Broadcast tensor indices via `torch.broadcast_tensors`
- Permute input so tensor-indexed dimensions come first when needed
- Compute `target_shape = broadcast_shape + slice_shape` correctly
- Use contiguous tensors for the kernel to avoid stride issues on XPU

## Test
All 36 previously failing `test_index_put_mixed_none_and_tensor` cases
now pass (80/80 total).

### DIFF
--- a/src/flag_gems/runtime/backend/_kunlunxin/ops/index_put.py
+++ b/src/flag_gems/runtime/backend/_kunlunxin/ops/index_put.py
@@ -320,21 +320,72 @@ def index_put(inp, indices, values, accumulate=False):
         for index in indices
     ]
 
-    target_shape = get_max_rank_shape(indices)
-    broadcast_indices(indices, target_shape)
-    target_shape += inp.shape[len(indices) :]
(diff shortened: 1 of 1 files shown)


up-to-star:
close

PR #2226, opened 2026-04-03
[Bug Fix] Fix_muti_backend_bugs

Galaxy1458:
### Description
change to (https://github.com/flagos-ai/FlagGems/pull/2699)

### Progress

### DIFF
--- a/src/flag_gems/runtime/backend/__init__.py
+++ b/src/flag_gems/runtime/backend/__init__.py
@@ -156,8 +156,12 @@ def _import_module_safe(module_name, vendor_name, module_type):
 def import_vendor_extra_lib(vendor_name=None):
     if _state.vendor_extra_lib_imported:
         return
-    _state.ops_module = _import_module_safe(f"_{vendor_name}.ops", vendor_name, "common")
-    _state.fused_module = _import_module_safe(f"_{vendor_name}.fused", vendor_name, "fused")
+    _state.ops_module = _import_module_safe(
+        f"_{vendor_name}.ops", vendor_name, "common"
+    )
+    _state.fused_module = _import_module_safe(
+        f"_{vendor_name}.fused", vendor_name, "fused"
+    )
     _state.vendor_extra_lib_imported = True
 
 
--- a/src/flag_gems/runtime/backend/__init__.py
+++ b/src/flag_gems/runtime/backend/__init__.py
@@ -129,8 +129,8 @@ class BackendArchEvent:
     def get_arch_ops(self):
         arch_specialized_ops = []
         sys.path.append(self.current_arch_path)
+        ops_module = getattr(self.arch_module, "ops", None)
         try:
-            ops_module = getattr(self.arch_module, "ops", None)
             if ops_module is None:
                 ops_module = importlib.import_module(f"{self.arch}.ops")
             arch_specialized_ops.extend(self.get_functions_from_module(ops_module))
@@ -145,12 +145,12 @@ def _import_module_safe(module_name, vendor_name, module_type):
         return importlib.import_module(module_name)
     except ModuleNotFoundError:
         print(
-            f"[Note] No specialized {module_type} operators were found in "
-            f"the {vendor_name} implementation, and general {module_type} operators are used by default."
+            f"[Note] No specialized {module_type} operators were found for "
+            f"the {vendor_name}, generic {module_type} operators will be used by default."
         )
         return None
     except Exception as e:
-        raise RuntimeError(f"Import vendor extra lib failed: {e}")
+        raise RuntimeError(f"Failed to import vendor extra lib: {e}")
 
 
 def import_vendor_extra_lib(vendor_name=None):
--- a/src/flag_gems/runtime/backend/__init__.py
+++ b/src/flag_gems/runtime/backend/__init__.py
@@ -9,17 +9,37 @@ from pathlib import Path
 from ..common import vendors
 from . import backend_utils
 
-vendor_module = None
-device_name = None
-torch_device_object = None
-torch_device_fn_device = None
-tl_extra_backend_module = None
-ops_module = None
-fused_module = None
-heuristic_config_module = None
-vendor_extra_lib_imported = False
-device_fn_cache = {}
-customized_ops = None
+
+class BackendState:
+    """Singleton class to manage backend state variables."""
+
+    _instance = None
+
+    def __new__(cls):
+        if cls._instance is None:
+            cls._instance = super().__new__(cls)
+            cls._instance._initialized = False
+        return cls._instance
+
+    def __init__(self):
+        if self._initialized:
+            return
+        self._initialized = True
+        self.vendor_module = None
+        self.device_name = None
+        self.torch_device_object = None
+        self.torch_device_fn_device = None
+        self.tl_extra_backend_module = None
+        self.ops_module = None
+        self.fused_module = None
+        self.heuristic_config_module = None
+        self.vendor_extra_lib_imported = False
+        self.device_fn_cache = {}
+        self.customized_ops = None
+
+
+# Global singleton instance
+_state = BackendState()
 
 
 class BackendArchEvent:
@@ -51,32 +71,29 @@ class BackendArchEvent:
         return inspect.getmembers(module, inspect.isfunction) if module else []
 
     def get_heuristics_configs(self):
-        heuristic_module = None
         try:
             heuristic_module = self.arch_module
-        except Exception:  # noqa E722
+        except Exception:
             sys.path.insert(0, str(self.current_arch_path))
             heuristic_module = importlib.import_module("heuristics_config_utils")
             sys.path.remove(str(self.current_arch_path))
-        if hasattr(heuristic_module, "HEURISTICS_CONFIGS"):
-            return heuristic_module.HEURISTICS_CONFIGS
-        return None
+        return getattr(heuristic_module, "HEURISTICS_CONFIGS", None)
 
     def get_autotune_configs(self):
         path = self.current_arch_path
         return backend_utils.get_tune_config(file_path=path)
 
     def get_arch(self, device=0):
-        if not hasattr(vendor_module, "ARCH_MAP"):
+        if not hasattr(_state.vendor_module, "ARCH_MAP"):
             return
-        arch_map = vendor_module.ARCH_MAP
+        arch_map = _state.vendor_module.ARCH_MAP
         arch_string = os.environ.get("ARCH", "")
         arch_string_num = arch_string.split("_")[-1][0] if arch_string else arch_string
         if not arch_string_num:
             try:
-                if not torch_device_object.is_available():
+                if not _state.torch_device_object.is_available():
                     return False
-                props = torch_device_object.get_device_properties(device)
+                props = _state.torch_device_object.get_device_properties(device)
                 arch_string_num = str(props.major)
             except Exception:
                 self.has_arch = False
@@ -89,16 +106,14 @@ class BackendArchEvent:
             return arch_map[arch_string_num]
 
     def _get_supported_archs(self, path=None):
-        path = path or vendor_module.__path__[0]
-        excluded = ("ops", "fused")
-        path = Path(path)
+        path = Path(path or _state.vendor_module.__path__[0])
         path = path.parent if path.is_file() else path
-        archs = {}
-        for p in path.iterdir():
-            name = str(p).split("/")[-1]
-            if p.is_dir() and name not in excluded and not name.startswith("_"):
-                archs.update({name: str(p)})
-        return archs
+        excluded = ("ops", "fused")
+        return {
+            p.name: str(p)
+            for p in path.iterdir()
+            if p.is_dir() and p.name not in excluded and not p.name.startswith("_")
+        }
 
     def get_supported_archs(self):
         return list(self.supported_archs.keys())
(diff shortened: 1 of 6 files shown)


src/flag_gems/runtime/backend/__init__.py
```diff
@@ -9,17 +9,37 @@
...
+        self.tl_extra_backend_module = None
+        self.ops_module = None
+        self.fused_module = None
+        self.heuristic_config_module = None
+        self.vendor_extra_lib_imported = False
+        self.device_fn_cache = {}
+        self.customized_ops = None
+
+
+# Global singleton instance
```
tengqm, line 41: 💯

src/flag_gems/runtime/backend/__init__.py
```diff
@@ -51,32 +71,29 @@ def get_functions_from_module(self, module):
...
             heuristic_module = self.arch_module
-        except Exception:  # noqa E722
+        except Exception:
             sys.path.insert(0, str(self.current_arch_path))
             heuristic_module = importlib.import_module("heuristics_config_utils")
             sys.path.remove(str(self.current_arch_path))
-        if hasattr(heuristic_module, "HEURISTICS_CONFIGS"):
-            return heuristic_module.HEURISTICS_CONFIGS
-        return None
+        return getattr(heuristic_module, "HEURISTICS_CONFIGS", None)
```
tengqm, line 80: 👍

src/flag_gems/runtime/backend/__init__.py
```diff
@@ -113,51 +128,37 @@ def get_arch_module(self):
...
-        ops_module = importlib.import_module(f"{self.arch}.ops")
         try:
-            ops_module = self.arch_module.ops
-            modules.append(ops_module)
-        except Exception:
-            try:
-                sys.path.append(self.current_arch_path)
+            ops_module = getattr(self.arch_module, "ops", None)
+            if ops_module is None:
                 ops_module = importlib.import_module(f"{self.arch}.ops")
```
tengqm, line 135: Please try shrink the try scope to this statement?

src/flag_gems/runtime/backend/__init__.py
```diff
@@ -113,51 +128,37 @@ def get_arch_module(self):
...
+    """Helper to import a module with proper error handling."""
     try:
-        ops_module = importlib.import_module(f"_{vendor_name}.ops")
+        return importlib.import_module(module_name)
     except ModuleNotFoundError:
         print(
-            f"[Note] No specialized common operators were found in"
-            f"the {vendor_name} implementation, and general common operators are used by default."
+            f"[Note] No specialized {module_type} operators were found in "
+            f"the {vendor_name} implementation, and general {module_type} operators are used by default."
```
tengqm:
```suggestion
            f"[Note] No specialized {module_type} operators were found for "
            f"the {vendor_name}, generic {module_type} operators will be used by default."
```

src/flag_gems/runtime/backend/__init__.py
```diff
@@ -113,51 +128,37 @@ def get_arch_module(self):
...
-    if vendor_extra_lib_imported is True:
-        return
-    global ops_module, fused_module
+def _import_module_safe(module_name, vendor_name, module_type):
+    """Helper to import a module with proper error handling."""
     try:
-        ops_module = importlib.import_module(f"_{vendor_name}.ops")
+        return importlib.import_module(module_name)
     except ModuleNotFoundError:
         print(
```
tengqm, line 147: Consider change this to logger.info ?

src/flag_gems/runtime/backend/__init__.py
```diff
@@ -113,51 +128,37 @@ def get_arch_module(self):
...
-            f"[Note] No specialized common operators were found in"
-            f"the {vendor_name} implementation, and general common operators are used by default."
+            f"[Note] No specialized {module_type} operators were found in "
+            f"the {vendor_name} implementation, and general {module_type} operators are used by default."
         )
+        return None
     except Exception as e:
         raise RuntimeError(f"Import vendor extra lib failed: {e}")
```
tengqm:
```suggestion
        raise RuntimeError(f"Failed to import vendor extra lib: {e}")
```

src/flag_gems/runtime/backend/__init__.py
```diff
@@ -182,43 +182,40 @@ def gen_torch_tensor_attr_res(tensor, attr_name):
...
 def set_torch_backend_device_fn(vendor_name=None):
-    global device_name, torch_device_fn_device
-    device_name = device_name or get_vendor_info(vendor_name).device_name
-    module_str = f"torch.backends.{device_name}"
-    if device_name in ("musa", "aipu", "npu", "txda", "ptpu", "gcu"):
-        torch_device_fn_device = None
+    _state.device_name = _state.device_name or get_vendor_info(vendor_name).device_name
+    module_str = f"torch.backends.{_state.device_name}"
+    if _state.device_name in ("musa", "aipu", "npu", "txda", "ptpu", "gcu"):
+        _state.torch_device_fn_device = None
```
tengqm, line 204: I don't like this logic ... it is hard to maintain.
Please consider using a `try ... except` struct here...

src/flag_gems/runtime/backend/__init__.py
```diff
@@ -233,73 +230,67 @@ def get_module(vendor_name):
     ):  # The purpose of a query is to provide the user with the instance that he wants to import
         return get_module(vendor_name)
 
-    global vendor_module
-    if vendor_module is None:
-        vendor_module = get_module("_" + vendor_name)
-    return vendor_module
+    if _state.vendor_module is None:
+        _state.vendor_module = get_module("_" + vendor_name)
```
tengqm, line 238: this `get_module` is exception free?

Galaxy1458: Yes, in principle, there is no exception. If there is an exception for unknown reasons, it must be reported.

src/flag_gems/runtime/backend/__init__.py
Galaxy1458: using a `try ... except` struct here doesn‘t conform to the original intention of the code. `torch_device_fn_device`  is not necessarily non-existent. It could be that the vendor doesn't want to use it, or it might not be needed at all

src/flag_gems/runtime/backend/__init__.py
Galaxy1458: If `except` is entered, this [note] need to be prompted to the user in any case.

src/flag_gems/runtime/backend/__init__.py
Galaxy1458: Done.

src/flag_gems/runtime/backend/__init__.py
Galaxy1458: Done

src/flag_gems/runtime/backend/__init__.py
tengqm: I mean 'except: pass'.
Don't enumerate the device name here ...
no one can remember that we planted some hardcoded device specific code here.

src/flag_gems/runtime/backend/__init__.py
tengqm: there are in general two approaches ...
either you catch all exceptions in the `get_module` so the function never throws an exception, and you may get an empty string here (check it);
or you allow it to throw an exception and you will catch that exception in the call chain.

src/flag_gems/runtime/backend/__init__.py
tengqm: Even in that case, the error message are not supposed to be printed on stdout.
This module is in the core of the project, and the whole project can be scripted.
I have noticed this unconditional print when writing test scripts.

By the way, it should be fine (not good though ...) if a vendor has no operators implemented.

src/flag_gems/runtime/backend/__init__.py
```diff
@@ -233,73 +234,67 @@ def get_module(vendor_name):
...
-        return heuristic_config_module.HEURISTICS_CONFIGS
-    return None
+    config_name = "heuristics_config_utils"
+    default_backend = "nvidia"
+    for backend in (vendor_name, default_backend):
+        mod_name = f"_{backend}.{config_name}"
+        try:
+            _state.heuristic_config_module = importlib.import_module(mod_name)
+        except Exception:
+            continue
```
tengqm, line 286: should be 'pass'?
```suggestion
            pass
```

src/flag_gems/runtime/backend/device.py
```diff
@@ -1,148 +1,104 @@
 import os
```
tengqm, line 1: File mode changed from 644 to 755, pls fix.

src/flag_gems/runtime/backend/device.py
```diff
@@ -1,148 +1,104 @@
...
-
-# A singleton class to manage device context.
-class DeviceDetector(object):
+from ..common import (
+    _VENDOR_TORCH_ATTR,
+    UNSUPPORT_BF16,
+    UNSUPPORT_FP64,
+    UNSUPPORT_INT64,
+    vendors,
+)
```
tengqm, line 15: cool. This should be extracted.

src/flag_gems/runtime/backend/device.py
```diff
@@ -1,148 +1,104 @@
...
-                self.vendor_name
-            ).device_count()
-            self.support_fp64 = self.vendor not in UNSUPPORT_FP64
-            self.support_bf16 = self.vendor not in UNSUPPORT_BF16
-            self.support_int64 = self.vendor not in UNSUPPORT_INT64
-
-    def get_vendor(self, vendor_name=None) -> tuple:
-        # Try to get the vendor name from a quick special command like 'torch.mlu'.
+        if hasattr(self, "initialized"):
+            return
```
tengqm, line 30: nit. personally I prefer your previous implementation that make the instance always have an `initialized` attribute. We only test if that attribute is true or not, rather than whether the instance has a specific attribute.

src/flag_gems/runtime/backend/device.py
```diff
@@ -1,148 +1,104 @@
...
+        self.initialized = True
+        self.vendor_list = vendors.get_all_vendors().keys()
+        self.info = self.get_vendor(vendor_name)
+        self.vendor_name = self.info.vendor_name
+        self.name = self.info.device_name
+        self.vendor = vendors.get_all_vendors()[self.vendor_name]
```
tengqm, line 36: always be careful when using `[]` ... it is disgusting that it may throw exceptions, use `get(key, default)` whenever possible.

src/flag_gems/runtime/backend/device.py
```diff
@@ -1,148 +1,104 @@
...
+        self.dispatch_key = self.info.dispatch_key or self.name.upper()
+        self.device_count = backend.gen_torch_device_object(
+            self.vendor_name
+        ).device_count()
+        self.support_fp64 = self.vendor not in UNSUPPORT_FP64
```
tengqm, line 41: sigh ... this feature should be pushed down to vendor layers ... let's leave it to a future PR

src/flag_gems/runtime/backend/device.py
```diff
@@ -1,148 +1,104 @@
...
-            "sunrise": "ptpu",
-            "enflame": "gcu",
-        }
-        for vendor_name, flag in cmd.items():
-            if hasattr(torch, flag):
+        for vendor_name, attr in _VENDOR_TORCH_ATTR.items():
+            if hasattr(torch, attr):
                 return vendor_name
         try:
             import torch_npu
```
tengqm, line 65: This damn import is still here ...

src/flag_gems/runtime/backend/device.py
```diff
@@ -1,148 +1,104 @@
...
-            thread.join()
-        if result_single_info.empty():
-            error.device_not_found()
-        else:
-            return result_single_info.get()
+                return info if result.returncode == 0 else None
+            except Exception:
+                return None
+
+        with ThreadPoolExecutor() as executor:
```
tengqm, line 89: :) Cannot imagine that we are using multiple threads to do this.

src/flag_gems/runtime/common.py
```diff
@@ -22,3 +22,45 @@ def get_all_vendors(cls) -> dict:
...
+    {
+        vendors.CAMBRICON,
+        vendors.ILUVATAR,
+        vendors.KUNLUNXIN,
+        vendors.MTHREADS,
+        vendors.AIPU,
+        vendors.ASCEND,
+        vendors.TSINGMICRO,
+        vendors.SUNRISE,
+        vendors.ENFLAME,
```
tengqm, line 37: mind sort the set by names?

src/flag_gems/runtime/common.py
```diff
@@ -22,3 +22,45 @@ def get_all_vendors(cls) -> dict:
...
+        vendors.SUNRISE,
+    }
+)
+
+UNSUPPORT_INT64 = frozenset(
+    {
+        vendors.AIPU,
+        vendors.TSINGMICRO,
+        vendors.SUNRISE,
+        vendors.ENFLAME,
```
tengqm, line 53: same here

src/flag_gems/runtime/common.py
```diff
@@ -22,3 +22,45 @@ def get_all_vendors(cls) -> dict:
...
+
+
+# Mapping from vendor name to torch attribute for quick detection
+_VENDOR_TORCH_ATTR = {
+    "cambricon": "mlu",
+    "mthreads": "musa",
+    "iluvatar": "corex",
+    "ascend": "npu",
+    "sunrise": "ptpu",
+    "enflame": "gcu",
```
tengqm, line 65: sort this dict?

src/flag_gems/runtime/configloader.py
```diff
@@ -52,15 +43,19 @@ def __init__(self):
...
             self.load_all()
 
+    def update_config_from_arch(self):
+        try:
+            archEvent = backend.BackendArchEvent()
+            if archEvent.has_arch:
+                self.arch_specialized_yaml_config = archEvent.autotune_configs
+                self.arch_heuristics_config = archEvent.heuristics_configs
+        except Exception as err:
+            print(f"[INFO] : {err}")
```
tengqm, line 57: Same suggestion here. Log an error or an warning please.

src/flag_gems/runtime/configloader.py
```diff
@@ -172,24 +167,37 @@ def to_gen_config(self, gen_config):
...
+
+    def _create_triton_config(self, single_config, current_config):
+        """Create a triton.Config with appropriate parameters."""
+        kwargs = {
+            "num_warps": current_config["num_warps"],
+            "num_stages": current_config["num_stages"],
+            "num_ctas": current_config["num_ctas"],
+        }
+        if self.device.vendor_name == "hygon":
+            kwargs["num_ldmatrixes"] = current_config["num_ldmatrixes"]
```
tengqm, line 189: Maybe a more generic ways is something like:

```suggestion
        kwargs["num_ldmatrixes"] = current_config.get("num_ldmatrixes", None)
```

The thing is that in this core module, we don't treat any vendor as a special beast.

src/flag_gems/runtime/configloader.py
```diff
@@ -200,23 +208,5 @@ def get_tuned_config(self, op_name):
...
-                )
-            else:
-                configs.append(
-                    triton.Config(
-                        single_config["META"],
-                        num_warps=current_config["num_warps"],
-                        num_stages=current_config["num_stages"],
-                        num_ctas=current_config["num_ctas"],
-                    )
-                )
```
tengqm, line 221: Nice refactoring ...

tengqm:
@Galaxy1458 Any progress on this? This should be a high-value, high-priority one for merge.

PR #2272, opened 2026-04-08
fused/DSA: add TLE bucket_sort_topk with internal dispatch

sunnycase:
# PR Summary

## What changed

This PR introduces a TLE-based implementation for `bucket_sort_topk` and keeps a safe Triton fallback path:

1. Added a TLE kernel path in `src/flag_gems/fused/DSA/bin_topk.py`:
   - `kernel_tle_bucket_sort_topk`
   - `tle_bucket_sort_topk`
2. Kept the original Triton implementation as an explicit baseline path:
   - `bucket_sort_topk_triton`
3. Updated the public entrypoint `bucket_sort_topk(inputs, starts, ends, topk)` to dispatch internally:
   - Prefer TLE when available on CUDA.
   - Fallback to the legacy Triton kernel on runtime failure.
4. Added benchmark coverage for large sequence length and large top-k cases in `build/bench_bucket_sort_topk.py`, and used Triton's built-in benchmark API (`triton.testing.do_bench`).

## Baseline definition

In the benchmark table below:

- `GEMS (ms)` is latency from the legacy Triton path (`bucket_sort_topk_triton`).
- `TLE (ms)` is latency from the new TLE path (`tle_bucket_sort_topk`).
- `TLE Speedup = GEMS(ms) / TLE(ms)`.

### GPU: NVIDIA H800

| B | S | K | GEMS(ms) | TLE(ms) | TLE Speedup |
|---:|---:|---:|---:|---:|---:|
| 1 | 8192 | 64 | 0.043780 | 0.012175 | 3.596x |
| 4 | 8192 | 256 | 0.038136 | 0.013314 | 2.864x |
| 8 | 16384 | 512 | 0.067173 | 0.019121 | 3.513x |
| 16 | 32768 | 1024 | 0.108865 | 0.024378 | 4.466x |
| 4 | 32768 | 2048 | 0.107768 | 0.024412 | 4.415x |
| 2 | 65536 | 2048 | 0.124436 | 0.036607 | 3.399x |
| 2 | 65536 | 4096 | 0.221248 | 0.036954 | 5.987x |
| 1 | 131072 | 4096 | 0.427550 | 0.056258 | 7.600x |
| **Avg** | - | - | **0.142370** | **0.027902** | **4.480x** (gmean **4.269x**) |

## Recall results

| B | S | K | Triton recall | TLE recall |
|---:|---:|---:|---:|---:|
| 1 | 8192 | 64 | 1.000000 | 1.000000 |
| 4 | 8192 | 256 | 1.000000 | 1.000000 |
| 8 | 16384 | 512 | 0.999756 | 1.000000 |
| 16 | 32768 | 1024 | 0.778015 | 1.000000 |
| 4 | 32768 | 2048 | 0.660767 | 1.000000 |
| 2 | 65536 | 2048 | 0.851562 | 1.000000 |
| 2 | 65536 | 4096 | 0.616699 | 1.000000 |
| 1 | 131072 | 4096 | 1.000000 | 1.000000 |

## Correctness / sanity checks

- Functional quality is reported in a separate local recall table (`torch.topk` reference).
- The TLE path shows strong speedup on H800 across both standard and large-shape cases.


src/flag_gems/fused/DSA/bin_topk.py
```diff
@@ -3,21 +3,55 @@
 import triton.language as tl
 
 
+def _triton_version_at_least(major: int, minor: int, patch: int = 0) -> bool:
```
tengqm: here is what Google suggests me to parse a versing string without introducing dependency on `semver` etc ...

```
def parse_simple(v_str):
    return tuple(map(int, v_str.split('.')))

if parse_simple("1.10.2") > parse_simple("1.2.10"):
    print("1.10.2 is newer")
```

tengqm (commented):
We are not registering this to `src/flag_gems/fused/__init__.py` ?

sunnycase:
Addressed the export point: `bucket_sort_topk` is now re-exported from `src/flag_gems/fused/__init__.py`, so it is visible through the fused public surface. I only exported the public entrypoint, not the internal `bucket_sort_topk_triton` / `tle_bucket_sort_topk` helpers.

src/flag_gems/fused/DSA/bin_topk.py
sunnycase: I’m not taking the `tuple(map(int, ...))` route here. This repo already depends on `packaging`, and the simple tuple parse is brittle for Triton version strings with prerelease/dev/local suffixes (for example `3.6.0rc1` / `3.6.0.dev...`). The current helper is intentionally tolerant of those forms. If we want to clean this up further, I’d rather switch to `packaging.version.parse(...)` than adopt the tuple-based parser.

tengqm: Makes sense.
Can we move this util function (I mean the logic that exacts major, minor, patch from version string) to src/flag_gems/utils/triton_version_utils.py and see if that we can avoid duplication while share this across the project?

src/flag_gems/fused/__init__.py
```diff
@@ -42,6 +43,7 @@
     "apply_repetition_penalties",
     "apply_rotary_pos_emb",
     "bincount",
+    "bucket_sort_topk",
```
tengqm, line 46: We need test cases even for fused operators. Please help add one?

src/flag_gems/fused/DSA/bin_topk.py
sunnycase: Done in `87a817a8`. I moved the shared version handling into `src/flag_gems/utils/triton_version_utils.py` and updated both `bin_topk.py` and `moe_align_block_size.py` to use it instead of carrying local parsers.

I also changed the TLE availability helper to be version-aware per caller: `has_triton_tle(3, 6, 0)` is used here, so the utility no longer bakes in a global `3.1` threshold. The version comparison uses `packaging.version.Version` with a numeric fallback for non-PEP440 strings.

src/flag_gems/fused/__init__.py
sunnycase: Done in `87a817a8`. I added `test_bucket_sort_topk_public_entrypoint_matches_torch_topk` in `tests/test_DSA/test_bin_topk.py`, covering the public `bucket_sort_topk` entrypoint against `torch.topk` for full-length and variable-length ranges.

The test is guarded with CUDA/TLE availability checks so unsupported environments skip it instead of failing during collection or launch.

tengqm (approved):

PR #2331, opened 2026-04-10
[kunlunxin]: fix addmm compute error

up-to-star:
### Problem

The `addmm` kernel for KunlunXin backend was missing boundary masks in
`tl.load` for matrices `a` and `b`. This caused out-of-bounds memory reads
when input shapes are not aligned to block sizes (e.g. `M=15`, `N=160`),
leading to garbage values being accumulated in `tl.dot` and producing
completely wrong results (~99% elements mismatched).

### Root Cause

```python
# Before: no mask, OOB memory read corrupts dot product
a = tl.load(a_ptrs)
b = tl.load(b_ptrs)


tengqm (approved):
Kicking this in.
We will do validation later...

PR #2344, opened 2026-04-10
[MTHREADS] Add w8a8_block_fp8_matmul for mthreads backend

Oslomayor:
Add w8a8_block_fp8_matmul op.


tengqm (commented):
Let's kick this in ...
we'll find out how many operators are actually working pretty soon.

tengqm (approved):

PR #2350, opened 2026-04-10
Feature/combination Add combination test

lovywu:
### PR Category
OP Test

### Type of Change
New Feature

### Description
Add combination test


tengqm (commented):
Thanks for this proposal. I'd suggest a step-by-step way to get this in.

1. Select one of the test cases as the first step, try run it as a normal FlagGems unit test by scheduling it to a Nvidia backend.
2. Reuse/adapt the existing workflow to drive this test as if it is a collection.
3. Add more models/test cases to this new flavor of tests.

Along the way, we will get familiar with what are actually existing features and what are new, valuable new testing scenarios/processes.

PR #2355, opened 2026-04-11
Mask out test_upsample_linear1d due to assertion failure

tengqm:
### PR Category

OP Test

### Type of Change

The `upsample_linear1d` is not stable.

### Description

This PR skips the testing and demote the operator to `beta`.


tengqm:
Update -- keep it in the test pipeline but demote it.

PR #2433, opened 2026-04-14
add precision log

YY4994:
### PR Category
Other

### Type of Change
New Feature

### Description
Add precision logging to compare FlagGems outputs with PyTorch CPU implementations, logging operator name, tensor shapes, and data types when mismatches occur.

### Progress


src/flag_gems/runtime/register.py
```diff
@@ -1,5 +1,14 @@
+import functools
 import warnings
 
+import torch
+
+from ..logging_utils import (
```
tengqm:
```suggestion
from flag_gems.logging_utils import (
```

src/flag_gems/runtime/register.py
```diff
@@ -148,3 +142,181 @@ def get_vendor_name(self):
...
+    Since FlagGems replaces the CUDA dispatch, the native implementation cannot be called on GPU,
+    so inputs are copied to CPU to compute the reference result. Performance overhead is controlled by:
+    - max_checks: only check the first N calls per operator (default 10)
+    - skip large tensors (over 1M elements)
+    - once a failure is logged, that operator is no longer checked
+    """
+    _call_count = 0
+
+    @functools.wraps(fn)
+    def wrapper(*args, **kwargs):
```
tengqm: I am not a big fan of this approach ... it adds overhead to all operators.

src/flag_gems/logging_utils.py
```diff
@@ -68,3 +71,102 @@ def teardown_flaggems_logging(logger: logging.Logger | None = None):
...
+    elif isinstance(fg_out, (tuple, list)) and isinstance(pt_out, (tuple, list)):
+        for i, (fg, pt) in enumerate(zip(fg_out, pt_out)):
+            ok, info = compare_outputs(fg, pt, rtol, atol)
+            if not ok:
+                info["index"] = i
+                return False, info
+    return True, {}
+
+
+def enable_precision_check(
```
tengqm, line 154: This and the above functions are suitable for performance checking ... they are not supposed to be on the normal execution paths, not even if there are only "a few" if-else checks.

w1120029931-bit (changes requested):
please modify the code according to the commit

huangyiqun:
https://github.com/flagos-ai/FlagGems/pull/3377

PR #2460, opened 2026-04-16
Added a dimension check at the top of the gather function

### LINKED ISSUE #2346: # gather 算子 input 和 index 维度不一致时崩溃
## 问题描述

当 `torch.gather(input, dim, index)` 中 `input` 是 3D tensor 而 `index` 是 2D tensor 时，FlagGems 的 gather 实现报错：

```
RuntimeError: mismatch in length of strides and shape
```

## 复现代码

```python
import torch
import flag_gems
flag_gems.enable()

x = torch.randn(2, 10, 64, device='cuda')
indices = torch.randperm(10, device='cuda').unsqueeze(0).expand(2, -1)  # shape [2, 10]
result = torch.gather(x, dim=1, index=indices)  # RuntimeError
```

## 错误位置

`src/flag_gems/ops/gather.py:190` → `src/flag_gems/utils/shape_utils.py:249`

```python
def restride_dim(src, dim, shape, step=0, storage_offset=None):
    strides = list(src.stride())  # 3个元素（3D tensor）
    strides[dim] *= step
    return src.as_strided(shape, strides, storage_offset)  # shape是2D，strides是3D → 报错
```

## 备注

按 PyTorch 文档，`torch.gather` 要求 `self` 和 `index` 的维度数必须相同。如果此为预期行为，建议添加明确的维度检查和错误提示，而非在 `as_strided` 处抛出难以理解的错误。

YY4994:
### PR Category
Operator

### Type of Change
Bug Fix

### Description
Added a dimension check at the top of the gather function

### Issue

- Resolves #2346

### Progress

### DIFF
--- a/src/flag_gems/ops/gather.py
+++ b/src/flag_gems/ops/gather.py
@@ -184,6 +184,11 @@ _gather_func = GatherFunction()
 
 def gather(inp, dim, index, out=None, sparse_grad=False):
     logger.debug("GEMS GATHER")
+    if inp.ndim != index.ndim:
+        raise IndexError(
+            f"self and index must have the same number of dimensions, "
+            f"got self.ndim = {inp.ndim} and index.ndim = {index.ndim}"
+        )
     if out is None:
         out = torch.empty_like(index, dtype=inp.dtype, device=inp.device)
     dim_stride = inp.stride(dim)
--- a/tests/test_reduction_ops.py
+++ b/tests/test_reduction_ops.py
@@ -1115,6 +1115,16 @@ def test_accuracy_trace(shape, dtype):
 @pytest.mark.parametrize("dim", [0, 1, 2])
 @pytest.mark.parametrize("dtype", FLOAT_DTYPES)
 def test_accuracy_gather(inp_shape, dim, dtype):
+    # Test ndim mismatch raises IndexError (only once to avoid redundant checks)
+    if dim == 0 and dtype == torch.float32:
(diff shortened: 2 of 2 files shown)


src/flag_gems/ops/gather.py
```diff
@@ -184,6 +184,11 @@ def arg_key(self, *args):
 
 def gather(inp, dim, index, out=None, sparse_grad=False):
     logger.debug("GEMS GATHER")
+    if inp.ndim != index.ndim:
+        raise IndexError(
+            f"self and index must have the same number of dimensions, "
+            f"got self.ndim = {inp.ndim} and index.ndim = {index.ndim}"
+        )
```
tengqm, line 191: Although this is a trivial condition check, we still expect an associated test case for this change.

YY4994: This change adds a dimension check to align with PyTorch's behavior. 
Since this is a validation check rather than core computation logic, 
I assume the correctness is already covered by existing PyTorch tests. 
However, if an explicit test case is required, I can add one.

Please let me know your preference.

tengqm: We anticipate (not strictly require) an accompany test case that `IndexError` would be thrown if dimension mismatch is the situation. It will help improve test coverage. If we don't do it today, no one would like to do it in the future... :)

tengqm (approved):

PR #2462, opened 2026-04-16
Rename skip rmsnorm test name

MARD1NO:
### Progress

### DIFF
--- a/tests/test_norm_ops.py
+++ b/tests/test_norm_ops.py
@@ -594,7 +594,7 @@ def test_accuracy_skip_layernorm(shape, dtype):
     gems_assert_close(res_out, ref_out, dtype)
 
 
-@pytest.mark.fused_add_rms_norm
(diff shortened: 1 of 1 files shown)


tests/test_norm_ops.py
```diff
@@ -594,7 +594,7 @@ def test_accuracy_skip_layernorm(shape, dtype):
     gems_assert_close(res_out, ref_out, dtype)
 
 
-@pytest.mark.fused_add_rms_norm
+@pytest.mark.skip_rms_norm
```
tengqm, line 597: The `pytest.mark` annotation is used for the team to identify the operator/kernel/function we support. For aten operators, we use the target operator name; for fused operators, we use the function name.

There is no reason to rename this to `skip_rms_norm`.
If you want to skip this particular test for a specific condition, you can do it using a `skipif()` condition.

huangyiqun:
This modification does not seem necessary.

PR #2483, opened 2026-04-16
fix the fill_scalar_ operator crashes when handling non-contiguous te…

wuhulalala:
…nsors (sliced views)

### Progress


tests/test_special_ops.py
```diff
@@ -1286,6 +1286,34 @@ def test_fill_out(value, shape, dtype):
...
 
+FILL_SLICE_CASES = [
+    # (shape, slice)
+    ((4, 128), (slice(None), slice(64, None))),
+    ((2, 1, 1, 512), (slice(None), slice(None), slice(None), slice(358, None))),
+    ((8, 32, 64), (slice(None), slice(16, None))),
+]
+
+
+@pytest.mark.fill
```
tengqm, line 1297: Please split the test case for scalar and tensor cases because those are TWO separate operators to test.

Also, please consider add test cases that cover the change you made elsewhere.

PR #2500, opened 2026-04-17
Fix moore threads setup script

tengqm:
### PR Category

CI/CD

### Type of Change

Other

### Description

This is NOT A BUG FIX.
It is weird that `uv` would uninstall `torch_musa` when installing `[mthreads]` group.
It is crazy.


tengqm:
This PR is solving the problem in the wrong direction.

PR #2523, opened 2026-04-18
[FlagGems Operator Development Competition]Add Triton gcd operator

goldenfox2025:
# [FlagGems Operator Development Competition] Add gcd operator with Triton implementation

## Summary

Implements `torch.gcd` in Triton and registers:

- `aten.gcd`
- `aten.gcd.out`

The operator is wired through:

- `src/flag_gems/ops/gcd.py`
- `src/flag_gems/ops/__init__.py`
- `src/flag_gems/__init__.py`

Correctness coverage is added in `tests/test_binary_pointwise_ops.py`, and benchmark coverage is added in:

- `benchmark/test_binary_pointwise_perf.py`
- `benchmark/core_shapes.yaml`

## Algorithm

The implementation does not fall back to PyTorch in the operator path.

For normal integer inputs, the kernel uses binary gcd (Stein's algorithm):

1. Remove common powers of two from both operands.
2. Reduce one operand to an odd integer.
3. Iteratively:
   - reduce the other operand to odd,
   - swap so the smaller odd value is first,
   - replace the larger value with the difference.
4. When the larger value becomes zero, shift the saved common power of two back.

This is more GPU-friendly than Euclid's modulo-heavy loop because the hot path mostly uses:

- `ffs/ctz`
- bit shifts
- subtraction
- comparisons

The Triton-specific structure is:

- one program handles a block of elements,
- each lane has an `active` mask,
- the loop exit is a block-level scalar condition: continue while any lane is still active.

This avoids the fixed-trip-count version and also avoids a per-element scalar kernel.

### `int16`

The normal path uses binary gcd.

`int16` has a small signed range, so the `min_int = -32768` overflow corner case is handled by a lookup table:

- if both inputs are `-32768`, result is `-32768`
- if exactly one input is `-32768`, the kernel looks up the PyTorch-equivalent result for `gcd(-32768, abs(other))`

The lookup table is built once per device and then read directly inside the Triton kernel.

### `int32` and `int64`

The normal path also uses binary gcd.

For the rare `min_int` lanes:

- `int32`: `-2147483648`
- `int64`: `-9223372036854775808`

the kernel switches to a signed Euclidean update to preserve PyTorch overflow semantics exactly.

## Accuracy Validation

Targeted operator validation:

```bash
export PATH=/usr/lib/wsl/lib:$PATH
export GEMS_VENDOR=nvidia
export PYTHONPATH=src
python -m pytest tests/test_binary_pointwise_ops.py -k 'gcd' -s --ref=cpu --mode=quick
```

Result:

```text
15 passed
```

Full binary pointwise validation:

```bash
python -m pytest -s -x tests/test_binary_pointwise_ops.py
python -m pytest -s -x tests/test_binary_pointwise_ops.py --ref=cpu --mode=quick
```

Results:

```text
7755 passed
1320 passed
```

Coverage includes:

- forward
- `out`
- zero / negative / `min_int`
- empty tensors
- non-contiguous + broadcast inputs

## Benchmark

Command:

```bash
export PATH=/usr/lib/wsl/lib:$PATH
export GEMS_VENDOR=nvidia
export PYTHONPATH=src
python -m pytest benchmark/test_binary_pointwise_perf.py -k 'gcd' -s --level core --warmup 5 --iter 10
```

Representative latest results:

| DType | Shape | PyTorch (ms) | FlagGems (ms) | Speedup |
|------|-------|--------------|---------------|---------|
| int16 | `[65536]` | 0.021504 | 0.015360 | 1.400x |
| int16 | `[1024, 1024]` | 0.093184 | 0.075776 | 1.230x |
| int16 | `[64, 256, 256]` | 0.319488 | 0.261120 | 1.224x |
| int32 | `[65536]` | 0.033792 | 0.021504 | 1.571x |
| int32 | `[1024, 1024]` | 0.154624 | 0.120832 | 1.280x |
| int32 | `[64, 256, 256]` | 0.538624 | 0.433152 | 1.243x |

## Checklist Alignment

- `ops/*.py`: added `src/flag_gems/ops/gcd.py`
- `ops/__init__.py`: import and export updated
- `src/flag_gems/__init__.py`: ATen registration updated
- correctness tests: added in `tests/test_binary_pointwise_ops.py`
- benchmark tests: added in `benchmark/test_binary_pointwise_perf.py`
- core benchmark shapes: added in `benchmark/core_shapes.yaml`
- special values / empty / non-contiguous: covered
- `pre-commit`: passed on changed files


bin913:
Please resolve cnflicts

goldenfox2025:
> Please resolve cnflicts

It's been resolved. Thanks for reminding me.

bin913 (approved):

huangyiqun (approved):

PR #2560, opened 2026-04-20
Fix fill operators crash on non-contiguous tensors (stride != 1) in hopper backend

wuhulalala:
### PR Category

Operator
### Type of Change

Bug fix
### Description

The hopper fill implementation uses flat offsets to index into tensor memory, which assumes stride=1 (contiguous layout). When the      
  tensor has non-unit strides — e.g. from slicing like x[:, ::2] — the kernel writes to wrong memory locations, producing results
  inconsistent with PyTorch's native fill_ behavior.                                                                                      
   
  The fix detects non-contiguous tensors, copies them to a contiguous buffer before running the fill kernel, then copies the result back  
  via copy_() which respects the original stride layout
### Issue
#2347 

### Progress


tengqm (commented):
Great that this passes on Hopper for real!
Huge thank you for quickly adapt the test cases.

Please note that we don't have a `fill` operator at the moment (see https://github.com/flagos-ai/FlagGems/blob/master/conf/operators.yaml#L1763-L1832). We have several variants for `fill`, each for a different scenario.

We are now marking operator tests stricter than before. Please use accurate and concise pytest mark for your test cases.

huangyiqun (dismissed):

tests/test_fill.py
```diff
@@ -83,6 +83,32 @@ def test_fill_scalar_out(value, shape, dtype):
...
 
+FILL_SLICE_CASES = [
+    # (shape, slice)
+    ((4, 128), (slice(None), slice(64, None))),
+    ((2, 1, 1, 512), (slice(None), slice(None), slice(None), slice(358, None))),
+    ((8, 32, 64), (slice(None), slice(16, None))),
+]
+
+
+@pytest.mark.fill
```
tengqm: This mark has to be one of `fill_scalar`, `fill_scalar_`, `fill_scalar_out`, `fill_tensor`, `fill_tensor_`, `fill_tensor_out`.
There is no `fill` operator in concept.

Marking the test case this way will lead to test case not found.
Please change.

tests/test_fill.py
```diff
@@ -100,6 +126,21 @@ def test_fill_scalar_(value, shape, dtype):
     utils.gems_assert_equal(x, ref_x)
 
 
+@pytest.mark.fill
```
tengqm: Same here.

huangyiqun (dismissed):

tests/test_fill.py
wuhulalala: fix it

src/flag_gems/runtime/backend/_nvidia/hopper/ops/fill.py
```diff
@@ -90,10 +104,13 @@ def fill_tensor_out(input, value, *, out=None):
...
         )
-    n_elements = out.numel()
+    out_contig, need_copy = _as_contiguous(out)
+    n_elements = out_contig.numel()
     grid = (triton.cdiv(n_elements, 1024),)
     with torch_device_fn.device(input.device):
-        fill_tensor_kernel[grid](out, value, n_elements, BLOCK_SIZE=1024)
+        fill_tensor_kernel[grid](out_contig, value, n_elements, BLOCK_SIZE=1024)
+    if need_copy:
+        out.copy_(out_contig)
```
tengqm, line 113: Not test case was added for this variant?

wuhulalala: No additional test is needed; the slice view test can already cover this change.

huangyiqun (approved):

huangyiqun (approved):

PR #2580, opened 2026-04-21
[MTHREADS] Fix device type checks of safe_softmax

Oslomayor:


w1120029931-bit (changes requested):
test_special_ops.py does not exist currently

Oslomayor:
Fine

PR #2593, opened 2026-04-21
[MTHREADS] Fix device type checks of zero

Oslomayor:


src/flag_gems/ops/zero.py
```diff
@@ -26,7 +28,8 @@ def zero_kernel(
 
 def _launch_zero_kernel(tensor: torch.Tensor):
     assert isinstance(tensor, torch.Tensor), "Expected a torch.Tensor"
-    assert tensor.is_cuda, "Tensor must be on CUDA device"
+    if tensor.device.type != flag_gems.device:
+        raise ValueError(f"Tensor must be on {flag_gems.device} device")
```
tengqm, line 32: Thanks for splitting the large PR!!!

Please now touch the test_zero function to trigger a test.

tests/test_tensor_constructor_ops.py
```diff
@@ -400,7 +400,7 @@ def test_accuracy_arange(end, dtype):
 
 
 @pytest.mark.zero
-@pytest.mark.parametrize("shape", [(2, 3), (128, 256), (512, 512)])
+@pytest.mark.parametrize("shape", [(512, 512), (128, 256), (2, 3)])
```
Oslomayor, line 403: Adjust the order of shape parameters to trigger CI tests.

tengqm: By "touching the test case", I didn't mean to change the actual logic or configuration. There is a risk that the changed parameters may fail on some backends.
This change is not harmful because it is only changing the parameter order.
A more lighter change would be just adding an empty line or remove one.

tengqm: In this case, a more appropriate test modification would be confirming that the `ValueError` is actually thrown when conditions are met.

bin913:
@Oslomayor  Please fix failing test case

bin913:
@Oslomayor Please resolve conflicts

Oslomayor:
I have been occupied with other priorities recently and discovered that the main branch has changed significantly due to refactoring. This has created substantial conflicts with my historical commits in this unmerged PR. Therefore, I have decided to close this PR. I will submit a new PR later to address these issues.

Issue #2310, opened 2026-04-09
# Bug: MatMul kernels fail with fp64 inputs due to accumulator type mismatch


lovywu:
## Summary

MatMul kernels fail when processing fp64 (double precision) inputs due to type mismatch between the accumulator initialization (fp32) and the loop-carried computation result (fp64).

**This is a confirmed bug but marked as LOW PRIORITY since fp64 is rarely used in deep learning training/inference scenarios.**

## Priority

**LOW** - Not a common use case in production deep learning environments. This can be addressed when time permits or if users report needing fp64 support.

### Rationale

| Factor | Reason |
|--------|--------|
| **Production usage** | fp64 almost never used in DL training/inference |
| **Hardware performance** | Consumer GPUs: fp64 is 1/64 of fp32 throughput |
| **Memory overhead** | fp64 doubles memory usage |
| **Model precision** | Production uses fp32, fp16, bf16, or lower |

## Error Message

```
AssertionError("Loop-carried variable accumulator has initial type <['128', '256'], fp32> 
but is re-assigned to <['128', '256'], fp64> in loop! Please make sure that the type stays consistent.")

/usr/local/lib/python3.12/dist-packages/triton/compiler/compiler.py:80: CompilationError
```

## Error Details

```python
# Kernel code (generated)
accumulator = tl.zeros((BLOCK_SIZE_M, BLOCK_SIZE_N), dtype=tl.float32)  # fp32

for k in range(0, tl.cdiv(K, BLOCK_SIZE_K)):
    a = tl.load(A + ...)  # fp64 when input is fp64
    b = tl.load(B + ...)  # fp64 when input is fp64
    a = a.to(C.dtype.element_ty)  # fp64
    b = b.to(C.dtype.element_ty)  # fp64
    accumulator += tl.dot(a, b)    # Type conflict: fp32 += fp64
```

### Minimum Reproducible Example

```python
import torch
import torch.nn as nn
import flag_gems

device = flag_gems.device

# SimpleLinear layer with fp64
model = nn.Linear(32, 16)
model = model.to(device).to(torch.float64)  # fp64

x = torch.randn(1, 32, device=device, dtype=torch.float64, requires_grad=True)

# This triggers the bug
output = model(x)  # FAILS during matmul
loss = output.sum()
loss.backward()
```

### Full Test Case

```bash
[...]

PR #2606, opened 2026-04-22
fix mix MatMul kernels fail with fp64 inputs due to accumulator type …

wuhulalala:
### PR Category

operator
### Type of Change

Bug fix
### Description

Fix fp64 type mismatch issue in matmul kernels by introducing a dtype-aware accumulator (via IS_FP64) to ensure type consistency during accumulation.
### Issue
#2310 

### Progress

### DIFF
--- a/src/flag_gems/ops/addmm.py
+++ b/src/flag_gems/ops/addmm.py
@@ -10,6 +10,21 @@ from flag_gems.runtime import torch_device_fn
 from flag_gems.utils import broadcastable_to, libentry, libtuner
 from flag_gems.utils import triton_lang_extension as tle
 
+_ordered_datatypes = [torch.float16, torch.bfloat16, torch.float32, torch.float64]
+
+
+def get_higher_dtype(a, b):
+    if a is b:
+        return a
+    assert a in _ordered_datatypes
+    assert b in _ordered_datatypes
+    for d in _ordered_datatypes:
+        if a is d:
+            return b
+        if b is d:
+            return a
+
+
 logger = logging.getLogger(__name__)
 
 
@@ -76,6 +91,9 @@ def addmm_kernel(
         if IS_FP64:
             a = a.to(tl.float32)
             b = b.to(tl.float32)
+        if a.dtype != b.dtype:
+            a = a.to(tl.float32)
+            b = b.to(tl.float32)
         accumulator += tl.dot(a, b, allow_tf32=False)
         a_ptrs += BLOCK_SIZE_K * stride_ak
         b_ptrs += BLOCK_SIZE_K * stride_bk
@@ -88,7 +106,7 @@ def addmm_kernel(
     bias = tl.load(i_ptrs, mask=c_mask, other=0.0)
 
     accumulator = accumulator * alpha + bias * beta
-    c = accumulator.to(bias.dtype)
+    c = accumulator.to(c_ptr.dtype.element_ty)
     tl.store(c_ptrs, c, mask=c_mask)
 
 
@@ -112,7 +130,8 @@ def addmm(bias, mat1, mat2, *, beta=1, alpha=1):
     )
     mat1 = mat1.contiguous()
     # mat2 = mat2.contiguous()
-    out = torch.empty((M, N), device=mat1.device, dtype=mat1.dtype)
+    c_dtype = get_higher_dtype(mat1.dtype, mat2.dtype)
+    out = torch.empty((M, N), device=mat1.device, dtype=c_dtype)
     bias = bias.broadcast_to(out.shape)
 
     grid = lambda META: (
@@ -151,7 +170,8 @@ def addmm_out(bias, mat1, mat2, *, beta=1, alpha=1, out=None):
     M, K = mat1.shape
     _, N = mat2.shape
     if out is None:
-        out = torch.empty((M, N), device=mat1.device, dtype=mat1.dtype)
+        c_dtype = get_higher_dtype(mat1.dtype, mat2.dtype)
+        out = torch.empty((M, N), device=mat1.device, dtype=c_dtype)
     else:
         assert out.shape == (M, N), "Incompatible output shape"
     logger.debug(
--- a/src/flag_gems/ops/addmm.py
+++ b/src/flag_gems/ops/addmm.py
@@ -10,21 +10,6 @@ from flag_gems.runtime import torch_device_fn
 from flag_gems.utils import broadcastable_to, libentry, libtuner
 from flag_gems.utils import triton_lang_extension as tle
 
-_ordered_datatypes = [torch.float16, torch.bfloat16, torch.float32, torch.float64]
-
-
-def get_higher_dtype(a, b):
-    if a is b:
-        return a
-    assert a in _ordered_datatypes
-    assert b in _ordered_datatypes
-    for d in _ordered_datatypes:
-        if a is d:
-            return b
-        if b is d:
-            return a
-
-
 logger = logging.getLogger(__name__)
 
 
@@ -91,9 +76,6 @@ def addmm_kernel(
         if IS_FP64:
             a = a.to(tl.float32)
             b = b.to(tl.float32)
-        if a.dtype != b.dtype:
-            a = a.to(tl.float32)
-            b = b.to(tl.float32)
         accumulator += tl.dot(a, b, allow_tf32=False)
         a_ptrs += BLOCK_SIZE_K * stride_ak
         b_ptrs += BLOCK_SIZE_K * stride_bk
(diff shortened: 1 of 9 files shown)


tengqm (commented):
Thanks for the PR. Please amend the test cases for validation and performance benchmark.

tengqm:
As a side note ... I'm expecting some failure for this change on some chips because not all backends support FP64. Some test cases for this is really a must.

wuhulalala:
> As a side note ... I'm expecting some failure for this change on some chips because not all backends support FP64. Some test cases for this is really a must.

I will add some FP64 test cases. Thanks for the suggestion.

wuhulalala:
> As a side note ... I'm expecting some failure for this change on some chips because not all backends support FP64. Some test cases for this is really a must.

The choice between FP64 and FP32 is determined at compile time. FP32 is used by default, and FP64 is only enabled when explicitly specified. I will add some FP64 test cases, but I don't think it's necessary to test FP64 on hardware that doesn't support it.

wuhulalala:
> > As a side note ... I'm expecting some failure for this change on some chips because not all backends support FP64. Some test cases for this is really a must.
> 
> The choice between FP64 and FP32 is determined at compile time. FP32 is used by default, and FP64 is only enabled when explicitly specified. I will add some FP64 test cases, but I don't think it's necessary to test FP64 on hardware that doesn't support it.

I can directly add FP64 support. FlagGems will determine whether the underlying hardware supports FP64 and decide whether to run FP64 tests accordingly.

tengqm:
> > > As a side note ... I'm expecting some failure for this change on some chips because not all backends support FP64. Some test cases for this is really a must.
> > 
> > 
> > The choice between FP64 and FP32 is determined at compile time. FP32 is used by default, and FP64 is only enabled when explicitly specified. I will add some FP64 test cases, but I don't think it's necessary to test FP64 on hardware that doesn't support it.
> 
> I can directly add FP64 support. FlagGems will determine whether the underlying hardware supports FP64 and decide whether to run FP64 tests accordingly.

Agreed. We will need to add `pytest.mark.skipif`.

wuhulalala:
> > > > As a side note ... I'm expecting some failure for this change on some chips because not all backends support FP64. Some test cases for this is really a must.
> > > 
> > > 
> > > The choice between FP64 and FP32 is determined at compile time. FP32 is used by default, and FP64 is only enabled when explicitly specified. I will add some FP64 test cases, but I don't think it's necessary to test FP64 on hardware that doesn't support it.
> > 
> > 
> > I can directly add FP64 support. FlagGems will determine whether the underlying hardware supports FP64 and decide whether to run FP64 tests accordingly.
> 
> Agreed. We will need to add `pytest.mark.skipif`.
ALL_FLOAT_DTYPES = FLOAT_DTYPES + [torch.float64] if fp64_is_supported else FLOAT_DTYPES
Maybe we can directly use ALL_FLOAT_DTYPES ? it will determine whether the machine supports FP64.

wuhulalala:
> > > > > As a side note ... I'm expecting some failure for this change on some chips because not all backends support FP64. Some test cases for this is really a must.
> > > > 
> > > > 
> > > > The choice between FP64 and FP32 is determined at compile time. FP32 is used by default, and FP64 is only enabled when explicitly specified. I will add some FP64 test cases, but I don't think it's necessary to test FP64 on hardware that doesn't support it.
> > > 
> > > 
> > > I can directly add FP64 support. FlagGems will determine whether the underlying hardware supports FP64 and decide whether to run FP64 tests accordingly.
> > 
> > 
> > Agreed. We will need to add `pytest.mark.skipif`.
> > ALL_FLOAT_DTYPES = FLOAT_DTYPES + [torch.float64] if fp64_is_supported else FLOAT_DTYPES
> > Maybe we can directly use ALL_FLOAT_DTYPES ? it will determine whether the machine supports FP64.

fp64_is_supported only reflects the general FP64 capability of the device, but it does not guarantee support for all operators. In particular, tl.dot used in matrix multiplication has stricter constraints and may not support FP64 even when the device does.

To address this, we have refined the test logic by adding more fine-grained skip conditions for mm-related operators, ensuring FP64 tests are only executed when fully supported.  
We will add pytest.mark.skipif to handle this properly. You are absolutely right, thanks for pointing this out.

tengqm:
> To address this, we have refined the test logic by adding more fine-grained skip conditions for mm-related operators, ensuring FP64 tests are only executed when fully supported.

Testing is only an approach to help validating that the software works. Passing tests doesn't necessarily mean that software won't break in production environments.  We can have many `skipif` annotations in the test cases so that they won't make noise.  However, all those `skipif` are also bombs that will break users some day.

A more defensive way to do this is to check if the logic is running on the targeted chip, without making assumptions about other backends.

src/flag_gems/ops/addmm.py
```diff
@@ -10,6 +10,21 @@
 from flag_gems.utils import broadcastable_to, libentry, libtuner
 from flag_gems.utils import triton_lang_extension as tle
 
+_ordered_datatypes = [torch.float16, torch.bfloat16, torch.float32, torch.float64]
+
+
+def get_higher_dtype(a, b):
```
tengqm: I can vaguely guess your intention in this function ...
Can you please add some comments about the logic?
Maybe this logic has some existing implementation somewhere we can invoke?
Type promotion is not an uncommon scenario in this community anyway.

tests/test_mm.py
```diff
@@ -66,47 +54,27 @@ def test_mm(M, N, K, dtype, b_column_major):
...
-    np.random.seed(0)
-    random.seed(0)
-
-    mat = torch.randn((K, M), dtype=dtype, device=flag_gems.device).t()
-    out = torch.empty((M, M), dtype=dtype, device=flag_gems.device)
-    ref_mat = utils.to_reference(mat, True)
-    ref_out = utils.to_reference(out, True)
+@pytest.mark.parametrize("M, N, K", MNK_SHAPES)
+@pytest.mark.parametrize("dtype_a, dtype_b", MIXED_DTYPE_PAIRS)
+def test_mm_mixed_dtype(M, N, K, dtype_a, dtype_b):
```
tengqm: 💯  This is exactly what we need for this change, i.e. make sure the code works for different scenarios.

wuhulalala: I believe this test is unnecessary, as mixed precision should be handled by the user. I plan to remove it.

src/flag_gems/ops/addmm.py
wuhulalala: it's unnecessary, i will remove it.

tests/test_mm.py
wuhulalala: > 

After discussing with @huangyiqun , we have decided not to add FP64 tests for now. This is because support for FP64 varies across different devices, and even though the check

ALL_FLOAT_DTYPES = FLOAT_DTYPES + [torch.float64] if fp64_is_supported else FLOAT_DTYPES

is in place, issues may still arise. In other words, even if the check passes, runtime errors can still occur. Such as A100. A comprehensive solution would require systematically identifying which devices truly support FP64, which would involve significant effort and low-level instruction analysis. Therefore, for now, we will focus on fixing the bug without adding FP64 tests.

tengqm: I still insist that we add test cases for behavior changes.
For nvidia specific changes, we can add a test case like

```
@pytest.mark.skipif(flag_gems.vendor_name != "nvidia", reason="blablah")
test_mm():
```

For other changes related to common implementations, a test change is also necessary.

tengqm: It is perfectly fine that something only works on a specific backend. For an operator that is customized for that backend, there are no global impacts.

For other changes, we have to cope with them carefully.

huangyiqun: I believe we should maintain a supported data type matrix for all chips connected to our backends going forward. This will facilitate unified management of unit tests in the long run. For example:
```
          fp32  fp64
chip1      1     0
chip2      1     1
```
Currently, our Nvidia 'backend' targets the Hooper architecture. If we introduce support for other Nvidia architectures in the future, maintaining test skip rules for each specific architecture would require significant effort. Furthermore, as we plan to add precisions like FP4 or FP8, continuing to manage test skips on a per-precision basis using this current approach feels inelegant and unsustainable.

tengqm: Totally agree to @huangyiqun's observations and visions.

Maybe it is time to add some device backend interface functions such as `fp32_supported()` and `fp64_supported()`. Then we can check the device function at runtime to determine the best execution path without breaking the common implementation.

huangyiqun: This is a great suggestion and should be added to backlog. However, this PR addresses a high-priority issue that needs immediate attention, so we should merg this PR soon.

huangyiqun (approved):

PR #2625, opened 2026-04-23
[FlagGems Operator Development Competition]  add gcd

LuyuZhang00:
### PR Category

Operator 
### Type of Change

New Feature

### Description

  This PR implements torch.gcd for FlagGems and adds full integration, validation, and benchmarking support. It covers the following aten schemas:

  - gcd(Tensor self, Tensor other) -> Tensor
  - gcd.out(Tensor self, Tensor other, *, Tensor(a!) out) -> Tensor(a!)

  ### Implementation Details

  - Kernel file: src/flag_gems/ops/gcd.py
  - Registration: src/flag_gems/__init__.py registers gcd and gcd.out
  - Exports: src/flag_gems/ops/__init__.py imports and exports gcd, gcd_out
  - Tests: tests/test_gcd.py adds both functional and error-path coverage
  - Benchmark: benchmark/test_binary_pointwise_perf.py adds gcd, gcd_distribution, and gcd_out benchmark entries

  ### Algorithm

  The implementation uses different strategies depending on dtype and value range, because gcd on signed integers is dominated by overflow-sensitive edge cases around INT_MIN.

  #### 1. Main path: binary GCD for regular values

  For the normal path, the kernel uses the binary GCD algorithm (Stein's algorithm).

  For nonzero a and b, it relies on:

  - gcd(a, b) = gcd(|a|, |b|)
  - factor out the common power of two
  - repeatedly subtract the smaller odd value from the larger odd value until one becomes zero

  This is implemented in _binary_gcd(ax, ay, normal):

  - ctz is computed with libdevice.ffs(x) - 1
  - common factors of two are removed first
  - active lanes iterate with subtraction rather than division/modulo
  - the final result is reconstructed by left-shifting back with the common factor

  This is the fast path for regular signed integers after converting them to nonnegative magnitudes.

  #### 2. Why INT_MIN needs special handling

  For signed integer dtypes, abs(INT_MIN) is not representable in the same signed dtype:

  - abs(-32768) does not fit in int16
  - abs(-(1 << 31)) does not fit in int32
  - abs(-(1 << 63)) does not fit in int64

  So a naive abs + gcd implementation is incorrect for boundary values. The kernel therefore splits execution into:

  - normal lanes: values whose absolute values are representable and can use the binary GCD path
  - special lanes: lanes where either input is INT_MIN

  #### 3. int16: lookup-table path for INT16_MIN

  For int16, the implementation uses a cached lookup table:

  - _get_i16_min_lut(device) materializes gcd(INT16_MIN, x) for all x in [0, INT16_MAX] using PyTorch reference computation
  - the LUT is cached per device in _I16_MIN_LUT_CACHE
  - in gcd_kernel_i16, lanes involving INT16_MIN bypass the regular path and read the answer directly from the LUT

  This keeps the int16 boundary handling simple and avoids overflow-sensitive control flow in the kernel.

  #### 4. int32 and int64: C-style remainder fallback for INT_MIN

  For int32 and int64, a LUT would be too large, so the implementation uses a fallback Euclidean-style reduction on the special lanes only.

  This path is built around:

  - _abs_u32 / _abs_u64: compute magnitudes in unsigned arithmetic, avoiding overflow on INT_MIN
  - _c_rem_i32 / _c_rem_i64: emulate C-style signed remainder, not Python modulo semantics

  For special lanes, the kernel iteratively applies Euclidean reduction using overflow-safe signed remainder until the remainder becomes zero.

  This preserves PyTorch-compatible behavior even when one or both operands are equal to the minimal signed value.

  #### 5. Type promotion and broadcasting

  The wrapper does not assume equal dtype or equal shape.

  _materialize_inputs(self, other) performs:

  - torch.promote_types(self.dtype, other.dtype) for the computation dtype
  - explicit casts when needed
  - torch.broadcast_tensors(lhs, rhs)
  - contiguous materialization before launching the Triton kernel

  So the operator follows the normal binary pointwise promotion-and-broadcast flow before flattening to a 1D launch.

  #### 6. out= behavior

  The gcd(..., out=...) path preserves PyTorch-style .out semantics rather than requiring a perfectly matched output tensor.

  The wrapper first checks whether out can be used directly:

  - same device
  - same dtype as the promoted result
  - same shape
  - contiguous

  If all of those hold, the kernel writes into out directly.

  Otherwise, it computes into a temporary result buffer and then copies into out via torch.ops.aten.copy_.default.redispatch(...).

  This supports cases such as:

  - aliased out
  - noncontiguous out
  - out tensors that require fallback copy semantics

  The follow-up commit adds one important guard here:

  - if out exists but torch.can_cast(promoted_dtype, out.dtype) is false, the operator now raises a RuntimeError

  This is necessary because invalid out dtypes should be rejected, not silently accepted.

  ### Tests

  tests/test_gcd.py covers:

  - basic correctness across pointwise shapes and integer dtypes
  - out= correctness
  - aliasing with out=inp1
  - noncontiguous out tensors
  - boundary-value correctness including 0, negative values, dtype.min, and dtype.max
  - representative distribution-style samples: equal inputs, unit inputs, power-of-two inputs
  - empty tensors
  - noncontiguous broadcast inputs

  The follow-up update also adds explicit negative-path coverage for:

  - invalid dtype
  - broadcast failure
  - out shape mismatch
  - incompatible out dtype

  ### Benchmark

  This PR adds three benchmark entry points in benchmark/test_binary_pointwise_perf.py:

  - gcd as a standard binary pointwise benchmark
  - gcd_out for .out performance
  
A dedicated GcdBenchmark class reports GBPS rather than TFLOPS, because gcd is a data-dependent iterative integer operator and memory throughput is more interpretable than a synthetic FLOPS metric.

  ### Files Changed

  | File | Change |
  |------|--------|
  | src/flag_gems/ops/gcd.py | New Triton implementation for gcd and gcd.out |
  | src/flag_gems/__init__.py | Register gcd / gcd.out |
  | src/flag_gems/ops/__init__.py | Import and export gcd, gcd_out |
  | tests/test_gcd.py | Add accuracy tests, boundary tests, out tests, and negative-path tests |
  | benchmark/test_binary_pointwise_perf.py | Add gcd, gcd_distribution, and gcd_out benchmark coverage |

  ### Issue

  New operator implementation and follow-up semantic hardening for the Operator Development Competition.

  ### Performance

  Environment: benchmark/test_binary_pointwise_perf.py, mode=kernel, level=core
  Speedup = Torch latency / Gems latency; > 1.0 means Gems is faster.

  #### int16

  | Shape | Torch (ms) | Gems (ms) | Speedup | GBPS |
  |-------|------------|-----------|---------|------|
  | (1073741824,) | 31.575 | 26.747 | 1.181 | 240.867 |
  | (64, 64) | 0.0143 | 0.00819 | 1.750 | 3.000 |
  | (4096, 4096) | 0.437 | 0.383 | 1.142 | 262.845 |
  | (64, 512, 512) | 0.437 | 0.378 | 1.157 | 266.407 |
  | (1024, 1024, 1024) | 31.295 | 26.745 | 1.170 | 240.886 |

  #### int32

  | Shape | Torch (ms) | Gems (ms) | Speedup | GBPS |
  |-------|------------|-----------|---------|------|
  | (1073741824,) | 55.096 | 47.089 | 1.170 | 273.631 |
  | (64, 64) | 0.0225 | 0.0113 | 2.000 | 4.364 |
  | (4096, 4096) | 0.740 | 0.684 | 1.082 | 294.323 |
  | (64, 512, 512) | 0.739 | 0.701 | 1.054 | 287.019 |
  | (1024, 1024, 1024) | 55.284 | 47.514 | 1.164 | 271.183 |

### DIFF
--- a/benchmark/test_binary_pointwise_perf.py
+++ b/benchmark/test_binary_pointwise_perf.py
@@ -10,7 +10,7 @@ from benchmark.attri_util import (
     FLOAT_DTYPES,
     INT_DTYPES,
 )
-from benchmark.performance_utils import Benchmark, generate_tensor_input
+from benchmark.performance_utils import Benchmark, Config, generate_tensor_input
 
 
 class BinaryPointwiseBenchmark(Benchmark):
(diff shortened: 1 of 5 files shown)


douxetpur:

PR #2643, opened 2026-04-24
[Mthreads] Optimized mm addmm bmm with FlagOSTune

Fattysand:
### PR Category

Operator
### Type of Change

Performance Optimization
### Description

1. USE FlagOSTune to optimized mthreads mm/addmm.bmm performance
2. Achieved 1.02x to 5.6x speedup for mm operator on MTT S5000;
3. Achieved 1x to 2.82x speedup for addmm operator on MTT S5000;
4. Achieved 1x to 1.19x speedup for bmm operator on MTT S5000;
### Progress


src/flag_gems/runtime/backend/_mthreads/addmm_mthreads_expand.yaml
```diff
@@ -0,0 +1,30 @@
+addmm_sqmma:
```
tengqm, line 1: I'd suggest we move this to `src/flag_gems/runtime/backend/_mthreads/config/` and named as `addmm.yaml` for simplicity.

src/flag_gems/runtime/backend/_mthreads/ops/mm.py
```diff
@@ -124,15 +137,19 @@ def mm_kernel(
...
+    configs=runtime.ops_get_configs("gemv", yaml_path=EXPAND_CONFIG_FILENAME)
+    if os.environ.get("USE_FLAGTUNE") == "1"
+    else [triton.Config({"BLOCK_M": 64, "BLOCK_K": 64})],
     key=["M", "K", "stride_am", "stride_bk"],
-    strategy=["align32", "align32", "align32", "default"],
+    strategy=runtime.get_expand_config("gemv", yaml_path=EXPAND_CONFIG_FILENAME)[
+        "strategy"
+    ]
+    if os.environ.get("USE_FLAGTUNE") == "1"
+    else ["align32", "align32", "align32", "default"],
```
tengqm, line 150: write this as

```
if cond:
    something
else:
    something_else
```

please.

Fattysand: Thanks for the suggestion! However, this code is inside the @libtuner(...) decorator arguments, which only accepts expressions, not statement-level if/else blocks. The inline conditional expression is the only way to keep the logic here without moving variables outside the decorator. I've kept the inline form for this reason.

src/flag_gems/runtime/backend/_mthreads/addmm_mthreads_expand.yaml
Fattysand: Thanks for the suggestion! The current path and naming (addmm_mthreads_expand.yaml) follows the same directory structure and naming convention used by all other ops' config files in this backend. Moving it to a different location would break consistency with the rest of the codebase. I'd prefer to keep it as-is for uniformity.

src/flag_gems/runtime/backend/_mthreads/ops/mm.py
tengqm: Got it. Does it make sense to wrap the following snippet into a function?

```
def _get_strategy():
  if os.environ.get("USE_FLAGTUNE") == "1":
    cfg = runtime.get_expand_config("gemv", yaml_path=EXPAND_CONFIG_FILENAME)
    return cfg.get("strategy", ...)

  return ["align32", "align32", "align32", "default"]
```

src/flag_gems/runtime/backend/_mthreads/addmm_mthreads_expand.yaml
tengqm: The `_mthreads_expand` part in the name is redundant, useless.
Whenever there is an opportunity to fine-tune the directory layout, we should do it **now** rather than worsening the situation.

benchmark/test_blas_perf_parallel.py
```diff
@@ -47,6 +67,376 @@
 DEEPGEMM_K_MULTIPLE = 128
 
 
+# ============================================================================
+# Blas benchmark classes (from test_blas_perf.py)
+# ============================================================================
+
+
+class BlasBenchmark(Benchmark):
```
tengqm, line 75: Please don't copy this class.
It has been moved into `performance_utils.py` for sharing.

Vincent-Xiao (approved):

Vincent-Xiao (approved):

huangyiqun (approved):

PR #2699, opened 2026-04-27
[Multibackend]fix_multibackend_bugs_and_polish_code

Galaxy1458:
### Progress


src/flag_gems/runtime/configloader.py
```diff
@@ -4,9 +4,11 @@
 
 import triton
 
-from . import backend
+from . import backend, common
 from .backend.device import DeviceDetector
 
+<<<<<<< HEAD
+=======
```
tengqm: remove

src/flag_gems/runtime/configloader.py
```diff
@@ -69,6 +71,7 @@
 }
 
 
+>>>>>>> 6fed5de86157e8148a36e3febb97f266ba5f8a36
```
tengqm: remove

src/flag_gems/runtime/common.py
```diff
@@ -22,3 +23,107 @@ def get_all_vendors(cls) -> dict:
...
+
+UNSUPPORT_INT64 = frozenset(
+    {
+        vendors.AIPU,
+        vendors.TSINGMICRO,
+        vendors.SUNRISE,
+        vendors.ENFLAME,
+    }
+)
+
```
tengqm, line 57: Instead of maintaining global sets, I'd suggest we push down these decisions to vendors. When implementing and/or testing an operator, we can simply check if the **current backend** supports FP64, for example. It will help make sure the operator is doing the right thing on all backends while avoiding to maintain this set. WDYT?

Galaxy1458:
@tengqm How should I run the full test?

src/flag_gems/runtime/common.py
Galaxy1458: This requires modifying the vendor adaptation architecture. The next PR can consider refactoring

tengqm:
> @tengqm How should I run the full test?

Emm ... we don't have a "full test" yet. The current CI logic is test case based. You have to change (add a new line for example) to a test case for CI to know that something needs a test. The same applies to backend tests. At least you can try change something in Ascend or Mthreads backend to see if the CI is okay with it.

Galaxy1458:
> > @tengqm How should I run the full test?
> 
> Emm ... we don't have a "full test" yet. The current CI logic is test case based. You have to change (add a new line for example) to a test case for CI to know that something needs a test. The same applies to backend tests. At least you can try change something in Ascend or Mthreads backend to see if the CI is okay with it.

This PR affects all performance and correctness tests of all backends. How can I enable the tests without changing the content of my PR, at least all the unit tests of nvidia.

tengqm:
> > > @tengqm How should I run the full test?
> > 
> > 
> > Emm ... we don't have a "full test" yet. The current CI logic is test case based. You have to change (add a new line for example) to a test case for CI to know that something needs a test. The same applies to backend tests. At least you can try change something in Ascend or Mthreads backend to see if the CI is okay with it.
> 
> This PR affects all performance and correctness tests of all backends. How can I enable the tests without changing the content of my PR, at least all the unit tests of nvidia.



> > > @tengqm How should I run the full test?
> > 
> > 
> > Emm ... we don't have a "full test" yet. The current CI logic is test case based. You have to change (add a new line for example) to a test case for CI to know that something needs a test. The same applies to backend tests. At least you can try change something in Ascend or Mthreads backend to see if the CI is okay with it.
> 
> This PR affects all performance and correctness tests of all backends. How can I enable the tests without changing the content of my PR, at least all the unit tests of nvidia.

For that kind of "full test", we are still developing. It is gonna take a loooong time to run all accuracy tests and all perf benchmarks. We are baby stepping towards that end. So far the benchmark split work has helped identified those tests that will never succeed. The next step is to enable per-operator, per-backend test, and still next step would be huge loops nightly tests for your purpose.

For the moment, I think it is okay to test just one backend.

If you want to run the tests locally on your machine (a given backend other than NV), you can do `tools/run_tests.py --gpus 0,1,2,3,4,5,6,7`. It will start parallel testing of all accuracy tests and all perf benchmarks. Still ... I have to warn you that it may fail in different ways. And that is the reason we are not yet exposing the interface at GitHub CI jobs.

huangyiqun (approved):

PR #2702, opened 2026-04-27
Fix mm benchmark shape

0x45f:
### PR Category

OP Test
### Type of Change

Bug Fix
### Description

Fix mm benchmakr shape
### Progress


huangyiqun (dismissed):

benchmark/base.py
```diff
@@ -663,10 +663,10 @@ def get_input_iter(self, dtype) -> Generator:
...
     def set_more_shapes(self):
         large_k_shapes = [
-            [8, 1848, 1536, 151936],
-            [8, 1848, 1536, 128256],
-            [8, 1848, 1536, 152064],
-            [8, 4096, 1, 152064],
+            (8, 1848, 1536, 151936),
+            (8, 1848, 1536, 128256),
+            (8, 1848, 1536, 152064),
+            (8, 4096, 1, 152064),
```
tengqm, line 648: What is the difference by changing lists to tuples?
We require that `set_more_shapes` to return list of lists.
https://github.com/flagos-ai/FlagGems/blob/master/benchmark/performance_utils.py#L237
Is this causing some test regression?

0x45f: using list will cause benchmark/test_mm.py fail

tengqm: okay ... need to check what happened.
let's go with tuples at the moment

tengqm: can you touch test_mm.py file to trigger a test, possibly by removing the skip condition?

tengqm: Please rebase to master.

`performance_utils.py` has been renamed to `base.py`.

huangyiqun (approved):

PR #2821, opened 2026-04-30
add mul.out, add.out, sub.out, div.out

taozhiwei:
add mul.out,add.out,sub.out,div.out


src/flag_gems/ops/add.py
```diff
@@ -93,6 +93,29 @@ def add(A, B, *, alpha=1):
         return torch.tensor(A + B * alpha)
 
 
+def add_out(A, B, *, alpha=1, out):
+    logger.debug("GEMS ADD_OUT")
+    if not (isinstance(A, torch.Tensor) and isinstance(B, torch.Tensor)):
+        raise TypeError("aten::add.out expects tensor operands")
```
tengqm, line 99: I'm not confident that this is correct.
See `add_` implementation.
We are dealing with tensor and scalar mixtures in that path.

tengqm (commented):
@taozhiwei Please consider splitting this PR into four PRs, each PR with implementation, accuracy tests and performance benchmarks.

PR #2870, opened 2026-05-01
[FlagGems Operator Development Competition] Add svd and linalg_svd operators

QuantumByte-01:
## Summary
Adds **ten::svd** and **ten::linalg_svd** with **PyTorch-identical numerics** using the same **CPU reference** pattern already used elsewhere when a custom kernel would recurse into itself (see also the median NaN CPU path): work is done on **	ensor.detach().cpu()**, then **U / S / V(h)** are moved back to the input device.

## API
- **svd(inp, some=True, compute_uv=True)** — batched legacy SVD, returns **(U, S, V)** like **	orch.svd**.
- **linalg_svd(A, full_matrices=True, *, driver=None)** — returns **(U, S, Vh)** like **	orch.linalg.svd**.

## dtypes
- **float32 / float64 / complex64** exercised in tests.
- **float16 / bfloat16**: promoted to **float32** on CPU for the solve, then **U** and **V/Vh** are cast back to the input dtype; **S** is returned in **float32** (real dtype for complex).

## Tests
- Parity vs CPU **	orch.svd** / **	orch.linalg.svd** under **use_gems()**.
- **some** True/False (non-quick).
- **compute_uv=False** for **	orch.svd**.
- Reconstruction check **U @ diag_embed(S) @ V[..., :k].mT** for **some=True**.

## Performance note
This is intentionally a **correctness-first** landing (competition baseline). A follow-up can add a **Triton / vendor** fast path while keeping this path as a fallback, similar to how other ops grow from reference to tuned kernels.

Made with [Cursor](https://cursor.com)


douxetpur:

PR #2935, opened 2026-05-06
Enhance run_cmd to log output and load KernelGen ops

tianxiao-baai:
Added functionality to log stdout and stderr to separate files in run_cmd function. Also ensured KERNELGEN_OPS is populated in subprocesses.

### Progress


tools/run_tests.py
```diff
@@ -194,19 +206,35 @@ def init():
...
+            stderr_target = stderr_fh
+
         p = subprocess.Popen(
             shlex.split(cmd),
             cwd=str(cwd),
             env=env,
-            stdout=subprocess.DEVNULL,
-            stderr=subprocess.DEVNULL,
+            stdout=stdout_target,
+            stderr=stderr_target,
```
tengqm, line 226: My recent research on timeout control gave me a different story.
We should be very cautious when piping stdout, stderr when we want to kill process that run out of control. Stability/predictability is more important than capturing the stdout and/or stderr. I'd suggest we revisit this change when we can handle timeout properly.

https://stackoverflow.com/questions/36952245/subprocess-timeout-failure
https://stackoverflow.com/questions/4789837/how-to-terminate-a-python-subprocess-launched-with-shell-true
https://bugs.python.org/issue30154
...

For a full range of testing, we only care about pass/fail/skip without further details. For problem diagnosis, a developer can always do a manual `pytest -m problematic_op tests ...` to get the stack trace.

tools/run_tests.py
```diff
@@ -488,6 +534,7 @@ def worker_proc(gpu_id, start, count):
         ]
         result = {
             "customized": op in customized_ops,
+            "kernelgen": op in KERNELGEN_OPS,
```
tengqm: I would not treat KernelGen operators as a special beast...
KernelGen is not an operator kind in the current operator inventory, it is only a label. It is a label among many others.
We can always to a post-testing process to filter out the generated operators.

However, if needed, we may add an option to the tool, say `run_tests.py --labels <label>`.

tools/run_tests.py
tianxiao-baai: My original intention here was to improve debuggability for long-running tests (~5 min). When a failure happens, having logs persisted can save a full rerun cycle, which is quite costly during large-scale testing.
Thanks for the detailed references, they were helpful. Just to clarify, in this change I'm not using PIPE, but redirecting stdout/stderr directly to file handles, so it shouldn't hit the typical pipe buffer deadlock issue.

That said, I do agree the current timeout/termination logic is not robust enough.

I can updated the implementation to:
- explicitly create a new process group (start_new_session=True)
- terminate the whole group on timeout
- add a SIGKILL fallback if SIGTERM doesn't work
- ensure the process is properly waited to avoid zombies

This should make the behavior both safe and predictable while still preserving logs for debugging long-running tests.

tools/run_tests.py
tianxiao-baai: Thanks. My intention of adding `"kernelgen"` here is mainly for acceptance testing,
where we need to distinguish operator sources (e.g. customized vs generated)
for reporting and validation purposes.

I think the `--labels` option sounds like a clean solution.
If you’re OK with that direction, I can help extend the tool to support it.

tools/run_tests.py
tengqm: The robustness improvement has been done in https://github.com/flagos-ai/FlagGems/pull/2940/. Initial result from mthreads backend is that there is only one test case **almost** run out of control. The process group has been successfully terminated at 10 minutes boundary. That was great! We can come back to this issue later if SIGTERM is not sufficient.

Some more thoughts on the capture output topic ...

Most of the time, the outputs are not interesting, especially for the 80+% successful runs. For buggy operators/test cases, the verbose output could be helpful for developers. For those "failed" (including "skipped") test cases, we are currently only provide a veeeeeery brief summary and that summary is from pytest. I'm not sure we want to zoom in those cases while we are focusing on the whole picture. When we do want to get more details, and **if** we still want to use the `run_tests.py` script, emmm ... we can add an option like `--dump-output`. When specified, we capture the stdout/stderr into `<op>_log.txt` and move it to the results directory. We detect whether this option is specified rather than passing many parameters around ... WDYT?

tools/run_tests.py
tengqm: Cool. Please go ahead with label-based operator selection then ...

We may need some protocols among the various options:

- if 'ops' is specified, it is the final list of operators to test, or
- if 'op-list-file' is specified, it contains the list of operators to test, period, or
- if 'labels' is selected, it is used as a filter, just like the case with 'stages'; in other words, 'labels' and 'stages' are not mutual exclusive ... 

Thoughts?

tools/run_tests.py
tianxiao-baai: That sounds reasonable to me.

I agree that for large-scale runs, pass/fail/skip summaries are usually sufficient, and always dumping logs may introduce unnecessary noise and storage overhead.

Adding an optional `--dump-output` mode sounds like a good balance:
- keep the default path lightweight and stable
- allow detailed stdout/stderr capture when debugging problematic operators

I can prepare a follow-up PR based on this direction.

tools/run_tests.py
```diff
@@ -194,22 +207,74 @@ def init():
...
-        return TIMEOUT
-    return p.returncode
 
+        try:
+            p.wait(timeout=timeout)
+        except subprocess.TimeoutExpired:
+            try:
+                os.killpg(p.pid, signal.SIGTERM)
+            except Exception:
+                p.terminate()
```
tengqm, line 236:
```suggestion
                os.killpg(os.getpgid(p.pid), signal.SIGKILL)
```

tools/run_tests.py
```diff
@@ -194,22 +207,74 @@ def init():
...
+
+            try:
+                p.wait(timeout=5)
+            except subprocess.TimeoutExpired:
+                try:
+                    os.killpg(p.pid, signal.SIGKILL)
+                except Exception:
+                    p.kill()
```
tengqm, line 244: Let's add this additional wait later when necessary.
So far the `killpg` above works fine.

tools/run_tests.py
```diff
@@ -439,9 +511,17 @@ def run_benchmark(gpu_id, start, index, count):
     if result_file.exists():
         result_file.unlink()
 
+    op_dir = OUTPUT_DIR.joinpath(op)
+    ensure_dir(op_dir)
+    stdout_log = str(op_dir / f"performance_{op}_stdout.log") if DUMP_OUTPUT else None
+    stderr_log = str(op_dir / f"performance_{op}_stderr.log") if DUMP_OUTPUT else None
```
tengqm, line 522: Cool. The logs are dumped to the target directory directly.

tools/run_tests.py
```diff
@@ -490,6 +572,7 @@ def worker_proc(gpu_id, start, count):
         ]
         result = {
             "customized": op in customized_ops,
+            "labels": OP_LABELS.get(op, []),
```
tengqm: Don't do this...
I mean ... don't replicate data else where unless absolutely necessary.

tools/run_tests.py
```diff
@@ -566,18 +649,26 @@ def get_ops_to_test(ops_file, ops_list, stages):
...
+                        help="Dump stdout/stderr of each test to log files")
     args = parser.parse_args()
 
+    DUMP_OUTPUT = args.dump_output
+
     # Probe environment setttings
     init()
 
+    # Load operator labels from inventory
+    load_op_labels()
```
tengqm: Please consider call this function inside `get_ops_to_test`.
We don't dump labels into results file ... 
If we want to filter out operator by labels as a post-run processing,
we can always consult the `conf/operators.yaml` by operator IDs.

tengqm (approved):
Let's kick it in and move it forward.

PR #2972, opened 2026-05-08
benchmark: support 'all' mode to output all time results

lyujheng:
### PR Category

[Benchmark ]
### Type of Change

[New Feature]
### Description

The 'all' mode is now the default benchmark option. It separately outputs kernel, operator, and wrapper latency without modifying the behavior of any existing modes.

### Before (--mode=kernel)
kernel

### After (--mode=all)
all

### Issue
N/A

### Progress


bin913:
Could you please provide the output results of all mode in the comment? Any operator is fine, thanks a lot.

lyujheng:
> Could you please provide the output results of all mode in the comment? Any operator is fine, thanks a lot.

Done. Screenshots of the --mode=all output have been added to the PR description.

benchmark/conftest.py
```diff
@@ -85,9 +85,9 @@ def pytest_addoption(parser):
             "--mode" if vendor_name != "kunlunxin" else "--fg_mode"
         ),  # TODO: fix pytest-* common --mode args
         action="store",
-        default="kernel",
+        default="all",
```
tengqm, line 88: It is okay to add a new mode for some dedicated investigation, but changing the default mode would be a disaster because we have simply too many tools built on this conftest.py file.

w1120029931-bit:
please resolve conversation

benchmark/conftest.py
lyujheng: May I ask, do these tools rely on a data structure like the one in consts.py?
image
 The new mode will modify this data structure, but for the existing modes — kernel, operator, wrapper — it will not change their log outputs. The log modifications will only be made for the all mode.

Existing modes:
image

ALL mode:
image

huangyiqun:
Are you still working on this PR？

PR #3028, opened 2026-05-11
[MTHREADS] Fix heuristic config overriding vendor-specific config

Oslomayor:
In get_heuristic_config(), the loop iterates over (vendor_name, "nvidia") without breaking on success, causing the nvidia fallback to always overwrite the vendor-specific heuristic config module. This made the mthreads backend unintentionally use nvidia heuristics which call torch.cuda.current_device() and fail on non-CUDA platforms. Add a break statement after successful import to preserve the vendor-specific config.


src/flag_gems/runtime/backend/__init__.py
```diff
@@ -288,6 +288,7 @@ def get_heuristic_config(vendor_name=None):
         mod_name = f"_{backend}.{config_name}"
         try:
             _state.heuristic_config_module = importlib.import_module(mod_name)
+            break
         except Exception:
             continue
```
tengqm: Thanks for this fix ... but ... can we go one step further ...
This logic is clearly an `if ... else ...` logic, there is not reason to do a `for...` loop. It hurts the maintainability.

Oslomayor: > Thanks for this fix ... but ... can we go one step further ... This logic is clearly an `if ... else ...` logic, there is not reason to do a `for...` loop. It hurts the maintainability.



> Thanks for this fix ... but ... can we go one step further ... This logic is clearly an `if ... else ...` logic, there is not reason to do a `for...` loop. It hurts the maintainability.

This suggestion is spot on.

src/flag_gems/runtime/backend/__init__.py
```diff
@@ -283,13 +283,15 @@ def get_unused_ops(vendor_name=None):
...
+    mod_name = f"_{vendor_name}.{config_name}"
+    try:
+        _state.heuristic_config_module = importlib.import_module(mod_name)
+    except Exception:
+        mod_name = f"_nvidia.{config_name}"
         try:
             _state.heuristic_config_module = importlib.import_module(mod_name)
         except Exception:
-            continue
+            pass
```
tengqm: This is not something you need a `try...except`, right?
The config for nvidia should be always there.

Oslomayor:
@tengqm This PR is ready for review now.

Vincent-Xiao (approved):

PR #3083, opened 2026-05-13
[FlagGems Operator Development Competition] Add ctc_loss

IveGotMagicBean:
# [FlagGems Operator Development Competition] Add ctc_loss

## Summary

This PR adds a Triton implementation of `torch.nn.functional.ctc_loss` with autograd support. The kernel pair implements the log-space Connectionist Temporal Classification forward-backward DP (Graves et al., 2006) and is registered into the FlagGems aten dispatcher.

The implementation outperforms PyTorch's native CUDA CTC across **every** benchmark case tested (72 of 72) on a 2080 Ti (sm_75 Turing), with the largest gains on the backward pass.

## What's new

- `src/flag_gems/ops/ctc_loss.py` — full implementation
  - `_ctc_forward_kernel`: log-space alpha DP, one program per batch element, state dim held as a single `BLOCK_S = next_pow2(2L+1)` register tile
  - `_ctc_backward_kernel`: log-space beta DP with a per-batch shifted-beta scratch buffer, posterior atomic-scattered into a `(T, N, C)` accumulator
  - `_ctc_grad_kernel`: fused `(T, N)`-grid kernel that assembles the input gradient from `(log_probs, posterior_sum, scale, input_lengths)` in a single pass, replacing 4 Python elementwise kernels
  - `_CtcLoss(torch.autograd.Function)`: orchestrates forward + backward
  - `ctc_loss(...)` public entry matching `F.ctc_loss` signature
- `tests/test_ctc_loss.py` — 54 pytest cases covering:
  - basic forward (fp32 / fp16 / bf16, padded + concatenated targets, mean / sum / none reductions)
  - non-zero blank index with repeated labels
  - some-empty and all-empty target rows
  - `zero_infinity` on impossible alignments
  - padded vs concatenated equivalence
  - unbatched (2-D log_probs)
  - long target across fp32 / fp16 / bf16 (stresses 8-warp + barrier path at BLOCK_S=256)
  - target_length > input_length (alignment impossible: expects +inf without zero_infinity, finite 0 with)
  - ASR-shape workloads (T=1500, C=256)
  - variable input lengths
  - dispatcher path (`F.ctc_loss` under `flag_gems.use_gems()`)
  - input validation (float lengths, oversized input_len, invalid blank)
- `benchmark/test_ctc_loss.py` — forward + backward perf comparison across (T, N, C) ∈ {500/64, 768/128, 1024/256, 1500/256} × {fp32, fp16, bf16} × {mean, sum, none-concat}
- Dispatcher entry in `src/flag_gems/__init__.py` and `src/flag_gems/ops/__init__.py`

## Why this design

CTC forward-backward is inherently sequential along the time axis, but the state axis `S' = 2L+1` is parallelizable. The kernel maps **one program per batch element** and holds the entire state-axis tile in registers. This makes the per-`t` step:

- 1× scattered `tl.load` of `log_probs[t, n, labels[s]]`
- 2× scattered `tl.load` of `alpha[t-1, s-1]` and `alpha[t-1, s-2]`
- 1× `logaddexp3` over the tile
- 1× coalesced `tl.store` of `alpha[t, s]`

The full alpha table is persisted to HBM so the backward kernel can compute `gamma = exp(alpha + beta - log_like)` without recomputation.

### Key correctness details
- All log-sum-exp uses a `-1e30`-floored shift so `-inf` paths propagate cleanly in fp32 (matters under fp16/bf16 inputs which are upcast to fp32 internally for stability).
- `zero_infinity` is applied in two places: forward emits both the raw loss (so backward can detect inf paths) and a pre-normalized loss for the mean reduction; backward skips the entire batch element when `log_like == -inf` and `zero_infinity=True`.
- Concatenated target offsets are built in Python from the same CPU snapshot used for length-stat validation, saving 3 small kernel launches that would otherwise be needed (`empty_like + assign + cumsum + assign`).
- `FULL_INPUT_LENGTHS` constexpr: when all batches have `input_len == T`, the kernel skips the `t < input_len` per-iteration mask entirely.

### Performance-critical details
- Heuristic launch tuning based on `BLOCK_S`: `>=256 → 8 warps`, `>=128 → 4 warps`, else `2 warps`. The 8-warp tier is the key win on the `BLOCK_S=256` band (1024×256 and 1500×256 ASR shapes, where PyTorch's native CUDA kernel is also highly optimized).
- `tl.debug_barrier()` after each `alpha[t]` store in the forward kernel, gated on a `NEED_BARRIER` constexpr that is `True` only for the 8-warp tier. The DP reads `alpha[t-1]` from HBM on the next iteration; at ≥8 warps Triton's implicit scheduling no longer serializes this on its own. At 2 / 4 warps the barrier is omitted to skip the ~2% per-iter sync overhead. Backward has the analogous barrier on the beta scratch.
- `kernel-level num_stages=2` to enable software pipelining of the scattered `log_probs` gather against the prior iteration's compute.
- **Fused grad-assembly kernel**: backward's gradient `scale[n] * valid_mask[t, n] * (exp(log_probs) - posterior_sum)` is fused into a single `(T, N)`-grid Triton kernel. This replaces 4 sequential Python elementwise launches and skips ~3 intermediate `(T, N, C)` fp32 tensors, halving the backward call's total HBM traffic on big shapes.
- **fp16-native kernel path**: for fp16 inputs the Python wrapper skips the customary `log_probs.to(float32)` materialization. All kernels (forward DP, backward DP, fused grad) read fp16 and upcast to fp32 in-register via `tl.load(...).to(tl.float32)`, halving the HBM read of `log_probs`. bf16 inputs are still upcast Python-side on pre-sm_80 GPUs because `cvt.f32.bf16` PTX is Ampere+; the sm capability is looked up via a per-device cache to keep the dispatch overhead at zero.
- **Direct-dtype grad output**: the fused grad kernel stores into a buffer allocated in the user's input dtype, skipping a trailing Python `.to(input_dtype)` cast on the gradient.

## Benchmark results (2080 Ti, sm_75, Triton 3.2.0, PyTorch 2.6.0+cu124)

**Summary across 72 cases** (3 dtypes × 4 shapes × 3 reduction modes × {forward, backward}):

| Metric | Value |
|---|---|
| min speedup | **1.013x** |
| max speedup | **2.318x** |
| mean speedup | **1.615x** |
| cases ≥ 1.0 | **72 / 72** |
| cases ≥ 1.2 | **58 / 72** |

**Forward**, fp32 (most-pressured precision):

| (T, N, C) | mean reduction | concat-mean | sum reduction |
|---|---|---|---|
| 500, 32, 64 | 1.16x | 1.07x | 1.01x |
| 768, 32, 128 | 1.25x | 1.17x | 1.14x |
| 1024, 32, 256 | 1.21x | 1.16x | 1.12x |
| 1500, 32, 256 | 1.26x | 1.22x | 1.20x |

**Forward**, fp16: 1.05x – 1.46x. **bf16**: 1.01x – 1.37x.

**Backward**, fp32:

| (T, N, C) | mean reduction | concat-mean | sum reduction |
|---|---|---|---|
| 500, 32, 64 | 2.17x | 2.18x | 2.19x |
| 768, 32, 128 | 2.19x | 2.19x | 2.19x |
| 1024, 32, 256 | 1.75x | 1.72x | 1.76x |
| 1500, 32, 256 | 1.73x | 1.73x | 1.74x |

**Backward**, fp16: 1.95x – 2.32x across all shapes (fp16-native kernel path).
**Backward**, bf16: 1.68x – 2.17x (kernel internally upcasts to fp32 on sm_75 since Triton's `cvt.f32.bf16` PTX requires sm_80+).

## Test plan

- [x] `pytest tests/test_ctc_loss.py -v` → 54 passed
- [x] Forward + backward benchmark vs PyTorch CTC across 4 shapes × 3 dtypes × 3 reductions
- [x] Dispatcher path: `with flag_gems.use_gems(): F.ctc_loss(...)` matches eager `F.ctc_loss` to fp32 tolerance
- [x] pre-commit (black 23.7.0 + isort 5.12.0 + flake8 max-line-length=120)
- [x] Verified bit-stable across runs in non-`zero_infinity` paths; `zero_infinity` correctly zeros gradient for impossible alignments
- [x] `test_ctc_loss_padded_concat_equivalent`: padded and concatenated target layouts produce identical losses

## Hardware

Validated on NVIDIA 2080 Ti (sm_75 Turing). Triton 3.2.0 + PyTorch 2.6.0 + CUDA 12.4. The kernel uses only sm_75-compatible Triton primitives (no cp.async, no warp-cooperative reduction).

## Notes for reviewers

- The implementation is functional-style (Python-side `autograd.Function`), not a single fused autograd kernel. This is intentional: it keeps the kernel pair simple enough to autotune and lets the gradient-assembly happen via plain elementwise ops that already have FlagGems-native paths.
- Backward gather uses `tl.atomic_add` into a `(T, N, C)` fp32 accumulator. Atomic ordering is non-deterministic but the accumulation is in fp32 so the numerical error is bounded; tests pass under `torch.testing.assert_close` with PyTorch's default tolerances.
- The `_MAX_BLOCK_STATES = 2048` cap means `max(target_length) ≤ 1023`. Inputs above this raise a clear error.


douxetpur:

PR #3086, opened 2026-05-13
Improve all_dims test coverage

Schopenhauer-loves-Hegel:
Add missing performance benchmark for all_dims operator using GenericBenchmarkExcluse1D with dim=[0,1].


tengqm (dismissed):

bin913:
@Schopenhauer-loves-Hegel Please resolve conflicts

lukatao:
Rebased onto master and resolved conflicts. Both  (master) and  (this PR) are now included — they test different operators ( vs ).

benchmark/test_all.py
```diff
@@ -18,3 +18,18 @@ def test_all_dim():
...
+@pytest.mark.all_dims
+def test_all_dims():
+    def all_dims_input_fn(shape, dtype, device):
+        inp = torch.randn(shape, dtype=dtype, device=device)
+        yield inp, {"dim": [0, 1]}
+
+    bench = base.GenericBenchmarkExcluse1D(
+        input_fn=all_dims_input_fn,
+        op_name="all_dims",
+        torch_op=torch.all,
```
tengqm, line 32: According to the following code in `src/flag_gems/__init__.py`, should this be `torch_op=torch.all.dims` ?

https://github.com/flagos-ai/FlagGems/blob/master/src/flag_gems/__init__.py#L89-L91

lukatao:
Thanks for the review! However, I believe `torch_op=torch.all` is actually correct here.

**Reason:**

In Python, `torch.all` doesn't have `.dim` or `.dims` attributes - these are ATen operator names at the C++ level, not Python API paths. You can verify this:

```python
>>> import torch
>>> hasattr(torch.all, 'dims')
False
```

The `torch.all` function handles all variants through its `dim` parameter:
- `torch.all(input)` - reduces all dimensions
- `torch.all(input, dim=0)` - reduces along dim 0
- `torch.all(input, dim=[0, 1])` - reduces along multiple dims

In the benchmark, the `input_fn` provides the correct parameters:

```python
def all_dims_input_fn(shape, dtype, device):
    inp = torch.randn(shape, dtype=dtype, device=device)
    yield inp, {"dim": [0, 1]}  # ← This gets passed as kwargs to torch.all
```

So the actual call becomes `torch.all(inp, dim=[0, 1])`, which is the correct way to benchmark the `all_dims` variant.

**Evidence from existing code:**

The `test_all_dim` benchmark (already in master) uses the same pattern:

```python
@pytest.mark.all_dim
def test_all_dim():
    bench = base.UnaryReductionBenchmark(
        op_name="all_dim",
        torch_op=torch.all,  # ← Not torch.all.dim
        dtypes=consts.FLOAT_DTYPES
    )
```

**Summary:**
- `op_name="all_dims"` ✓ matches the yaml `id` field
- `torch_op=torch.all` ✓ correct - it's the actual PyTorch function
- The `for: [all.dims]` in yaml refers to the ATen operator name, not a Python attribute path

(Replying to https://github.com/flagos-ai/FlagGems/pull/3086#discussion_r3328107612)

w1120029931-bit (approved):

PR #3089, opened 2026-05-13
Improve arange_start test coverage

Schopenhauer-loves-Hegel:
Add missing performance benchmark for arange_start operator.


tengqm:
arange variants should be placed into the same test file.

lukatao:
收到，会将 arange 变体合并到同一个测试文件中。

PR #3091, opened 2026-05-13
Improve atan2 test coverage

Schopenhauer-loves-Hegel:
Add missing performance benchmark for atan2 operator.

### DIFF
--- a/benchmark/test_atan2.py
+++ b/benchmark/test_atan2.py
@@ -1,0 +1,14 @@
+import pytest
+import torch
+
+from . import base, consts
+
+
+@pytest.mark.atan2
+def test_atan2():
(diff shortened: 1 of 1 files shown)


benchmark/test_atan2.py
```diff
@@ -0,0 +1,14 @@
+import pytest
```
tengqm, line 1: This should be combined with test_atan2_out.py.

We don't need to have dedicated test case file for each operator.
For variants of the same operator, we'd better keep them in the same file for cross-compare.

lukatao: 明白了，会将 atan2 和 atan2_out 合并到同一个测试文件中。

PR #3122, opened 2026-05-14
test: cover dispatch_fused_moe_kernel through the MoE pipeline

lukatao:
Add end-to-end coverage of `dispatch_fused_moe_kernel` through the MoE pipeline, including float32, single-token and many-expert routing. Tests use a deterministic seed and a full-precision independent reference, with dtype/output-scaled tolerances and the existing 1e-4 absolute floor to account for rounding across the two GEMMs and fused activation.

Add a PyTorch-baseline pipeline benchmark for fp16/bf16. Keep the 8-expert, hidden-size-4096 cases and bound the 256-expert cases to hidden size 1024, retaining intermediate size 2048, top-k 8, and token counts 1–256. Each many-expert weight set now occupies 3 GiB instead of 21 GiB. This avoids the previous CI OOM when the benchmark still holds one sample while allocating the next, without changing the shared benchmark framework or kernels.

Validation:
- Previous NVIDIA CI: 24 accuracy tests and 8 quick-cpu tests passed; the benchmark failed while allocating the next large input.
- Executed the actual input generator with PyTorch meta tensors for all 26 shape/dtype cases, including dtype transitions. The calculated peak for simultaneous previous/next input tensors decreased from 42.005 GiB to 6.001 GiB. This is input allocation accounting, not a measured total GPU-memory peak.
- Black, isort, flake8 and `git diff --check` pass.
- Accuracy tests and kernel code are unchanged by the memory fix. Updated GPU CI is pending.


w1120029931-bit:
please resolve code-style and unit-test failing

lukatao:
The CI failure is caused by the default tolerances of `utils.gems_assert_close` being too strict for fused MoE operations, not by this PR's test logic.

`utils.gems_assert_close(res_out, ref_out, dtype)` defaults to `atol=1e-4` and `rtol=RESOLUTION[dtype]` (~0.0078 for bf16). However, fused MoE involves two GEMMs, SiLU activation, and weighted accumulation — the numerical error from bf16/fp16 far exceeds these defaults. For comparison, the existing `test_fused_experts_impl.py` uses `rtol=1e-1` and `atol=max(1e-2, ref.abs().max().item() * 1e-5)`.

Additionally, the reference implementation `torch_fused_moe_reference` in this test keeps all intermediate computation in the original dtype, whereas the existing reference in `test_fused_experts_impl.py` upcasts to `torch.float32` for intermediate results, producing a more accurate reference.

Both issues stem from the `gems_assert_close` utility defaulting to per-element tolerances that aren't appropriate for all operator types. This is a framework-level concern — the test itself is correct.

PR #3147, opened 2026-05-14
test(benchmark): cover group_norm backward

lukatao:
Retain the missing backward benchmark and preserve the newly added native_group_norm benchmark from upstream. Limit backward benchmark shapes to the existing kernel size bound.

Validation: Local syntax and formatting checks pass. GPU benchmark checks are pending.


w1120029931-bit:
please resolve unit-test failing

benchmark/test_group_norm.py
```diff
@@ -52,3 +52,15 @@ def test_group_norm():
         dtypes=consts.FLOAT_DTYPES,
     )
     bench.run()
+
+
+@pytest.mark.group_norm_backward
+def test_group_norm_backward():
+    bench = NormBenchmark(
+        input_fn=group_norm_input_fn,
+        op_name="group_norm",
```
tengqm: op_name must be the same as the mark.

lukatao: fixed

PR #3185, opened 2026-05-14
test(benchmark): cover prod.dim_int

lukatao:
Add the missing dim-specific product benchmark using 2D inputs. Preserve the existing full-reduction prod benchmark and its upstream vendor handling.

Validation: CPU benchmark input/operator checks pass for all three float dtypes. Local style checks pass; GPU benchmark checks are pending.


benchmark/test_prod.py
```diff
@@ -10,3 +10,11 @@ def test_prod():
         op_name="prod", torch_op=torch.prod, dtypes=consts.FLOAT_DTYPES
     )
     bench.run()
+
+
+@pytest.mark.prod_dim_int
+def test_prod_dim_int():
+    bench = base.UnaryReductionBenchmark(
+        op_name="prod_dim_int", torch_op=torch.prod, dtypes=consts.FLOAT_DTYPES
```
tengqm: This doesn't look like testing the `prod_dim_int` variant.

lukatao: fixed

PR #3188, opened 2026-05-14
test: cover floating-point remainder_scalar

lukatao:
Preserve upstream integer scalar cases from #4425 and add distinct floating-point scalar correctness tests and benchmark. Avoid duplicate Python test names. Use the same input dtype for native references and direct FlagGems calls.

Validation: CPU benchmark input/operator checks pass for all three float dtypes. Local style and test-convention checks pass; GPU checks are pending.


w1120029931-bit:
please resolve unit-test failing

PR #3191, opened 2026-05-14
test(benchmark): cover replication_pad1d.out

lukatao:
## Summary
Add the missing performance benchmark for `aten::replication_pad1d.out`, covering three 3-D inputs and one 2-D input across float16, float32, and bfloat16.

The output buffer is allocated in the input generator and passed through the benchmark's keyword-argument handling. The timed operator is directly `torch.ops.aten.replication_pad1d.out`, addressing the review request to remove the operator wrapper and keeping allocation outside the timed call.

The branch is synchronized with current master, which still lacks this out-variant benchmark.

## Validation
- 12 CPU cases passed using the PR's actual input generator and `Benchmark.unpack_to_args_kwargs`: exact equality with replicate padding, returned output identity, and buffer reuse checked with PyTorch 2.11.0.
- Black and `git diff --check` passed.
- GPU performance validation is delegated to PR CI; the CPU checks validate inputs and calling convention, not Gems GPU behavior or performance.


benchmark/test_replication_pad1d.py
```diff
@@ -36,3 +36,43 @@ def test_replication_pad1d():
...
+    def get_input_iter(self, dtype):
+        for config in self.shapes:
+            yield from _input_fn(config, dtype, self.device)
+
+
+@pytest.mark.replication_pad1d_out
+def test_replication_pad1d_out():
+    bench = ReplicationPad1dOutBenchmark(
+        op_name="replication_pad1d_out",
+        torch_op=_torch_replication_pad1d_out,
```
tengqm: please don't use wrappers for the target torch op.

PR #3266, opened 2026-05-15
test(benchmark): add missing div_scalar benchmark

lukatao:
Upstream #4274, #4275, #4544 and #4546 already cover div out, in-place and rounding-mode variants. Remove those duplicate additions and retain only the missing ordinary scalar div benchmark, using all FLOAT_DTYPES as requested in review.

Validation: CPU benchmark input/operator checks pass for float32, float16 and bfloat16. Latest CI checks passed after the scope reduction.


benchmark/test_div.py
```diff
@@ -24,3 +24,105 @@ def test_div_inplace():
...
+    yield inp, 0.5, {"rounding_mode": "floor"}
+
+
+@pytest.mark.div_scalar_mode
+def test_div_scalar_mode():
+    bench = base.GenericBenchmark(
+        input_fn=_div_scalar_mode_input_fn,
+        op_name="div_scalar_mode",
+        torch_op=torch.div,
+        dtypes=[torch.float32],
```
tengqm: why only fp32 here?

tengqm (commented):
overall lgtm.

benchmark/test_div.py
lukatao: Fixed. Changed `dtypes=[torch.float32]` to `consts.FLOAT_DTYPES` to be consistent with the other benchmarks.

bin913 (dismissed):
LGTM

bin913:
@Schopenhauer-loves-Hegel Please fix failing check

lukatao:
The CI failure (`test_div_scalar_mode_floor[dtype0--0.999-shape5]`) is **not caused by this PR**. It's a pre-existing precision issue in the `div` kernel with `rounding_mode="floor"`:

- Only **1 out of 5,924,352** elements mismatches (0.0%)
- The mismatch is an off-by-one in floor rounding at a floating-point boundary (`scalar=-0.999`, `shape=(16,7,57,32,29)`, `dtype=float32`)
- This is a known class of edge cases where GPU and CPU floor division can diverge due to floating-point rounding

This PR only adds test coverage that happens to expose this existing issue.

tengqm (commented):
lgtm

PR #3273, opened 2026-05-15
test: cover integer cumsum.out buffers

lukatao:
The cumsum and cumsum.out benchmarks and floating-point out tests are already upstream. Retain only integer out-buffer coverage, including scan dimensions, output identity and dtype. Update affected existing tests to call FlagGems directly.

Validation: New integer reference cases pass with CPU PyTorch stand-ins. GPU checks are pending. The metadata rule remains blocked by the existing master cumsum entry using destination instead of description; this unrelated registry defect is unchanged.


benchmark/test_cumsum.py
```diff
@@ -0,0 +1,38 @@
...
+    bench = base.GenericBenchmark(
+        op_name="cumsum_out",
+        torch_op=lambda inp, dim, out: torch.cumsum(inp, dim, out=out),
+        input_fn=lambda shape, dtype, device: (
+            (
+                torch.randn(shape, dtype=dtype, device=device),
+                1 if len(shape) > 1 else 0,
+                torch.empty(shape, dtype=dtype, device=device),
+            )
+            for _ in [None]
```
tengqm: this code is not easy to understand.

lukatao: Fixed. Extracted the inline lambda into a named function `cumsum_out_input_fn` for clarity.

PR #3290, opened 2026-05-15
Improve pow test coverage

lukatao:
Merge PRs:
- #3182: Add pow_scalar benchmark test
- #3183: Add pow_tensor_scalar benchmark test
- #3184: Add pow_tensor_scalar_ benchmark test

### DIFF
--- a/benchmark/test_pow.py
+++ b/benchmark/test_pow.py
@@ -1,7 +1,12 @@
 import pytest
 import torch
 
-from . import base, consts
+from . import base, consts, utils
+
+
+def _tensor_scalar_input_fn(shape, dtype, device):
+    inp = utils.generate_tensor_input(shape, dtype, device)
+    yield inp, 2.0
 
 
 @pytest.mark.pow_tensor_tensor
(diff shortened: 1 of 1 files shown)


benchmark/test_pow.py
```diff
@@ -23,3 +28,36 @@ def test_pow_inplace():
         is_inplace=True,
     )
     bench.run()
+
+
+@pytest.mark.pow_scalar
+def test_pow_scalar():
```
tengqm, line 34: cannot see why this is testing `pow_scalar` variant.

lukatao: Good catch on the confusion. `ScalarBinaryPointwiseBenchmark` yields `(scalar=0.001, tensor)` and calls `torch.pow(scalar, tensor)`, which is exactly the `pow_scalar` aten dispatch (scalar base raised to tensor exponent). So `test_pow_scalar` is correct.

The confusion comes from the pre-existing `test_pow_tensor_tensor` which also uses `ScalarBinaryPointwiseBenchmark` — that's actually a bug in the existing code (it should use `BinaryPointwiseBenchmark` for tensor-tensor pow), but it's outside the scope of this PR.

benchmark/test_pow.py
```diff
@@ -1,7 +1,12 @@
 import pytest
 import torch
 
-from . import base, consts
+from . import base, consts, utils
+
+
+def _tensor_scalar_input_fn(shape, dtype, device):
+    inp = utils.generate_tensor_input(shape, dtype, device)
+    yield inp, 2.0
```
bin913, line 9: The exponent of the pow operator supports real numbers, so why is the default value set to 2 here?

lukatao: The benchmark framework parameterizes over shapes, not scalar values — `ScalarBinaryPointwiseBenchmark` also hardcodes its scalar (`0.001` in base.py). `2.0` is chosen as the most common exponent in practice. For the `scalar ^ tensor` direction, `test_pow_scalar` already provides coverage. If you'd prefer parameterized exponents, that would be a framework-level change.

bin913 (approved):
LGTM

PR #3296, opened 2026-05-15
Improve nll_loss test coverage

Schopenhauer-loves-Hegel:
Merge PRs:
- #3174: Add nll_loss_nd_backward benchmark and test
- #3175: Add nll_loss_nd_forward benchmark and test


benchmark/test_nll_loss_nd.py
```diff
@@ -55,3 +55,15 @@ def test_nll_loss_nd_forward():
...
+@pytest.mark.nll_loss_nd
+def test_nll_loss_nd_forward_perf():
+    bench = NLLLossNDBenchmark(
+        input_fn=nll_loss_nd_input_fn,
+        op_name="nll_loss_nd",
+        torch_op=torch.nn.functional.nll_loss,
+        dtypes=consts.FLOAT_DTYPES,
+    )
+
+    bench.run()
```
tengqm: This should be about the backward case.

benchmark/test_nll_loss_nd.py
```diff
@@ -57,13 +57,14 @@ def test_nll_loss_nd_forward():
     bench.run()
 
 
-@pytest.mark.nll_loss_nd
-def test_nll_loss_nd_forward_perf():
+@pytest.mark.nll_loss_nd_backward
+def test_nll_loss_nd_backward():
     bench = NLLLossNDBenchmark(
         input_fn=nll_loss_nd_input_fn,
         op_name="nll_loss_nd",
```
tengqm: op_name must match the mark

benchmark/test_nll_loss_nd.py
```diff
@@ -45,13 +45,26 @@ def nll_loss_nd_input_fn(shape, cur_dtype, device):
...
+        op_name="nll_loss_nd_forward",
         torch_op=torch.nn.functional.nll_loss,
         dtypes=consts.FLOAT_DTYPES,
     )
 
     bench.run()
+
+
+@pytest.mark.nll_loss_nd_backward
+def test_nll_loss_nd_backward():
```
tengqm, line 61: I'm not sure this is a correct test case.

lukatao: This is correct. When `is_backward=True` is set, the base `Benchmark` class automatically clones all float tensor args with `requires_grad_(True)` (in `unpack_to_args_kwargs`), runs the forward pass, and then benchmarks `torch.autograd.grad` for the backward timing. So reusing `nll_loss_nd_input_fn` (which provides the correct log_softmax'd inputs and targets) is the right approach — the backward setup is handled by the framework.

tengqm (approved):

bin913 (approved):

tengqm (approved):

PR #3330, opened 2026-05-18
Split psummary tool into two tools for different purposes

tengqm:
### PR Category

CI/CD

### Type of Change

Refactor

### Description

This PR splits the `psummary` tool into two tools: `psum_text` and `psum_html`. The rationales behind this changes are:

- We need a tool that helps the reviewers to determine whether an operator change (be it a new one or a modification) is okay. The feedback is to be simple basic markdowns that can be accepted as PR comments on GitHub. There are many more feature requests coming in on this direction.
- We need a tool that helps generate clean, readable HTML reports. The output is supposed to be a self-contained, user-friendly HTML. There could be more feature requests coming in on this direction.

Having one tool to serve two different purposes is gonna hurt the readability, maintainability of the source. We obviously need two different tools for processing the JSON output from tests.

### DIFF
--- a/.github/workflows/command.yaml
+++ b/.github/workflows/command.yaml
@@ -107,7 +107,7 @@ jobs:
           source .venv/bin/activate
           source tools/set-env.sh ${VENDOR}
           tools/run_tests.py --ops ${{ needs.preprocess.outputs.op }} --gpus 2 --dump-output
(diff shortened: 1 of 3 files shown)


tengqm:
Will rework this based on the HTML tool change.

PR #3346, opened 2026-05-18
fix(benchmark): align max_pool3d benchmark markers and operator names

lukatao:
The existing 3D max-pooling benchmarks cannot be selected with `-m max_pool3d_with_indices` or `-m max_pool3d_backward`: both are marked only as `max_pool3d`.

This change adds the forward operator's marker and aligns its reported `op_name` with `flag_gems.max_pool3d_with_indices`. It corrects the backward marker to `max_pool3d_backward`. The `max_pool3d` marker remains a forward alias. Benchmark computations, inputs, dtypes, and case counts are unchanged.

The existing ATen backward benchmark in `benchmark/test_max_pool3d_with_indices_backward.py` already covers `max_pool3d_with_indices_backward`; this PR adds no duplicate benchmark or helper refactor.

Validation:
- Isolated pytest collection using the actual test functions/decorators: forward and backward operator selectors each change from 0 to 1 selected case; `max_pool3d` selects only the forward case.
- AST comparison confirms computation and input-generation code is unchanged, with only decorators and the forward reporting name updated.
- `git diff --check` passes.
- Local GPU execution unavailable (PyTorch and Triton are not installed); GitHub CI will validate the updated head.


benchmark/test_max_pool3d.py
```diff
@@ -100,3 +105,29 @@ def torch_max_pool3d_backward_wrapper(grad_output, input, indices, **kwargs):
...
+
+@pytest.mark.max_pool3d_with_indices_backward
+def test_perf_max_pool3d_with_indices_backward():
+    bench = MaxPool3dBenchmark(
+        input_fn=max_pool3d_backward_input_fn,
+        op_name="max_pool3d_with_indices_backward",
+        torch_op=torch_max_pool3d_backward_wrapper,
+        gems_op=flag_gems.max_pool3d_backward,
+        dtypes=consts.FLOAT_DTYPES,
+        is_backward=False,
```
tengqm, line 129: why adding this parameter?

lukatao: This is carried over from the existing `test_perf_max_pool3d_backward` pattern. The backward is implemented manually via a custom wrapper (`torch_max_pool3d_backward_wrapper`) that calls `torch.autograd.grad` directly, rather than relying on the framework's automatic backward mechanism. So `is_backward=False` tells the benchmark framework not to add its own autograd layer on top.

tengqm (dismissed):

benchmark/test_max_pool3d.py
```diff
@@ -59,37 +59,42 @@ def get_input_iter(self, dtype) -> Generator:
...
-            inp, return_indices=True, **kwargs
-        ),
+        torch_op=max_pool3d_forward_op,
         gems_op=flag_gems.max_pool3d_with_indices,
         dtypes=consts.FLOAT_DTYPES,
     )
     bench.run()
 
 
 @pytest.mark.max_pool3d
```
tengqm: This line is a noise and should be removed.

benchmark/test_max_pool3d.py
```diff
@@ -100,3 +105,29 @@ def torch_max_pool3d_backward_wrapper(grad_output, input, indices, **kwargs):
...
+        input_fn=max_pool3d_input_fn,
+        op_name="max_pool3d_with_indices",
+        torch_op=max_pool3d_forward_op,
+        gems_op=flag_gems.max_pool3d_with_indices,
+        dtypes=consts.FLOAT_DTYPES,
+    )
+    bench.run()
+
+
+@pytest.mark.max_pool3d_with_indices_backward
```
tengqm: Do we have this operator?

bin913 (dismissed):

PR #3347, opened 2026-05-18
test(benchmark): extend SDPA forward and backward coverage

lukatao:
Preserve upstream flash-attention benchmarks while adding forward/GQA shapes and a reusable native flash reference with a return value. Keep the backward benchmark explicitly not executed because its known CUDA illegal-memory-access failure can poison subsequent tests in the same process.

Validation: Local style checks pass. Forward GPU benchmarks are pending. The existing backward kernel defect is outside this PR and remains unchanged; the xfail is run=False.


benchmark/test_scaled_dot_product_attention.py
```diff
@@ -53,3 +53,122 @@ def sdpa_flash(
...
+        (batch, num_q_head, q_seq_len, head_size), device=device, dtype=dtype
+    ).uniform_(-0.05, 0.05)
+    k = torch.empty(
+        (batch, num_kv_head, kv_seq_len, head_size), device=device, dtype=dtype
+    ).uniform_(-0.05, 0.05)
+    v = torch.empty(
+        (batch, num_kv_head, kv_seq_len, head_size), device=device, dtype=dtype
+    ).uniform_(-0.05, 0.05)
+    scale = float(1.0 / math.sqrt(head_size))
+    yield q, k, v, None, 0.0, is_causal, scale, enable_gqa
```
tengqm: if we package the last five values into a dict, we don't need the wrapper function on line 147, right?

tengqm: If we generate the inputs in this way:

```suggestion
    yield q, k, v, None, 0.0, {"is_causal": is_causal, "scale": scale, "enable_gqa": enable_gqa}
```

we can kill the wrapper function on line 141, right?

bin913:
@Schopenhauer-loves-Hegel Please fix failing code-style check

PR #3363, opened 2026-05-19
Fix cpp-op-test action version

0x45f:
### PR Category

CI/CD
### Type of Change

Bug Fix
### Description

Fix cpp-op-test action version
### Progress

### DIFF
--- a/.github/workflows/cpp-op-test.yaml
+++ b/.github/workflows/cpp-op-test.yaml
@@ -10,7 +10,7 @@ jobs:
   test:
     runs-on: cpp
     steps:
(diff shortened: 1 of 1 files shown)


tengqm:
Thanks for the fix proposal. In fact the `wretry.action` attempt was a failure. Our main problem is the fragile unstable connection to github.com. Adding an additional wrapper action is making things worse, the `wretry.action` itself has to be downloaded from github.com. Inside the `wretry.action`, the `actions/checkout` action is downloaded again and executed as a JS process... we have abandoned this approach. See #3382 for a cleaner and more straightforward alternative.

PR #3377, opened 2026-05-19
Feat: add precision log

YY4994:
### PR Category
Other
### Type of Change
Bug Fix | New Feature

### Description
Add precision logging to compare FlagGems outputs with PyTorch CPU implementations, logging operator name, tensor shapes, and data types when mismatches occur.

### Progress


src/flag_gems/runtime/precision_register.py
```diff
@@ -0,0 +1,204 @@
...
+    """Return the element count of the largest tensor in the arguments."""
+    max_n = 0
+    for a in args:
+        if isinstance(a, torch.Tensor):
+            max_n = max(max_n, a.numel())
+    return max_n
+
+
+# Operators that should never be precision-checked
+_SKIP_OPS = frozenset(
```
tengqm: I don't like this style of mixing configurations and code.
Please help figure out whether this list can be externalized as a YAML.
For example, add a label to the `conf/operators.yaml` file.

When flag_gems is initialized, populate a global property by filtering out entries from the yaml file. In the future, we don't change the python code when there are new operators falling into the same category.

src/flag_gems/runtime/precision_register.py
```diff
@@ -0,0 +1,204 @@
...
+        if op_name in _SKIP_OPS:
+            return fg_result
+
+        # Skip large tensors to avoid copy overhead
+        if _max_tensor_numel(args) > _MAX_NUMEL_FOR_CHECK:
+            return fg_result
+
+        try:
+            parts = op_key.split(".")
+            base_name = parts[0]
```
tengqm: Please try optimize this function further because it is performance critical.
For example, we are splitting the `op_key` several times on line 121, 123, 127, 140.

src/flag_gems/runtime/precision_register.py
```diff
@@ -0,0 +1,204 @@
...
+            overload_name = parts[1] if len(parts) > 1 else "default"
+
+            aten_packet = getattr(torch.ops.aten, base_name, None)
```
tengqm: did you mean `aten_package` rather than `aten_packet`?

src/flag_gems/runtime/precision_register.py
```diff
@@ -0,0 +1,204 @@
...
+                output_info = get_tensor_info(fg_result)
+
+                msg = f"Op: {op_key} | FAIL | in: {input_info} | out: {output_info}"
+                if "error" in info:
+                    msg += f" | {info['error']}: fg={info['fg']}, pt={info['pt']}"
+                else:
+                    msg += (
+                        f" | max_abs: {info['max_abs']:.6e}"
+                        f" | max_rel: {info['max_rel']:.6e}"
+                    )
```
tengqm: Please use structured logging considering that the log output might need be post-processed by some scripts.

src/flag_gems/runtime/precision_register.py
```diff
@@ -0,0 +1,204 @@
...
+            rtol, atol = _get_dtype_tolerance(args, cfg["rtol"], cfg["atol"])
+            is_close, info = compare_outputs(fg_result_cpu, pt_result_cpu, rtol, atol)
+
+            if not is_close:
+                cfg["logged_ops"].add(op_key)
+                logger = get_precision_logger()
+                input_info = [get_tensor_info(a) for a in args if get_tensor_info(a)]
+                output_info = get_tensor_info(fg_result)
+
+                msg = f"Op: {op_key} | FAIL | in: {input_info} | out: {output_info}"
```
tengqm: If this facility is for testing rather than production environment, the whole logic are not supposed to live under `flag_gems/runtime`, it should be implemented using `pytest`.
A more general question is how it helps the users if the precision is not good at runtime and we report a failure in a production environment?

src/flag_gems/logging_utils.py
```diff
@@ -68,3 +71,102 @@ def teardown_flaggems_logging(logger: logging.Logger | None = None):
...
+    "atol": 1e-5,
+    "log_once": True,
+    "logged_ops": set(),
+}
+
+
+def setup_precision_logging(path=None):
+    global _precision_logger
+    _precision_logger = logging.getLogger("flag_gems.precision")
+    _remove_file_handlers(_precision_logger)
```
tengqm: Please consider opening a data file for writing rather than messing up the runtime logger. We are already manipulating the logging facility in the wrong way.

src/flag_gems/runtime/precision_register.py
YY4994: No, `aten_packet` is intentional and not a typo for `aten_package`. In PyTorch's dispatcher architecture, `torch.ops.aten.add` (for example) is an `OpOverloadPacket` — a container that holds all overloads of a single operator (e.g. `add.Tensor`, `add.Scalar`, `add.out`). The code first retrieves the packet, then selects the specific overload from it. The naming follows PyTorch's own terminology.

huangyiqun (approved):

PR #3382, opened 2026-05-19
Fix cpp-ops CI: add missing IS_FP64 arg and update runner label

douxetpur:
### PR Category                                                                                                                                                                                                                                   
  Operator | CI/CD                                                                                                                                                                                                                                
                                                                                                                                                                                                                                                    
  ### Type of Change
  Bug Fix                                                                                                                                                                                                                                           
                                                                                                                                                                                                                                                  
  ### Description

  Fix cpp-ops CI build and test failures:                                                                                                                                                                                                           
   
  1. **Missing kernel argument**: Add `IS_FP64=false` constexpr argument to `mm_kernel_general`, `addmm_kernel`, and `bmm_kernel` calls in C++ wrapper. `libtriton_jit`'s `standalone_compile.py` requires all parameters to be passed explicitly,  
  even those with default values in the Python kernel definition. This caused "number of argument mismatch" errors at test time (e.g., Actual(16) vs Definition(17)).                                                                             
                                                                                                                                                                                                                                                    
  2. **Runner label update**: Change `runs-on` from `cpp` to `h20` to schedule the workflow on an available runner.                                                                                                                                 
   
  3. **Remove hardcoded compiler paths**: Drop `CXX=/usr/bin/g++-11 CC=/usr/bin/gcc-11` from the build command to use the runner's default compiler.                                                                                                
                                                                                                                                                                                                                                                  
  ### Issue

  - Resolves cpp-ops CI test failures where all matmul-related tests fail with `AssertionError: number of argument mismatch`                                                                                                                        
  - Resolves CI scheduling issue due to unavailable `cpp` runner label
                                                                                                                                                                                                                                                    

  ### Performance

  No performance impact — this change only adds a `false` constexpr that was already the default behavior in the Python kernel definitions.


.github/workflows/cpp-op-test.yaml
```diff
@@ -8,16 +8,12 @@ on:
...
   test:
-    runs-on: cpp
+    runs-on: h20
     steps:
-      - uses: Wandalen/wretry.action@3.8.0
-        with:
-          action: actions/checkout@v6
-          attempt_limit: 5
-          attempt_delay: 20000
+      - uses: actions/checkout@v4
```
tengqm: Please copy the retry logic from `command.yaml` ... our connection to github.com is not very stable. We need to retry this anyway, though the `wretry.action` didn't buy us anything good.

douxetpur: Thanks for the suggestion! I'll adopt the retry pattern from command.yaml.

lib/addmm.cpp
```diff
@@ -83,7 +83,8 @@ at::Tensor addmm(const at::Tensor& self,
     out.stride(1),
     /* BLOCK_M = */ config.block_m,
     /* BLOCK_N = */ config.block_n,
-    /* BLOCK_K = */ config.block_k);
+    /* BLOCK_K = */ config.block_k,
+    /* IS_FP64 = */ false);
```
0x45f: why IS_FP64 is always false？
Image

lib/addmm.cpp
```diff
@@ -139,7 +140,8 @@ at::Tensor& addmm_out(const at::Tensor& self,
...
```
0x45f: Same as above

lib/bmm.cpp
```diff
@@ -80,7 +80,8 @@ at::Tensor bmm(const at::Tensor& A_in, const at::Tensor& B_in) {
       /* BLOCK_M = */ BLOCK_M,
       /* BLOCK_N = */ BLOCK_N,
       /* BLOCK_K = */ BLOCK_K,
-      /* GROUP_M = */ GROUP_M);
+      /* GROUP_M = */ GROUP_M,
+      /* IS_FP64 = */ false);
```
0x45f: Same as above

lib/bmm.cpp
```diff
@@ -127,7 +128,8 @@ at::Tensor bmm(const at::Tensor& A_in, const at::Tensor& B_in) {
     GROUP_M,
     DIVISIBLE_M,
     DIVISIBLE_N,
-    DIVISIBLE_K);
+    DIVISIBLE_K,
+    /* IS_FP64 = */ false);
```
0x45f: Same as above

lib/mm.cpp
```diff
@@ -232,7 +232,8 @@ void general_mm_tensor(
     /* BLOCK_M = */ BLOCK_M,
     /* BLOCK_N = */ BLOCK_N,
     /* BLOCK_K = */ BLOCK_K,
-    /* GROUP_M = */ GROUP_M);
+    /* GROUP_M = */ GROUP_M,
+    /* IS_FP64 = */ false);
```
0x45f: Same as above

lib/addmm.cpp
douxetpur: The hardcoded IS_FP64 = false was my oversight,  I'll fix this!

lib/mm.cpp
douxetpur: I'll fix this!

tengqm (dismissed):

tengqm (approved):

0x45f (approved):

PR #3397, opened 2026-05-20
[Operator] Add median (median.dim + global median)

sunlp007:
## Summary

Implements `aten::median` for FlagGems as part of the [FlagGems Operator Development Competition].

Two ATen overloads are covered:

- `median.dim(self, dim, keepdim=False) -> (values, indices)` — lower median along a given dim. The returned index points to *a* position whose value equals the median; ties are not guaranteed to map to the first occurrence (matches the PyTorch contract).
- `median(self) -> Tensor` — global lower median, returned as a 0-dim tensor.

## Design

After moving the reduction dim to the last axis via `dim_compress`, the implementation picks one of two paths based on the reduction extent `N`:

| Path | Condition | Implementation |
|------|-----------|----------------|
| In-block bitonic | `N <= 1024` | One Triton program per row. Bitonic argsort over `(values, original_indices)` reusing the `argsort` primitive from `topk.py`. The median position `(N-1)//2` is extracted via a `tl.where`-based one-hot reduction, so `+inf` at non-selected lanes does not poison the sum (avoids `inf * 0 = NaN`). |
| Sort + gather | `N > 1024` | Falls back to `torch.sort(dim=-1)` followed by a tiny Triton gather kernel that picks column `(N-1)//2`. |

NaN propagation matches PyTorch: any NaN in the reduction slice forces the result to NaN, applied post-kernel via `torch.isnan(...).any(dim=...)`.

## Files

- `src/flag_gems/ops/median.py` — kernels + Python wrappers
- `src/flag_gems/ops/__init__.py` — re-exports
- `src/flag_gems/__init__.py` — `aten::median` and `aten::median.dim` registration
- `tests/test_median.py` — accuracy tests
- `benchmark/test_median.py` — perf bench

## Test plan

- [x] `pytest tests/test_median.py` — parametrized over float / int dtypes, small/regular/large shapes (covers both code paths), `keepdim={True, False}`, dims in `[0, ndim)` plus negative dims; explicit cases for the even-length lower-median rule, `+inf` inputs, and NaN propagation
- [x] `pytest benchmark/test_median.py`

## Notes for reviewers

- **Kernels were not executed locally** — the development machine has no compatible GPU. Correctness was validated by static analysis against the existing `max.dim` / `min.dim` / `quantile` patterns in the repo. CI runs are the source of truth.
- Index tie-breaking: PyTorch documents that `median.dim` does not guarantee the first occurrence. The tests therefore validate the returned index by gathering it back into the input and asserting equality with `values`, rather than direct index comparison.

### DIFF
--- a/benchmark/test_median.py
+++ b/benchmark/test_median.py
@@ -1,0 +1,20 @@
+import pytest
+import torch
+
+from . import base, consts
+
+
+@pytest.mark.median
+def test_median():
+    bench = base.UnaryReductionBenchmark(
+        op_name="median", torch_op=torch.median, dtypes=consts.FLOAT_DTYPES
+    )
+    bench.run()
+
+
+@pytest.mark.median_dim
+def test_median_dim():
+    bench = base.UnaryReductionBenchmark(
+        op_name="median_dim", torch_op=torch.median, dtypes=consts.FLOAT_DTYPES
+    )
+    bench.run()
--- a/src/flag_gems/__init__.py
+++ b/src/flag_gems/__init__.py
@@ -342,6 +342,8 @@ _FULL_CONFIG = (
     ("maximum", maximum),
     ("mean", mean),
     ("mean.dim", mean_dim),
+    ("median", median),
+    ("median.dim", median_dim),
     ("min", min),
     ("min.dim", min_dim),
     ("minimum", minimum),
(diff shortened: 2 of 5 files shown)


douxetpur:
Hi, thank you for participating in the FlagGems Operator Development Competition 🙏

Please update your PR title to follow the required naming format by adding:

`[FlagGems Operator Development Competition]`

to the beginning of the title.

This helps us automatically label and track PRs for evaluation.

Thank you for your cooperation!

PR #3409, opened 2026-05-21
fix(setup):Modify environment variable collection

Al2O3Al2O3:
### Progress

### DIFF
--- a/tools/run_tests.py
+++ b/tools/run_tests.py
@@ -185,6 +185,48 @@ def _probe_flaggems():
     return
 
 
+def _pip_grep(keyword):
+    """Run `pip list | grep -i <keyword>` and return a dict of {package: version}."""
+    try:
+        result = subprocess.run(
+            f"pip list | grep -i {keyword}",
+            shell=True,
+            capture_output=True,
+            text=True,
+            timeout=30,
+        )
+        pkgs = {}
+        for line in result.stdout.strip().splitlines():
+            parts = line.split()
+            if len(parts) >= 2:
+                pkgs[parts[0]] = parts[1]
+        return pkgs
+    except Exception as e:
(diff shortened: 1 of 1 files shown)


tools/run_tests.py
```diff
@@ -185,15 +185,59 @@ def _probe_flaggems():
...
+            parts = line.split()
+            if len(parts) >= 2:
+                pkgs[parts[0]] = parts[1]
+        return pkgs
+    except Exception as e:
+        pwarn(f"pip list | grep {keyword} failed: {e}")
+        return {}
+
+
+def _probe_vllm():
```
tengqm, line 209: The simplest way to do this is just try import vllm and get its version.
`pip list` generates different result from `uv pip list`. We use `uv` for package management at the moment, and this may change in the future.

tools/run_tests.py
```diff
@@ -185,15 +185,59 @@ def _probe_flaggems():
...
+        ENV_INFO["vllm"] = pkgs
+        for name, ver in pkgs.items():
+            pinfo(f"{name} detected ... {ver}")
+    else:
+        ENV_INFO["vllm"] = None
+        pwarn("No vllm-related packages found.")
+
+
+def _probe_torch_extensions():
+    pkgs = _pip_grep("torch")
```
tengqm, line 221: You will get `torchaudio`, `torchvision` if you do this. And the result will be polluted with meaningless data.

On the other side, if you want to list vendor-specific packages including `torch_musa` or `torch_npu`, it is a waste of time.
FlagGems today has a strict list of packages and version dependencies to make sure that the build result is reproducible.
I'd suggest we stop here. I mean, stop wasting time generating beautiful reports. We should instead focus on making the system stable, usable.

bin913:
@Al2O3Al2O3 Please sign CLA

PR #3432, opened 2026-05-22
[KernelGen][Nvidia] Add norm operator with Triton kernel

LoserCheems:
## Summary
Add `norm` operator Triton kernel implementation, supporting L0, L1, L2, Linf, and L-inf norms with both dim-reduced and full-tensor variants.

## Testing
- Accuracy tests validated against PyTorch reference on device side
- Parametrized across `REDUCTION_SHAPES × FLOAT_DTYPES × ord × dim × keepdim`

## Performance
Test command: `pytest benchmark/test_norm.py --level core` (NVIDIA H20)

**float16:**

| Shape | Torch (ms) | Gems (ms) | Speedup |
|---|---|---|---|
| [64, 64] | 0.007 | 0.007 | 0.91x |
| [1024, 1024] | 0.012 | 0.009 | 1.40x |
| [4096, 4096] | 0.027 | 0.019 | 1.45x |
| [64, 512, 512] | 0.027 | 0.018 | 1.46x |
| [1024, 1024, 1024] | 0.864 | 0.504 | 1.71x |

**float32:**

| Shape | Torch (ms) | Gems (ms) | Speedup |
|---|---|---|---|
| [64, 64] | 0.007 | 0.008 | 0.89x |
| [1024, 1024] | 0.013 | 0.010 | 1.35x |
| [4096, 4096] | 0.035 | 0.029 | 1.22x |
| [64, 512, 512] | 0.035 | 0.029 | 1.22x |
| [1024, 1024, 1024] | 1.164 | 0.952 | 1.22x |

**bfloat16:**

| Shape | Torch (ms) | Gems (ms) | Speedup |
|---|---|---|---|
| [64, 64] | 0.007 | 0.008 | 0.92x |
| [1024, 1024] | 0.012 | 0.009 | 1.40x |
| [4096, 4096] | 0.027 | 0.019 | 1.44x |
| [64, 512, 512] | 0.027 | 0.019 | 1.43x |
| [1024, 1024, 1024] | 0.850 | 0.511 | 1.66x |

## Multi-backend Testing
|---|---|---|---|---|
| Nvidia (H20) | PASS | PASS (15 cases, --level core) | 1.31 | Primary |
| Tianshu | FAIL | PASS (11 cases) | 4.062 | 177 failed: AssertionError on bfloat16 + norm_Scalar |
| Muxi | PASS | PASS (11 cases) | 1.463 | — |
| Ascend | FAIL | N/A | — | Device does not support double dtype |
| Hygon | PASS | PASS (33 cases) | 1.585 | — |

## Files Changed
- `src/flag_gems/ops/norm.py`: Triton kernel implementation
- `src/flag_gems/__init__.py`: Register in `_FULL_CONFIG`
- `tests/test_norm.py`: Accuracy tests
- `benchmark/test_norm.py`: Performance benchmarks
- `conf/operators.yaml`: Operator entry (3 standalone entries: norm, norm_scalar, norm_scalaropt_dim)


tests/test_norm.py
```diff
@@ -0,0 +1,44 @@
...
+    ref_inp = to_reference(inp, True)
+
+    ref_out = torch.norm(ref_inp, ord, dim, keepdim)
+    with flag_gems.use_gems():
+        res_out = torch.norm(inp, ord, dim, keepdim)
+
+    gems_assert_close(res_out, ref_out, dtype)
+
+
+@pytest.mark.norm
```
Dongxu-H: norm_scalar

src/flag_gems/ops/__init__.py
```diff
@@ -710,6 +711,9 @@
     "nll_loss_nd_backward",
     "nonzero",
     "nonzero_numpy",
+    "norm",
```
Dongxu-H, line 1039: here are 3 kernels

src/flag_gems/__init__.py
```diff
@@ -381,6 +381,8 @@ def torch_ge(v):
     ("nll_loss2d_forward", nll_loss2d_forward),
     ("nonzero", nonzero),
     ("nonzero_numpy", nonzero_numpy),
+    ("norm.Scalar", norm_Scalar),
```
Dongxu-H: here are 2 kernels

src/flag_gems/ops/norm.py
```diff
@@ -0,0 +1,327 @@
...
+            elif p == 0:
+                l0_norm_kernel[grid](x, out, M, N)
+            else:
+                v_norm_kernel[grid](x, out, M, N, p)
+    if not keepdim:
+        out = out.squeeze(dim=dim)
+    return out
+
+
+def norm_ScalarOpt_dim(x, p, dim, keepdim=False):
```
tengqm:
```suggestion
def norm_scalaropt_dim(x, p, dim, keepdim=False):
```

src/flag_gems/ops/norm.py
```diff
@@ -0,0 +1,327 @@
...
+    logger.debug("GEMS NORM_SCALAR_OPT_DIM")
+    return norm(x, p=p, dim=dim, keepdim=keepdim)
+
+
+def norm_Scalar(x, p=2):
```
tengqm:
```suggestion
def norm_scalar(x, p=2):
```

benchmark/test_norm.py
```diff
@@ -0,0 +1,15 @@
+import pytest
+import torch
+
+from . import base, consts, utils
+
+
+@pytest.mark.norm
```
tengqm, line 21: We need test cases for all variants of this operator since they are separately dispatched by PyTorch.

conf/operators.yaml
```diff
@@ -4170,6 +4170,21 @@ ops:
       - Tensor
     stages:
       - alpha: '5.1'
+  - id: norm
+    description: |
+      Computes the norm of vectors, matrices, and tensors.
+    for:
+      - norm
+      - norm.Scalar
+      - norm.ScalarOpt_dim
```
tengqm, line 5958: There is no single entry for all three target operations. These operators have to be registered separately, treated as three standalone operators.

An alternative is to register only `norm`, but in that case, the implementation functions have to be combined.

src/flag_gems/ops/norm.py
```diff
@@ -0,0 +1,327 @@
...
+            elif p == 0:
+                l0_norm_kernel[grid](x, out, M, N)
+            else:
+                v_norm_kernel[grid](x, out, M, N, p)
+    if not keepdim:
+        out = out.squeeze(dim=dim)
+    return out
+
+
+def norm_scalaropt_dim(x, p, dim, keepdim=False):
```
tengqm, line 328: Will this code work at all?
You changed the spelling here but not the __init__.py files.

tests/test_norm.py
tengqm: We need at least three test cases each testing the different variants, with `pytest.mark` set to different IDs.

benchmark/test_norm.py
```diff
@@ -0,0 +1,37 @@
...
+    bench = base.GenericBenchmarkExcluse1D(
+        input_fn=utils.unary_input_fn,
+        op_name="norm",
+        torch_op=torch.norm,
+        dtypes=consts.FLOAT_DTYPES,
+    )
+    bench.run()
+
+
+@pytest.mark.norm_scalar
```
tengqm, line 32: The input function is the same as the previous one ... 
are you sure this is the right test case?

benchmark/test_norm.py
```diff
@@ -0,0 +1,37 @@
...
+    bench = base.GenericBenchmarkExcluse1D(
+        input_fn=utils.unary_input_fn,
+        op_name="norm_scalar",
+        torch_op=torch.norm,
+        dtypes=consts.FLOAT_DTYPES,
+    )
+    bench.run()
+
+
+@pytest.mark.norm_scalaropt_dim
```
tengqm, line 43: same question

benchmark/test_norm.py
LoserCheems: The three benchmarks intentionally use the same `unary_input_fn` because:

1. **`torch.norm` signature differs from `torch.all`** — for `torch.all(inp, 1)`, the second arg is `dim`. But for `torch.norm(inp, 1)`, the second arg is `p` (norm order), not `dim`. So `UnaryReductionBenchmark` (which passes `(inp, 1)` for dim) doesn't work here.

2. **All three variants share the same underlying Triton kernel** — `norm_scalar` and `norm_scalaropt_dim` are thin wrappers that both call `norm()`. The kernel performance is identical regardless of which dispatch entry is triggered.

3. **This follows the `vector_norm` pattern** — the upstream-merged `benchmark/test_vector_norm.py` also uses `GenericBenchmarkExcluse1D` + `unary_input_fn` for the same reason.

The marks are different so CI can select them independently, but the measured kernel throughput is the same since it's the same code path.

bin913:
@LoserCheems Please fix failing check

LoserCheems:
The CI failure is **not introduced by this PR**. It's a pre-existing compatibility issue in `src/flag_gems/ops/copy.py:89`:

```
AttributeError: module 'torch' has no attribute 'float8_e8m0fnu'
```

**Call chain:**
```
torch.norm(inp, 2, dim=0, keepdim=True)       # our test
→ torch.linalg.vector_norm()                   # PyTorch internal redirect
→ flag_gems vector_norm → dim_compress → .contiguous()
→ flag_gems copy_ (copy.py:89)                 # pre-existing code
→ checks torch.float8_e8m0fnu → AttributeError
```

The `copy_.py` file references `torch.float8_e8m0fnu` which doesn't exist in the CI's PyTorch version. This only triggers when `.contiguous()` is called on a permuted tensor (specific shape + dim=0 combination).

55 out of 56 tests pass. The single failure is caused by the existing `copy_` operator, not our norm implementation.

Should I add a workaround (skip this specific case), or will the `copy_.py` compatibility fix be handled separately?

conf/operators.yaml
```diff
@@ -4170,6 +4170,43 @@ ops:
       - Tensor
     stages:
       - alpha: '5.1'
+  - id: norm
+    description: |
+      Computes the norm of vectors, matrices, and tensors.
+    for:
+      - norm
```
Dongxu-H, line 5934: norm_tensor

benchmark/test_norm.py
tengqm: The thing is ... we have three variants here, and they, as you have said, share the same underlying kernel. That is fine.
There must be something different, and that "something" has to be the invocation protocol, the interface.
However, from the test cases, I failed to see any difference in the invocation either. Are we registering the same operator three times with different names? I don' hope so.

LoserCheems:
@bin913 Done🤗

Dongxu-H (changes requested):
## Code Review: Norm Operator Implementation

### Overview
This PR adds `norm`, `norm_scalar`, and `norm_scalaropt_dim` operators for computing vector/matrix/tensor norms. Implements L0, L1, L2, and infinity norm variants using Triton kernels.

### 🔴 Critical Issues

1. **Numerical Stability in `v_norm_kernel`**
   - File: `src/flag_gems/ops/norm.py`, line ~280
   - `out = pow(sum, 1 / ord)` can produce NaN for negative `sum` values with fractional orders
   - **Recommendation**: Add explicit handling for edge cases where sum is negative or zero

2. **Missing Input Validation for `p` Parameter**
   - File: `src/flag_gems/ops/norm.py`, function `norm()`
   - No validation for invalid `p` values (e.g., negative values other than `-inf`, NaN)
   - **Recommendation**: Add validation: `if p is not None and p != 0 and p != float('inf') and p != float('-inf') and p < 0: raise ValueError(...)`

3. **Memory Leak in Multi-Kernel Path**
   - When `dim is None`, the code creates intermediate `mid` tensor but doesn't use context manager for cleanup
   - **Recommendation**: Consider using `torch.empty` with proper memory management

### 🟡 Suggestions

4. **Performance: Two-Pass Algorithm Overhead**
   - For `dim=None` case, using two separate kernels adds synchronization overhead
   - Consider single-kernel approach for smaller tensors

5. **Code Duplication**
   - Significant code duplication between `l2_norm_kernel_1/2`, `max_norm_kernel_1/2`, `min_norm_kernel_1/2`, `l0_norm_kernel_1/2`, `l1_norm_kernel_1/2`
   - **Recommendation**: Consider consolidating into a single parameterized kernel

6. **Test Coverage Gaps**
   - Missing tests for:
     - Empty tensors
     - Very large tensors (memory edge cases)
     - Mixed precision scenarios
     - Non-contiguous tensors

7. **Benchmark Test Issues**
   - File: `benchmark/test_norm.py`
   - The benchmark uses `torch.norm` as reference but doesn't test all `ord` values
   - **Recommendation**: Add `ord` parameter to benchmarks

8. **Documentation**
   - Missing docstrings for public functions `norm`, `norm_scalar`, `norm_scalaropt_dim`
   - Should document supported `p` values and their behavior

### ✅ Strengths

- Clean separation of different norm types (L0, L1, L2, inf, -inf)
- Proper use of `@libentry()` decorator
- Good autotune configuration usage
- Comprehensive handling of `keepdim` parameter

### Questions

- Why is `p=0` (L0 norm) implemented? PyTorch's `torch.norm` doesn't support p=0. Is this a custom extension?
- Is there a performance benefit to having separate kernels for each norm type vs a unified kernel?
- The `do_not_specialize=["ord"]` annotation - why not specialize? Performance reasons?

### Testing Recommendations

Add test cases for:
```python
# Edge cases
torch.norm(torch.zeros(10), p=2)  # All zeros
torch.norm(torch.tensor([1e-10, 1e-10]), p=2)  # Very small values
torch.norm(x, p=float('nan'))  # Should raise error
```

LoserCheems:
Addressed all review comments:

1. **Separate operators**: `norm`, `norm.Scalar`, and `norm.ScalarOpt_dim` are registered as 3 standalone operators in `operators.yaml`, `__init__.py`, and `ops/__init__.py`.
2. **Function naming**: `norm_scalaropt_dim` and `norm_scalar` spelling is consistent across all files.
3. **Benchmark differentiated**: Each variant now has its own input function reflecting the actual invocation protocol:
   - `test_norm`: passes `(inp, p)`
   - `test_norm_scalar`: passes only `(inp,)`
   - `test_norm_scalaropt_dim`: passes `(inp, p, dim, keepdim)`
4. **Tests**: 3 separate test cases with distinct `pytest.mark` IDs, using `to_reference` for CPU-based ref output.
5. **Stage** updated to `alpha: '5.4'`.
6. Rebased on latest upstream/master, pre-commit all passed.

Dongxu-H (commented):
LGTM

Dongxu-H (dismissed):

tengqm:
/hold

Please address the previous comments.

LoserCheems:
@tengqm All previous comments have been addressed in commit d10b0c12:

1. **Spelling consistency**: `norm_scalaropt_dim` and `norm_scalar` are now consistent across `norm.py`, `ops/__init__.py`, `__init__.py`, and `operators.yaml`. No import mismatch — CI `python-op` passes.

2. **3 standalone operators**: Registered separately in `operators.yaml` with distinct `for:` entries (`norm`, `norm.Scalar`, `norm.ScalarOpt_dim`).

3. **Benchmark invocation differentiated**:
   - `test_norm`: passes `(inp, p)`
   - `test_norm_scalar`: passes `(inp,)` only
   - `test_norm_scalaropt_dim`: passes `(inp, p, dim, keepdim)`

4. **3 separate test cases** with distinct `pytest.mark` IDs.

Could you please re-review and `/unhold`? Thanks!

Caeruleann:
It looks like some of the previous review comments are still unresolved.

LoserCheems:
All previous review comments have been addressed — please see the detailed summary in my Jul 14 comment above. The inline threads were not formally resolved on GitHub but the code changes are all in place. Happy to re-explain any specific item if needed.

Caeruleann (approved):

PR #3439, opened 2026-05-23
[KMCompiler]Add indexer k quant and cache kernel

chenkc2025:
## Summary

Add a Triton implementation for `indexer_k_quant_and_cache`, aligned with the behavior of the vLLM CUDA custom op / CUDA kernel.

This operator quantizes input K tensors into FP8 format and writes them into the paged KV cache according to `slot_mapping`. It also stores the per-quant-block scales in the cache layout used by vLLM.

## Changes

- Add `indexer_k_quant_and_cache_triton`
- Add correctness tests against the vLLM CUDA custom op when available, with a PyTorch reference fallback
- Add a benchmark comparing the Triton implementation with the vLLM CUDA custom op

## Benchmark Scope

The benchmark covers multiple combinations of token count, cache block count, block size, and head dimension to evaluate the quantization-and-cache-write kernel under different input sizes.

- `Input K shape`: shape of the input K tensor
- `KV cache shape`: shape of the target FP8 KV cache
- `slot_mapping shape`: mapping from each input token to its physical cache slot
- `head_size`: K head dimension
- `scale_fmt`: scale format used by the quantized cache

## Benchmark Results

The Triton implementation shows comparable latency to the vLLM CUDA custom op on smaller cases, while larger cases are currently slower than the CUDA implementation.

### `indexer_k_quant_and_cache_triton_vs_vllm`

`dtype = torch.float16`

| Status | Input K shape | KV cache shape | slot_mapping shape | quant_block_size | scale_fmt | vLLM CUDA Latency (ms) | Gems Latency (ms) | Gems Speedup |
| --- | --- | --- | --- | ---: | --- | ---: | ---: | ---: |
| SUCCESS | `[128, 128]` | `[16, 16, 132]` | `[128]` | 128 | `ue8m0` | 0.006368 | 0.006048 | 1.053 |
| SUCCESS | `[512, 128]` | `[64, 16, 132]` | `[512]` | 128 | `ue8m0` | 0.006432 | 0.006368 | 1.010 |
| SUCCESS | `[1024, 512]` | `[128, 16, 528]` | `[1024]` | 128 | `ue8m0` | 0.007008 | 0.008928 | 0.785 |
| SUCCESS | `[2048, 512]` | `[256, 16, 528]` | `[2048]` | 128 | `ue8m0` | 0.008032 | 0.012320 | 0.652 |

`dtype = torch.bfloat16`

| Status | Input K shape | KV cache shape | slot_mapping shape | quant_block_size | scale_fmt | vLLM CUDA Latency (ms) | Gems Latency (ms) | Gems Speedup |
| --- | --- | --- | --- | ---: | --- | ---: | ---: | ---: |
| SUCCESS | `[128, 128]` | `[16, 16, 132]` | `[128]` | 128 | `ue8m0` | 0.006368 | 0.006080 | 1.047 |
| SUCCESS | `[512, 128]` | `[64, 16, 132]` | `[512]` | 128 | `ue8m0` | 0.006400 | 0.006304 | 1.015 |
| SUCCESS | `[1024, 512]` | `[128, 16, 528]` | `[1024]` | 128 | `ue8m0` | 0.007008 | 0.008896 | 0.788 |
| SUCCESS | `[2048, 512]` | `[256, 16, 528]` | `[2048]` | 128 | `ue8m0` | 0.008000 | 0.012000 | 0.667 |

## Testing

- `python -m py_compile src/flag_gems/fused/sparse_attn_indexer/indexer_k_quant_and_cache.py tests/test_indexer_k_quant_and_cache.py benchmark/test_indexer_k_quant_and_cache.py`


src/flag_gems/fused/sparse_attn_indexer/indexer_k_quant_and_cache.py
```diff
@@ -0,0 +1,113 @@
...
+    dst_scale_ptr = (
+        kv_cache_scale_ptr
+        + block_id * kv_cache_scale_stride
+        + block_offset * num_quant_blocks
+        + quant_block_id
+    )
+    tl.store(dst_scale_ptr, scale)
+
+
+def indexer_k_quant_and_cache_triton(
```
tengqm: Please consider register this into `src/flag_gems/fused/__init__.py` and `conf/operators.yaml` for tracking.

tests/test_indexer_k_quant_and_cache.py
```diff
@@ -0,0 +1,213 @@
...
+
+
+def _make_slot_mapping(num_tokens, num_blocks, block_size, device):
+    slot_mapping = torch.randperm(num_blocks * block_size, device=device)[:num_tokens]
+    slot_mapping = slot_mapping.to(torch.long)
+    slot_mapping[-1] = -1
+    return slot_mapping
+
+
+@pytest.mark.indexer_k_quant_and_cache_triton
```
tengqm: If the operator is not registered in `conf/operators.yaml`, it won't be tested in the CI pipeline.

tests/test_indexer_k_quant_and_cache.py
```diff
@@ -0,0 +1,213 @@
+import os
+
+import pytest
+import torch
+from packaging.version import InvalidVersion, Version
+
+from flag_gems.fused.sparse_attn_indexer.indexer_k_quant_and_cache import (
+    indexer_k_quant_and_cache_triton,
+)
```
tengqm: If the operator is registered in `src/flag_gems/fused/__init__.py`, the lines above can be simplified as:

```suggestion
from flag_gems.fused import indexer_k_quant_and_cache_triton
```

The key point is that we provide an unified interface for users to import this operator and others, instead of having them to know exactly where the implementation code lives.

chenkc2025: Updated as suggested, thanks.

Vincent-Xiao (approved):

PR #3504, opened 2026-05-26
[MTHREADS] fix assert_async

Oslomayor:
replace torch.cuda.synchronize with torch_device_fn.synchronize, skip native ref on non-CUDA

### DIFF
--- a/tests/test_assert_async.py
+++ b/tests/test_assert_async.py
@@ -2,6 +2,7 @@ import pytest
 import torch
 
 import flag_gems
+from flag_gems.runtime import torch_device_fn
 
 
 @pytest.mark.assert_async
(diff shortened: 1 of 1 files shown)


tests/test_assert_async.py
```diff
@@ -22,17 +23,18 @@ def test_assert_async(shape, value, expected_err, match_str):
...
+                    torch_device_fn.synchronize()
     else:
         with flag_gems.use_gems():
             flag_gems._assert_async(inp_triton, msg)
-            torch.cuda.synchronize()
+            torch_device_fn.synchronize()
 
-    if expected_err:
-        with pytest.raises(expected_err, match=match_str):
+    if torch.cuda.is_available():
```
huangyiqun: What's the behavior if the CUDA check fails?

Oslomayor: torch._assert_async is a CUDA-only API and will fail on non-CUDA devices. When torch.cuda.is_available() is False, the native reference comparison is skipped, but the flag_gems triton-based path still runs and is validated. This is the expected behavior since there is no non-CUDA equivalent for torch._assert_async.

huangyiqun (approved):

PR #3551, opened 2026-05-28
fix torch init error

0x45f:
### PR Category

Other
### Type of Change

Bug Fix
### Description

init torch before get device name
### Issue
https://github.com/flagos-ai/FlagGems/issues/3523

### Progress

### DIFF
--- a/src/flag_gems/utils/libentry.py
+++ b/src/flag_gems/utils/libentry.py
@@ -26,6 +26,7 @@ from typing import (
     overload,
 )
 
+import torch
 import triton
 
 from flag_gems import runtime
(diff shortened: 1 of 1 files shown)


huangyiqun (dismissed):

bin913 (approved):

JosephNew:
Verified with the following setup:

- Machine: Hygon BW1000 x8, driver 6.3.30-V1.4.1a, DTK 26.04
- Clean base image `ubuntu:22.04`, FlagGems 5.0.2 installed via `setup.sh hygon` (master branch, includes PR #3551)
- Container with `-v /opt:/opt`, sourced `/opt/dtk-26.04/env.sh`

Result:
```
>>> import flag_gems
>>> print('OK')
OK
```

The `RuntimeError: No HIP GPUs are available` is gone. PR #3551 fix confirmed.

One minor note: DTK `env.sh` does not include the HSA library path (`/opt/dtk-26.04/.hyhal/hsa/lib/libhsa-runtime64.so`), which causes a few UserWarnings in device_info. Does not block import but would be nice to fix on the DTK side. Nothing else from my side, thanks for the fix.

PR #3607, opened 2026-05-29
scaled_dot_product_attention_backward support non-square tuning, non-causal, and unequal lengths of q and kv sequences

### LINKED ISSUE #2848: The `scaled_dot_product_attention_backward` operator fails accuracy tests
### Describe the bug

See `tests/test_scaled_dot_product_attention.py` for details.

### LINKED ISSUE #3395: Calling FlagGems scaled_dot_product_attention_forward/backward can trigger training failures on NVIDIA and Tsingmicro backends
### Describe the bug

I hit a training-level grad-norm explosion while running a Qwen3 0.6B
baseline-aligned training check in May 2026. The path that showed the issue was
using the external FlagGems scaled dot product attention ops:

```text
flag_gems.scaled_dot_product_attention_forward
flag_gems.scaled_dot_product_attention_backward
```

We have observed this class of failure on both NVIDIA and Tsingmicro backends.
The detailed evidence and source-level chain below focus on the NVIDIA path,
where the failure could be tied to the `attention_bwd` tuning config and dQ
coverage logic.

The concrete run was:

```text
model: Qwen3 0.6B
sequence length: 4096
global batch size: 2048
micro batch size: 4
train_samples: 29297664
TE-FL: 2.9.0+9d1c48a8
FlagGems: 5.0.1rc0
torch: 2.7.0+cu126
triton: 3.3.0
flash-attn: 2.8.1
vendor: nvidia\tsingmicro
```

In this configuration, the training log showed an abnormal first-step grad norm:

```text
iteration 1
lm loss:   1.194416E+01
grad norm: 1468098560.000
```

After narrowing the call path, the relevant runtime markers were:

```text
Op 'get_flash_attention_class' using 'default.flagos'
[DEBUG] flag_gems.ops.attention.scaled_dot_product_attention_forward: GEMS SCALED DOT PRODUCT ATTENTION FORWARD
[DEBUG] flag_gems.ops.attention.scaled_dot_product_attention_backward: GEMS SCALED DOT PRODUCT ATTENTION BACKWARD
```

The issue appears to be in the backward tuning config and dQ coverage logic.

### Source-level issue

In FlagGems 5.0.1rc0, the external SDPA backward path reads its backward launch
candidates from the tuned config:

```python
config_backward = runtime.get_tuned_config("attention_bwd")
```

The NVIDIA tuning config I observed contains these `attention_bwd` candidates:

```yaml
attention_bwd:
  metas:
    - {BLOCK_M1: 32, BLOCK_N1: 128, BLOCK_M2: 128, BLOCK_N2: 32}
    - {BLOCK_M1: 64, BLOCK_N1: 128, BLOCK_M2: 64,  BLOCK_N2: 64}
    - {BLOCK_M1: 64, BLOCK_N1: 128, BLOCK_M2: 64,  BLOCK_N2: 32}
```

The suspicious part is that two candidates have:

```text
BLOCK_N1 = 128
BLOCK_M2 = 64
```

The backward wrapper allocates global `dq` with `empty_like`:

```python
dq = torch.empty_like(query).contiguous()
```

Then the launch grid is sized from the maximum `BLOCK_N1` across the backward
configs:

```python
max_block_n1 = max([cfg.kwargs["BLOCK_N1"] for cfg in config_backward])
grid = (triton.cdiv(Q_CTX, max_block_n1), 1, BATCH * Q_HEAD)
```

Inside the backward kernel, the dQ writeback advances by `BLOCK_M2`:

```python
start_m = pid * BLOCK_M2
offs_m = start_m + tl.arange(0, BLOCK_M2)
tl.store(dq_ptrs, dq, mask=offs_m_mask[:, None])
```

So, for the Qwen3 training shape:

```text
Q_CTX = 4096
max_BLOCK_N1 = 128
grid_x = ceil(4096 / 128) = 32
```

If autotune selects a candidate with `BLOCK_M2=64`, then the launched programs
can write at most:

```text
32 * 64 = 2048 rows
```

But full dQ needs:

```text
4096 rows
```

That leaves:

```text
dQ[2048:4096]
```

unwritten. Since global `dq` is allocated with `torch.empty_like`, the unwritten
region is not guaranteed to be zero. It can contain dirty memory, which can then
flow into parameter gradients and produce the observed grad-norm explosion.

In short, the problem chain appears to be:

```text
attention_bwd config allows BLOCK_M2 < max(BLOCK_N1)
-> grid count is computed from max(BLOCK_N1)
-> dQ writeback stride is BLOCK_M2
-> selected BLOCK_M2=64 candidate cannot cover full dQ
-> part of global dq remains uninitialized
-> dirty dQ contaminates gradients
-> training grad norm jumps to 1e9+
```

### Local workaround I tested

As a short-term mitigation, I only changed the two NVIDIA `attention_bwd`
candidates with `BLOCK_M2=64` to `BLOCK_M2=128`:

```diff
- {BLOCK_M1: 64, BLOCK_N1: 128, BLOCK_M2: 64,  BLOCK_N2: 64}
- {BLOCK_M1: 64, BLOCK_N1: 128, BLOCK_M2: 64,  BLOCK_N2: 32}
+ {BLOCK_M1: 64, BLOCK_N1: 128, BLOCK_M2: 128, BLOCK_N2: 64}
+ {BLOCK_M1: 64, BLOCK_N1: 128, BLOCK_M2: 128, BLOCK_N2: 32}
```

With only this config change, the same path recovered normal grad norm.

Stop1 result:

```text
iteration 1
lm loss:   1.194577E+01
grad norm: 12.285
skipped iterations: 0
nan iterations: 0
```

Stop10 result:

```text
iter01 grad norm: 13.051
iter02 grad norm: 19.228
iter03 grad norm: 16.426
iter04 grad norm:  8.019
iter05 grad norm:  4.355
iter06 grad norm:  3.153
iter07 grad norm:  2.791
iter08 grad norm:  2.644
iter09 grad norm:  2.569
iter10 grad norm:  2.549
```

I also extended the same patched path to 500 optimizer steps. It reached
iteration 500 cleanly:

```text
iteration 500/14305
lm loss:   4.203157E+00
grad norm: 0.277
skipped iterations: 0
nan iterations: 0
```
Also my colleague using tsingmicro BE reported this change can fix GN issue.
### Why I think this is a config/kernel invariant issue

The local workaround is not meant to be the final fix. It just avoids the
unsafe candidate family in this environment.

The more general invariant seems to be:

```text
For every attention_bwd candidate:
  BLOCK_M2 >= max(candidate.BLOCK_N1 for candidate in all attention_bwd configs)
```

or, more conservatively for this kernel shape:

```text
Use the same sequence stride for the N-side grid and dQ writeback:
  BLOCK_N1 == BLOCK_M2
```

In the current NVIDIA config, that means using `BLOCK_M2=128` for the two
candidates that currently use `BLOCK_M2=64`.
### Possible long-term fixes

Some possible fixes are:

1. Change the NVIDIA `attention_bwd` candidates so `BLOCK_M2 >= max(BLOCK_N1)`.
2. Add a config validation check that rejects backward candidates whose dQ
   coverage is impossible under the current grid rule.

### Caveats

- I am not claiming this explains every TE-FL / FlagGems grad-norm anomaly.
- This issue is only about the external FlagGems SDPA backward path and the
  NVIDIA `attention_bwd` config.
- I do not yet have a standalone tiny tensor repro. The current evidence is a
  training repro plus a source-level dQ coverage analysis and a config-only
  mitigation that keeps the same external FlagGems SDPA path active.
- The `BLOCK_M2=128` change is a short-term workaround. A stable upstream fix
  likely needs a config invariant check and/or launch-grid logic change.

103yiran:
### PR Category

Operator

### Type of Change

 Bug Fix

### Description

1. Added `mask=offs_m_mask[:, None], other=0.0` to the do load to prevent out‑of‑bounds reads when `Q_CTX` is not aligned.
2. Added `IS_CAUSAL: tl.constexpr = True` and removed `tl.device_assert(Q_CTX % BLOCK_M1 == 0)`.
3.  dK/dV:
    -   For causal: first execute a masked diagonal phase, then set `start_m`.
    -   For non‑causal: directly set `start_m = 0`.
    -   Both share the same unmasked `_attn_bwd_dkdv` call to process the remaining Q rows.
    -   Added an `if start_m < Q_CTX` guard to prevent executing an invalid masked phase when `start_n >= Q_CTX`.
    -   Replaced the fixed `BLOCK_N1 // MASK_BLOCK_M1` with `end_m = min(start_m + BLOCK_N1, Q_CTX)` to compute the actual number of steps.        
4. dQ:
    -   For causal: first execute a masked diagonal phase, then compute `stage2_end_n = min(diag_n, KV_CTX)` (fixes out‑of‑bounds when `Q_CTX > KV_CTX`). 
    -   For non‑causal: directly iterate over all KV columns.   
    -   Both share the same unmasked `_attn_bwd_dq` call.
    -   Used `diag_n = min(start_m, KV_CTX)` to ensure the diagonal index does not exceed the KV range.
    -   The number of steps for the unmasked phase is now computed directly based on `diag_n`.    
        
5. Changed the grid to a lambda (to support autotuning) and passed `IS_CAUSAL=is_causal`. 
6. Added three new kernel arguments: `dk_stride_z`, `dk_stride_h`, `dk_stride_tok`. Used a separate `dk_adj` offset for DK/DV pointers. DK/DV stores now use `dk_stride_tok` instead of `stride_tok`.        
6.  Changed `start_n` from `stage2_end_n - stage2_num_steps * BLOCK_N2` (which could be negative) to `0`, and restored ceiling division to ensure full coverage. `_attn_bwd_dq` already has an `offs_n < KV_CTX` mask to protect out‑of‑bounds accesses.
7. Fix gradient test and relax dv threshold.

### Issue

- Resolves #3395
- Related to #2848

### Progress

### DIFF
--- a/src/flag_gems/ops/attention.py
+++ b/src/flag_gems/ops/attention.py
@@ -672,91 +672,90 @@ def _attn_bwd(
         dk_ptrs = DK + offs_n[:, None] * stride_tok + offs_k[None, :] * stride_d
         tl.store(dk_ptrs, dk, mask=offs_n_mask[:, None])
 
-    # THIS BLOCK DOES DQ:
+    # dQ: only execute when this pid covers a valid Q block
     MASK_BLOCK_N2: tl.constexpr = BLOCK_N2 // BLK_SLICE_FACTOR
     start_m = pid * BLOCK_M2
-    end_n = min(start_m + BLOCK_M2, KV_CTX)  # Ensure end_n does not exceed N_CTX
-    num_steps = (end_n - start_n + MASK_BLOCK_N2 - 1) // MASK_BLOCK_N2
 
-    offs_m = start_m + tl.arange(0, BLOCK_M2)
-    offs_m_mask = offs_m < Q_CTX
+    if start_m < Q_CTX:
+        # Causal boundary: for Q rows [start_m, start_m+BLOCK_M2), the diagonal
+        # is at column start_m (not pid*BLOCK_N1 which differs when BLOCK_N1 != BLOCK_M2)
+        diag_n = start_m
+        end_n = min(start_m + BLOCK_M2, KV_CTX)
+        num_steps = (end_n - diag_n + MASK_BLOCK_N2 - 1) // MASK_BLOCK_N2
 
-    query = tl.load(
-        Q + offs_m[:, None] * stride_tok + offs_k[None, :] * stride_d,
-        mask=offs_m_mask[:, None],
-        other=0.0,
-    )
-    dq = tl.zeros([BLOCK_M2, BLOCK_DMODEL], dtype=tl.float32)
-    do = tl.load(
-        DO + offs_m[:, None] * stride_tok + offs_k[None, :] * stride_d,
-        mask=offs_m_mask[:, None],
-        other=0.0,
-    )
+        offs_m = start_m + tl.arange(0, BLOCK_M2)
+        offs_m_mask = offs_m < Q_CTX
 
-    m = tl.load(M + offs_m, mask=offs_m_mask, other=float("inf"))
-    m = m[:, None]
-
-    # Stage 1 - Compute dQ for masked (diagonal) blocks.
-    # NOTE: This code scans each row of QK^T backward (from right to left,
-    # but inside each call to _attn_bwd_dq, from left to right), but that's
-    # not due to anything important.  I just wanted to reuse the loop
-    # structure for dK & dV above as much as possible.
-
-    if num_steps > 0:
-        dq = _attn_bwd_dq(
-            dq,
-            query,
-            K,
-            V,  #
-            do,
-            m,
-            D,  #
-            stride_tok,
-            stride_d,  #
-            H,
-            Q_CTX,  #
-            KV_CTX,  #
-            BLOCK_M2,
-            MASK_BLOCK_N2,
-            BLOCK_DMODEL,  #
-            start_m,
-            start_n,
-            num_steps,  #
-            MASK=True,  #
+        query = tl.load(
+            Q + offs_m[:, None] * stride_tok + offs_k[None, :] * stride_d,
+            mask=offs_m_mask[:, None],
+            other=0.0,
+        )
+        dq = tl.zeros([BLOCK_M2, BLOCK_DMODEL], dtype=tl.float32)
+        do = tl.load(
+            DO + offs_m[:, None] * stride_tok + offs_k[None, :] * stride_d,
+            mask=offs_m_mask[:, None],
+            other=0.0,
         )
 
-    # Stage 2 - non-masked blocks
-    stage2_end_n = start_n
-    stage2_num_steps = (stage2_end_n + BLOCK_N2 - 1) // BLOCK_N2
+        m = tl.load(M + offs_m, mask=offs_m_mask, other=float("inf"))
+        m = m[:, None]
 
-    if stage2_num_steps > 0:
-        dq = _attn_bwd_dq(
-            dq,
-            query,
-            K,
-            V,  #
-            do,
-            m,
-            D,  #
-            stride_tok,
-            stride_d,  #
-            H,
-            Q_CTX,  #
-            KV_CTX,  #
-            BLOCK_M2,
-            BLOCK_N2,
-            BLOCK_DMODEL,  #
-            start_m,
-            stage2_end_n - stage2_num_steps * BLOCK_N2,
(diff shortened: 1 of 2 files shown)


huangyiqun (dismissed):

huangyiqun (dismissed):

huangyiqun (commented):
Plz fix CI test `unit-test / python op`.

huangyiqun (approved):

PR #3723, opened 2026-06-02
[Track 1-B] Refactor log1p_ to @pointwise_dynamic

AdamPlatin123:
## Summary
Refactor `log1p_` from KernelGen manual kernel (57 lines) to `@pointwise_dynamic` (19 lines).

## Changes
- `src/flag_gems/ops/log1p_.py`: Replaced manual kernel with @pointwise_dynamic

## Improvements
- **-38 lines** (-67% code reduction)
- **Non-contiguous tensor support**: Original fell back to `torch.ops.aten.log1p_` for non-contiguous tensors, now handled natively
- **Removed manual dtype guard**: Original had dtype allowlist check, now uses INT_TO_FLOAT promotion for all numeric types
- **Consistent with log1p.py**: Matches existing non-inplace implementation pattern

🤖 Generated with [FlagGems Operator Development Competition]


w1120029931-bit:
What is the reason for such a modification?
What is the performance after modification?
Please provide performance data

PR #412, opened 2025-01-12
[Operator] Add Conv3d forward function

Gxiandy:
### PR Category

Operator 
### Type of Change

New Feature
### Description

Add Conv3d forward function and related tests
### Progress

- [ ] Change is properly reviewed (1 reviewer required, 2 recommended).
- [ ] Change is responded to an issue.
- [ ] Change is fully covered by a UT.

### Performance

Operator: conv3d  Performance Test (dtype=torch.float16, mode=cuda,level=core)
|Status|       Torch Latency (ms) |   Gems Latency (ms) |        Gems Speedup |         Size Detail|
|----------|---------------------------|--------------------|--------------|------------------------|
|SUCCESS|               6.790144   |         6.793216              | 1.000 |         {'input': torch.Size([104, 16, 32, 32, 32]), 'weight': torch.Size([32, 16, 4, 4, 4]), 'bias': None, 'groups': 1, 'stride': 1, 'padding': 0}|
|SUCCESS  |             4.712448     |       4.705280              | 1.002   |       {'input': torch.Size([64, 32, 18, 180, 18]), 'weight': torch.Size([32, 32, 5, 5, 5]), 'bias': None, 'groups': 1, 'stride': 2, 'padding': 1}|
|SUCCESS    |           0.943104       |     0.941056            |   1.002     |     {'input': torch.Size([4, 32, 110, 110, 10]), 'weight': torch.Size([64, 32, 5, 5, 5]), 'bias': None, 'groups': 1, 'stride': 2, 'padding': 1}|
|SUCCESS      |         2.338816         |   2.342912          |     0.998       |   {'input': torch.Size([4, 64, 110, 110, 10]), 'weight': torch.Size([16, 64, 5, 5, 5]), 'bias': None, 'groups': 1, 'stride': 2, 'padding': 1}|
|SUCCESS        |       0.208896           | 0.208896        |       1.000         | {'input': torch.Size([16, 32, 120, 12, 12]), 'weight': torch.Size([24, 32, 3, 3, 3]), 'bias': None, 'groups': 1, 'stride': 2, 'padding': 1}|
|SUCCESS          |     7.403520           | 7.422976      |         0.997          |{'input': torch.Size([16, 32, 240, 24, 24]), 'weight': torch.Size([24, 16, 3, 3, 3]), 'bias': None, 'groups': 2, 'stride': 1, 'padding': 1}|
[...]

PR #3750, opened 2026-06-03
[KernelGen] Fix --ref cpu cross-device tensor comparison

Schopenhauer-loves-Hegel:
## Summary
- Fix `_maybe_move_to_cpu()` in `src/flag_gems/testing/__init__.py` to handle cross-device tensor comparison by moving both tensors to CPU when they are on different devices
- Add root-level `conftest.py` to patch `gems_assert_close` / `gems_assert_equal` for `--ref cpu` compatibility when tests compute reference on GPU

Fixes internal issues #411, #412, #413 (`act_quant_triton`, `addmm_dtype`, `addmm_dtype_out` accuracy test failures under `--ref cpu` mode).

## Test plan
- [x] `pytest -m "act_quant_triton" tests/ --ref cpu -vs` — 112 passed
- [x] `pytest -m "addmm_dtype" tests/ --ref cpu -vs` — 3 passed
- [x] `pytest -m "addmm_dtype_out" tests/ --ref cpu -vs` — 3 passed
- [x] All 114 addmm-related tests pass with and without `--ref cpu`
- [x] Benchmarks pass

## Issue
- WEEKTEST-411
- WEEKTEST-412
- WEEKTEST-413


conftest.py
```diff
@@ -0,0 +1,65 @@
...
+tests/accuracy_utils.to_cpu fails for operators whose test manually
+computes the reference on CPU and then moves it to the accelerator
+device.  We relax the assertion to instead move *ref* to CPU, which
+is the safe and correct thing to do for comparison purposes.
+"""
+
+import torch
+
+
+def pytest_configure(config):
```
tengqm: don't do this.

If you are confident that this is the right approach for accuracy testing, update the tests/conftest.py directly. We don't want any .py files in the root directory.

tengqm (approved):

PR #3753, opened 2026-06-03
[KernelGen] Restore leaky_relu_ input during autotuning

lukatao:
## Summary
Use `restore_value=["input_ptr"]` on the existing leaky-ReLU autotuner, as suggested in review. This restores input data after tuning trials when the input and output share storage. Remove the duplicate fixed-configuration in-place kernel from the previous proposal and retain the existing shared kernel and autotune configurations.

Add an all-negative, uncommon-size input with slope 0.5 to exercise an initial autotune call followed by a cached call, checking exact output values and returned tensor identity. Update this operator's existing accuracy tests to compare Gems directly with PyTorch, following the current test convention.

## Validation
- The new regression test collects for float16, float32, and bfloat16.
- Refreshed PR GPU CI validates correctness; local GPU execution is unavailable. No new cross-vendor performance claim is made.

Related work item: WEEKTEST-417.


w1120029931-bit:
The reason for the problem is that when inplace=True, the input/output points to the same piece of video memory. Triton.autotune will launch the kernel multiple times for warmup/benchmark, causing leaky_relu to be repeatedly applied in-place, and negative numbers to be multiplied by negative_slope multiple times, resulting in incorrect results?
Is it possible
In @triton.autotune, use restore_value=['input_ptr'] to enable autotune to restore the input data after each test.
Or the reason for using other better methods

tengqm:
Verified the proposal suggested by @w1120029931-bit, this approach is the cleanest one. It works on Nvidia H20, MetaX and MooreThreads. There are still failures on Ascend due to other reasons. The one line change is something worth a try.

lukatao:
Applied the suggested restore_value=["input_ptr"] to the shared forward autotuner and removed the separate fixed inplace-kernel approach. Added an all-negative, uncommon-shape regression covering the first/autotuned call and cached call. Current head passed 75 GPU tests and 5 quick CPU-reference tests.

PR #3777, opened 2026-06-04
fix to_copy benchmark test

bin913:
### PR Category
[OP Test | Benchmark]

### Type of Change
[ Bug Fix | Refactor]

### Description
fix error of no performance result for to_copy benchmark 

### Progress


tengqm (commented):
Thanks for the diligent work, however, I'm a little bit concerned over this approach. We are tweaking the base benchmark framework and the test class to make the benchmark happy.

If the idea is to make sure that the dtypes we feed the operator are valid. In other words, we are testing all valid dtype combinations for the operator. We then can split the `test_to_copy()` function to several functions, each focus on a specify group of type combinations.

tengqm (approved):

benchmark/test_to_copy.py
```diff
@@ -47,71 +73,24 @@ def get_tflops(self, op, *args, **kwargs):
...
+
+        if not hasattr(self, "_type_pairs"):
+            return super().run()
+
+        self.init_user_config()
+        for src_dtype, dst_dtypes in self._type_pairs:
+            self.src_dtype = src_dtype
+            self.to_bench_dtypes = dst_dtypes
+            self._bench_dtype_loop()
+
```
tengqm, line 88: Why this change?

tengqm: We can easily split the existing test cases into smaller ones, each testing a different aspect of the operator.

Don't invent new wheels in this space.

PR #3859, opened 2026-06-08
[Advanced Compiler] Fix conv_depthwise2d test with --ref cpu

### LINKED ISSUE #3021: The `conv_depthwise2d` operator fails accuracy test
### Describe the bug

It is thus skipped and tracked for investigation.

AdvancedCompiler:
### PR Category

OP Test
### Type of Change

Bug Fix
### Description

This failure occurs because aten::_conv_depthwise2d is a CUDA‑only implementation of depthwise convolution in PyTorch. It is not available for the CPU backend, as confirmed by the error message listing only [CUDA, Meta, ...] backends.

To resolve this, when the test is run with --ref cpu (which moves reference tensors to CPU), we now use torch.nn.functional.conv2d with groups = ref_inp.shape[1] (the number of input channels) to compute the reference result. This avoids the unsupported CPU call and provides a correct baseline for comparison. The actual operator under test (aten::_conv_depthwise2d) still runs on CUDA as before.

### Issue
Fix #3021 : conv_depthwise2d fails under --ref cpu mode

### Progress

### Performance

A100 test:
image
image


tengqm:
Thanks for looking into this. If this operator hates CPU, we can make it an exception.
I mean we can patch tools/test-op.sh by adding `tests/test_conv_depthwise2d.py` to the `NO_QUICK_CPU_TESTS` list. Compute results using a different operator for comparison doesn't smell like a right fix to me.

tests/test_conv_depthwise2d.py
```diff
@@ -45,15 +44,20 @@ def test_conv_depthwise2d(
...
-    ref_out = torch.ops.aten._conv_depthwise2d(
-        ref_inp,
-        ref_weight,
-        kernel,
-        ref_bias,
-        stride,
-        padding,
-        dilation,
-    )
+    if utils.TO_CPU:
```
0x45f: please add `tests/test_conv_depthwise2d.py` to `NO_QUICK_CPU_TESTS `

bin913 (commented):
## FlagGems Automated Review

**PR Type:** Platform/Backend change
**Operator:** N/A

This PR modifies infrastructure/backend files. The operator-level format checks are not applicable. No format issues to report.

---
<sub>Automated review by FlagGems PR Review Skill</sub>

tengqm (approved):

0x45f (approved):

PR #3932, opened 2026-06-09
[Performance Optimization] Add QUICK_MODE support to reduce test suite runtime

git-flyer:
Add QUICK_MODE test support for a series of operators and optimize test parameters to accelerate quick-mode verification.


tests/accuracy_utils.py
```diff
@@ -216,19 +216,23 @@ def SkipVersion(module_name, skip_pattern):
...
-    PRIMARY_FLOAT_DTYPES + [torch.bfloat16]
-    if bf16_is_supported
-    else PRIMARY_FLOAT_DTYPES
+    [torch.float32]
+    if QUICK_MODE
+    else (
+        PRIMARY_FLOAT_DTYPES + [torch.bfloat16]
+        if bf16_is_supported
+        else PRIMARY_FLOAT_DTYPES
+    )
```
huangyiqun, line 237: Why is only fp32 tested in QUICK_MODE?

zhzhcookie: Should be [torch.bfloat16] if bf16_is_supported else [torch.float32] when QUICK_MODE

huangyiqun (approved):

PR #3956, opened 2026-06-10
[KernelGen] Keep tile offset arithmetic in int64

lukatao:
Upstream #4289 already widened generated kernel arguments. Retain the missing int64 conversions for program IDs, program counts and the initial tile offsets. Exercise the operator directly and add a non-contiguous-input regression.

Validation: Executed the actual generated offset code with CPU tensor operations: master wraps at 2^31, while all 256 offsets in [2^31, 2^31+256) are correct with this change. Full GPU tests are pending; this check does not allocate a multi-billion-element GPU tensor.


w1120029931-bit:
lgtm

PR #3966, opened 2026-06-10
Fix per-channel INT8 W8A16 fused_moe: in-kernel dequant instead of ho…

zhan4808:
…st-side

The use_int8_w8a16 path host-dequantized ALL expert weights on every call (w1.to(dtype) * scale), moving ~10 GB per call on Mixtral-8x7B shapes -- a flat ~8.5 ms overhead on H100 that made W8A16 6-22x slower than bf16 at every token count.

The non-fused fused_moe_kernel already supports in-kernel W8A16 (b.to(compute_type) in the K-loop, per-N scale applied once on the fp32 accumulator); only the FUSE_SILU fused gate/up path was missing that support, which is why the host fallback existed. This change:

- adds W8A16 handling to the FUSE_SILU two-pass path (int8 tile conversion in both K-loops, per-channel gate/up scales applied on the accumulators)
- disables PAIR_GATE_UP_DOT for W8A16 (the paired dot has no conversion/scale handling) in favor of the two-pass path
- routes per-channel INT8 W8A16 through the kernel, keeping the host dequant fallback only for INT4 and grouped-scale INT8 (the WNA16 path expects offset-binary packed layouts not prepared here)

Measured on H100 (Mixtral-8x7B shapes, E=8 H=4096 I=14336 top-k=2, fp16), W8A16 vs bf16: 1.07x at T=1, 1.70x at T=16-64, 1.20x at T=128, 1.03x at T=256, 2.63x at T=512 -- previously 0.05-0.16x (always slower). Output rel err vs fp16 reference ~0.012, consistent with per-channel INT8 weight quantization error.

### Progress

### DIFF
--- a/src/flag_gems/fused/fused_moe.py
+++ b/src/flag_gems/fused/fused_moe.py
@@ -1184,6 +1184,16 @@ def fused_moe_kernel(
             offs_k[:, None] * stride_bk + offs_bn_up[None, :] * stride_bn
         )
 
+        if use_int8_w8a16:
+            # Per-channel weight scales for the gate and up halves; applied once
+            # on the fp32 accumulators after each K-loop.
+            b_scale_gate = tl.load(
+                b_scale_ptr + off_experts * stride_bse + offs_bn_gate[None, :] * stride_bsn
+            )
+            b_scale_up = tl.load(
+                b_scale_ptr + off_experts * stride_bse + offs_bn_up[None, :] * stride_bsn
+            )
+
         if use_fp8_w8a8 or use_int8_w8a8:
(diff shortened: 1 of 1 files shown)


huangyiqun:
Plz fix code style by `pre-commit run --all-files`.

huangyiqun:
Will this PR performance degradation of the fused_moe operator?

huangyiqun (commented):
Please add a targeted regression test for the newly introduced FUSE_SILU + use_int8_w8a16 path.

huangyiqun:
Are you still working on this PR?

PR #4005, opened 2026-06-11
[KernelGen] Define masked reduction tails and add boundary coverage

### LINKED ISSUE #3716: Potential undefined values from masked tl.load without `other` used in reductions
Hi, I noticed a few Triton kernels where `tl.load(..., mask=...)` is used without an explicit `other=...`, and the loaded tensor is later consumed by reductions or nonlinear math.

According to the Triton `tl.load` documentation, when `mask[idx]` is false and `other is None`, the masked-out value is undefined. This can be risky when the loaded vector is subsequently used in reductions such as `tl.sum`, `tl.max`, `tl.cumsum`, softmax/log-softmax backward, dot products, or activation/normalization formulas.

Example:
https://github.com/flagos-ai/FlagGems/blob/c7e6a3e2eeb156030759de3668b34e4c490d0369/src/flag_gems/ops/log_softmax.py#L81

https://github.com/flagos-ai/FlagGems/blob/c7e6a3e2eeb156030759de3668b34e4c490d0369/src/flag_gems/ops/cumsum.py#L48

lukatao:
## Summary
Provide explicit zero values for masked loads used by cumsum and log-softmax backward. In the backward gradient sum, inactive tail lanes must contribute zero. Explicit padding also makes scan/reduction tail values defined.

Address the review request for tests with 12 cumsum boundary cases spanning row scans, non-inner scans, and multi-block tails, plus 9 log-softmax backward cases with odd reduction lengths. Update these operators' existing tests to compare Gems directly with PyTorch, following the current test convention.

## Validation
- The 21 added boundary cases collect successfully.
- GPU CI passed for the affected test modules. [Test run](https://github.com/flagos-ai/FlagGems/actions/runs/34463730312/job/102827546147). These are boundary-coverage tests; unspecified masked-load values do not necessarily cause a reproducible failure on every compiler/device, so no deterministic before/after GPU failure is claimed.

Related work item: WEEKTEST-421; related issue #3716.

## Existing base-branch blocker
The operator metadata check fails because master defines cumsum documentation under `destination` instead of the required `description`. This PR does not modify that pre-existing metadata issue; it remains a merge blocker outside the masked-load/test changes.


w1120029931-bit:
lgtm

bin913:
@Schopenhauer-loves-Hegel Please add test  for this change

lukatao:
Added the requested masked-tail regression tests: cumsum boundary shapes and odd-sized log_softmax backward reductions. Current head passed 107 GPU tests and 24 quick CPU-reference tests. The remaining rule failure is the pre-existing cumsum registry entry missing description (it uses destination in master); no unrelated registry change is included.

PR #4024, opened 2026-06-12
[KUNLUNXIN]Fix crash when running repeat test cases caused by a confl…

RobertLuobo:
…ict between the repeat operator and the pytest environment

### Progress


tests/test_repeat.py
```diff
@@ -8,7 +8,7 @@
 REPEAT_SIZES = [(2, 3, 4, 5), (5, 0, 4)]
 
 
-@pytest.mark.repeat
+@pytest.mark.repeat_op
```
huangyiqun, line 11: Why change mark or to `repeat_op`?

RobertLuobo: "repeat" would be crashed when running repeat test cases caused by a conflict between the repeat operator and the pytest environment
image

huangyiqun: The current implementation works fine on Nvidia and other backends - this appears to be an issue specific to the KunLunXin backend. This seems to be related to the pytest_repeat.py plugin you're using.

RobertLuobo:
close,plz uninstall pytest-repeat

PR #4062, opened 2026-06-15
Debug Ascend chip check

tengqm:
### PR Category

CI/CD

### Type of Change

Debug

### Description

The CI runner fails to execute the `npu-smi` command. Not sure what the error message is.


huangyiqun (commented):
Other backends use `2>/dev/null`. Why is the Ascend backend being modified this way?

tengqm:
> Other backends use `2>/dev/null`. Why is the Ascend backend being modified this way?

Good question. See https://github.com/flagos-ai/FlagGems/actions/runs/27521997775/job/81343912226
I'm getting `npu-smi` command error in the workflow. However, if I'm logging in to the machine, I can execute `npu-smi` and I can run this `gpu_check_ascend.sh` script.
Dropping `2>/dev/null` may be helpful for debugging.

PR #4097, opened 2026-06-16
[KernelGen][Iluvatar] Avoid stale specialization in std

lukatao:
Retain the backend std override with do_not_specialize for N/correction. Preserve current backend exports, exercise std directly, and add a sequential dim=0, dim=1, dim=0 regression on shape (200, 40999).

Validation: The new reference/input case passes with CPU PyTorch stand-ins. Local style/export/test-convention checks pass; generic and Iluvatar GPU checks are pending.

Latest target-backend result: Iluvatar fails before test collection because triton.language.map_elementwise is missing. This is outside the changed operator; no framework/runner workaround is included.


lukatao:
The sunrise performance failure is unrelated to this PR — `aten::std.correction` is not supported on the `ptpu` device, which is a pre-existing backend limitation. This PR only modifies Iluvatar-specific files (`_iluvatar/ops/std.py`) and should not be blocked by this.

lukatao:
> /test | std:ascend

this PR is unrelated to ascend, why should we check this?

tengqm:
> > /test | std:ascend
> 
> this PR is unrelated to ascend, why should we check this?

You are right.

src/flag_gems/runtime/backend/_iluvatar/ops/std.py
```diff
@@ -0,0 +1,171 @@
...
+
+    denom = tl.maximum(N - correction, 1.0).to(tl.float32)
+    var = var / denom
+    safe_var = tl.maximum(var, 0.0)
+    std_dev = tl.sqrt(safe_var)
+    tl.store(out_ptr, std_dev.to(Out.dtype.element_ty))
+
+
+def std(x, dim=None, *, correction=None, keepdim=False):
+    logger.debug("GEMS STD")
```
tengqm:
```suggestion
    logger.debug("GEMS_ILUVATAR STD")
```

tengqm (dismissed):

PR #4112, opened 2026-06-16
[KernelGen][Iluvatar] Restore shared var/var_mean performance and isolate backend handling

lukatao:
Upstream #4118 already provides the Iluvatar var override, so preserve that implementation. Retain restoration of shared autotuned Welford kernels and the still-missing Iluvatar var_mean override. Resolve exports and add direct var/var_mean regression coverage.

Validation: Local style/export/test-convention checks pass. New var_mean reference cases pass with CPU PyTorch stand-ins; generic and Iluvatar GPU checks are pending.

Latest target-backend result: Iluvatar fails before test collection because torch cannot load cudaProfilerInitialize from CUDART. This is outside the changed operator; no framework/runner workaround is included.


w1120029931-bit:
lgtm

PR #4118, opened 2026-06-17
[KernelGen][Iluvatar] Fix var_correction: accuracy_fail

Schopenhauer-loves-Hegel:
## Summary
- **Operator:** var_correction
- **Root cause:** The shared var_welford_kernel uses tl.reduce with a custom Welford combine function to merge partial results across BLOCK_N chunks. On the Iluvatar backend, this pattern produces NaN values when BLOCK_M > 1, and Triton kernel specialization causes compiled kernels to produce incorrect results when reused with different M/N values across sequential calls.
- **Fix:** Created an Iluvatar-specific override (var.py) using a two-pass approach (compute mean via tl.sum, then compute variance from squared deviations) instead of the Welford tl.reduce pattern. Marked M, N, correction as do_not_specialize to prevent kernel specialization issues. Reused the shared var_kernel_1/var_kernel_2 for full-reduce (dim=None) which uses 1D tl.reduce that works correctly on Iluvatar.
- **Files modified:**
  - `src/flag_gems/runtime/backend/_iluvatar/ops/var.py`
  - `src/flag_gems/runtime/backend/_iluvatar/ops/__init__.py`

## Verification
- Accuracy test: 180/180 passed
- Benchmark: passed
- Format check: N/A

## Test Plan
- [ ] `pytest -m 'var_correction' tests/ --ref cpu -vs`
- [ ] `pytest -m 'var_correction' benchmark/ --level core --record log`

## Issue
- WEEKTEST-550

### DIFF
--- a/src/flag_gems/runtime/backend/_iluvatar/ops/var.py
+++ b/src/flag_gems/runtime/backend/_iluvatar/ops/var.py
@@ -1,0 +1,122 @@
+import logging
+
+import torch
+import triton
+import triton.language as tl
+
+from flag_gems.runtime import torch_device_fn
+from flag_gems.utils import dim_compress, libentry
+from flag_gems.utils import triton_lang_extension as ext
+
+logger = logging.getLogger(__name__)
+
+
+@libentry()
+@triton.jit(do_not_specialize=["M", "N", "correction"])
+def iluvatar_var_twopass_kernel(
+    X,
+    Var,
+    M,
+    N,
+    correction,
+    BLOCK_M: tl.constexpr,
+    BLOCK_N: tl.constexpr,
+):
+    """Two-pass variance kernel for the Iluvatar backend.
+
+    Avoids the tl.reduce + welford combine_fn pattern which produces
+    incorrect results (NaN) on Iluvatar when BLOCK_M > 1 due to a
+    backend-specific issue with multi-row reductions using custom
+    combine functions.
+    """
+    # Map the program id to the row of X it should compute.
+    pid = ext.program_id(0) * BLOCK_M + tl.arange(0, BLOCK_M)[:, None]
+    X = X + pid * N
+    Var = Var + pid
+    row_mask = pid < M
+
+    # Pass 1: compute mean via sum
+    _sum = tl.zeros([BLOCK_M, BLOCK_N], dtype=tl.float32)
+    for off in range(0, N, BLOCK_N):
+        cols = off + tl.arange(0, BLOCK_N)[None, :]
+        col_mask = cols < N
+        mask = row_mask and col_mask
+        x = tl.load(X + cols, mask, other=0.0).to(tl.float32)
+        _sum += x
+
+    row_mean = tl.sum(_sum, axis=1) / N  # [BLOCK_M]
+    row_mean_2d = row_mean[:, None]  # [BLOCK_M, 1]
+
+    # Pass 2: compute sum of squared deviations from mean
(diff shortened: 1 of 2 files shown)


src/flag_gems/runtime/backend/_iluvatar/ops/var.py
```diff
@@ -0,0 +1,122 @@
...
+                x, var, M, N, correction, BLOCK_M=BLOCK_M, BLOCK_N=BLOCK_N
+            )
+
+    if not keepdim:
+        var = var.squeeze(dim=dim)
+    return var
+
+
+def var_dim(x, dim=None, *, correction=None, keepdim=False):
+    logger.debug("GEMS ILUVATAR VAR_DIM")
```
tengqm:
```suggestion
    logger.debug("GEMS_ILUVATAR VAR_DIM")
```

src/flag_gems/runtime/backend/_iluvatar/ops/var.py
```diff
@@ -0,0 +1,122 @@
...
+    return var(x, dim=dim, correction=correction, keepdim=keepdim)
+
+
+def var_correction(x, dim=None, *, correction=None, keepdim=False):
+    logger.debug("GEMS ILUVATAR VAR_CORRECTION")
```
tengqm:
```suggestion
    logger.debug("GEMS_ILUVATAR VAR_CORRECTION")
```

src/flag_gems/runtime/backend/_iluvatar/ops/var.py
```diff
@@ -0,0 +1,122 @@
...
+
+def var(x, dim=None, *, correction=None, keepdim=False):
+    logger.debug("GEMS ILUVATAR VAR")
+    if correction is None:
+        correction = 1.0
+
+    if dim is None or len(dim) == x.ndim:
+        # Full reduce: use the shared kernel_1 + kernel_2 path from ops.var
+        # (these use 1D tl.reduce which works correctly on Iluvatar)
+        from flag_gems.ops.var import var_kernel_1, var_kernel_2
```
tengqm: don't do import in the middle of your program.

src/flag_gems/runtime/backend/_iluvatar/ops/var.py
```diff
@@ -0,0 +1,122 @@
...
+        _acc += diff * diff
+
+    row_acc = tl.sum(_acc, axis=1)  # [BLOCK_M]
+    var = row_acc / (N - correction)
+    var = var[:, None]
+    tl.store(Var, var, row_mask)
+
+
+def var(x, dim=None, *, correction=None, keepdim=False):
+    logger.debug("GEMS ILUVATAR VAR")
```
tengqm:
```suggestion
    logger.debug("GEMS_ILUVATAR VAR")
```

lukatao:
This PR is not about var_dim. It only fixes the var_correction operator on the Iluvatar backend. The var_dim CI result is expected to be NotFound since this PR does not modify var_dim.

tengqm:
> This PR is not about var_dim. It only fixes the var_correction operator on the Iluvatar backend. The var_dim CI result is expected to be NotFound since this PR does not modify var_dim.

It should be fine to `/test | var_dim:<vendor>` or `/test | var_correction:<vendor>`. Because this PR didn't touch any test cases, the automatic CI check job won't be triggered. It has to be manually tested.

tengqm (approved):

PR #4119, opened 2026-06-17
[KernelGen][Iluvatar] Avoid quantization overflow with float32 fallback

lukatao:
Retain the backend-specific (x / absmax) * fp8_max formulation for non-ue8m0 scales, avoiding division by a flushed subnormal scale. Resolve backend export conflicts, enable direct operator tests and use the equivalent stable reference formula for the float32 fallback.

Validation: Local syntax, formatting, export ordering and test-convention checks pass. Generic and Iluvatar GPU checks are pending; the old CI failure was an external package-download failure.

Latest target-backend result: Iluvatar fails before test collection because triton.language.map_elementwise is missing. This is outside the changed operator; no framework/runner workaround is included.


w1120029931-bit:
lgtm

PR #4120, opened 2026-06-17
[KernelGen][Iluvatar] Fix flip negative-stride accuracy

lukatao:
Retain the Iluvatar flip override that computes flipped coordinates without negative block-pointer strides. Preserve all current backend exports and add regressions for row widths 17, 24, 31 and 32.

Validation: 12 new reference cases pass with CPU PyTorch stand-ins. Generic GPU checks are pending. The latest Iluvatar run fails while importing shared code because triton.language.map_elementwise is unavailable, before any flip test executes; no unrelated compatibility patch is included.


w1120029931-bit:
lgtm

PR #4121, opened 2026-06-17
[KernelGen][Iluvatar] Match topk_softmax reference API and normalization

lukatao:
Use the four-argument vLLM reference and int32 reference indices while retaining FlagGems index-dtype coverage. Perform requested renormalization in both the accuracy reference and benchmark reference; benchmark gating inputs remain float32.

Validation: Local syntax, formatting and test-convention checks pass. Updated GPU checks are pending.


w1120029931-bit:
lgtm

PR #4127, opened 2026-06-17
[Benchmark] Remove unnecessary clone in inplace backward

stringl1l1l1l:
### PR Category
Benchmark

### Type of Change
Performance Optimization

### Description
For inplace backward benchmarks, floating tensor inputs were previously prepared with `clone().requires_grad_()`. This extra clone is unnecessary for inplace cases and can cause avoidable additional device memory usage, especially on machines with limited GPU memory.

This change uses `detach().requires_grad_()` for non-inplace backward benchmark inputs while keeping the existing `clone().requires_grad_()` behavior for inplace backward benchmarks. Unit tests passed after the change, confirming this does not introduce side effects.

### Issue
N/A

### Progress

### Performance
No benchmark score comparison added. This change reduces unnecessary peak memory pressure in inplace backward benchmark input setup by removing an avoidable tensor clone.


w1120029931-bit:
Could the inplace branch trigger the autograd leaf restriction?
For example:
a.detach().clone().requires_grad_()
A leaf tensor with requires_grad=True may be generated. If an inplace operation is directly performed on it later, an error may be reported:
a leaf Variable that requires grad is being used in an in-place operation

stringl1l1l1l:
> Could the inplace branch trigger the autograd leaf restriction? For example: a.detach().clone().requires_grad_() A leaf tensor with requires_grad=True may be generated. If an inplace operation is directly performed on it later, an error may be reported: a leaf Variable that requires grad is being used in an in-place operation

Thanks for pointing this out.

Currently, there is no benchmark case where `is_inplace` and `is_backward` are both set to `True`. The existing backward benchmarks all use out-of-place forward ops, so this leaf in-place restriction should not be triggered by the current test coverage.

If we add support for inplace backward benchmarks in the future, I think the fix should be handled in `get_latency()` rather than in `unpack_to_args_kwargs()`. The reason is that `unpack_to_args_kwargs()` only prepares the input arguments, while `get_latency()` is where the backward benchmark actually builds the autograd graph:
```python
def get_latency(self, op, *args, **kwargs):
        fn = lambda: op(*args, **kwargs)
        if self.is_backward:
            out = fn()
            dout = torch.randn_like(out)
            # fn = lambda: out.backward(dout, retain_graph=True)
            xs = list(filter(lambda x: torch.is_tensor(x) and x.requires_grad, args))
            fn = lambda: torch.autograd.grad(
                (out,), xs, grad_outputs=(dout,), retain_graph=True
            )
```

For an inplace backward benchmark, we would need to keep separate tensors for the inplace forward call and for the gradient target. This is part of the autograd graph construction in `get_latency()`, and cannot be handled properly in `unpack_to_args_kwargs()`, which only prepares the input arguments.

awayzjj:
@w1120029931-bit @tengqm Please review the PR, thanks!

PR #4133, opened 2026-06-17
[SiliconFlow] Fix scatter_reduce reduce argument and dtype configuration

ShawnXuan:
## Summary

- Fix `scatter_reduce` / `scatter_reduce_` benchmark failures caused by passing `reduce` as a positional argument.
- Restrict legacy reduce-scatter benchmark dtypes to `float16` and `float32`, matching the currently supported implementation and existing accuracy coverage.
- Keep `scatter_reduce.two` benchmarks unchanged, since they cover the newer `torch.scatter_reduce(...)` path.

Fixes internal issue 570.

## Test plan

- [x] `python -m compileall benchmark/test_scatter_reduce.py`
- [x] `git diff --check`
- [x] `pre-commit run --files benchmark/test_scatter_reduce.py`
- [x] CUDA H20 GPU perf: `pytest -s benchmark/test_scatter_reduce.py -m scatter_reduce --level core --record log`
- [x] CUDA H20 GPU perf: `pytest -s benchmark/test_scatter_reduce.py -m scatter_reduce_ --level core --record log`

## Issue

- WEEKTEST-123
- WEEKTEST-188


tengqm:
The removing of BF16 is justified?

ShawnXuan:
> The removing of BF16 is justified?

Yes, removing BF16 is justified.

The benchmark targets the legacy torch.scatter(..., reduce=...) / Tensor.scatter_(..., reduce=...) path, which explicitly asserts against bfloat16 in the implementation:

[src/flag_gems/ops/scatter.py#L334-L337](https://github.com/flagos-ai/FlagGems/blob/master/src/flag_gems/ops/scatter.py#L334-L337)

```python
if reduce is not None:
    assert inp.dtype not in (
        torch.bfloat16,
    ), "Unsupported operation: reduce scatter bfloat tensors."
```
Using FLOAT_DTYPES (which includes bfloat16) for this path would unconditionally trigger this assertion. Narrowing to [float16, float32] aligns the benchmark coverage with the operator's actual supported dtypes.

PR #4154, opened 2026-06-18
[KernelGen][Metax] Use the available flash attention benchmark reference

lukatao:
Retain a dedicated Metax flash_attn wrapper with 4D paged KV inputs and supported keyword arguments. Preserve the separate Iluvatar wrapper. The reviewer accepted that the missing vLLM reference API is a baseline issue. Sync current upstream test setup so the requested Metax run can start.

Validation: The original generic CI passed, but target Metax execution still needs validation. The first new on-demand attempt used the old branch and failed before tests because setup-flaggems was missing. The branch now includes current upstream actions; no new infrastructure workaround was added.


tengqm (commented):
I don't think we should fix the test case to make it work on metax. We may want to fix the implementation directly.

lukatao:
> I don't think we should fix the test case to make it work on metax. We may want to fix the implementation directly.

The operator implementation in FlagGems works correctly on metax, but the benchmark baseline reference comes from  vllm.flash_attn, which is implemented by metax developer. This is a vllm-metax plugin gap, not a FlagGems ops issue. This PR only changes the benchmark file to use an alternative baseline (the flash_attn package directly) for metax, similar to what's already done for iluvatar.

tengqm:
> > I don't think we should fix the test case to make it work on metax. We may want to fix the implementation directly.
> 
> The operator implementation in FlagGems works correctly on metax, but the benchmark baseline reference comes from vllm.flash_attn, which is implemented by metax developer. This is a vllm-metax plugin gap, not a FlagGems ops issue. This PR only changes the benchmark file to use an alternative baseline (the flash_attn package directly) for metax, similar to what's already done for iluvatar.

That makes sense. Thanks for the clarification.

PR #4156, opened 2026-06-18
Fix int32 offset overflow in index/index_put for large tensors

### LINKED ISSUE #4153: [NVIDIA | General] index.py : index kernel overflows int32 offset for large tensors with int32 indices, causing CUDA illegal memory access
## Description

  When running vLLM with FlagGems enabled, inference can crash consistently under high-concurrency long-context workloads with:

  ```CUDA illegal memory access```

  The async CUDA error may initially appear in later ops such as cat.py, index_put, or nonzero. However, with CUDA_LAUNCH_BLOCKING=1, the actual crash is traced to:

  flag_gems/ops/index.py

  ## Triggering scenario

  Observed in a vLLM inference workload with:

  - USE_FLAGGEMS=1
  - around 1000 concurrent requests
  - 4k input / 1k output
  - model code executes something like:

  ssm_state[non_spec_state_indices_tensor]

  In this case:

  ssm_state shape = [24901, 8, 128, 128]
  ssm_state stride0 = 131072
  non_spec_state_indices_tensor dtype = int32

  The crash happens when non_spec_state_indices_tensor contains an index greater than or equal to 16384.

  ## Root cause

  The generated Triton index kernel computes the input memory offset like this:

  cur_index0 = tl.load(indices0_ptr + ..., mask=mask0, other=0)
  input_offset = cur_index0 * input_stride0 + ...

  When the index tensor is int32, cur_index0 is also int32.

  For the tensor above:

  input_stride0 = 8 * 128 * 128 = 131072
  index = 16384
  index * input_stride0 = 16384 * 131072 = 2147483648

  2147483648 exceeds the maximum signed int32 value, 2147483647. As a result, the multiplication overflows in int32, producing an invalid offset. The kernel then accesses an invalid memory
  address and crashes with CUDA illegal memory access.

  ## Minimal reproduction logic

  A minimal shape that exposes the issue:

  input tensor shape = [20000, 8, 128, 128]
  dtype = bf16
  stride0 = 131072
  index dtype = int32
  index value = 16384

  Observed behavior:

  int32 index = 16382 -> PASS
  int32 index = 16383 -> PASS
  int32 index = 16384 -> CRASH

  int64 index = 16384 -> PASS
  int64 index = 19999 -> PASS

  This strongly suggests that the issue is caused by int32 arithmetic in the offset computation.

  ## Expected behavior

  FlagGems index should match PyTorch behavior. Indexing a large tensor should work even when the computed offset exceeds the int32 range.

  Native PyTorch indexing works correctly in the same scenario.

  ## Suggested fix

  The generated Triton kernel should promote values used for address offset computation to int64, for example:

  cur_index0 = tl.load(...).to(tl.int64)

  Alternatively, all variables involved in input_offset computation, such as cur_index, input_idx, and related offset expressions, should be explicitly promoted to tl.int64 before
  multiplication/addition.

  This avoids:

  int32 * int32 -> int32

  overflow.

  ## Related code

  flag_gems/ops/index.py

  Problematic generated logic:

  cur_index{i} = tl.load(indices{i}_ptr + ..., mask=mask0, other=0)
  input_offset = cur_index{i} * input_stride{i} + ...

  ## Additional notes

  This issue may not appear in small-batch or small-cache workloads. It is more likely to happen with high concurrency and long-context inference, where cache block ids can become large
  enough to exceed the int32 offset range.

*this bug occurs in the 5.0.2 FlagGems
*bug happened while vllm running Qwen3-Next-80B-A3B-Instruct(BF16) and doing vllm bench serve (4k-1k-1000 concurrency)

JiaryCoder:
### PR Category
Operator

### Type of Change
Bug Fix

### Description
The Triton `index` / `index_put` kernels compute element offsets as
`offset = index * stride` in int32. When the indexed tensor has more than
2**31 elements (e.g. a GDN/linear-attention `ssm_state` of shape
[24901, 8, 128, 128], stride0 = 131072), any index >= 16384 makes
`index * stride0 >= 2**31`, overflowing signed int32 to a negative offset.
The pointer then lands before the tensor base -> CUDA "illegal memory access".
(In vLLM this surfaced under high concurrency and was often reported from a
later `cat`/`nonzero` launch, because the CUDA error is asynchronous;
`CUDA_LAUNCH_BLOCKING=1` localizes it to the index kernel.)

This adopts the same `INT32_OFFSET` dual-path that `scatter.py` already uses:
- the wrapper computes `int32_offset = can_use_int32_index(...)` over every
  participating tensor and passes it to the kernel as a `tl.constexpr`;
- when `int32_offset` is False, the kernel promotes the index / coordinate
  variables to int64 before the offset multiply.

Because `INT32_OFFSET` is a constexpr, the small-tensor path is unchanged: the
int64 branch is eliminated at compile time, so the generated int32 kernel is
byte-identical to before. Large tensors take the int64 path.

### Changed
- src/flag_gems/ops/index.py        (fix)
- src/flag_gems/ops/index_put.py    (fix, symmetric to the above)
- tests/test_index.py               (regression test for the >2**31-element case)
- tests/test_index_put.py           (regression test for the >2**31-element case)

### Issue

Resolves #4153

### Progress

### Performance
No change on the existing (small-tensor) path: with `INT32_OFFSET=True` the
generated kernel is identical to the current one (constexpr dead-code
elimination), so no extra branch / register / ALU cost. For tensors above
2**31 elements the int64-offset path is taken; since this op is a
memory-bandwidth-bound gather/scatter, the extra int64 offset arithmetic is
hidden behind memory traffic and the throughput impact is negligible. Added
two regression tests (`tests/test_index.py`, `tests/test_index_put.py`) that
exercise the >2**31-element case with int32 indices.
### Reproduction & Verification
Before — vLLM `bench serve` (4k input / 1k output, 1000 concurrency) crashes in
the Qwen3-Next GDN linear-attention index/index_put path:
After this fix — the same benchmark completes with 0 failed requests
(1000 successful / 0 failed)

### DIFF
--- a/src/flag_gems/ops/index.py
+++ b/src/flag_gems/ops/index.py
@@ -41,6 +41,7 @@ def generate_imports(code: IndentedBuffer) -> IndentedBuffer:
     code.writeline("from flag_gems.utils import libentry, libtuner")
     code.writeline("from flag_gems import runtime")
     code.writeline("from flag_gems.utils.shape_utils import volume")
+    code.writeline("from flag_gems.utils.shape_utils import can_use_int32_index")
     code.writeline("from flag_gems.utils import triton_lang_extension as ext")
 
     code.newline()
@@ -76,6 +77,7 @@ def generate_index_kernel(
         args += [
             "M,",
             "N,",
+            "INT32_OFFSET: tl.constexpr,",
             "BLOCK_SIZE0: tl.constexpr,",
             "BLOCK_SIZE1: tl.constexpr,",
         ]
@@ -111,6 +113,18 @@ def generate_index_kernel(
             code.writeline(
                 f"cur_index{i} = tl.load(indices{i}_ptr + {' + '.join(comp)}, mask=mask0, other=0)"
             )
+        code.newline()
+
+        # Promote to int64 when INT32_OFFSET is False to prevent overflow
+        code.writeline("if not INT32_OFFSET:")
+        with code.indent():
+            for i in range(indices_len):
+                code.writeline(f"cur_index{i} = cur_index{i}.to(tl.int64)")
+            for i in range(index_rank):
+                code.writeline(f"indices_idx{i} = indices_idx{i}.to(tl.int64)")
+            for i in range(indices_len, inp_rank):
+                code.writeline(f"input_idx{i} = input_idx{i}.to(tl.int64)")
+
         code.newline()
         index_mask = [
             f"(cur_index{i} >= 0) & (cur_index{i} < input_shape{i})"
@@ -160,6 +174,15 @@ def generate_index_wrapper(
         code.writeline("M = indices[0].numel()")
         code.writeline(f"N = volume(input_shape[{indices_len}: ])")
         code.newline()
+        # Determine if int32 offsets are safe for all tensors
+        code.writeline(
+            "int32_offset = can_use_int32_index(input) and can_use_int32_index(out)"
+        )
+        for i in range(indices_len):
+            code.writeline(
+                f"int32_offset = int32_offset and can_use_int32_index(indices[{i}])"
+            )
+        code.newline()
         code.writeline("grid = lambda meta: (")
         with code.indent():
             code.writeline("triton.cdiv(M, meta['BLOCK_SIZE0']), ")
@@ -180,7 +203,7 @@ def generate_index_wrapper(
             args += [
                 f"out_stride[{i}]," for i in range(index_rank + inp_rank - indices_len)
             ]
-            args += ["M,", "N,"]
+            args += ["M,", "N,", "int32_offset,"]
             code.writelines(args)
         code.writeline(")")
         code.writeline("return input")
--- a/src/flag_gems/ops/index_put.py
+++ b/src/flag_gems/ops/index_put.py
@@ -41,6 +41,7 @@ def generate_imports(code: IndentedBuffer) -> IndentedBuffer:
     code.writeline("from flag_gems.utils import libentry")
     code.writeline("from flag_gems import runtime")
     code.writeline("from flag_gems.utils.shape_utils import volume")
+    code.writeline("from flag_gems.utils.shape_utils import can_use_int32_index")
     code.writeline("from flag_gems.utils import triton_lang_extension as ext")
 
     code.newline()
@@ -71,6 +72,7 @@ def generate_index_put_kernel(
             "M,",
             "N,",
             "IS_ACCUMULATE: tl.constexpr,",
+            "INT32_OFFSET: tl.constexpr,",
             "BLOCK_SIZE0: tl.constexpr = 2,",
             "BLOCK_SIZE1: tl.constexpr = 2048,",
         ]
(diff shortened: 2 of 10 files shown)


JiaryCoder:
Thanks for running these! H20 (common backend) passes. The metax failure is because those vendor backends ship their own copy of index/index_put into their backend/ops , which don't pick up the common fix. 
I will keep working on those vendors~

JiaryCoder:
Update: extended the fix to the vendor backends that ship their own copies of these ops.

_metax / _kunlunxin (index.py + index_put.py) and _mthreads (index_put.py) are code-generator copies, so they get the exact same INT32_OFFSET dual-path as the common backend — the small-tensor path is unchanged (the int64 branch is constexpr-eliminated), only >2**31-element tensors take the int64 path.
_ascend (index.py) is a hand-written Triton kernel rather than a generator, so there's no can_use_int32_index plumbing. I applied the equivalent minimal fix — casting the index/offset to int64 before the * stride multiply. The logic is identical, but it's unconditional and I don't have an Ascend device to verify it. So if ascend failed, emmm, maybe we can drop it and keeping the other backends in this PR.

Could you re-run the vendor CI :D? /test | index:metax, /test | index_put:metax, /test | index:kunlunxin, /test | index_put:kunlunxin, /test | index_put:mthreads, /test | index:ascend

JiaryCoder:
@tengqm Thanks for running~ kunlunxin/metax update — both failures are in the pre-existing test_index_put_mixed_none_and_tensor, neither caused by this PR's int64 change:

metax: RuntimeError at torch.broadcast_to(values, target_shape) in _metax/ops/index_put.py::index_put — host side, before any kernel launch. target_shape = get_max_rank_shape(indices) + inp.shape[len(indices):] drops the size of None-indexed dims: for indices=[idx(30), None] on (1024,1024) it gives [30] instead of [30,1024], so broadcasting values(30,1024) fails. Kernel change in this PR doesn't touch this host-side logic.
kunlunxin: accuracy mismatch on indices_config=(1,0,0,1). The tensor is 16^4 = 65536 elements → can_use_int32_index is True → INT32_OFFSET True → the int64 branch is constexpr-eliminated and the generated kernel is byte-identical to pre-PR, so the wrong result is pre-existing in the non-contiguous mixed-None path.
mthreads: index + index_put both pass.
ascend: looks like a triton-ascend / CANN toolchain mismatch on the runner (vendor ascend-cann850). The first test_index case passes; the failure is triton-ascend building its own npu_utils.cpp (RT_LIMIT_TYPE_SIMT_WARP_STACK_SIZE is not a member of rtLimitType_t in this CANN's headers), before my kernel is ever reached. This one needs the Ascend backend owner; happy to drop the _ascend/ops/index.py change (and skip the new test on Ascend) to keep this PR clean.

@0x45f @huangyiqun All of the above reproduce on the unmodified vendor files. How would you like to proceed — separate issue(s) for the metax/kunlunxin index_put bugs, or xfail that case for those two? I'd prefer to keep this PR scoped to the int32 fix, but I'm happy to take the metax shape bug here if you want (its logic should use broadcast_shape + slice_shape like index_put_ does).

PR #4248, opened 2026-06-22
[SiliconFlow] Fix mode operator on Mthreads backend

winfan1314:
## Summary

This PR adds a Moore Threads MUSA implementation for `torch.mode` in the MThreads backend.

The previous path used the generic `mode` implementation. On Moore Threads S5000, the generic `mode_kernel` scans the sorted reduction dimension element by element and triggers an asynchronous MUSA kernel failure during autotune on shapes such as `(4096, 256), dim=0`. This PR adds a fixed-config chunked scan under `_mthreads/ops/mode.py` and exports it through the MThreads backend registry, so MThreads dispatch no longer uses the generic unsafe kernel.

This addresses internal issue #574.

## Test plan

- [x] `python -m py_compile src/flag_gems/runtime/backend/_mthreads/ops/mode.py src/flag_gems/runtime/backend/_mthreads/ops/__init__.py`
- [x] `pre-commit run isort black flake8`
- [x] MThreads backend-test

## Issue

- WEEKTEST-123
- WEEKTEST-188

### DIFF
--- a/src/flag_gems/runtime/backend/_mthreads/ops/__init__.py
+++ b/src/flag_gems/runtime/backend/_mthreads/ops/__init__.py
@@ -22,6 +22,7 @@ from .log_softmax import (
 )
 from .max import max, max_dim
 from .min import min, min_dim
+from .mode import mode
 from .normal import normal_
 from .one_hot import one_hot
 from .ones import ones
@@ -81,6 +82,7 @@ __all__ = [
     "max_dim",
     "min",
     "min_dim",
+    "mode",
     "normal_",
     "one_hot",
     "ones",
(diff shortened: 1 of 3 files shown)


winfan1314:
@tengqm the current backend test won't execute if test files are not included in changed files. 
e.g.: In this PR I only added an op but did not modify test files.

winfan1314:
> @tengqm the current backend test won't execute if test files are not included in changed files. 
> e.g.: In this PR I only added an op but did not modify test files.

It seems like this backend test is passed but there is no actual run in test-op section.

tengqm (approved):

PR #4249, opened 2026-06-22
Drop experimental operator `fix`

### LINKED ISSUE #4222: Revisit the `fix` experimental operator for adoption
The adoption criteria are as follows:

- pass the accuracy tests used for the existing implementation;
- exhibit performance improvements in terms for speedup;
- works at least on NVIDIA and two other backends.

tengqm:
### PR Category

Operatgor

### Type of Change

Refactor

### Description

The experimental version of the `fix` operator is exactly the same one with the in-tree `fix` operator. The in-tree source code has copyright lines and that is the only difference.

### Issues

closes: #4222


huangyiqun (approved):
LGTM

PR #4255, opened 2026-06-22
[KernelGen][Nvidia] Add linalg_slogdet operator with Triton kernel

Dingxingdi:
## Summary
Adds a Triton kernel for `linalg_slogdet`. Computes the sign and natural logarithm of the absolute value of the determinant of a square matrix using LU decomposition with partial pivoting.

## Testing
- Validated against PyTorch reference on device side

## Performance
Test command: `pytest benchmark/test_linalg_slogdet.py --level core` (NVIDIA H20)

| Configuration | Torch Latency (ms) | Gems Latency (ms) | Speedup |
|---|---|---|---|
| [2, 3, 3] | 0.046 | 0.010 | 4.809 |
| [4, 4] | 0.057 | 0.010 | 5.862 |
| [8, 8] | 0.061 | 0.018 | 3.446 |
| [16, 16] | 0.072 | 0.081 | 0.884 |
| [32, 32] | 0.091 | 0.628 | 0.144 |
| **Arithmetic Mean** | — | — | **3.029** |

## Multi-backend Testing
|---|---|---|---|---|
| Nvidia (H20) | PASS | PASS (5 cases, --level core) | 3.029 | Primary |
| Tianshu | FAIL | PASS (5 cases) | 4.915 | AssertionError: Scalars are not close! |
| Muxi | FAIL | FAIL | — | sum_dim_kernel_inner[grid]( |
| Ascend | FAIL | N/A | — | AssertionError: Tensor-likes are not close! |
| Hygon | FAIL | PASS (5 cases) | 2.156 | AssertionError: Scalars are not close! |

## Files Changed
- `src/flag_gems/ops/linalg_slogdet.py`: Triton kernel implementation
- `tests/test_linalg_slogdet.py`: Accuracy test
- `benchmark/test_linalg_slogdet.py`: Performance benchmark


benchmark/test_linalg_slogdet.py
```diff
@@ -0,0 +1,37 @@
...
+from . import base
+
+# Custom non-pointwise shapes are required because shared pointwise benchmark shapes do not apply.
+SLOGDET_SHAPES = [
+    (2, 3, 3),
+    (4, 4),
+    (8, 8),
+    (16, 16),
+    (32, 32),
+]
```
tengqm, line 17: How did you choose these shapes?

benchmark/test_linalg_slogdet.py
```diff
@@ -0,0 +1,37 @@
...
+
+    def get_input_iter(self, cur_dtype):
+        for shape in self.shapes:
+            A = torch.randn(shape, dtype=cur_dtype, device=self.device)
+            yield (A,)
+
+
+@pytest.mark.linalg_slogdet
+def test_linalg_slogdet():
+    from flag_gems.ops.linalg_slogdet import linalg_slogdet
```
tengqm: Change this line to the following form and move it to the head of the file.
```suggestion
    from flag_gems import linalg_slogdet
```

conf/operators.yaml
```diff
@@ -3763,6 +3763,18 @@ ops:
     stages:
       - beta: '5.0'
       - stable: '5.3'
+  - id: linalg_slogdet
+    description: |
+      Computes the sign and natural logarithm of the absolute value of the determinant of a square matrix using LU decomposition with partial pivoting.
+    for:
+      - linalg_slogdet
```
tengqm:
```suggestion
      - linalg.slogdet
```

src/flag_gems/ops/__init__.py
```diff
@@ -936,6 +933,8 @@
     "sinh_",
     "slice_backward",
     "slice_scatter",
+    "slogdet",
+    "slogdet_",
```
tengqm: Either you don't implement the in-place version, or you register the in-place version to the registry as well.

src/flag_gems/ops/linalg_slogdet.py
```diff
@@ -0,0 +1,233 @@
...
+        A: tensor of shape (*, n, n) where * is zero or more batch dimensions.
+
+    Returns:
+        A tuple (sign, logabsdet, LU, pivots) where:
+        - sign has the same dtype as A
+        - logabsdet is always real-valued
+        - LU is the LU decomposition
+        - pivots is the pivot indices (not computed in this implementation)
+    """
+    logger.debug("GEMS slogdet")
```
tengqm:
```suggestion
    logger.debug("GEMS SLOGDET")
```

src/flag_gems/ops/linalg_slogdet.py
```diff
@@ -0,0 +1,233 @@
...
+            return 2
+
+    return SlogdetResult(sign, logabsdet)
+
+
+def slogdet_(A):
+    """
+    In-place version of slogdet (not supported, raises error).
+    """
+    raise NotImplementedError("In-place slogdet is not supported")
```
tengqm: You add a non-implemented operator?

src/flag_gems/__init__.py
```diff
@@ -57,6 +57,7 @@ def torch_ge(v):
     ("_grouped_mm", group_mm),
     ("_index_put_impl_", _index_put_impl_),
     ("_is_all_true", _is_all_true),
+    ("_linalg_slogdet", slogdet),
```
tengqm: why the leading underscore?

src/flag_gems/ops/linalg_slogdet.py
```diff
@@ -129,7 +129,7 @@ def slogdet(A):
         - LU is the LU decomposition
         - pivots is the pivot indices (not computed in this implementation)
     """
-    logger.debug("GEMS slogdet")
+    logger.debug("GEMS SLOGDET")
```
tengqm: Be consistent here ...
When you decided to use `linalg_slogdet` as the operator ID, you will be looking for `GEMS LINALG_SLOGDET` in the debug output.

This is not a big issue. But if we don't pay attention to it, we may run into big troubles in the future.

benchmark/test_linalg_slogdet.py
Dingxingdi: Added a comment explaining the shape choices. The batched 3x3 case covers the `(*, n, n)` interface, and 4x4 through 32x32 covers the small/medium square matrices targeted by this single-program LU implementation.

src/flag_gems/ops/linalg_slogdet.py
tengqm: Okay, this one is no longer published (see __init__.py), then we don't dump operator ID to the debug stream.

src/flag_gems/ops/linalg_slogdet.py
```diff
@@ -195,39 +195,7 @@ def slogdet(A):
...
-
-    return SlogdetResult(sign, logabsdet)
-
-
-def slogdet_(A):
-    """
-    In-place version of slogdet (not supported, raises error).
-    """
-    raise NotImplementedError("In-place slogdet is not supported")
+    sign, logabsdet, _lu, _pivots = slogdet(A)
```
tengqm, line 200: We need a `logger.debug` at the beginning of this function.

src/flag_gems/ops/__init__.py
Dingxingdi: Done. Removed the unimplemented in-place variant and only export the implemented `linalg_slogdet` operator.

src/flag_gems/ops/linalg_slogdet.py
Dingxingdi: Done. Removed the non-implemented `slogdet_` path and simplified the wrapper to return only `(sign, logabsdet)` for `aten::linalg_slogdet`.

src/flag_gems/__init__.py
Dingxingdi: Done. Registered the public `linalg_slogdet` aten op instead of the internal `_linalg_slogdet` name.

src/flag_gems/ops/linalg_slogdet.py
Dingxingdi: Done. Removed the debug log from the internal slogdet helper since it is no longer the published operator entry.

src/flag_gems/ops/linalg_slogdet.py
Dingxingdi: Done. Added logger.debug("GEMS LINALG_SLOGDET") at the beginning of the public linalg_slogdet entrypoint.

tengqm (commented):
Code clean from my view point. Let's see how the operator behaves on some backends.

tengqm (dismissed):

conf/operators.yaml
```diff
@@ -4356,6 +4356,18 @@ ops:
...
+  - id: linalg_slogdet
+    description: |
+      Computes the sign and natural logarithm of the absolute value of the determinant of a square matrix using LU decomposition with partial pivoting.
+    for:
+      - linalg.slogdet
+    labels:
+      - aten
+      - KernelGen
+    kind:
+      - Tensor
```
tengqm:
```suggestion
      - LinearAlg
```

tengqm:
Everything looks good except for the failing accuracy tests on KLX.

Dingxingdi:
Thanks. I investigated the KLX failures. The generic linalg_slogdet Triton kernel of FlagGems was being registered on Kunlunxin, but this implementation was generated/validated for NVIDIA and produces incorrect results on KLX.

I added linalg_slogdet to Kunlunxin's CUSTOMIZED_UNUSED_OPS so KLX will skip this generic implementation and fall back to its native torch/torch_klx path.

I believe this time /test | linalg_slogdet:kunlunxin could produce correct result.

However, I think maybe we should add specific linalg_slogdet in FlagGems for KLX. That's the goal of FlagGems.

tengqm:
> However, I think maybe we should add specific linalg_slogdet in FlagGems for KLX. That's the goal of FlagGems.

Totally agree. There are two options here. A. Merge this and file an issue about that the operator doesn't work on Kunlunxin. B. Take a closer look at the failure, maybe there is a simple precision and/or dtype conversion issue which only surfaced on Kunlunxin. Both options are acceptable to me.

tengqm:
> this implementation was generated/validated for NVIDIA and produces incorrect results on KLX.

I want to take this opportunity to remind you and maybe other contributors/reviewers. The core value of FlagGems is not that we can do a better job on NVIDIA than the NVIDIA engineers. The core value, instead, is that we are delivering operators that be written once and then run on all backends we support. An operator only that works only on NVIDIA is nothing, at least from my viewpoint. Users don't have to use FlagGems to achieve far better performance.

Caeruleann:
> * Expected 0.6414832472801208 but got 28.608354568481445.

I agree with option A. The KLX test failures show significant correctness issues, it doesn't look like a simple precision issue.

bin913 (approved):

PR #4267, opened 2026-06-23
[KMCompiler]Optimize dequantize_and_gather_k_cache for DeepSeek V4 attention

littlekid-zxt:
### PR Category
Operator | Benchmark

### Type of Change
Performance Optimization

### Description
Optimize the `dequantize_and_gather_k_cache` Triton kernel for DeepSeek V4 attention:

1. **Skip the trailing padding FP8 scale slot** when `nope_dim` is a multiple of `quant_block` (e.g. `448 % 64 == 0`). The original implementation iterated over `scale_slots` which contains one extra padding slot, causing one redundant FP8 load/scale load/store per token.
2. **Remove unnecessary masks for full FP8 quantization blocks**. When `nope_dim % 64 == 0`, all blocks are full 64-element blocks, so the per-block `mask` is eliminated, allowing the compiler to generate simpler, better-vectorized load/store instructions.
3. **Widen the rope copy granularity from 16 to 64 elements**, reducing loop overhead and improving vectorization for the common `rope_dim = 64` case.

The output layout and numerical results remain unchanged; only the kernel efficiency is improved.

### Performance

Benchmark was run on an **H20** GPU using `benchmark/test_deepseek_v4_attention_dequantize_and_gather_k_cache.py` (`mode=kernel`, dtype `torch.bfloat16`). The table below compares the speedup against the vLLM baseline before and after the optimization.

```bash
pytest benchmark/test_deepseek_v4_attention_dequantize_and_gather_k_cache.py -v --record log
```

| shape (batch, seq_len, gather_len, dim, nope, rope) | before speedup | after speedup | absolute gain | relative gain |
| --- | ---: | ---: | ---: | ---: |
| (1, 512, 128, 512, 448, 64) | 0.9968 | 1.0813 | +0.0845 | +8.48% |
| (2, 1024, 256, 512, 448, 64) | 0.9956 | 1.1275 | +0.1320 | +13.25% |
| (4, 2048, 512, 512, 448, 64) | 1.0000 | 1.1827 | +0.1827 | +18.27% |
| (4, 2048, 2048, 512, 448, 64) | 0.9977 | 1.2194 | +0.2217 | +22.22% |
| (8, 4096, 1024, 512, 448, 64) | 1.0000 | 1.1935 | +0.1935 | +19.35% |
| (1, 8192, 8192, 512, 448, 64) | 0.9982 | 1.2118 | +0.2135 | +21.39% |
| (1, 16384, 16384, 512, 448, 64) | 0.9992 | 1.2139 | +0.2147 | +21.49% |
| (1, 32768, 32768, 512, 448, 64) | 0.9985 | 1.2169 | +0.2184 | +21.87% |
| (1, 65536, 65536, 512, 448, 64) | 0.9994 | 1.2186 | +0.2192 | +21.93% |
| (16, 1024, 1024, 512, 448, 64) | 0.9991 | 1.2043 | +0.2052 | +20.54% |
| (32, 512, 512, 512, 448, 64) | 0.9982 | 1.1976 | +0.1994 | +19.98% |
| (64, 256, 256, 512, 448, 64) | 1.0025 | 1.1892 | +0.1866 | +18.61% |
| (8, 8192, 640, 512, 448, 64) | 1.0046 | 1.1868 | +0.1822 | +18.14% |
| (16, 32768, 640, 512, 448, 64) | 0.9973 | 1.1939 | +0.1966 | +19.71% |
| (4, 65536, 8320, 512, 448, 64) | 1.0001 | 1.2328 | +0.2327 | +23.27% |
| (8, 65536, 8320, 512, 448, 64) | 0.9993 | 1.2244 | +0.2251 | +22.53% |
| (16, 4096, 128, 512, 448, 64) | 0.9913 | 1.1165 | +0.1252 | +12.63% |
| (32, 4096, 128, 512, 448, 64) | 0.9986 | 1.1309 | +0.1324 | +13.26% |

- **Average relative gain**: ~18.7%
- **Min / Max relative gain**: +8.48% / +23.27%
- Larger `gather_len` and long-context shapes benefit the most.


benchmark/test_deepseek_v4_attention_dequantize_and_gather_k_cache.py
```diff
@@ -64,6 +64,19 @@ def set_shapes(self, shape_file_path=None):
...
+            (1, 65536, 65536, 512, 448, 64),
+            (16, 1024, 1024, 512, 448, 64),
+            (32, 512, 512, 512, 448, 64),
+            (64, 256, 256, 512, 448, 64),
+            (8, 8192, 640, 512, 448, 64),
+            (16, 32768, 640, 512, 448, 64),
+            (4, 65536, 8320, 512, 448, 64),
+            (8, 65536, 8320, 512, 448, 64),
+            (16, 4096, 128, 512, 448, 64),
+            (32, 4096, 128, 512, 448, 64),
```
tengqm, line 79: Are these extra shapes absolutely necessary?
I asked because I'm a little bit concerned with the time needs to benchmark this oeprator. We have a hard constraint to finish a benchmark with in 10 minutes.

littlekid-zxt: Thanks for the feedback — the concern about benchmark runtime is totally valid.

I measured the impact of those 4 extra shapes:

Before adding shapes: ~5.5–5.78 s
After adding all extra shapes: ~21.19–23.02 s
If we remove the last 4 shapes: ~16.06–16.21 s
So the last 4 shapes alone add roughly 5–7 seconds (about 25–35% of the current total). They are not strictly required for correctness, but they cover the DeepSeek V4 HCA layer and decode-only SWA window paths that the earlier shapes don’t exercise.

Given the hard 10-minute limit, the current runtime is still far below the budget. But if the goal is to keep each operator benchmark lightweight, we could drop those 4 shapes or keep just a minimal representative subset (e.g., one HCA + one decode case).

What target runtime per operator benchmark would you consider appropriate?

tengqm: Thanks for the confirmation. Less than 30 seconds is totally acceptable, provided that you see these extra shapes necessary.

littlekid-zxt: Thanks for confirming the 30-second budget. Yes, I think the extra shapes are necessary.

The existing shapes don't cover two DeepSeek V4-specific paths:

HCA (c128a) prefill: the gather buffer is much larger (8192 compressed tokens + 128 SWA window) compared to CSA (512 + 128), so it exercises a different memory footprint.
Decode SWA-only gather: decode only gathers the recent 128 tokens, which has different access patterns from prefill.

The current benchmark runs in ~21–23 seconds, so it's still comfortably below the 30-second limit. I'll keep these shapes.

Vincent-Xiao (approved):

PR #4294, opened 2026-06-23
[KernelGen] Fix i0_out mixed-dtype computation and rounding

lukatao:
## Summary
Fix `i0_out` when input and output dtypes differ. Preserve the input before polynomial evaluation, then round the result to the input dtype before storing into the output dtype, matching ATen's result conversion semantics. Integer input promotion remains unchanged.

Address the review with six mixed float16/bfloat16/float32 combinations and retain the `act_out is out_act` assertion. The reference now calls `aten.i0.out` on the original input dtype directly; the previous fp32 reference conversion has been removed because it hid input-dtype rounding when the output was wider.

Update this operator's existing tests to call Gems directly and compare with PyTorch, following the current test convention.

## Validation
- A CPU harness executing the actual wrapper and kernel expressions with tensor equivalents reproduced mismatches in 5 of 6 mixed-dtype pairs before the fix; all six passed afterward. This does not substitute for GPU execution.
- All six new test combinations collect successfully.
- Refreshed PR CI validates the accuracy tests on GPU.

Related work item: WEEKTEST-541.


w1120029931-bit:
The kernel-side change looks reasonable: out dtype should not force input dtype before computation.
But the current test does not cover the main changed behavior, because out_act is created with empty_like(inp), so input dtype and out dtype are still the same.
Please add tests for mismatched input/output dtypes, keep the `act_out is out_act` assertion, and clarify why fp16/bf16 reference is computed via fp32.

lukatao:
Addressed the review request: added all six mismatched float16/bfloat16/float32 input/output dtype pairs and retained output identity checks. The native reference now uses the actual input dtype. The kernel computes internally in fp32, rounds to the input dtype, then stores to the requested out dtype. Current head passed 42 GPU tests and 3 quick CPU-reference tests (1 skipped).

PR #4295, opened 2026-06-23
[KernelGen] Accumulate gather_backward low-precision gradients in float32

lukatao:
Accumulate float16/bfloat16 gradients in float32 before casting back. Test the FlagGems backward function directly against native autograd; repeated-index floating sums use numerical tolerance rather than bitwise equality. Add a deterministic 257-way fractional-gradient regression and bfloat16 coverage.

Validation: Six new reference/input cases pass with CPU PyTorch stand-ins. GPU checks are pending.


bin913:
@Schopenhauer-loves-Hegel Please fix failing check

PR #4300, opened 2026-06-23
[KernelGen] Fix trunc test discovery and benchmark markers

lukatao:
## Summary
Rename the trunc accuracy and benchmark modules to `test_trunc.py`, as requested in review. Correct benchmark selection so `-m trunc` selects the ordinary operator and `-m trunc_` selects the in-place operator.

Synchronize with master and update the existing accuracy tests to compare `flag_gems.trunc` / `flag_gems.trunc_` directly against PyTorch references, following the current rule against using `use_gems()` inside tests. No kernel implementation is changed.

## Validation
- Isolated pytest collection selects exactly one benchmark for each operator marker.
- Updated-test convention check, Black, and `git diff --check` passed.
- Accuracy and benchmark execution are delegated to the refreshed GPU CI.

Related work item: WEEKTEST-533.


benchmark/test_trunc.py
```diff
@@ -4,7 +4,7 @@
 from . import base, consts
```
tengqm, line 20: The file should be renamed to 'test_trunc.py'.

tengqm (dismissed):

tengqm (dismissed):

huangyiqun (approved):
Markers and direct API calls look good.

PR #4325, opened 2026-06-24
Add special_ndtri operator with Triton kernel

Eclipse-Arrebol:
### PR Category
Operator

### Type of Change
New Feature

### Description
Adds a Triton kernel for `special_ndtri`, the inverse of the standard normal CDF (quantile function), together with the `special_ndtri.out` variant.

The kernel is implemented using the Cephes rational polynomial approximation, split into three regions for numerical stability:
- Central region (tail probability > exp(-2)): rational approximation P0/Q0 centered at 0.5.
- Near tail (sqrt(-2*ln y) < 8): change of variable to sqrt(-2*ln y) with rational correction P1/Q1.
- Far tail (sqrt(-2*ln y) >= 8): rational correction P2/Q2 for extreme probabilities.

Distribution symmetry `ndtri(p) = -ndtri(1-p)` is used to fold the upper tail onto the lower tail. Edge cases are handled explicitly: `p=0 -> -inf`, `p=1 -> +inf`, and `p<0 or p>1 -> NaN`.

Note: PyTorch's native `ndtri` on CUDA does not support fp16/bf16. This Triton implementation supports fp16/bf16/fp32 via an fp32 intermediate computation, which is a functional improvement over the native operator.

Files changed:
- `src/flag_gems/ops/special_ndtri.py`: Triton kernel implementation
- `tests/test_special_ndtri.py`: accuracy tests
- `benchmark/test_special_ndtri.py`: performance benchmark
- `src/flag_gems/ops/__init__.py`: register import and `__all__`
- `src/flag_gems/__init__.py`: register to `_FULL_CONFIG`
- `conf/operators.yaml`: add operator entry

### Issue
N/A

### Progress

### Performance
Accuracy validated against `torch.special.ndtri` on fp16/bf16/fp32 (18/18 tests pass), over shapes (2,3)/(128,256)/(512,512). fp32 max abs error ~2.4e-7 (machine epsilon).

Benchmark is fp32-only because PyTorch's native CUDA `ndtri` does not support fp16/bf16 — consistent with the existing `special_gammainc` / `special_chebyshev` benchmarks, which are also fp32-only.

Test command: `pytest benchmark/test_special_ndtri.py --level core -s` (RTX 4090, float32)

**special_ndtri**
| Size | Torch Latency (ms) | Gems Latency (ms) | Speedup |
| --- | --- | --- | --- |
| [1073741824] | 9.131 | 9.134 | 1.000 |
| [64, 64] | 0.006 | 0.004 | 1.500 |
| [4096, 4096] | 0.153 | 0.145 | 1.049 |
| [64, 512, 512] | 0.151 | 0.143 | 1.056 |
| [1024, 1024, 1024] | 9.126 | 9.143 | 0.998 |

**special_ndtri.out**
| Size | Torch Latency (ms) | Gems Latency (ms) | Speedup |
| --- | --- | --- | --- |
| [1073741824] | 9.132 | 9.138 | 0.999 |
| [64, 64] | 0.006 | 0.005 | 1.200 |
| [4096, 4096] | 0.152 | 0.144 | 1.050 |
| [64, 512, 512] | 0.152 | 0.143 | 1.057 |
| [1024, 1024, 1024] | 9.122 | 9.136 | 0.998 |

**Multi-backend Testing**
| --- | --- | --- | --- | --- |
| Nvidia (RTX 4090) | PASS | PASS (--level core) | ~1.05 | Primary |

Only NVIDIA hardware (RTX 4090) was available for testing. The kernel is backend-neutral Triton and should run on other FlagGems-supported backends, but those were not verified due to lack of hardware access.


conf/operators.yaml
```diff
@@ -7107,6 +6982,24 @@ ops:
     stages:
       - beta: '5.0'
       - stable: '5.3'
+  - id: special_ndtri
+    description: |
+      Computes the argument x for which the area under the standard normal
+      probability density function (integrated from minus infinity to x)
+      equals the input, elementwise. Inverse of the normal CDF.
+    for:
+      - special_ndtri
```
tengqm, line 7033:
```suggestion
      - special.ndtri
```

conf/operators.yaml
```diff
@@ -7107,6 +6982,24 @@ ops:
...
+    labels:
+      - aten
+      - pointwise
```
tengqm, line 7036: Add kind and stages

conf/operators.yaml
```diff
@@ -7107,6 +6982,24 @@ ops:
...
+  - id: special_ndtri_out
+    description: |
+      A variant of `special_ndtri()` with output saved to provided `out`.
+    for:
+      - special_ndtri.out
```
tengqm, line 7045:
```suggestion
      - special.ndtri.out
```

conf/operators.yaml
```diff
@@ -7107,6 +6982,24 @@ ops:
...
+    labels:
+      - aten
+      - pointwise
```
tengqm, line 7048: Same here

src/flag_gems/ops/special_ndtri.py
```diff
@@ -0,0 +1,151 @@
...
+    out = torch.empty_like(x)
+    return _run_special_ndtri_kernel(x, out)
+
+
+def special_ndtri_out(x: torch.Tensor, out: torch.Tensor):
+    """ATen wrapper: special_ndtri.out(Tensor self, Tensor out) -> Tensor"""
+    if x.shape != out.shape:
+        x = x.expand(out.shape)
+    _run_special_ndtri_kernel(x, out)
+    return out
```
tengqm, line 157: Missing logger.debug

src/flag_gems/__init__.py
```diff
@@ -604,6 +601,8 @@ def torch_ge(v):
     ("special_i0e.out", special_i0e_out),
     ("special_i1", special_i1),
     ("special_i1.out", special_i1_out),
+    ("special_ndtri", special_ndtri),
+    ("special_ndtri.out", special_ndtri_out),
```
tengqm, line 609:
```suggestion
    ("special.ndtri", special_ndtri),
    ("special.ndtri.out", special_ndtri_out),
```

conf/operators.yaml
Eclipse-Arrebol: Done.

conf/operators.yaml
Eclipse-Arrebol: Added kind: Math and stages: alpha '5.4'.

conf/operators.yaml
Eclipse-Arrebol: Done.

conf/operators.yaml
Eclipse-Arrebol: Added kind: Math and stages: alpha '5.4'.

src/flag_gems/ops/special_ndtri.py
Eclipse-Arrebol: Added logger.debug to both wrappers.

src/flag_gems/__init__.py
Eclipse-Arrebol: Thanks for the suggestion! I tried the dotted form special.ndtri / special.ndtri.out, but it causes the operator registration to fail at runtime:
Skipped registering operator: expected eof but found '.' here: special.ndtri.out
When registration fails, the op falls back to PyTorch's native ndtri_cuda (no fp16/bf16 support), so 12 half-precision tests break.
The existing working ops like special_i0e and special_i1 use the underscore form in both the registration table and the yaml for: field, so I reverted to special_ndtri / special_ndtri.out for consistency. FYI special.gammainc.out in the current code shows the same warning, so it may have the same issue.
Happy to adjust if you prefer a different approach.

tengqm: Encountered the same situation somewhere else...
So it means that in Torch, there is not operator named 'special.ndtri.out', right?
We are not replacing existing ops, we are attempting to invent a new one?

My understanding is that if we use the `special_i1` form, we are also adding new operators to Torch rather than reimplementing existing one in Triton. I may be wrong though.

Eclipse-Arrebol: Right — `special.ndtri.out` (the dotted name) does not exist in Torch. But the operator itself does exist; its real name just uses the underscore: `special_ndtri.out`. So it's not that the op is missing — we're reimplementing an existing ATen op, just under its actual (underscore) name.

Here's the evidence:

```python
>>> torch.ops.aten.special_ndtri.default._schema
aten::special_ndtri(Tensor self) -> Tensor
>>> torch.ops.aten.special_ndtri.out     # exists
aten.special_ndtri.out
>>> torch.ops.aten.special.ndtri         # does NOT exist
AttributeError
```

`special` isn't a namespace — the `.` in `torch.special.ndtri` is just the Python module path, which dispatches to the ATen op `aten::special_ndtri`. That's why all the working `special_*` ops use the underscore form.

**Why the dotted form fails to register.** In `op_registrar.py`, the registrar passes the config key straight to PyTorch's `Library.impl`:

```python
def register_impl(self, key, fn, extra_dispatch_keys=()):
    ...
    self.lib.impl(key, fn, device_key)   # key = the string in _FULL_CONFIG
```

`self.lib` is `torch.library.Library("aten", "IMPL")`, so the key has to be a valid ATen name that PyTorch's schema parser accepts (`op_name` or `op_name.overload`). A three-segment dotted key like `special.ndtri.out` makes PyTorch's parser stop at the first `.`, raising `expected eof but found '.'`. And in `for_each` the failure is caught and only logged, not raised:

```python
def for_each(self):
    for key, func, extra_dispatch_keys in self.config:
        try:
            self.register_impl(key, func, extra_dispatch_keys)
        except Exception as e:
            error.register_error(e)   # logged as a warning, not raised
```

So the op is silently skipped and falls back to the native kernel — which is why the fp16/bf16 tests broke (native `ndtri_cuda` doesn't support half types).

**Reproduced on the existing `special.gammainc.out` entry** (`__init__.py` line 599). Just enabling FlagGems triggers it:

```
Skipped registering operator: expected eof but found '.' here:
special.gammainc.out
```

So I've reverted to `special_ndtri` / `special_ndtri.out`, consistent with all the other working `special_*` ops (18/18 tests pass). It looks like `special.gammainc.out` has the same latent issue — happy to file a separate PR to fix it if useful.

tengqm: file a separate PR and fix it then.

PR #4361, opened 2026-06-25
update vendor's debug logger

huangyiqun:
### PR Category
Other

### Type of Change
Other

### Description
Standardize debug log formats across different vendors.

### Issue
N/A

### Progress

### Performance
N/A


src/flag_gems/runtime/backend/_arm/ops/index_select.py
```diff
@@ -39,7 +39,7 @@ def index_select_kernel(
 
 
 def index_select(inp, dim, index):
-    logger.debug("GEMS INDEX SELECT")
+    logger.debug("GEMS_ARM INDEX SELECT")
```
tengqm: This is a great cleanup. Since we are making a global fix up, why don't we take one step further, change this to `GEMS_ARM INDEX_SELECT`? `INDEX_SELECT` maps to the operator ID, but `INDEX SELECT` is nothing.

huangyiqun: great suggestion

PR #4364, opened 2026-06-25
[MTHREADS] fix soft_margin_loss performance with two-phase reduction

Oslomayor:
The original _soft_margin_loss_sum_kernel used tl.atomic_add on a single scalar output for every thread block, causing severe atomic contention when grid size is large (e.g., ~1M blocks for 1G elements). This made the kernel 50-70x slower than the PyTorch native implementation.

Fix: adopt the two-phase reduction pattern (already used by smooth_l1_loss and vdot in this repo):

Phase 1 - _soft_margin_loss_partial_sum_kernel:
  Each block computes its partial loss sum and writes to mid[pid] via
  tl.store (independent addresses, zero atomic contention).

Phase 2 - _soft_margin_loss_final_reduce_kernel:
  A single block loads the mid array and reduces it to the final scalar
  via tl.sum.

Performance improvement: speedup goes from 0.014-0.022x to 2.4-3.2x across all large shapes (150-230x relative improvement). All 27 accuracy tests pass.

### DIFF
--- a/src/flag_gems/ops/soft_margin_loss.py
+++ b/src/flag_gems/ops/soft_margin_loss.py
@@ -32,8 +32,8 @@ def _soft_margin_loss_elementwise_kernel(
 
 
 @triton.jit
-def _soft_margin_loss_sum_kernel(
-    x_ptr, y_ptr, out_ptr, n_elements, BLOCK_SIZE: tl.constexpr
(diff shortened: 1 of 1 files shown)


tengqm:
This is BEAUTIFUL!!!

tengqm (approved):

PR #4369, opened 2026-06-25
[MTHREADS] reduce one_hot latency 

Oslomayor:
## Problem
The one_hot operator had three nearly-identical comparison kernels (`one_hot_kernel_16/32/64`) with hard-coded class dimensions, plus a separate scatter path.  Benchmark speedup vs. PyTorch native was only 0.012–0.122× for core-level shapes [64,64] with `num_classes` in {16,64}.

Root causes:
1. Redundant GPU validation — independent `.any()` calls each triggered a kernel launch + device sync, dominating the critical path for small tensors (~70% of total latency).
2. Code duplication — three comparison kernels differing only in the hard-coded `tl.arange(N)`, making maintenance fragile.
3. Fixed block sizes — `BLOCK_M=128` (comparison) and `BLOCK_SIZE=128` (scatter) were hard-coded, leaving GPU occupancy on the table for larger inputs.

## Solution

1. **Unified comparison kernel** — Replace the three comparison kernels with a single templated `one_hot_comparison_kernel` parameterized by `BLOCK_N` (the class tile size).  The dispatch threshold (`num_classes <= 64` for comparison, `> 64` for scatter) is preserved because:
   - Small `num_classes`: a single coalesced write pass has lower overhead than `torch.zeros` + scatter (avoiding an extra kernel launch and the memset traffic).
   - Large `num_classes`: scatter (pre-zero + write-1s) reduces the write volume from O(n×k) to O(n), where `torch.zeros` uses the device's optimized memset.

2. **Fused validation** — Combine the two validation `.any()` calls into one fused check, eliminating one kernel launch + sync per call.  The error branch still provides the specific message via a secondary individual check (only executed on failure).

3. **Dynamic block sizing** — `BLOCK_M` scales 32→128→256→512 based on `num_elements` for the comparison kernel, and `BLOCK_SIZE` scales 256→1024→2048 for the scatter kernel, improving GPU occupancy on larger tensors.

4. **Device-side assert** — Add `tl.device_assert` inside both kernels as a belt-and-suspenders check.  Note: on MUSA this API is a no-op; Python-level GPU validation is retained as the primary correctness mechanism.

## Benchmark (core level, shape [64,64], MUSA)

| num_classes | Original (ms) | Optimized (ms) | Delta  | Speedup vs. PyTorch |
|-------------|--------------|---------------|--------|---------------------|
| 16          | 0.3533       | 0.2727        | -22.8% | 0.033               |
| 64          | 0.3537       | 0.2758        | -22.0% | 0.015               |
| -1 (→16)    | 0.5476       | 0.5084        | -7.2%  | 0.121               |
| -1 (→64)    | 0.5488       | 0.5138        | -6.4%  | 0.120               |

On larger tensors ([1024,1024], `num_classes=256`) the speedup reaches 0.89× vs. PyTorch native.

The remaining gap on small tensors comes from the unavoidable GPU validation sync (~0.06ms) and Triton kernel launch overhead.  An attempt to eliminate validation via `tl.device_assert` was reverted: MUSA treats `tl.device_assert` as a no-op, silently producing incorrect output for invalid indices, making Python-level validation the only reliable correctness safeguard on this backend.


tengqm:
please always rebase your PR to the master head.

Oslomayor:
The speedup did not meet expectations. Let me convert this to a draft.

Oslomayor:
I close this PR for now since #4468 is merged.

PR #3909, opened 2026-06-09
[KernelGen][Nvidia] Add amin operator with Triton kernel

XDYuanzhuLee:
## Summary
Adds a Triton kernel for `amin`. Returns the minimum value of each row of the input tensor in the given dimension dim.

## Testing
- Validated against reference on device via `to_reference(inp, True)`
- Tested on: Nvidia, Tianshu, Muxi, Ascend, Hygon

## Performance
Test command: `pytest benchmark/test_amin.py --level core` (NVIDIA H20)

### amin

| dtype | Size | Torch Latency (ms) | Gems Latency (ms) | Speedup |
|-------|------|--------------------|-------------------|---------|
| float16 | [1048576] | 0.013376 | 0.009616 | 1.391 |
| float16 | [64, 64] | 0.006720 | 0.005856 | 1.148 |
| float16 | [4096, 4096] | 0.026880 | 0.016928 | 1.588 |
| float16 | [64, 512, 512] | 0.049472 | 0.151904 | 0.326 |
| float16 | [1024, 1024, 1024] | 1.160192 | 9.517792 | 0.122 |
| float32 | [1048576] | 0.013024 | 0.010224 | 1.274 |
| float32 | [64, 64] | 0.006944 | 0.005920 | 1.173 |
| float32 | [4096, 4096] | 0.033472 | 0.027648 | 1.211 |
| float32 | [64, 512, 512] | 0.036288 | 0.190176 | 0.191 |
| float32 | [1024, 1024, 1024] | 1.051552 | 10.009920 | 0.105 |
| bfloat16 | [1048576] | 0.013184 | 0.009696 | 1.360 |
| bfloat16 | [64, 64] | 0.006912 | 0.005984 | 1.155 |
| bfloat16 | [4096, 4096] | 0.027360 | 0.018496 | 1.479 |
| bfloat16 | [64, 512, 512] | 0.049792 | 0.159056 | 0.313 |
| bfloat16 | [1024, 1024, 1024] | 1.227040 | 9.713424 | 0.126 |

| Operator | Arithmetic Mean Speedup |
|----------|------------------------|
| amin | **0.864** |

### amin_ (in-place)

| dtype | Size | Torch Latency (ms) | Gems Latency (ms) | Speedup |
|-------|------|--------------------|-------------------|---------|
| float16 | [1048576] | 0.017376 | 0.012256 | 1.418 |
| float16 | [64, 64] | 0.009504 | 0.008384 | 1.134 |
[...]

PR #4391, opened 2026-06-26
[KernelGen][Nvidia] Add reduce_min operator with Triton kernel

bwbwzzz:
## Summary

Add `reduce_min` operator with Triton kernel implementation. This operator computes the minimum value over all elements of a tensor, equivalent to `torch.amin(input)`. Supports float16, float32, bfloat16, and integer dtypes.

### float16

| Shape | Torch (ms) | Gems (ms) | Speedup |
|-------|-----------|----------|---------|
| [1048576] | 0.013 | 0.013 | 1.00x |
| [64, 64] | 0.007 | 0.008 | 0.98x |
| [4096, 4096] | 0.029 | 0.029 | 0.99x |
| [64, 512, 512] | 0.029 | 0.029 | 0.99x |
| [1024, 1024, 1024] | 1.013 | 0.924 | 1.10x |

**Average: 1.01x (5/5 cases passed)**

### float32

| Shape | Torch (ms) | Gems (ms) | Speedup |
|-------|-----------|----------|---------|
| [1048576] | 0.013 | 0.014 | 0.95x |
| [64, 64] | 0.007 | 0.007 | 1.02x |
| [4096, 4096] | 0.036 | 0.036 | 1.01x |
| [64, 512, 512] | 0.036 | 0.036 | 1.00x |
| [1024, 1024, 1024] | 1.191 | 1.157 | 1.03x |

**Average: 1.00x (5/5 cases passed)**

### bfloat16

| Shape | Torch (ms) | Gems (ms) | Speedup |
|-------|-----------|----------|---------|
| [1048576] | 0.013 | 0.013 | 0.98x |
| [64, 64] | 0.008 | 0.007 | 1.05x |
| [4096, 4096] | 0.029 | 0.029 | 1.01x |
| [64, 512, 512] | 0.029 | 0.029 | 0.98x |
| [1024, 1024, 1024] | 0.973 | 0.934 | 1.04x |

**Average: 1.01x (5/5 cases passed)**

- **Overall**: Average speedup **1.01x** (15/15 cases passed)

## Multi-backend Testing

| Backend | Accuracy | Speedup | Notes |
|---------|----------|---------|-------|
| NVIDIA (H20) | PASS | 1.01x | 15 cases, 3 dtypes |
| Tianshu | — | — | Not tested |
| Muxi | — | — | Not tested |
| Ascend | — | — | Not tested |
| Hygon | — | — | Not tested |

## Files Changed

- `src/flag_gems/ops/reduce_min.py` — Triton kernel implementation
- `src/flag_gems/ops/__init__.py` — Register operator
- `src/flag_gems/__init__.py` — Add to _FULL_CONFIG
- `tests/test_reduce_min.py` — Accuracy tests (float + int dtypes)
- `benchmark/test_reduce_min.py` — Performance benchmark
- `conf/operators.yaml` — Operator metadata


src/flag_gems/__init__.py
```diff
@@ -37,6 +37,7 @@ def torch_ge(v):
 
 
 _FULL_CONFIG = (
+    ("reduce_min", reduce_min),
```
tengqm: why put it here?

src/flag_gems/ops/__init__.py
```diff
@@ -502,6 +503,7 @@
 from flag_gems.ops.zeros_like import zeros_like
 
 __all__ = [
+    "reduce_min",
```
tengqm: You know the list should be sorted, right?

conf/operators.yaml
```diff
@@ -1,4 +1,15 @@
 ops:
+  - id: reduce_min
```
tengqm: The ops array should be sorted in terms of op IDs.

src/flag_gems/__init__.py
```diff
@@ -37,6 +37,7 @@ def torch_ge(v):
```
tengqm, line 38: Is this a fused operator?

```
>>> torch.reduce_min
  File "/home/tengqm/FlagGems/.venv/lib/python3.12/site-packages/torch/__init__.py", line 2757, in __getattr__
    raise AttributeError(f"module '{__name__}' has no attribute '{name}'")
AttributeError: module 'torch' has no attribute 'reduce_min'
```

tests/test_reduce_min.py
```diff
@@ -0,0 +1,25 @@
...
+def test_reduce_min(shape, dtype):
+    if dtype in utils.FLOAT_DTYPES:
+        inp = torch.randn(shape, dtype=dtype, device=flag_gems.device)
+    else:
+        inp = torch.randint(-10000, 10000, shape, dtype=dtype, device="cpu").to(
+            flag_gems.device
+        )
+    ref_inp = utils.to_reference(inp)
+
+    ref_out = torch.amin(ref_inp)
```
tengqm, line 21: Are we reimplementing `amin` and giving it a new name `reduce_min`?

src/flag_gems/__init__.py
bwbwzzz: have revised， thanks

tests/test_reduce_min.py
bwbwzzz: Thanks for pointing this out. I agree that reduce_min is essentially amin here. I checked the relevant documentation, and it does seem to be the case.

Could you please clarify whether this means I should drop this operator from the PR, or whether I misunderstood the intended functionality of reduce_min?

src/flag_gems/__init__.py
```diff
@@ -516,6 +516,7 @@ def torch_ge(v):
     ("randperm", randperm),
     ("reciprocal", reciprocal),
     ("reciprocal_", reciprocal_),
+    ("reduce_min", reduce_min),
```
tengqm, line 519: what is the dispatch key for this operator?

tests/test_reduce_min.py
tengqm: Looks to me that 'reduce_min' is a tensorflow operator rather than a torch operator.

bwbwzzz: he key "reduce_min" in _FULL_CONFIG is not a valid ATen operator. PyTorch uses aten::amin (or aten::min) instead of aten::reduce_min, which aligns with the test file where torch.amin(ref_inp) is used for the reference.
Thank you for your guidance and suggestions. I think I should probably close this PR?

tengqm: Then why don't we retarget this implementation to `aten::amin`?
What's the difference between `aten::min` (we already have it?) and `atan::amin`?

bwbwzzz: Thank you so much for your guidance! Your point is completely correct.

According to the documentation, torch.min(input, dim, ...) returns a namedtuple of (values, indices), whereas amin/amax supports reduction over multiple dimensions and does not return indices.

Since the amin operator has already been implemented with a Triton kernel in PR #3909 (by XDYuanzhuLee), there is no need to register its interception logic again in full_config. Doing so would cause redundant interception and could potentially introduce syntax or runtime errors due to the signature mismatch.

Thank you again for steering me in the right direction

PR #4402, opened 2026-06-26
[HYGON] Add C++ wrapper support on HCU

fangzexian:
### Progress

### DIFF
--- a/cmake/BackendHCU.cmake
+++ b/cmake/BackendHCU.cmake
@@ -1,0 +1,86 @@
+# ==============================================================================
+# HCU (Hygon) Backend Configuration
+# ==============================================================================
+message(STATUS "Configuring HCU backend...")
+
+# HCU uses HIP runtime. Find HIP package — provided by DTK (Hygon Developer ToolKit).
+#
+# Users can set the DTK_PATH environment variable to point to DTK installation:
+#   export DTK_PATH=/opt/dtk-26.04
+#
+# If DTK_PATH is not set, the build will fall back to common DTK installation
+# paths (/opt/dtk, /opt/dtk-*) and ROCm-compatible paths for flexibility.
+if(DEFINED ENV{DTK_PATH})
+    list(APPEND CMAKE_PREFIX_PATH
+        "$ENV{DTK_PATH}"
+        "$ENV{DTK_PATH}/hip"
+    )
+    message(STATUS "Using DTK_PATH (env): $ENV{DTK_PATH}")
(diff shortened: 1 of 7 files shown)


cmake/BackendHCU.cmake
```diff
@@ -0,0 +1,50 @@
+# ==============================================================================
+# HCU (Hygon) Backend Configuration
+# ==============================================================================
+message(STATUS "Configuring HCU backend...")
+
+# HCU uses HIP runtime. Find HIP package — typically provided by DTK installation
+# (Hygon Developer ToolKit) at /opt/dtk or /opt/dtk-<version>.
```
tengqm: It might be a common scenario that users install DTK to, say, `/opt/dtk-26.04` without creating a symbol link `/opt/dtk`.

A related question, is 'ROCM_PATH' or 'HIP_PATH' a famous configuration option on Hygon machines?

The net effect could be that users get confusing error messages.

Suggestion, expose a single environment variable for users, say 'DTK_PATH' or 'DTK_HOME' or 'DTK_ROOT', whatever. Derive all other environment variables from this single knob. If the hooking root is not specified, ask users to specify it.

lib/CMakeLists.txt
```diff
@@ -28,8 +28,7 @@ add_library(operators
             rwkv_ka_fusion.cpp
             copy.cpp
             accuracy_utils.cpp
-            fp8_matmul.cpp
-            fp8_matmul_direct.cpp)
```
tengqm, line 32: is this a bug fix?

cmake/BackendHCU.cmake
fangzexian: Thanks for the suggestion. Updated BackendHCU.cmake:

1. Added DTK_PATH as the primary configuration knob, supporting both
   environment variable (export DTK_PATH=...) and cmake variable
   (-DDTK_PATH=...).

2. When DTK_PATH is not set, the build falls back to scanning
   common paths: /opt/dtk, /opt/dtk-* (versioned directories like
   /opt/dtk-26.04), and legacy ROCM_PATH/HIP_PATH from env.sh.

3. If HIP is still not found, a clear error message tells users to
   set DTK_PATH or source the DTK environment.

lib/CMakeLists.txt
fangzexian: This was a leftover from local testing. The change has been reverted, fp8 files are back in the source list.

tengqm (approved):

0x45f (approved):

PR #4411, opened 2026-06-27
Merge two in-place operators

tengqm:
### PR Category

Operator

### Type of Change

Refactor

### Description

Merge two in-place operators which makes no sense to have standalone source files.


tengqm:
not working.

PR #4426, opened 2026-06-29
[MTHREADS] fix log10 perf

Oslomayor:
Add MTHREADS log10 backend reusing log_kernel with a SCALE parameter.

The generic log10_ operator used @pointwise_dynamic which scored 0.27x-0.45x speedup (2-4x slower than PyTorch MUSA baseline).

Instead of duplicating the log kernel, this patch adds a SCALE constexpr to the existing autotuned log_kernel so that log10 can reuse the same code path (including tl_extra_shim.log for fp32 and fast bit-manipulation approximation for fp16/bf16).

Changes:
- log.py: add SCALE: tl.constexpr=1.0 to log_kernel, pass through _launch_log (zero overhead for existing log)
- log10.py (new): call _launch_log with scale=1/ln(10)
- \_\_init\_\_.py: register log10/log10_/log10_out for dispatch

Before -> After (speedup, Gems/Torch):
fp16  [4096,4096]  0.33x -> 1.15x  
fp16  [1024,65536] 0.27x -> 1.02x  
fp32  [1024,65536] 0.44x -> 1.02x  
bf16  [1024,65536] 0.28x -> 1.01x

### DIFF
--- a/src/flag_gems/runtime/backend/_mthreads/ops/__init__.py
+++ b/src/flag_gems/runtime/backend/_mthreads/ops/__init__.py
@@ -14,6 +14,7 @@ from .index_add import index_add, index_add_
 from .index_put import _index_put_impl_, index_put, index_put_
 from .index_select import index_select
 from .log import log
+from .log10 import log10, log10_, log10_out
 from .log_softmax import (
     log_softmax,
     log_softmax_backward,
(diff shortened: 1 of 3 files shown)


huangyiqun (approved):
LGTM

PR #4457, opened 2026-06-30
[MTHREADS] simplify reflection_pad1d_backward: single kernel launch

Oslomayor:
- Remove float32 intermediate buffer and copy kernels -- write directly to output dtype via tl.atomic_add, eliminating 2 extra kernel launches.
- Simplify reflection index formula.
- Use triton.next_power_of_2(W_out) for adaptive BLOCK_W.
- Simplify no-padding fast path to a direct clone.

Benchmark:  
  fp16  2x3:   0.76x -> 0.99x vs Torch  
  fp16  8x128: 0.66x -> 0.76x vs Torch  
  fp32:        0.62x -> 0.62x (unchanged, was already single-launch)  

  All 48 accuracy tests pass.

### DIFF
--- a/src/flag_gems/ops/reflection_pad1d_backward.py
+++ b/src/flag_gems/ops/reflection_pad1d_backward.py
@@ -30,22 +30,16 @@ def reflection_pad1d_backward_kernel(
     base_out = pid_b * W_out
     base_in = pid_b * W_in
 
-    # Load gradient from output and cast to float32 for accumulation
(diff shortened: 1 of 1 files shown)


Oslomayor:
Local tests pass on my side with the MTHREADS GPU backend, but the failing CI is reflection_pad1d_backward:hygon, and I don’t have access to a HYGON backend to reproduce or validate a fix. To avoid leaving an unsupported optimization open, I’m closing this PR for now.

PR #4468, opened 2026-07-01
[MTHREADS] Optimize one_hot

Oslomayor:
Optimize one_hot via fused validation, dynamic block sizing

Changes:
- Fused two .any() validation calls into one single pass (saves 1 device sync for the common non-error path).
- When num_classes is inferred from max (-1 path), skip the upper-bound validation entirely since max() already guarantees it.
- Dynamic BLOCK_SIZE for dense kernels: scales 64-512 based on tensor size, with per-block output capped at ~32 kB for L1 cache residency.
- Dynamic BLOCK_SIZE for scatter kernel: scales 256/512/1024.


tengqm:
Although the absolute speedup is not good, the relative improvement is impressive.

tengqm (approved):

PR #2616, opened 2026-04-23
Fix broadcast tensor (stride=0) crash in Hopper mm TMA path

wuhulalala:
### PR Category
<!-- [ Operator | OP Test | Model Test | Benchmark | CI/CD | User Experience | Other] -->
Operator
### Type of Change
<!-- [ Bug Fix | New Feature | Performance Optimization | Refactor | Documentation Update | Other] -->
Bug fix
### Description
<!-- Briefly describe the changes and the purpose of the changes.-->
Fix broadcast tensor (stride=0) crash in Hopper mm TMA path                                                                             
                                                                                                                                        
  Tensors with stride=(0,0) from expand()/broadcast (e.g. sum().backward()                                                                
  gradients) bypassed the existing contiguity check (stride > 1) and hit
  TMA TensorDescriptor's "last dimension must be contiguous" assertion.                                                                   
                                                                                                                                          
  Add stride=0 detection in general_mm() to materialize broadcast tensors                                                                 
  via .contiguous() before entering TMA or non-TMA kernel paths.  
### Issue
#2338 
<!--
List any related issues that this PR resolves, if applicable, for example:
- Resolves #123
- Associated with Feature #456
-->

### Progress

- [ ] Change is properly reviewed (1 reviewer required, 2 recommended).
- [ ] Change is responded to an issue.
- [ ] Change is fully covered by a UT.

### Performance
<!-- Please describe any performance tests you have added or the results of any benchmarks. -->

PR #4471, opened 2026-07-01
fix(mm): copy unaligned stride/base operands to contiguous on Hopper (#2489)

### LINKED ISSUE #2489: mm fails on non-16B-aligned stride tensors in real workloads (e.g. vLLM)
### Describe the bug

`flag_gems.mm` fails when input tensors have strides that are not 16-byte aligned.

This issue is observed not only with manually constructed tensors, but also in real-world inference workloads (e.g. vLLM), where such tensor layouts can naturally occur.

Image

This indicates that the current implementation assumes all input tensors are 16B aligned, which is not always true in practice.

---

### Case : Minimal synthetic example

```python
import torch
import flag_gems

device = "cuda"
dtype = torch.float16

base = torch.randn(5000, device=device, dtype=dtype)

# Construct a tensor with non-16B-aligned stride
a = torch.as_strided(
    base,
    size=(64, 64),
    stride=(65, 1)   # 65 * 2 bytes = 130 bytes (not 16B aligned)
)

b = torch.randn(64, 64, device=device, dtype=dtype)

print("stride:", a.stride())
print("stride * elem_bytes:", [s * a.element_size() for s in a.stride()])

# This triggers the error
c = flag_gems.mm(a, b)

### PR DESCRIPTION
**Pull request by whycoming**
## What & why

On Hopper (SM90), `general_mm` builds a TMA `TensorDescriptor` from each operand and from the output. The descriptor requires **both** the outer stride and the base address to be 16-byte aligned; when they are not, triton raises `strides/base must be 16-byte aligned` and `flag_gems.mm` crashes. Fixes #2489.

The reported case is a gapped view such as `torch.as_strided(base, (64, 64), (65, 1))` (65 * 2B = 130B), a layout that arises naturally in real vLLM inference. The old guard only copied *fully*-strided operands (`stride(0) > 1 and stride(1) > 1`), so such a view slipped through.

While validating the fix on an H20 I found that **five more layouts crash on master with the same error**, all for the same reason — nothing validates the stride that actually reaches the descriptor:

| layout | why it crashes on master |
|---|---|
| gapped view, unaligned outer stride | the reported #2489 case |
| stride-contiguous view, odd `storage_offset` | 16B-aligned strides, but an unaligned **base** |
| **clean transpose whose M is not 16B-aligned** | `general_mm` builds a column-major operand's descriptor from `a.T.stride() == (M, 1)`, and `is_tma_compatible` validates only N and K — **never M**. `torch.mm(x.t(), w)` (e.g. `grad_w = x.t() @ grad_out`) with `M % 8 != 0` crashes. |
| **an `M == 1` operand** | `x.t()` of a `(K, 1)` input is `(1, K)` strided `(1, 1)`. A size-1 dim never constrains contiguity, so `is_contiguous()` is True and *both* `.contiguous()` and `.clone()` preserve the stride — the descriptor gets a 2-byte outer stride. |
| **unaligned `out`** | `c_desc` is built from the caller's `out`, which nothing checks. |
| **column-major `out`** | unlike an operand, `c` is fed to the descriptor *untransposed*, so TMA needs its innermost stride to be 1. |
| **`router_gemm` with an unaligned `x`** | it calls the same `general_mm`. No existing test reaches it: every shape in `ROUTER_SHAPES` has `K >= 4096`, so `router_gemm` sends them all to `splitk_mm`. |

## Fix

The fixup happens **inside `general_mm`'s TMA branch** — the only place a descriptor is built, and the only place it is known whether TMA is used at all (`is_tma_compatible` depends on the dtypes and on N/K).

`_is_tma_safe()` checks the operand's *non-unit* stride — the one that actually reaches the descriptor, so a column-major operand is checked on M rather than on K — plus its base address. `_tma_safe()` rebuilds the operand with `empty_like(contiguous_format).copy_()` when it fails; the outer stride then becomes K (resp. N), which `is_tma_compatible` has already established is a multiple of 16 bytes. (`.contiguous()` and `.clone()` are both stride-preserving for an already stride-contiguous operand, which is why they cannot be used here — see the `M == 1` row above.) `out` cannot simply be replaced, so an unaligned or column-major one is computed into an aligned row-major buffer and copied back.

Doing it here rather than in `mm()` matters:

- **No dispatch change.** `mm()`, `mm_out()` and `router_gemm` are byte-identical to master, so no operand layout changes the kernel it is dispatched to. Copying operands *before* dispatch would flip `is_contiguous()` and with it `cluster_remote_mm_scenario` / `streamk_scenario` — e.g. a gapped view with `K % 8 != 0` never uses TMA and runs fine on the strided kernel today, but a pre-dispatch copy would reroute it into the experimental TLE cluster kernel.
- **Nothing is copied that does not need to be.** Operands headed for `gemv`, `splitk`, `streamk`, the cluster kernel, or `general_mm`'s non-TMA branch keep their layout: none of them build a TMA descriptor.
- **`router_gemm` is covered too**, since it calls the same `general_mm`.

## Validation (H20, SM90, torch 2.9.1+cu128, triton 3.5.1)

Same repro run against master and against this branch:

| case | master | this PR |
|---|---|---|
| clean transpose `a`, M=495 (outer stride 990B) | 💥 `strides must be 16-byte aligned` | ✅ |
| gapped view `a`, stride (65,1) = 130B | 💥 `strides must be 16-byte aligned` | ✅ |
| unaligned base `a` (`storage_offset=1`) | 💥 `base must be 16-byte aligned` | ✅ |
| unaligned `out` (`storage_offset=1`) | 💥 `base must be 16-byte aligned` | ✅ |
| gapped `out`, stride (65,1) | 💥 `strides must be 16-byte aligned` | ✅ |
| `M == 1` transpose, stride (1,1) | 💥 `strides must be 16-byte aligned` | ✅ |
| column-major `out` | 💥 `strides must be 16-byte aligned` | ✅ |
| `router_gemm`, unaligned `x`, K=2048 | 💥 `strides must be 16-byte aligned` | ✅ |
| gapped `a`, K=70 (never TMA-compatible) | ✅ (runs on the strided kernel) | ✅ (untouched — must not be perturbed) |

- `tests/test_mm.py` + `tests/test_router_gemm.py` on this branch: **118 passed**.
- Every new regression test fails against master's `mm.py` with a 16-byte-alignment assertion, and passes here; the deliberately *aligned* control cases (`pad=8`, and the aligned-`M` transpose) pass on both, since they must stay on the fast path with no copy.

## Tests

`tests/test_mm.py`: unaligned row-major stride (two sizes × misaligned and aligned gaps), unaligned column-major stride, unaligned `mm_out`, unaligned base (`storage_offset`), a clean transpose with unaligned M, an `M == 1` transpose, an unaligned `out` (odd base and gapped stride) and a column-major `out`. `tests/test_router_gemm.py`: an unaligned `x` with `K < 4096`, so it reaches `general_mm` rather than `splitk_mm`.

### CODE REVIEW

**Comment by huangyiqun** [comment]
Please note that the CI tests are failing.

**Comment by whycoming** [comment]
Heads-up on CI. After the latest push, `python-op / test` is now **11 passed, 1 failed**, and the single remaining failure looks unrelated to this fix:

```
FAILED tests/test_mm.py::test_mm[False-dtype0-495-5333-71]
E   triton.compiler.errors.CompilationError: ...
E   'triton._C.libtriton.ir.builder' object has no attribute 'make_nv_mma_shared_encoding_attr'
    mm.py: in mm -> cluster_remote_mm -> _run_cluster_remote -> _cluster_remote_gemm_kernel (tle_exp.gpu.alloc)
```

This case has **contiguous** inputs (`M,N,K = 495, 5333, 71`, fp16), so `mm()` routes it through the experimental TLE cluster path (`cluster_remote_mm_scenario` → `_cluster_remote_gemm_kernel`), which fails to compile on the CI runner's Triton (missing `make_nv_mma_shared_encoding_attr`). The change in this PR (`_tma_safe`) is a no-op for contiguous operands, so the dispatch here is identical to `main` — it doesn't touch this path.

For transparency on the earlier red: the previous revision of this PR *did* introduce a regression — `_tma_safe` copied a column-major operand to contiguous, which flipped `cluster_remote_mm_scenario` (it requires both operands contiguous) and pushed `test_mm[True-...]` into the same cluster kernel. That's fixed now (clean transposes and aligned views are left untouched), and `test_mm[True-495-5333-71]` passes again.

Could you advise on the `_cluster_remote_gemm_kernel` / Triton incompatibility on the CI runner? It appears to affect any contiguous, N>=256, fp16 mm on SM90, independent of this #2489 change. Happy to adjust anything on this PR's side if useful.

**Comment by whycoming** [comment]
Quick analysis of the two failing checks after the branch update — both appear unrelated to this PR's change:

**`python-op`** — infra issue on the runner: `coverage` is not installed there, so `tools/test-op.sh` exits before pytest even starts ([job log](https://github.com/flagos-ai/FlagGems/actions/runs/28568748219/job/84701656553)):

```
tools/test-op.sh: line 85: coverage: command not found
Running unit tests for tests/test_mm.py
tools/test-op.sh: line 90: coverage: command not found
##[error]Process completed with exit code 1.
```

No tests were executed at all. The same job also needed ~4 minutes of retries to reach github.com during checkout, so that particular worker looks unhealthy. Other PRs' `python-op` jobs today pass on a different worker with the same script, so a re-run should clear this one.

**`backend-tests (nvidia-cuda133) / test`** — the same pre-existing failure as before ([job log](https://github.com/flagos-ai/FlagGems/actions/runs/28568748219/job/84701656698)): `test_mm[False-dtype0-495-5333-71]` (contiguous inputs) routes to the experimental TLE cluster kernel, which fails to compile on the runner's triton (`'triton._C.libtriton.ir.builder' object has no attribute 'make_nv_mma_shared_encoding_attr'`). Contiguous operands don't go through this PR's `_tma_safe` change (it is a no-op for them), so this is the same issue described in my comment above.

One note: since pytest stops at that first failure, the 27 new regression tests added by this PR (92 collected in total) are never reached in CI. They pass locally on an H20.

Happy to rebase/re-run once the runner-side issues are sorted out.

**Comment by bin913** [comment]
@whycoming Please fix failing unit-test

**Comment by whycoming** [comment]
@bin913 The failing case is a CI **environment** problem, and master has already fixed it — I've merged master into this branch to pick the fix up (`f42a3ed`).

**What actually breaks.** `setup.sh` installs FlagTree over Triton, and the next step of the `setup-flaggems` action then reinstalls stock Triton on top of it. From this PR's own h20 log ([job 84766199416](https://github.com/flagos-ai/FlagGems/actions/runs/28588558791/job/84766199416)):

```
Replacing Triton with FlagTree ... [OK]
Installing FlagTree ... [OK]
uv pip install vllm==0.21.0 --torch-backend=cu130
 + tokenspeed-triton==3.7.10.post20260531
 + triton==3.6.0
```

That leaves FlagTree's TLE Python front-end in place — so `HAS_TLE` is `True` and `mm()` dispatches into `_cluster_remote_gemm_kernel` — sitting on stock Triton's `libtriton.so`, which has no `make_nv_mma_shared_encoding_attr`. Hence the `AttributeError` at compile time. This is exactly what #4535 ("safe vllm install to protect torch/triton/flagtree from being overwritten") fixed on 2026-07-06, four days after that run; the step itself had only been introduced on 2026-07-01. Since composite actions are read from the PR's own checkout, a plain re-run would have reused the old, unsafe step — so merging master was the way to pick it up.

**On whether this PR causes it.** `test_mm[False-dtype0-495-5333-71]` uses contiguous fp16 operands, and `_tma_safe()` returns them unchanged: `t.is_contiguous()` short-circuits the `or` before the alignment check is evaluated. `cluster_remote_mm_scenario()` gates only on dtype / contiguity / `M>=64` / `N>=256` / `K>=64`, so it sees exactly what it sees on main — the previous `if a.stride(0) > 1 and a.stride(1) > 1` guard was equally a no-op for that shape. And the error is raised while *compiling* the kernel on module-level constexprs, so no operand transformation could influence it either way.

**One thing I did get wrong, now fixed.** While checking the above I found a real problem in my own patch: `_tma_safe()` was copying zero-stride broadcast operands (from `expand()`) that the old guard left alone. `general_mm()` already copies those itself (#2616), and materializing them in `mm()` made them contiguous — which newly routed `test_mm_broadcast_stride_zero` (fp16, 128x256x256) into the TLE cluster kernel, a path `mm()` never selects for it on main. Zero-stride operands are now returned untouched, so dispatch matches main for every operand layout except the unaligned stride/base views this PR is about.

**Request.** The new run is sitting at `action_required` — could you approve the workflow run so it can execute? Also worth noting: as far as I can tell this is the first PR to run `tests/test_mm.py` on the flagtree-based `nvidia-cuda133` backend, so if the cluster kernel still fails to compile after the environment fix, the gate itself is the suspect — `has_triton_tle()` is an import probe and cannot detect a `libtriton` that lacks the TLE builder ops. I'd be happy to send a capability guard for `cluster_remote_mm_scenario()` as a separate PR, rather than scope-creeping this one.

whycoming:
I rented an H20 and validated this properly, and it turned up two things I got wrong in my previous comment. Correcting both, and the PR has been rewritten as a result (the description is updated too).

**1. My "dispatch is unchanged" claim was wrong.** I wrote that dispatch matched `main` for every layout except the unaligned ones this PR targets. That is not true. `_tma_safe()` ran in `mm()` *before* dispatch, so it could not know whether a TMA descriptor would be built at all — `is_tma_compatible` depends on the dtypes and on N/K. A gapped view with `K % 8 != 0` is never TMA-compatible, runs fine on the strided kernel on master, and my copy turned it contiguous — which newly satisfied `cluster_remote_mm_scenario` and rerouted it into the experimental TLE cluster kernel. That is exactly the kind of change I claimed not to be making.

**2. The fix was incomplete — it did not actually close #2489.** `_tma_safe()` whitelisted clean transposes, but `general_mm` builds a column-major operand's descriptor from `a.T.stride() == (M, 1)`, and `is_tma_compatible` validates only N and K — never M. So `torch.mm(x.t(), w)` (e.g. `grad_w = x.t() @ grad_out`) with `M % 8 != 0` still raised `strides must be 16-byte aligned`. The same applies to `out`: `c_desc` is built from the caller's tensor, which nothing checked. Both still crashed on the previous revision of this PR — I confirmed it on the H20.

**What changed.** The fixup now happens inside `general_mm`'s TMA branch, which is the only place a descriptor is built and the only place it is known whether TMA is used. `_is_tma_safe()` checks the *non-unit* stride — the one that actually reaches the descriptor, so a column-major operand is checked on M — plus the base address. `mm()`, `mm_out()` and `router_gemm` are now byte-identical to `main`, so no operand layout can change the kernel it dispatches to; and `router_gemm`, which calls the same `general_mm`, is covered for free.

**Validation** (H20, SM90, torch 2.9.1+cu128, triton 3.5.1). Same repro against `main` and this branch:

| case | main | this PR |
|---|---|---|
| clean transpose `a`, M=495 (outer stride 990B) | 💥 `strides must be 16-byte aligned` | ✅ |
| gapped view `a`, stride (65,1) = 130B | 💥 `strides must be 16-byte aligned` | ✅ |
| unaligned base `a` (`storage_offset=1`) | 💥 `base must be 16-byte aligned` | ✅ |
| unaligned `out` (`storage_offset=1`) | 💥 `base must be 16-byte aligned` | ✅ |
| gapped `out`, stride (65,1) | 💥 `strides must be 16-byte aligned` | ✅ |
| gapped `a`, K=70 (never TMA-compatible) | ✅ strided kernel | ✅ untouched |

`tests/test_mm.py` on this branch: **101 passed**. The new regression tests against `main`'s `mm.py`: **27 failed** (9 × `base must be 16-byte aligned`, 18 × `strides must be 16-byte aligned`) and 9 passed — the 9 being the deliberately *aligned* control cases, which must stay on the fast path.

So `main` has two more instances of this crash class beyond the one #2489 reports, and this PR now fixes all of them.

On CI: I see #4733/#4734/#4735 are reworking the h100 runner. The `backend-tests` failure on this PR's earlier runs was `_cluster_remote_gemm_kernel` failing to compile (`'triton._C.libtriton.ir.builder' object has no attribute 'make_nv_mma_shared_encoding_attr'`), which needs FlagTree's TLE Python layer sitting on a stock `libtriton` — the overwrite that #4535 fixed. With `COMPILER=triton` the TLE path is simply not taken (`HAS_TLE` is False, as on my H20, where that same contiguous 495x5333x71 case passes). The workflow run on this push is at `action_required`; could someone approve it?

whycoming:
Following up on my own comment above: I audited the fix again and found that "this PR fixes all of them" was still not true. Two more layouts hit the same illegal descriptor, and both slipped through the revision I described. Fixed in `31642a8`, and the description now lists the full set.

**`M == 1` operands.** `x.t()` of a `(K, 1)` input is `(1, K)` with stride `(1, 1)`. A size-1 dim never constrains contiguity, so `is_contiguous()` is True — which means `.contiguous()` *and* `.clone()` are both stride-preserving, and `_tma_safe()` was handing back a tensor its own `_is_tma_safe()` still rejects. The descriptor then got a 1-element (2-byte) outer stride. `empty_like(contiguous_format).copy_()` normalizes the strides unconditionally.

**Column-major `out`.** `_is_tma_safe()` checks the *non-unit* stride, which is right for an operand — `general_mm` transposes a column-major one before building its descriptor — but wrong for `c`, which is always fed untransposed. TMA cannot express a non-unit innermost stride, so `out` additionally has to be row-major; a transposed `out` now goes through the aligned buffer that was already there for the unaligned case.

**`router_gemm` had no test coverage of this path at all.** It calls the same `general_mm`, so I claimed it was covered for free. That is true in code, but every shape in `ROUTER_SHAPES` has `K >= 4096` and therefore short-circuits into `splitk_mm` — `general_mm` is never reached, so the existing tests proved nothing about it. On master, `router_gemm` with an unaligned `x` and `K = 2048` raises `strides must be 16-byte aligned`; there is now a test for it.

Updated validation on the H20 (SM90, torch 2.9.1+cu128, triton 3.5.1) — every row crashes on `main` and passes here:

| case | main | this PR |
|---|---|---|
| clean transpose `a`, M=495 | 💥 `strides must be 16-byte aligned` | ✅ |
| `M == 1` transpose, stride (1,1) | 💥 `strides must be 16-byte aligned` | ✅ |
| gapped view `a`, stride (65,1) | 💥 `strides must be 16-byte aligned` | ✅ |
| unaligned base `a` (`storage_offset=1`) | 💥 `base must be 16-byte aligned` | ✅ |
| unaligned `out` (`storage_offset=1`) | 💥 `base must be 16-byte aligned` | ✅ |
| gapped `out`, stride (65,1) | 💥 `strides must be 16-byte aligned` | ✅ |
| column-major `out` | 💥 `strides must be 16-byte aligned` | ✅ |
| `router_gemm`, unaligned `x`, K=2048 | 💥 `strides must be 16-byte aligned` | ✅ |
| gapped `a`, K=70 (never TMA-compatible) | ✅ strided kernel | ✅ untouched |

`tests/test_mm.py` + `tests/test_router_gemm.py`: **118 passed**. Every new test fails against `main`'s `mm.py` with an alignment assertion.

Two things I should be straight about rather than let a reviewer find them:

- The claim that the *previous* revision would have rerouted a `K % 8 != 0` gapped view into the TLE cluster kernel is an argument from the source, not a measurement: my H20 has stock triton, so `HAS_TLE` is False and the cluster path never executed there. The current revision cannot reroute anything regardless — `mm()`/`mm_out()`/`router_gemm` are byte-identical to `main`.
- I have no perf numbers for copy-to-contiguous versus simply skipping TMA for these operands (i.e. adding `and _is_tma_safe(a) and _is_tma_safe(b)` to the branch predicate and letting `mm_kernel_general` take them with no copy). Happy to benchmark and switch if you would rather not pay a copy on an otherwise-legal layout.

`backend-tests` still has not re-run on this head — the workflow run is at `action_required`. Could someone approve it?

huangyiqun:
@whycoming 
Thanks for the detailed investigation and for updating the fix. I’ve approved the workflow run. Let’s wait for the latest CI results, and I’ll take another look once the checks finish.

huangyiqun:
plz mention ci failed

huangyiqun (commented):
The changed MM and router_gemm tests still use use_gems, which current master rejects. Please migrate them during the rebase.

PR #4537, opened 2026-07-06
[Kunlunxin] Add complex true_divide path for div

llaboon:
Route complex inputs through the generic pointwise_dynamic codegen (which supports register_complex) using Smith's method cross-kernel, since the kunlunxin codegen has no register_complex hook and Triton 3.0 cannot mangle complex64. Skip the four complex div test cases on kunlunxin where the xdnn reference path itself fails.

### Progress


tests/test_div.py
```diff
@@ -80,6 +80,11 @@ def test_div_tensor_mode_float(shape, rounding_mode, dtype):
 @pytest.mark.parametrize("shape", utils.POINTWISE_SHAPES)
 @pytest.mark.parametrize("rounding_mode", ["trunc", "floor"])
 @pytest.mark.parametrize("dtype", utils.INT_DTYPES)
+@pytest.mark.skipif(
+    flag_gems.vendor_name == "kunlunxin",
+    reason="xdnn does not implement integer div with rounding_mode on Kunlunxin (XPU); "
```
tengqm, line 85: Please file an issue and reference that issue in the `reason` field so that the problem can be tracked formally.

tests/test_div.py
```diff
@@ -201,6 +211,14 @@ def test_div_scalar_(shape, scalar, dtype):
 def test_div_scalar_tensor(shape, scalar, dtype):
     if flag_gems.vendor_name == "tsingmicro" and dtype == torch.float16:
         pytest.skip("Issue #3796: not working")
+    if flag_gems.vendor_name == "kunlunxin" and dtype in (
+        torch.float16,
+        torch.bfloat16,
+    ):
+        pytest.skip(
+            "reference torch.div(scalar, low-precision-float tensor) fails on Kunlunxin (XPU); "
```
tengqm, line 219: Maybe we need to file separate issues for these different scenarios.

PR #4541, opened 2026-07-06
Fix cumsum ZeroDivisionError on empty tensor

### LINKED ISSUE #4543: Op cumsum raises ZeroDivisionError on empty tensor
## Describe the bug

`cumsum_wrapper` in `src/flag_gems/ops/cumsum.py` crashes when the input tensor is empty (`numel() ==0`).

The failure occurs at `K = inp.numel() // M // N` when `shape[dim] == 0` (e.g. input shape `[0]` with`dim=0`).

## Expected behavior

Same as PyTorch native `torch.cumsum`:

- No exception
- Output shape equals input shape
- `int32` input → output shape `(0,)`, dtype `int64`

```python
import torch

torch.cumsum(torch.empty(0, dtype=torch.int32), dim=0)
# tensor([], dtype=torch.int64)
```

## Actual behavior

```
```

## Steps to reproduce

```python
import torch
import flag_gems

flag_gems.enable()
inp = torch.empty(0, dtype=torch.int32, device=flag_gems.device)
torch.cumsum(inp, dim=0)
```

## Traceback

```
  File ".../flag_gems/ops/cumsum.py", line XXX, in cumsum
    return cumsum_wrapper(inp, dim, dtype)
  File ".../flag_gems/ops/cumsum.py", line XXX, in cumsum_wrapper
    K = inp.numel() // M // N
```

DannyP0:
### PR Category
Operator

### Type of Change
Bug Fix

### Description
Fix `cumsum_wrapper` crashing with `ZeroDivisionError` on empty input tensors.
When the reduction dimension has size 0, `N` is 0 and `K = inp.numel() // M // N` fails.

Return early after allocating the output tensor when `inp.numel() == 0`.

Add `test_cumsum_empty` to cover this case.

### Issue

Resolves #4543

### Progress

### Performance
N/A

### DIFF
--- a/src/flag_gems/ops/cumsum.py
+++ b/src/flag_gems/ops/cumsum.py
@@ -211,15 +211,8 @@ def scan_then_fan(inp, out, A, B, C, dtype):
 
 def cumsum_wrapper(inp, dim=1, dtype=None, out=None):
     assert dim >= -inp.ndim and dim < inp.ndim, "Invalid dim"
-    shape = inp.shape
     dim = dim % inp.ndim
-    M = 1
-    N = shape[dim]
-    for i in range(dim):
-        M *= shape[i]
     inp = inp.contiguous()
-    K = inp.numel() // M // N
-
     if dtype is None:
         dtype = inp.dtype
         if is_integer_dtype(dtype) or is_boolean_dtype(dtype):
@@ -227,6 +220,15 @@ def cumsum_wrapper(inp, dim=1, dtype=None, out=None):
     if out is None:
         out = torch.empty_like(inp, dtype=dtype)
 
+    if inp.numel() == 0:
+        return out
+
+    shape = inp.shape
+    M = 1
+    N = shape[dim]
+    for i in range(dim):
+        M *= shape[i]
+    K = inp.numel() // M // N
     compute_dtype = out.dtype
     if inp.dtype == torch.float16 or inp.dtype == torch.bfloat16:
         compute_dtype = torch.float32
--- a/tests/test_cumsum.py
+++ b/tests/test_cumsum.py
@@ -18,6 +18,20 @@ else:
 
 random.seed(time.time() // 100)
 
+# Backends that ship their own cumsum copy still carrying the empty-tensor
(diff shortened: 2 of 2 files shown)


zheng1:
This looks correct to me. Since the early return fires whenever `inp.numel() == 0`, it also covers the `M == 0` shapes, not just `N == 0` — e.g. `(0, 5)` with `dim=1` and `(3, 0, 4)` with `dim=2` both hit the same `// M // N` division. Might be worth parametrizing `test_cumsum_empty` over a few of those shapes (plus a float dtype) so that coverage is locked in.

FYI, one thing this class of fix doesn't reach: several backends ship their own `cumsum` copy (registered via `SpecOpRegistrar`) that still carries the same div-by-zero — I filed #4602 with the list. Nothing this PR needs to handle, just flagging so it's tracked.

DannyP0:
> This looks correct to me. Since the early return fires whenever `inp.numel() == 0`, it also covers the `M == 0` shapes, not just `N == 0` — e.g. `(0, 5)` with `dim=1` and `(3, 0, 4)` with `dim=2` both hit the same `// M // N` division. Might be worth parametrizing `test_cumsum_empty` over a few of those shapes (plus a float dtype) so that coverage is locked in.
> 
> FYI, one thing this class of fix doesn't reach: several backends ship their own `cumsum` copy (registered via `SpecOpRegistrar`) that still carries the same div-by-zero — I filed #4602 with the list. Nothing this PR needs to handle, just flagging so it's tracked.

Thanks @zheng1! Addressed in f21536b:

* test_cumsum_empty parametrized over (0,) dim=0, (0, 5) dim=1, (3, 0, 4) dim=2, plus a couple of extra empty shapes
* Both int32 and float32, with shape/dtype checked against torch.cumsum
* Vendor skips for the backend overrides tracked in #4602

(Early return in edc97388b.) Backend overrides left out of scope as discussed 

Verified locally: 
```pytest tests/test_cumsum.py --ref cpu``` — 50 passed. Ready for review.

bin913 (approved):

PR #4561, opened 2026-07-06
[KernelGen][Nvidia] Add narrow operator with Triton kernel

Yukun-Cui:
## Summary
Adds a Triton kernel for `narrow`. Returns a new tensor that is a narrowed version of the input tensor along the given dimension.

## Testing

## Performance
Test command: `pytest benchmark/test_narrow.py --level core` (NVIDIA H20)

### narrow

| float16 | [64, 64] | 0.003616 | 0.005728 | 0.631 |
| float16 | [256, 256] | 0.003552 | 0.006208 | 0.572 |
| float16 | [1024, 1024] | 0.003552 | 0.007072 | 0.502 |
| float16 | [4096, 4096] | 0.003520 | 0.016256 | 0.217 |
| float16 | [1024, 65536] | 0.003520 | 0.046208 | 0.076 |
| float32 | [64, 64] | 0.003520 | 0.006272 | 0.561 |
| float32 | [256, 256] | 0.003520 | 0.006272 | 0.561 |
| float32 | [1024, 1024] | 0.003520 | 0.007872 | 0.447 |
| float32 | [4096, 4096] | 0.003520 | 0.023680 | 0.149 |
| float32 | [1024, 65536] | 0.003520 | 0.076096 | 0.046 |
| bfloat16 | [64, 64] | 0.003520 | 0.005728 | 0.615 |
| bfloat16 | [256, 256] | 0.003520 | 0.005888 | 0.598 |
| bfloat16 | [1024, 1024] | 0.003520 | 0.006912 | 0.509 |
| bfloat16 | [4096, 4096] | 0.003520 | 0.016192 | 0.217 |
| bfloat16 | [1024, 65536] | 0.003520 | 0.045088 | 0.078 |

| narrow | **0.385** |

## Multi-backend Testing
|---|---|---|---|
| Nvidia (H20) | PASS (15 cases) | 0.385 | Primary |
| Tianshu | N/A | — | — |
| Muxi | N/A | — | — |
| Ascend | N/A | — | — |
| Hygon | N/A | — | — |

## Files Changed
- `src/flag_gems/ops/narrow.py`: Triton kernel implementation
- `tests/test_narrow.py`: Accuracy test
- `benchmark/test_narrow.py`: Performance benchmark
- `conf/operators.yaml`: Add operator entry (kind: Tensor, stage: alpha 5.1)


benchmark/test_narrow.py
```diff
@@ -0,0 +1,53 @@
+import pytest
+import torch
+
+import flag_gems
+from flag_gems.utils import shape_utils
+
+from . import base, consts, utils
+
+
+class TensorSelectBenchmark(base.GenericBenchmark2DOnly):
```
tengqm: Why this name?

benchmark/test_narrow.py
```diff
@@ -0,0 +1,53 @@
...
+    def set_more_metrics(self):
+        return ["gbps"]
+
+    def set_more_shapes(self):
+        # Speed Up Benchmark Test, Big Shape Will Cause Timeout
+        if flag_gems.vendor_name == "kunlunxin":
+            return []
```
tengqm: Why?

benchmark/test_narrow.py
```diff
@@ -0,0 +1,53 @@
...
+
+        shapes = super().set_more_shapes()
```
tengqm: Am I supposed to guess what are the shapes?
Why don't we just enumerate the shapes directly in this function?

benchmark/test_narrow.py
Yukun-Cui: Good catch — that was a leftover generic name from the template. Renamed it to `NarrowBenchmark` to match the operator, consistent with the naming used by other benchmarks. Fixed in f5f5b0c.

benchmark/test_narrow.py
Yukun-Cui: This was copied over from another benchmark and isn't relevant to this Nvidia-focused PR, so I removed it (along with the now-unused `flag_gems` import). Fixed in f5f5b0c.

benchmark/test_narrow.py
Yukun-Cui: Agreed — relying on the base shapes plus a filter made the actual coverage hard to see. I now enumerate the shapes directly: `[(10000, 256), (10000, 4096), (10000, 65536)]`. Fixed in f5f5b0c.

Caeruleann (commented):
`toch.narrow` is a view operation, so the output must share the underlying storage with the input. Implementing it by copying data into a new buffer with a Triton kernel would be incorrect.

Yukun-Cui:
@Caeruleann Good catch, you're right — `torch.narrow` is a view op and the output must alias the input's storage, so copying into a fresh buffer with a Triton kernel was incorrect.

I've reimplemented `narrow` as a **zero-copy view** using `torch.as_strided` (adjusting only the size along `dim` and the storage offset), following the existing `expand` op, which is likewise a pure layout operation. The kernel is gone.

Additional changes:
- `conf/operators.yaml`: marked as a pure layout op and added `skip_precision_check` (matching `expand`).
- `benchmark/test_narrow.py`: remodeled on `ExpandBenchmark`, dropped the `gbps` metric (meaningless for a metadata-only op).
- `tests/test_narrow.py`: added an assertion that the output **shares storage** with the input, locking in the view semantics.

Verified on GPU that the result matches `torch.narrow` and that in-place mutation of the view propagates back to the input (`x.narrow(...).add_(...)` now mutates `x`), including non-contiguous inputs. Fixed in 0e2fc338; CI is green.

src/flag_gems/ops/narrow.py
```diff
@@ -0,0 +1,54 @@
...
+    assert length >= 0 and start + length <= inp.size(
+        dim
+    ), f"Invalid narrow range: start={start}, length={length}, dim_size={inp.size(dim)}"
+
+    size = list(inp.shape)
+    size[dim] = length
+    stride = list(inp.stride())
+    storage_offset = inp.storage_offset() + start * inp.stride(dim)
+
+    return torch.as_strided(inp, size, stride, storage_offset)
```
bin913, line 54: FlagGems is a Triton library; please do not use Torch

huangyiqun: This `torch.as_strided` call doesn't seem to involve any GPU/NPU kernel, so it should be fine here.

bin913 (approved):

Issue #3023, opened 2026-05-10
The `conv3d_padding` operator doesn't work on Kunlunxin platform.


tengqm:
### Describe the bug

The accuracy test for the operator is thus currently skipped.

Issue #3022, opened 2026-05-10
The `conv1d_padding` operation fails on Kunlunxin platform.


tengqm:
### Describe the bug

The accuracy test is currently skipped.

PR #4566, opened 2026-07-06
[Kunlunxin] Support 'same'/'valid' string padding for conv3d

llaboon:
Add string padding mode handling ('same'/'valid') to the kunlunxin conv3d op. 'same' computes asymmetric per-side padding from the kernel and dilation and pre-pads the input via F.pad; 'valid' maps to zero padding. Re-enable the conv1d_padding (#3022) and conv3d_padding (#3023) tests for kunlunxin now that string padding is supported.

### Progress


bin913:
@llaboon Please fix failing check

PR #4572, opened 2026-07-06
Tune Triton-TLE topk paths

zheng1:
### PR Category
Operator, OP Test, Benchmark

### Type of Change
Performance Optimization

### Description
- Tune the generic `topk` Triton-TLE radix launch config for large fp32 `k=256` rows by using a wider radix pass and 1024-column tiles.
- Prefer radix final selection in `top_k_per_row_prefill` for large full-vocabulary rows (`vocab_size >= 65536`).
- Add focused config tests and make the `top_k_per_row_prefill` benchmark runnable without vLLM by using the existing FlagGems non-TLE fallback as the baseline.

### Issue
N/A

### Progress

### Performance
A100-PCIE-40GB local results:

- Direct generic `topk` TLE kernel comparison for `fp32, N=32768, k=256` showed the tuned config faster than the previous default on representative batch sizes (`64`: 1.36x, `128`: 1.25x, `256`: 1.34x).
- `benchmark/test_topk.py -m topk`: current `fp32` row `x=[128, 32768], k=256` reports Gems latency `0.242688 ms`.
- `benchmark/test_top_k_per_row_prefill.py -m top_k_per_row_prefill`: new full-vocab shape `[64, 129280], k=1024` reports non-TLE baseline `0.353280 ms`, TLE `0.130048 ms`, speedup `2.717x`.
- Directed prefill comparison for `[2048, 129280], k=1024` showed radix-final path `1.881600 ms` vs sort-final path `2.044928 ms` (`1.079x`).

Validation run:

```bash
python -m isort --check-only --profile black src/flag_gems/ops/topk.py src/flag_gems/fused/top_k_per_row_prefill.py tests/test_topk.py tests/test_top_k_per_row_prefill.py benchmark/test_top_k_per_row_prefill.py
python -m black --check --target-version py312 src/flag_gems/ops/topk.py src/flag_gems/fused/top_k_per_row_prefill.py tests/test_topk.py tests/test_top_k_per_row_prefill.py benchmark/test_top_k_per_row_prefill.py
pytest tests/test_topk.py -m topk -q --tb=short
pytest tests/test_top_k_per_row_prefill.py -m top_k_per_row_prefill -q --tb=short
pytest benchmark/test_topk.py -m topk -q --tb=short
pytest benchmark/test_top_k_per_row_prefill.py -m top_k_per_row_prefill -q --tb=short
```

### DIFF
--- a/benchmark/test_top_k_per_row_prefill.py
+++ b/benchmark/test_top_k_per_row_prefill.py
@@ -3,14 +3,17 @@
 Shapes match DeepSeek V4 production config:
     - vocab_size=129280: DeepSeek V4 vocabulary size (full BPE vocab)
     - top_k=1024: number of KV cache slots selected per token in sparse attention
-    - num_rows=1: single-token decode (latency-critical path)
+    - num_rows=1: single-token prefill/decode-like latency-sensitive path
     - num_rows=32: typical prefill micro-batch
     - num_rows=64: larger prefill batch
     - num_rows=2048: max prefill sequence length
 
 The baseline uses vLLM's persistent_topk CUDA kernel when available,
-falling back to torch.topk for the selection-only theoretical minimum.
+falling back to FlagGems' non-TLE implementation so the benchmark can run
+in plain Triton-TLE development environments.
 """
+
+from importlib import import_module
 
 import pytest
 import torch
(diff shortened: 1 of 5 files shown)


huangyiqun (commented):
please fix code-style

zheng1:
Thanks! The code-style issue has been fixed in the latest head -- `pre-commit run --all-files` now passes clean. The remaining CI just needs a maintainer approval to re-run. Please take another look when you have a chance.

huangyiqun (commented):
Could we include tuning results for a few more NVIDIA architectures, or move these NVIDIA-specific launch configurations into a YAML file, similar to [`fused_moe_config.yaml`,](https://github.com/flagos-ai/FlagGems/blob/master/src/flag_gems/utils/configs/fused_moe_config.yaml) so they’re easier to maintain and specialize per architecture?

zheng1:
I moved the tuned radix-TLE launch config out of the inline conditional into a small data table `_TOPK_RADIX_TLE_TUNED_CONFIGS` keyed by `(dtype, k)`, with the base heuristic as the fallback. Adding a newly tuned shape is now a one-row edit, and the selected `(BLOCK_N, RADIX_BITS, num_warps)` is unchanged for every input (checked exhaustively across dtype/k/elem).

Two notes on the specifics you raised:

- I kept it as an in-module table rather than a `fused_moe_config.yaml`-style file for now, since there's a single measured row and a YAML plus loader felt heavier than the payoff. If you'd rather it live in the per-vendor `tune_configs.yaml` mechanism so it dispatches per architecture, I'm happy to move it there as a follow-up.
- On more architectures: I only have an A100 to tune on, so I can't add the H-series or others right now, but the table makes that a one-row addition once someone can measure them, and untuned architectures fall back to the heuristic.

huangyiqun (approved):
LGTM.

PR #4604, opened 2026-07-08
Fix dispatch keys registered under non-existent ATen overloads

zheng1:
## Summary

Three operators were registered in `_FULL_CONFIG` under overload names that do not exist in the ATen schema. `torch.library` accepts such registrations silently but never dispatches to them, so the operators fell back to eager and were **not accelerated by FlagGems**:

| Registered (broken) | Correct key | Reason |
|---|---|---|
| `greater.out` | `greater.Tensor_out` | `greater_out` handles Tensor x Tensor; `aten.greater` has no `out` overload (only `Tensor`/`Scalar`/`Tensor_out`/`Scalar_out`). `greater.Scalar_out` is already registered separately. |
| `geometric.float` | `geometric` | `aten.geometric` has only `default`/`out`. |
| `geometric_.float` | `geometric_` | `aten.geometric_` has only `default`. |

This is the same class of bug fixed in #4522.

## Verification (NVIDIA A100)

Confirmed via dispatch logging that before the fix these calls silently fell back to eager (no `GEMS ...` log), and after the fix they dispatch to the FlagGems kernel:

- `torch.greater(a, b, out=out)` -> `GEMS GREATER_OUT` (result matches `a > b`)
- `x.geometric_(p)` -> `GEMS GEOMETRIC_` (result valid)

Existing accuracy tests did not catch this because a silent eager fallback still returns correct results.

## Testing

- Added `tests/test_dispatch_keys.py` asserting the corrected keys are registered. It **fails on the pre-fix keys and passes after** the fix.
- `pytest tests/test_greater.py tests/test_geometric.py -q` -> 93 passed.
- `pre-commit run --files ...` clean.


src/flag_gems/__init__.py
```diff
@@ -372,14 +371,14 @@ def torch_ge(v):
     ("gelu", gelu),
     ("gelu_", gelu_),
     ("gelu_backward", gelu_backward),
-    ("geometric.float", geometric),
-    ("geometric_.float", geometric_),
+    ("geometric", geometric),
+    ("geometric_", geometric_),
```
tengqm, line 381: pls help check the values of `for` in the `conf/operators.yaml` for these operators.

tests/test_dispatch_keys.py
```diff
@@ -0,0 +1,24 @@
+import pytest
```
tengqm: This smells like an over-engineering. We don't change dispatch keys accidentally, so there is virtually nothing to guide against.

On the other hand, such a test is a good candidate for CI job. Maybe we can add a generic job tool under `tools` directory that checks that all dispatch keys in `src/flag_gems/__init__.py` are valid.

zheng1:
Thanks for the review, @tengqm.

**1. operators.yaml `for:` fields.** Fixed in `713f81a`. The inventory carried the same stale overload names, now updated to match:
- `geometric.float` -> `geometric`
- `geometric_.float` -> `geometric_`
- `greater.out` -> `greater.Tensor_out`

**2. The test.** You're right that guarding three specific keys is thin, so I went with your generic-checker idea. Replaced the pytest with `tools/check_dispatch_keys.py` (`e3335da`). It checks every key in `_FULL_CONFIG` against the ATen schema: a dotted key has to name a real overload (that's what catches `greater.out`), a bare key is fine once the base op exists, and the custom ops plus the deliberate forward-compat keys like `addmm.dtype` are allowlisted.

`python tools/check_dispatch_keys.py --check` exits 1 on any invalid key. Right now all 668 pass; revert the fix in this PR and it fails pointing at the exact key.

I didn't wire it into CI because it imports torch to read `op.overloads()`, so it can't live in the code-style job. Tell me where you want it (a step in unit-test, or its own job) and I'll add it.

tools/check_dispatch_keys.py
```diff
@@ -0,0 +1,142 @@
...
+    "nll_loss_nd_backward",
+    "nll_loss_nd_forward",
+    "normed_cumsum",
+    "scaled_softmax_backward",
+    "scaled_softmax_forward",
+    "special_chebyshev_polynomial_w_out",
+    "special_i0e_out",
+    "special_i1_out",
+    "special_shifted_chebyshev_polynomial_u_",
+}
```
tengqm, line 66: This list reveals the discrepancies we have in the operator registry.
We need to kill this list eventually. I mean, there should be no exceptions.
For example, the recently added `matmuladd` operator is actually a *fused* operator. You can check if an operator carries the `fused` label in the `conf/operators.yaml` file.
For another example, if an operator was invented out of thin air, and we don't have proofs that they are used by any existing frameworks/models, maybe we should delete it or move it to the `fused` category.

Make senses?

tengqm (commented):
Thank you for coming up with this consistency checking tool.

zheng1:
Good point. I switched the checker to read the `fused` label from `conf/operators.yaml`, so `matmuladd`, `beam_search_score`, and `beam_search_score_` are exempted automatically now instead of by name (`2e7d69c`).

What is left in the hardcoded set is the real registry dirt: 16 entries registered under a base that is neither a real `torch.ops.aten` op nor labeled `fused`. Two kinds:
- custom composites mislabeled `aten` (`scaled_softmax_forward`/`backward`, `normed_cumsum`, `nll_loss_nd_forward`/`backward`, `add_rms_norm`, ...) that should probably be `fused`, or removed if nothing uses them;
- non-standard names for a real ATen op (`special_i0e_out` for `special_i0e.out`, `grid_sample` for `grid_sampler`, `max_pool3d_backward` for `max_pool3d_with_indices_backward`, ...) that should be renamed to the correct overload.

Cleaning each of these touches its operator and needs case-by-case checking, so I would rather not fold it into this PR. I renamed the set to `KNOWN_REGISTRY_DISCREPANCIES` with a comment that it should shrink to empty. Want me to open a tracking issue with the full list so we can burn it down separately?

tengqm (approved):

zheng1:
Thanks for the review. Addressed the first two, and a question on the third.

1. `for` values in `operators.yaml`: checked and synced with the registration keys. `geometric.float` -> `geometric`, `geometric_.float` -> `geometric_`, and `greater.out` -> `greater.Tensor_out`, so the yaml `for` field now matches what `__init__.py` registers.

2. Over-engineered test: agreed, dropped `tests/test_dispatch_keys.py`. The logic now lives only in `tools/check_dispatch_keys.py` with a `--check` mode that exits non-zero, so it can run as a CI job instead of a unit test.

3. Killing `KNOWN_REGISTRY_DISCREPANCIES`: the tool already reads the `fused` label and auto-exempts those ops via `load_fused_ops()`, so `matmuladd` and other fused ops never land in the list. What remains are ops still labeled `aten` in the yaml (`add_rms_norm`, `normed_cumsum`, `scaled_softmax_forward/backward`, and the `special_*` ones). Relabeling each as `fused` or deleting an invented-from-thin-air op is a per-op call about whether a framework actually uses it, so I would rather not fold that into this PR. Happy to open a follow-up that walks the list one op at a time. Does that split work for you?

zheng1:
Hi @tengqm, this has been approved and green for a few days. Anything blocking the merge?

tengqm:
@zheng1 Nothing left to do from your side. Just be a little patient for other eyes to screen this again and merge it. I'd suggest you shift to other open issues if possible.

PR #4609, opened 2026-07-08
feat: record test start/end/total time in run_tests.py summary and log

wuwentao:
### PR Category

CI/CD

### Type of Change

New Feature

### Description

`tools/run_tests.py` runs each operator's accuracy test (`tests/`) and benchmark
(`benchmark/`) and writes a combined `results/summary.json`. This PR adds wall-clock
timing to that run:

- Captures the **start time** before environment probing and prints it via `pinfo`.
- Captures the **end time** at the end of the run, computes the **total_duration time**,
  and prints it just before `Test completed.`
- Writes `total_duration` (H:MM:SS)  into `summary.json`.

#### Why we need it

Full accuracy + benchmark sweeps over the operator inventory can run for a long time across
many GPUs, but the runner previously printed no start marker and no total duration — only a
bare completion timestamp buried in the JSON. There was no easy way to answer "when did this
run start?" and "how long did the whole run take?", which matters for CI dashboards,
run-time regression tracking, and capacity planning.

#### Advantages

- **Run duration at a glance** — start / end / total are printed to the console and
  persisted in `summary.json` for every run.
- **Clearer schema** — `end_time` is self-documenting where `timestamp` was ambiguous.
- **No behavior change to tests** — timing only; accuracy/benchmark execution is untouched.

### Issue

N/A — not derived from a specific issue.

### Progress

### Performance

N/A — this is test-runner / reporting tooling; it measures timing but does not affect
operator or runtime performance.


tools/psum_html
```diff
@@ -192,7 +192,11 @@ def _render_log_tabs(summary_text, log_files, uid):
 
 
 def parse_env():
-    timestamp = DATA.get("timestamp", "?")
```
tengqm, line 195: `timestamp` is an indicator when the summary data was generated.
It is meant to be used across many JSON files for historical data tracking.
Don't change it.

tools/psum_html
```diff
@@ -192,7 +192,11 @@ def _render_log_tabs(summary_text, log_files, uid):
...
+    start_time = DATA.get("start_time", "?")
+    # `end_time` replaces the former `timestamp`; fall back to it for
+    # summaries produced before the rename.
+    end_time = DATA.get("end_time", DATA.get("timestamp", "?"))
+    elapsed = DATA.get("elapsed", "?")
```
tengqm: It is probably okay to add a field that indicates how many seconds the whole test consumed, for example, `total_duration`.

With `timestamp` and `total_duration`, you can easily deduce when the test was started. Do not include redundant fields in the resulting JSON. We may want to add more fields later.

tools/psum_html
tengqm: Also, there are downstream tools already consuming this field (not checked into the repo). We have to maintain data compatibility when possible.

wuwentao: thanks for your review comments.
in current `run_test.py`, just rename this `timestamp` args to new name with `end_time`, as this is the same value, then rename this arg name in `psum_text/html/web`, also confirmed with @tianxiao-baai , `op stat web` can match with this changes.

if there is more downstream tool, we should confirm `timestap` name rule:
- option 1.  remove `end_time`  and use keep old `timestap` ,
- option 2. use both name at the same time.

in addition, this changes only apply to the  value in `summary.json`
please help to confirm the result, then I can update the code with your review comments, thanks

tools/psum_html
wuwentao: yes, rename `elapsed` to `total_duration`, and recovery  `end_time` to `timestamp` arg name ?

tools/psum_html
tengqm: - Keep the `timestamp` field if it is not breaking anything;
- Avoid redundant fields which can be easily deduced by downstream tools. That means adding a `total_duration` would be enough.

bin913 (approved):

tengqm (approved):

tianxiao-baai (approved):

PR #4616, opened 2026-07-08
Optimize amin single-dim reduction with inner/non-inner kernels

zheng1:
### PR Category

ops

### Type of Change

Performance Optimization

### Description

Optimize `torch.amin(..., dim=...)` for a single reduction dimension by adding native inner and non-inner Triton kernels, the same split already used by `amax`, `sum`, and `std`.

The old dim path always called `dim_compress`, which permutes the reduced dimension to the innermost position and copies the whole tensor before reducing. For a non-inner reduction (a middle or outer dim) that copy reads and writes the entire tensor once just to reorder it, and it was the dominant cost.

Global reduction and multi-dim reduction keep their existing paths. The single-dim path also guards empty tensors (return a valid empty output when a spectator dim is empty, raise for a zero-size reduction dim) and validates the dim range, matching torch. The in-place `amin_` inherits the speedup since its dim path delegates to `amin(..., keepdim=True)`.

### Performance

| (4096, 4096) | 0 | 0.28x | 0.97x |
| (8192, 2048) | 0 | 0.20x | 1.01x |
| (1024, 16384) | 0 | 0.16x | 0.83x |
| (4096, 4096) | 1 | 0.54x | 0.87x |

Non-inner reductions go from roughly 4-6x slower than eager to near parity. The inner path improves too, and global / multi-dim are untouched.

### Testing

- `pytest tests/test_amin.py`: 24 passed.
- Checked results against `torch.amin` across float32/float16/bfloat16, several 1D/2D/3D shapes, every single dim, multi-dim, and both keepdim settings: all match. Empty-tensor and out-of-range dim cases match torch (valid empty output or IndexError). All `+inf` rows and the `amin_` in-place variant verified.
- `pre-commit run --files src/flag_gems/ops/amin.py`: clean.

### Progress


src/flag_gems/ops/amin.py
```diff
@@ -130,30 +212,65 @@ def amin(inp, dim=None, keepdim=False):
...
+    out = torch.empty(shape, dtype=dtype, device=inp.device)
+    grid = lambda meta: (triton.cdiv(M, meta["BLOCK_M"]),)
+    with torch_device_fn.device(inp.device):
+        amin_kernel[grid](inp, out, M, N)
+    if not keepdim:
+        out = out.squeeze(dim=dim)
+    return out
+
 
 def amin_(inp, dim=None, keepdim=False):
```
huangyiqun, line 275: Are the optimizations for the two kernels specifically targeting the `amin` interface? Can the `amin_` interface also benefit from these optimizations?

huangyiqun (commented):
Do the accuracy tests cover these two newly added kernels?

zheng1:
Good questions.

1. Coverage: the existing tests hit the non-inner kernel on their `dim=0` shapes, but not the multi-tile path where the reduced dim spans more than one tile. Added `test_amin_dim_multi_tile` (`062ad8c`) with N=4096 and K=4096 shapes to cover both reduction loops.

2. Yes, `amin_` benefits automatically. Its dim path calls `amin(..., keepdim=True)`, so it goes through the same new kernels. I confirmed the in-place call dispatches to the non-inner kernel.

zheng1:
Hi @huangyiqun, on your question: `amin_` with a dim goes through `amin(...)`, so it already uses the new kernels. This has been green for a few days, anything else I should do?

huangyiqun:
> Hi @huangyiqun, on your question: `amin_` with a dim goes through `amin(...)`, so it already uses the new kernels. This has been green for a few days, anything else I should do?

Thanks for the clarification. I have approved the workflow.

huangyiqun:
There is no benchmark tests marked "amin", could you add your benchmark testcases?

zheng1:
Added `test_amin_dim` (marked `amin`) following the `test_sum_dim_out` pattern — it benchmarks `torch.amin(inp, dim=...)`, which is the path this PR optimizes (the existing `test_amin` only covered the full reduction). `/test | amin` now collects both, verified on the CI image (torch 2.11 / triton 3.6). I also added the matching dim benchmarks to my other single-dim optimization PRs (amax / var / std / var_mean) so they're covered the same way.

huangyiqun (commented):
The added cases cover both inner and non-inner multi-tile paths, and `amin_` picks up the same path through `amin`. Looks good.

src/flag_gems/ops/amin.py
```diff
@@ -62,6 +63,87 @@ def amin_kernel_2(mid, out, mid_size, BLOCK_MID: tl.constexpr):
     tl.store(out, amin_val)
 
 
+@libentry()
+@triton.heuristics(runtime.get_heuristic_config("softmax_non_inner"))
```
huangyiqun, line 67: Cambricon's softmax_non_inner heuristic only supplies TILE_MODE, but this kernel requires TILE_N, TILE_K, and ONE_TILE_PER_CTA; there is no Cambricon amin override. This path won't compile there. Please add an amin-specific heuristic or keep that backend on the old path.

src/flag_gems/ops/amin.py
```diff
@@ -130,30 +212,65 @@ def amin(inp, dim=None, keepdim=False):
...
-        dim = [d % inp.ndim for d in dim]
-        inp = dim_compress(inp, dim)
-        N = 1
-        for i in dim:
-            N *= shape[i]
-            shape[i] = 1
-        M = inp.numel() // N
+    if isinstance(dim, int):
+        dim = [dim]
+    if not all(-inp.ndim <= i < inp.ndim for i in dim):
```
huangyiqun, line 218: PyTorch accepts dim=0 and dim=-1 for a 0-D tensor, but this rejects both. Please preserve scalar behavior and add a regression case.

huangyiqun (commented):
This PR changes tests/test_amin.py, and current master checks the whole changed test file. The existing use_gems calls will fail check-kernelgen-tests after rebase; please switch them to the public API.

huangyiqun (commented):
Follow-up: the Cambricon, scalar-dim, and test-rule issues above are blockers. Please treat my earlier path-only note as superseded.

PR #4618, opened 2026-07-08
Optimize prod single-dim reduction with inner/non-inner kernels

zheng1:
### PR Category

ops

### Type of Change

Performance Optimization

### Description

Optimize `torch.prod(..., dim=...)` for a single reduction dimension by adding native inner and non-inner Triton kernels, the same split already used by `amax`, `amin`, `sum`, and `std`.

The old dim path always called `dim_compress`, which permutes the reduced dimension to the innermost position and copies the whole tensor before reducing. For a non-inner reduction (a middle or outer dim) that copy reads and writes the entire tensor once just to reorder it, and it was the dominant cost.

Both kernels multiply into a float32 accumulator and reduce with `tl.reduce(combine_fn=reduce_mul)`, matching the existing kernel. Unlike max and min, a product over an empty dimension has an identity, so an empty reduction dim returns 1 rather than raising, and an empty spectator dim returns an empty tensor, both matching torch.

### Performance

| (4096, 4096) | 0 | 0.22x | 0.86x |
| (8192, 2048) | 0 | 0.19x | 0.85x |
| (1024, 16384) | 0 | 0.15x | 0.67x |
| (4096, 4096) | 1 | 0.54x | 0.83x |

Non-inner reductions go from roughly 4-6x slower than eager to near parity. The inner path improves too.

### Testing

- `pytest tests/test_prod.py`: 53 passed (45 existing plus 8 new multi-tile cases).
- Added `test_prod_dim_int_large` covering the multi-N-tile and multi-K-tile paths (N=4096, K=4096), which the small-shape tests did not reach.
- Checked results against `torch.prod` across float32/float16/bfloat16 and int64, several 1D/2D/3D shapes, every single dim, and both keepdim settings. Empty-tensor behavior matches torch: identity 1 for an empty reduction dim, empty output for an empty spectator dim; out-of-range dim raises IndexError. nan/inf/zero propagate as in torch.
- `pre-commit run --files ...`: clean.

### Progress

### DIFF
--- a/tests/test_prod.py
+++ b/tests/test_prod.py
@@ -53,3 +53,26 @@ def test_prod_dim_int(shape, dim, keepdim, dtype):
         res_out = torch.prod(inp, dim=dim, keepdim=keepdim)
 
     utils.gems_assert_close(res_out, ref_out, dtype)
+
+
+@pytest.mark.prod_dim_int
+@pytest.mark.parametrize(
+    "shape, dim",
+    [
+        ((4, 8, 4096), 1),  # non-inner: K = 4096 spans multiple K tiles
+        ((4, 4096, 8), 1),  # non-inner: N = 4096 exercises the reduction loop
+        ((8, 4096), 1),  # inner: N = 4096 exercises the reduction loop
+        ((4096, 8), 0),  # non-inner via the outer dim
+    ],
+)
+@pytest.mark.parametrize("keepdim", [False, True])
+def test_prod_dim_int_large(shape, dim, keepdim):
+    # Values near 1 keep the product finite over large reduction sizes.
+    inp = torch.rand(shape, dtype=torch.float32, device=flag_gems.device) * 0.4 + 0.8
+    ref_inp = utils.to_reference(inp, True)
+
+    ref_out = torch.prod(ref_inp, dim=dim, keepdim=keepdim)
+    with flag_gems.use_gems():
+        res_out = torch.prod(inp, dim=dim, keepdim=keepdim)
+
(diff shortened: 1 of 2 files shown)


tests/test_prod.py
```diff
@@ -53,3 +53,26 @@ def test_prod_dim_int(shape, dim, keepdim, dtype):
         res_out = torch.prod(inp, dim=dim, keepdim=keepdim)
 
     utils.gems_assert_close(res_out, ref_out, dtype)
+
+
+@pytest.mark.prod_dim_int
```
huangyiqun: The test marks should match the test interface names. It appears that the `prod_dim_int` interface no longer exists. We will handle the remaining tests marked with `prod_dim_int` later.

zheng1:
Thanks, done in `c2b2906`. Switched the new test to the `prod` mark and renamed it to `test_prod_dim_multi_tile`. I left the existing `prod_dim_int`-marked tests alone since you said you'll retag those separately.

huangyiqun:
> Thanks, done in `c2b2906`. Switched the new test to the `prod` mark and renamed it to `test_prod_dim_multi_tile`. I left the existing `prod_dim_int`-marked tests alone since you said you'll retag those separately.

GET

huangyiqun (approved):
LGTM

zheng1:
Hi @huangyiqun, this one is approved and CI is green. Anything else you need from me, or is it good to merge?

huangyiqun:
> Hi @huangyiqun, this one is approved and CI is green. Anything else you need from me, or is it good to merge?

Thanks for your contribution!

PR #4644, opened 2026-07-09
Optimize var single-dim reduction with inner/non-inner kernels

zheng1:
### PR Category

ops

### Type of Change

Performance Optimization

### Description

Optimize `torch.var(..., dim=...)` for a single reduction dimension by adding native inner and non-inner Triton kernels, the same split already used by `amax`, `amin`, `sum`, `prod`.

The old single-dim path always called `dim_compress`, which permutes the reduced dimension to the innermost position and copies the whole tensor before running the welford kernel. For a non-inner reduction (a middle or outer dim) that copy reads and writes the entire tensor once just to reorder it, and it was the dominant cost.

Each kernel reduces in a single read of the input with a per-lane streaming Welford: every lane keeps a running `(count, mean, M2)` with no cross-lane reduction inside the loop, and the lanes are merged once at the end with a numerically stable Chan combine. An earlier revision of this PR used a two-pass mean/variance kernel that read the reduced dim twice; on large tensors that doubled global-memory traffic and left the reduction ~2x slower than eager (e.g. `1024^3` dim=1 fp32 at 0.52x). The single-read Welford removes that second pass while keeping the two-pass numerics.

Global reduction and multi-dim reduction keep the existing welford paths. NaN for `N - correction <= 0` and empty-tensor handling match torch.

### Performance

A100, `torch/gems` latency ratio (higher is better, 1.0 = parity with eager), measured by interleaving torch and gems so the ratio is robust to machine load. `two-pass` is the earlier revision of this PR; `single-read` is the current kernel:

| shape | dim | dtype | two-pass | single-read |
|-------|-----|-------|---------:|------------:|
| (1024, 1024, 1024) | 1 | fp16 | 0.52x | 0.88x |
| (1024, 1024, 1024) | 1 | fp32 | 0.60x | 0.72x |
| (64, 512, 512) | 1 | fp32 | 0.57x | 0.64x |
| (4096, 4096) | 0 | fp16 | 0.71x | 1.05x |
| (4096, 4096) | 0 | fp32 | 0.84x | 0.94x |
| (4096, 4096) | 1 | fp16 | 1.08x | 1.07x |
| (8192, 8192) | 1 | fp16 | 1.72x | 1.72x |

Both revisions already beat the old `dim_compress` copy path (which was ~3-4x slower than eager on non-inner reductions). The single-read kernel additionally removes the large-tensor two-pass penalty, and inner reductions are unchanged.

### Testing

- `pytest tests/test_var.py`: 376 passed (default CUDA), 20 passed (`--ref=cpu --quick`).
- `test_accuracy_var_dim_multi_tile` covers the multi-N-tile and multi-K-tile paths (N=4096, K=4096), which the small-shape tests did not reach.
- Checked against `torch.var` across float32/float16/bfloat16, 1D/2D/3D shapes, every single dim, both keepdim settings, and correction in {0, 1, 2}. Edge cases match torch: NaN when `N - correction <= 0` (including an empty reduction dim with non-empty output), empty output for an empty spectator dim, non-negative variance for constant inputs.
- Numerics verified on hostile inputs for the single-read kernel: large data mean (shift up to 1e5), outlier first element, tiny variance with N up to 2^25, all matching torch.
- `pre-commit` and `flake8` clean.

### Progress

### DIFF
--- a/src/flag_gems/ops/var.py
+++ b/src/flag_gems/ops/var.py
@@ -179,7 +179,7 @@ def _var_dim_kernel_inner(
         sq_sum = tl.sum(sq_acc, axis=0)
 
     denom = N - correction
-    var = sq_sum / tl.maximum(denom, 1e-12)
+    var = sq_sum / tl.maximum(denom, 0.0)
     tl.store(Out + pid_m, var.to(Out.dtype.element_ty), mask=pid_m < M)
 
 
@@ -238,7 +238,7 @@ def _var_dim_kernel_non_inner(
         sq_sum = tl.sum(sq_acc, axis=0, keep_dims=True)
 
     denom = N - correction
-    var = sq_sum / tl.maximum(denom, 1e-12)
+    var = sq_sum / tl.maximum(denom, 0.0)
     out_offsets = pid_m * K + k_offsets
     tl.store(Out + out_offsets, var.to(Out.dtype.element_ty), mask=k_offsets < K)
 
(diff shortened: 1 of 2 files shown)


zheng1:
Pushed a correctness fix to the correction handling. While working on the sibling var_mean optimization I noticed torch clamps the variance divisor at 0, so `correction >= N` (with `N > 0`) returns `+inf`, not NaN. The previous version filled NaN for every `N - correction <= 0`, which was wrong for the `N > 0` case.

Now the kernel clamps the divisor at 0 (matching torch), so `denom <= 0` gives `+inf`, and only an empty reduction dim (`N == 0`, a genuine 0/0) is filled with NaN. Verified against torch across correction in {0,1,2}, correction >= N (inf), and N == 0 (nan); test_var still passes 376.

PR #4655, opened 2026-07-09
Optimize min single-dim reduction with inner/non-inner kernels

zheng1:
### PR Category

ops

### Type of Change

Performance Optimization

### Description

Optimize `torch.min(x, dim=d)` for a single reduction dimension by adding native inner and non-inner Triton kernels, the same split already merged for `sum`, `mean`, `std`, and the amax/amin work. Unlike those, `min` returns both values and indices, so the kernels produce both in one pass.

The old single-dim path called `dim_compress`, which permutes the reduced dimension to the innermost position and copies the whole tensor before reducing. For a non-inner reduction (a middle or outer dim) that copy reads and writes the entire tensor once just to reorder it, and it dominated the runtime.

Both the min value and its index come from `tl.min(..., return_indices=True, return_indices_tie_break_left=True)`, and ties resolve to the lowest index to match torch. Across N tiles the update uses a strict comparison so the earlier index wins on ties.

A note on the heuristics: this reuses the `softmax_inner`/`softmax_non_inner` tile configs (as amax/amin do) rather than `argmax_*`, because those give power-of-2 tile sizes for every N.

### Performance

| (4096, 4096) | 0 | 0.40x | 0.59x |
| (8192, 2048) | 0 | 0.47x | 0.94x |
| (8, 512, 512) | 1 | 0.34x | 0.60x |
| (4, 4096, 8) | 1 | 0.13x | 0.24x |
| (4096, 4096) | 1 | 0.40x | 0.43x |

Every shape improves. The non-inner reductions that were 3-8x slower than eager move toward parity.

### Testing

- `pytest tests/test_min.py`: all pass, including a new `test_min_dim_multi_tile` covering the multi-N-tile and multi-K-tile paths (N=4096, K=4096) that the small default shapes do not reach.
- Checked both values and indices against `torch.min` across float16/bfloat16/float32/int32, 2D and 3D shapes, every single dim, and both keepdim settings. Tie-breaking matches torch (lowest index on equal values), negative dims work, and empty-dim behavior matches torch (a zero-size reduction dim raises, an empty spectator dim returns an empty result).
- `flake8` clean.

### Progress

### DIFF
--- a/src/flag_gems/ops/min.py
+++ b/src/flag_gems/ops/min.py
@@ -8,7 +8,7 @@ import triton.language as tl
 
 from flag_gems import runtime
 from flag_gems.runtime import torch_device_fn
-from flag_gems.utils import dim_compress, libentry, libtuner
+from flag_gems.utils import libentry, libtuner
 from flag_gems.utils import triton_lang_extension as ext
 from flag_gems.utils.limits import get_dtype_max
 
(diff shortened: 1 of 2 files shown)


zheng1:
Folding this into #4654. max and min are exact mirrors, so I combined them into one PR to keep the review in one place. The min kernels and tests are now part of #4654.

PR #4658, opened 2026-07-10
Fix argmax crash on non-power-of-2 reduction dim <= 128 (non-inner)

### LINKED ISSUE #4657: argmax(dim) crashes on non-power-of-2 reduction dim <= 128 (non-inner)
### Describe the bug

`torch.argmax(x, dim=d)` crashes with `ValueError: arange's range must be a power of 2` when the reduced dimension `N` is `<= 128` and not a power of 2, on a non-inner reduction (the reduced dim is a middle or outer dim, so `K > 1`).

### Reproduce

```python
import torch
import flag_gems

x = torch.randn(37, 101, 53, device="cuda")
with flag_gems.use_gems():
    torch.argmax(x, dim=1)   # N=101, K=53 -> CompilationError
```

Fails for N in {101, 100, 65, 127, ...} (any non-power-of-2 <= 128 with K > 1). Works when N is a power of 2 (64, 128) or when K == 1 (inner reduction).

```
triton.compiler.errors.CompilationError: at 25:19:
    ...
    n_offset = tl.arange(0, TILE_N)
               ^
ValueError('arange's range must be a power of 2')
```

### Root cause

`argmax_heur_tile_n_non_inner` in `src/flag_gems/runtime/backend/_nvidia/heuristics_config_utils.py` returns `N` directly as `TILE_N` when `N <= 128`:

```python
def argmax_heur_tile_n_non_inner(args):
    n = args["N"]
    ...
    if n <= 128:
        return n   # <-- non power of 2 when N is e.g. 101
```

`argmax_kernel_non_inner` then builds `tl.arange(0, TILE_N)`, which requires a power-of-2 length.

The kernel already masks padding lanes (`n_offset < N`, `other=min_value`), so rounding `TILE_N` up to the next power of 2 is safe and does not change the result.

`argmin` is not affected: it uses a different `BLOCK_M`/`BLOCK_K` kernel that does not build `tl.arange` over the reduced dim.

### Environment

- FlagGems: current master
- GPU: NVIDIA A100
- torch + triton as pinned in the repo

### Fix

Round up to the next power of 2 in that branch. PR incoming.

zheng1:
### PR Category

ops

### Type of Change

Bug Fix

### Description

Fixes #4657.

`torch.argmax(x, dim=d)` crashed with `ValueError: arange's range must be a power of 2` when the reduced dimension `N` was `<= 128` and not a power of 2, on a non-inner reduction (the reduced dim is a middle or outer dim, so `K > 1`).

`argmax_heur_tile_n_non_inner` returned `N` directly as `TILE_N` in the `n <= 128` branch, but `argmax_kernel_non_inner` builds `tl.arange(0, TILE_N)`, which requires a power-of-2 length.

The fix rounds `TILE_N` up to the next power of 2 in that branch. The kernel already masks the padding lanes (`n_offset < N`, `other=min_value`), so an oversized `TILE_N` is safe and does not change the result: padding lanes sit at indices above every real element and carry the max-identity, so they can never become the argmax or shift the returned index.

`argmin` is not affected. It uses a different `BLOCK_M` / `BLOCK_K` kernel that does not build `tl.arange` over the reduced dim, and it does not crash on the same shapes.

### Reproduce (before this PR)

```python
import torch
import flag_gems

x = torch.randn(37, 101, 53, device="cuda")
with flag_gems.use_gems():
    torch.argmax(x, dim=1)   # N=101, K=53 -> CompilationError
```

### Testing

- Added `test_argmax_non_power_of_2_non_inner` covering the previously-crashing path: non-inner (`K > 1`) reductions with non-power-of-2 `N <= 128`, including the tightest padding ratios (N=2, 3), a 4D tensor, and both keepdim settings. These shapes crash on master and pass with this fix, and the test asserts index equality against `torch.argmax`, not just absence of a crash.
- `pytest tests/test_argmax.py`: all pass (126 existing plus the new cases).
- `flake8` clean.

### Progress


huangyiqun (commented):
The heuristic fix looks good. Current master now rejects use_gems() in changed tests, so please call flag_gems.argmax directly in both cases before rebasing.

PR #4659, opened 2026-07-10
Optimize var_mean and count_nonzero single-dim reduction with inner/non-inner kernels

zheng1:
### PR Category

ops

### Type of Change

Performance Optimization

### Description

Optimize `torch.var_mean(x, dim=d)` and `torch.count_nonzero(x, dim=d)` for a single reduction dimension. Both are single-dim reductions that used `dim_compress`, so they share one direction and one PR.

`dim_compress` permutes the reduced dimension to the innermost position and copies the whole tensor before reducing. For a non-inner reduction (a middle or outer dim) that copy reads and writes the entire tensor once just to reorder it, and it dominated the runtime.

Both new paths split the input into `(M, N, K)` and reduce over `N` in place:

`var_mean` reduces in a single read of the input with a per-lane streaming Welford: each lane keeps a running `(count, mean, M2)` with no cross-lane reduction in the loop, and the lanes are merged once at the end with a numerically stable Chan combine. Welford produces the mean directly, so both outputs are stored with no extra pass. (An earlier revision used a two-pass mean/variance kernel that read the reduced dim twice, which doubled global-memory traffic on large tensors.) The divisor is clamped at 0, so a correction `>= N` yields `+inf`, matching torch. `count_nonzero` sums the nonzero mask into an int64 count. Multi-dim reductions keep the original paths.

### Performance

A100, `torch/gems` latency ratio (higher is better, 1.0 means parity with eager):

**var_mean** (fp32):

| (4096, 4096) | 0 | 0.35x | 1.12x |
| (8, 512, 512) | 1 | 0.36x | 1.14x |
| (4, 4096, 8) | 1 | 0.17x | 0.52x |
| (4096, 4096) | 1 | (inner) | 2.18x |

**count_nonzero** (fp32):

| (4096, 4096) | 0 | 0.93x | 2.75x |
| (8, 512, 512) | 1 | 0.15x | 0.62x |
| (4, 4096, 8) | 1 | 0.09x | 0.33x |
| (4096, 4096) | 1 | (inner) | 3.12x |

Every non-inner shape improves, several past parity.

### Testing

- `pytest tests/test_var_mean.py`: 204 passed; `pytest tests/test_count_nonzero.py`: 48 passed. Both include a new multi-tile test covering the multi-N-tile and multi-K-tile paths (N=4096, K=4096).
- Checked var and mean against `torch.var_mean` across float16/float32, 2D/3D shapes, every single dim, keepdim, and correction in {0,1,2}; and count against `torch.count_nonzero` across float/int/bool. Both match torch on empty dims: var_mean returns NaN for an empty reduction dim and an empty result for an empty spectator dim; count_nonzero counts 0 over an empty reduction dim. `correction >= N` gives `+inf` for the variance, matching torch. Negative dims work.
- `flake8` and `pre-commit` clean.

### Progress


src/flag_gems/ops/var_mean.py
```diff
@@ -144,6 +259,48 @@ def var_mean(x, dim=None, *, correction=None, keepdim=False):
             var_mean_kernel_2[(1,)](
                 acc, average, count, var, mean, N, correction, BLOCK_NUM
             )
+    elif len(dim) == 1:
```
tengqm, line 288: better double check the test coverage for new big branches here, thx.

zheng1: Good call. The multi-tile test already exercised the inner (K==1) and non-inner (K>1) kernels, but the empty branches were only covered by manual checks, not the committed tests. I added `test_var_mean_dim_empty` and `test_count_nonzero_dim_empty` that hit both empty cases: an empty reduction dim (N==0, which fills NaN for var and mean, or counts 0), and an empty spectator dim (M==0 or K==0, which returns an empty output), with keepdim variants, all asserted against torch.

So the new single-dim path is now covered on all its branches: inner vs non-inner, single-tile vs multi-tile, empty reduction dim, and empty spectator dim. 293 pass across the two test files.

bin913:
@zheng1 Please fix failing unit-test

zheng1:
The failing job is a CI setup network timeout, not a unit-test failure. The log shows `Failed to download cuda-bindings==13.3.1 ... network timeout` and `Installing FlagGems [nvidia-cuda133] [FAILED]`, so test-operator never ran the ops.

Re-ran locally on a CI-aligned A100 env (torch 2.11 / triton 3.6): test_var_mean.py 240 passed, all green. Could you re-trigger once the mirror is stable? Thanks.

PR #4676, opened 2026-07-11
Add retry logic to gpu_check_ascend.sh

tengqm:
## Problem

`gpu_check_ascend.sh` calls `npu-smi info` once at startup. If it fails (driver init delay, transient error), the script exits immediately with code 1, failing the entire CI job.

## Fix

Retry `npu-smi info` up to 3 times with 5-second intervals before giving up.


tengqm:
Wrong direction.

PR #4679, opened 2026-07-11
Only run post_install in triton mode

tengqm:
## Problem

On ascend-cann900, `post_install` installs `triton_ascend==3.2.1` which conflicts with FlagTree's triton backend:

```
ModuleNotFoundError: No module named 'triton._C.libtriton.ascend';
'triton._C.libtriton' is not a package
```

FlagTree replaces the triton package entirely — installing triton_ascend on top of it breaks the import chain.

## Fix

- Only execute `post_install` when `COMPILER=triton` (default is flagtree, which doesn't need triton_ascend)
- Also add `--index ${MIRROR}` so transitive deps (e.g. `attrs`) can be resolved from the public mirror

### DIFF
--- a/setup.sh
+++ b/setup.sh
@@ -199,11 +199,11 @@ if [ "${COMPILER}" != "flagtree" ] && [ "${COMPILER}" != "triton" ]; then
   exit 1
 fi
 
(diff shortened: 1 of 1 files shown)


tengqm:
wrong direction.

PR #4732, opened 2026-07-11
Promote silu.out from experimental; local tile tuning; drop experimental silu

### LINKED ISSUE #4694: Revisit the `silu` experimental operator for adoption
Part of #4686 (Batch 2 — exact duplicates).

The adoption criteria are as follows:

- pass the accuracy tests used for the existing implementation;
- exhibit performance improvements in terms for speedup;
- works at least on NVIDIA and two other backends.

tengqm:
## Promote `silu.out`; local tile tuning; drop experimental `silu`

The experimental `silu` is early KernelGen legacy code. Its only new surface is the `.out` overload. This PR adds `silu.out` backed by the **existing** `pointwise_dynamic` kernel, applies a small local tuning fix, and removes the experimental file.

### Why (data)

Both kernels were run through the same accuracy + benchmark suite on 5 backends (H20, Iluvatar, MetaX, Mthreads, Hygon).

- **Accuracy:** identical — all pass on every backend.
- **Performance:** parity on 4 backends. The one exception was **H20**, where the experimental kernel was clearly faster on large tensors.

Root cause of the H20 gap: NVIDIA's default `CodeGenConfig.max_tile_size = 512` is too small for this memory-bound elementwise op on H20. For contiguous inputs `pointwise_dynamic` uses a 1D fast-path with tile=512; the experimental kernel simply used `BLOCK_SIZE=1024`. Bumping the tile to 1024 makes the mature kernel match the experimental — so the experimental is not actually a better kernel.

H20, `silu` Gems Speedup vs torch (existing → +local config → experimental):

| dtype | shape | existing (tile 512) | local config (tile 1024) | experimental |
|-------|-------|---------------------|--------------------------|--------------|
| fp16 | 1024³ | 1.21 | **1.35** | 1.40 |
| fp16 | 4096² | 1.01 | **1.22** | 1.21 |
| bf16 | 1024³ | 1.19 | **1.35** | 1.38 |

### What this PR does
- adds `silu.out` backed by the existing kernel;
- applies a **local, per-op** config to `silu_forward`/`silu_backward`: `max_tile_size = max(1024, vendor_default)`. NVIDIA goes 512 → 1024; vendors already ≥1024 (Iluvatar/MetaX/Mthreads/Hygon) are unchanged; the global config and all other ops are untouched;
- removes the experimental implementation.

### Note for reviewers
The local config raises the tile to 1024 for **all** NVIDIA models for `silu` only. It was validated on H20; the CI runner is A100, so the CI benchmark gives us a second NVIDIA data point — please check that A100 does not regress before merging.

### Changes
- `ops/silu.py`: add `silu_out`; local config bump; normalize `logger.debug` tags to `GEMS <OPERATOR_ID>`
- `ops/__init__.py`, `__init__.py`: export + register `silu.out`
- `conf/operators.yaml`: add `silu_out` (sorted by id; `alpha: '5.4'`)
- `tests/test_silu.py`, `benchmark/test_silu.py`: add `silu_out`
- remove `experimental_ops/silu.py` and its tests

Closes #4694

---


tengqm:
## Cross-backend A/B data — existing vs experimental `silu`

Both kernels patched into `ops/` and run through the **same** accuracy + benchmark suite (`-m silu`, forward). Experimental = legacy KernelGen (`tl.exp`, fixed `BLOCK_SIZE=1024`), `is_cuda` guard removed so it runs on all backends. Swap verified on each node (`grep BLOCK_SIZE=1024`).

### Accuracy
All backends: **existing 18/18, experimental 18/18** — identical, no differences.

### Performance — Gems Speedup vs torch (higher = faster), large tensors

`existing → experimental`:

| Backend | fp16 1024³ | bf16 1024³ | fp32 1024³ | verdict |
|---------|-----------|-----------|-----------|---------|
| NVIDIA H20 | 1.21 → **1.40** | 1.19 → **1.38** | 0.97 → **1.05** | experimental notably faster |
| Iluvatar BI-V150 | 1.05 → **1.07** | 1.05 → **1.07** | 1.01 → 1.02 | experimental ~+2.5% |
| Mthreads MTT S5000 | ~parity | 0.89 → 0.92 | ~parity | parity |
| Hygon DCU BW2000 | 1.21 / 1.21 | 1.07 / 1.07 | 0.97 / 0.97 | parity (≡ to 3 decimals) |
| MetaX C500 | **0.80** → 0.77 | ~parity | 0.98 → **1.01** | fp16 slightly favors existing |

Small shapes (64×64) are launch-overhead noise, excluded.

### Root cause of the H20 gap (not a better kernel)
`silu` is memory-bandwidth bound. For contiguous inputs `pointwise_dynamic` collapses to a 1D fast-path; on NVIDIA it uses `CodeGenConfig.max_tile_size = 512`, while the experimental kernel used `BLOCK_SIZE=1024`. Bumping the tile to 1024 makes the **existing** kernel match the experimental — confirmed by a sweep on H20:

| fp16 speedup | tile=512 (default) | tile=1024 | tile=2048 | experimental |
|---|---|---|---|---|
| 1024³ | 1.21 | **1.35** | 1.35 | 1.40 |
| 4096² | 1.01 | **1.22** | 1.19 | 1.21 |
| 1G flat | 0.95 | 1.11 | 1.03 | 1.04 |

`num_warps` (4) and block-pointer vs plain load were **not** the cause — only tile size.

### What this PR ships (verified on H20)
Keeps the mature FlagGems 1.0 kernel and adds a **local, per-op** config `max_tile_size = max(1024, vendor_default)` on `silu_forward`/`silu_backward` (NVIDIA 512→1024; Iluvatar/MetaX/Mthreads/Hygon already ≥1024, unchanged; global config untouched). Result on H20 — `silu` / `silu.out`:

| dtype | shape | before (tile 512) | this PR (local config) |
|---|---|---|---|
| fp16 | 1024³ | 1.21 | **1.35** |
| fp16 | 4096² | 1.01 | **1.22** |
| bf16 | 1024³ | 1.19 | **1.35** |
| silu.out fp16 | 1024³ | — | **1.36** |

Accuracy after the change: **72/72 pass** on H20.

> Caveat: the tile bump applies to all NVIDIA models for `silu`. Validated on H20; the CI runner is A100 — please confirm A100 does not regress. (We also plan to probe/route to an H100 runner, see #4733.)

src/flag_gems/ops/silu.py
```diff
@@ -1,23 +1,36 @@
...
+from flag_gems.utils.codegen_config_utils import get_codegen_config
 from flag_gems.utils.triton_lang_extension import div_rn
 
 logger = logging.getLogger(__name__)
 
+# Local, per-op tuning. The default NVIDIA max_tile_size (512) is too small for
+# this memory-bound elementwise op on newer GPUs (e.g. H20), leaving it slower
+# than torch on large tensors. Bump the tile to >=1024 for silu only, starting
+# from the current device's default config so vendors whose default is already
+# >=1024 (Iluvatar/MetaX/Mthreads/Hygon/...) are unchanged, and the global config
```
tengqm, line 31: To reviewers: This is a finding when profiling the operator on multiple backends. We found that the experimental operator performs well in certain situation. This is an indication of improvement opportunity. After digging into the implementation, we found that relaxing max tile size in the experimental version happens to be the root cause. After revising the config, we tested the change on H100 as well (see CI job output), the operator gets improved not only on H20, but also on H100. This confirms that the revision is worthy for consideration.

huangyiqun:
Plz resolve conflicts.

PR #4737, opened 2026-07-12
fix: support SQLAlchemy 1.4 in SQL persistent model

shiptux:
## Motivation

`src/flag_gems/utils/models/sql.py` uses `sqlalchemy.orm.DeclarativeBase` and `mapped_column()`, which are SQLAlchemy >= 2.0 APIs. LTS distributions still ship SQLAlchemy 1.4 (`python3-sqlalchemy` is 1.4.31 on Ubuntu 22.04 and 1.4.50 on Ubuntu 24.04), so when flag_gems is installed as a distro package and its `Depends` resolves to the system SQLAlchemy, `import flag_gems` crashes at import time:

```
AttributeError: module 'sqlalchemy.orm' has no attribute 'DeclarativeBase'
```

## Changes

- Detect `DeclarativeBase` availability; on SQLAlchemy 1.4 fall back to the legacy `declarative_base()` API.
- `build_sql_model_by_py`: annotation-driven `Mapped`/`mapped_column()` mapping is also 2.0-only, so on 1.4 build the table columns explicitly with `sqlalchemy.Column`, mapping the python types (`bool`/`int`/`float`/`str`) to the corresponding SQL types.
- The rest of the ORM usage (`Session.get`, automap `prepare`, `CreateTable(if_not_exists=True)`, `RollbackSession`) is already 1.4-compatible.

## Verification

In an `ubuntu:24.04` container with apt `python3-sqlalchemy` 1.4.50:
- `import flag_gems` and `flag_gems.enable()` succeed;
- `SQLPersistantModel` config/benchmark round trips and re-loading an existing db via automap (fresh engine) all pass.

Regression-checked with pip SQLAlchemy 2.x (same test script): no behavior change, the 2.0 code path is untouched.


huangyiqun (commented):
`pyproject.toml` still pins `sqlalchemy==2.0.48`, so the new SQLAlchemy 1.4 fallback may not be exercised in a normal installation. Would it make sense to relax the dependency range and add a small CI case for SQLAlchemy 1.4 to make sure this compatibility path stays covered?

shiptux:
@huangyiqun Thanks for the review — addressed in 83a242259:

- **pyproject.toml**: relaxed `sqlalchemy==2.0.48` to `sqlalchemy>=1.4.31,<2.1` (floor matches the `python3-sqlalchemy` shipped by Ubuntu 22.04), so the 1.4 path is installable normally.
- **tests/test_sql_persistent_model.py**: device-independent, sqlite-backed round-trip tests for `SQLPersistantModel` (config, benchmark, and reload-via-automap — the automap case covers the 1.4 fallback's explicit-`Column` table creation).
- **.github/workflows/sqlalchemy-compat.yaml**: runs those tests on a CPU-only `ubuntu-latest` runner against both `sqlalchemy>=1.4.31,<1.5` and `>=2.0,<2.1` (path-filtered so it only triggers on changes to the models code, the test, or dependency metadata). Verified both matrix legs locally in a container before pushing (3 passed under 1.4.50, 3 passed under 2.0.x).

Also completed a real-GPU verification of this branch in the meantime (A100, Ubuntu 22.04 container, apt `python3-sqlalchemy` **1.4.31**): real device detection (no `GEMS_VENDOR` override), `import flag_gems` + `flag_gems.enable()`, the SQL persistent-model round trips, and add/softmax/mm dispatched through flag_gems on CUDA all pass, matching torch references.

huangyiqun:
> @huangyiqun Thanks for the review — addressed in [83a2422](https://github.com/flagos-ai/FlagGems/commit/83a242259a9d7eba0bb8f15f2070d8571e125232):
> 
> * **pyproject.toml**: relaxed `sqlalchemy==2.0.48` to `sqlalchemy>=1.4.31,<2.1` (floor matches the `python3-sqlalchemy` shipped by Ubuntu 22.04), so the 1.4 path is installable normally.
> * **tests/test_sql_persistent_model.py**: device-independent, sqlite-backed round-trip tests for `SQLPersistantModel` (config, benchmark, and reload-via-automap — the automap case covers the 1.4 fallback's explicit-`Column` table creation).
> * **.github/workflows/sqlalchemy-compat.yaml**: runs those tests on a CPU-only `ubuntu-latest` runner against both `sqlalchemy>=1.4.31,<1.5` and `>=2.0,<2.1` (path-filtered so it only triggers on changes to the models code, the test, or dependency metadata). Verified both matrix legs locally in a container before pushing (3 passed under 1.4.50, 3 passed under 2.0.x).
> 
> Also completed a real-GPU verification of this branch in the meantime (A100, Ubuntu 22.04 container, apt `python3-sqlalchemy` **1.4.31**): real device detection (no `GEMS_VENDOR` override), `import flag_gems` + `flag_gems.enable()`, the SQL persistent-model round trips, and add/softmax/mm dispatched through flag_gems on CUDA all pass, matching torch references.

Thanks for addressing this so thoroughly. I’ll approve workflows to run.

shiptux:
Additional regression check on a ROCm stack (the second environment named in the original report): torch 2.10.0+rocm7.2.4, venv SQLAlchemy **2.0.51**, AMD Radeon 780M.

- `import flag_gems` + `enable()`, SQL persistent-model config/benchmark round trips, and automap reload all pass with this branch;
- a real op dispatched through flag_gems on the AMD GPU matches the torch reference (clean-state runs of this branch and of master behave identically, confirming the sqlalchemy change has no effect on the kernel path — under SQLAlchemy ≥ 2.0 the code path is unchanged).

huangyiqun:
Plz resolve conflicts.

bin913:
@shiptux Please resolve conflicts

PR #4745, opened 2026-07-13
Split C++ extension into a separate per-vendor wheel

tengqm:
Split the C++ extension into a separate per-vendor wheel so the main package can ship as pure Python.

- Move `csrc/`, `CMakeLists.txt`, `cmake/`, `include/`, `lib/`, `ctests/` into `cpp/`.
- Main `pyproject.toml`: setuptools backend, add `cpp-*` extras.
- New `cpp/pyproject.toml`: scikit-build-core; the `.so` files install into the `flag_gems/` namespace; version follows the main package (setuptools_scm `root=".."`). Vendor name is injected at build time by `tools/set_cpp_vendor.sh`.
- `setup.sh`: `ENABLE_CPP=1` now builds and installs the extension from `cpp/`.
- Also fixes the NPU build (`torch::custom_class_available` does not exist; use `c10_npu::device_count()`).

Install:
- `pip install flag-gems` — pure Python
- `pip install flag-gems[cpp-cuda]` — pulls the CUDA native package

Test results per backend in a comment below.


tengqm:
End-to-end test on each backend (build wheel → install → import → check op dispatch):

| Backend | Node | Result |
|---|---|---|
| CUDA | h20 | pass |
| MUSA | mthreads | pass |
| NPU | hw25 | pass |
| GCU | enflame | builds; needs libtriton_jit#32 (string constexpr fix) |
| IX | ix15 | blocked: corex ships a fake nvcc, Torch still forces the CUDA language. Reported to vendor. |

CUDA/MUSA/NPU: `.so` files land in `flag_gems/`, `from flag_gems import c_operators` works, main and native package report the same version, and `aten::nonzero` is overridden by the native op.

Also verified from the Nexus index on h20: `pip install flag-gems[cpp-cuda]` pulls both packages, and `pytest tests/test_abs.py tests/test_add.py` passes (1094 cases).

Note: each vendor needs `source tools/env.sh <backend>` first, otherwise torch_npu/torch_musa fails to load.

chen98038 (approved):

PR #4780, opened 2026-07-13
[KernelGen][Mthreads] Fix reglu: accuracy_fail

xzliu-opt:
## Summary
- **Operator:** reglu
- **Root cause:** The FlagGems reglu kernel is correct; the failures came from the reference path. TransformerEngine's reglu kernel on the mthreads backend only accepts 2-D inputs (its cast_gated kernel asserts input.data.shape.size() == 2), so calling tex.reglu on 1-D/3-D/4-D/5-D test shapes crashed, failing 12 of 15 accuracy cases and the benchmark.
- **Fix:** Wrapped the TE reference call in the accuracy test (te_reglu_ref helper) and in the benchmark (te_reglu_op wrapper) to flatten leading dims into a 2-D tensor before invoking tex.reglu and restore the original output shape afterwards. No operator source code was changed.
- **Files modified:**
  - `tests/test_reglu.py`
  - `benchmark/test_reglu.py`

## Verification
- Accuracy test: 15/15 passed
- Benchmark: passed
- Format check: passed

## Test Plan
- [ ] `pytest -m 'reglu' tests/ --ref cpu -vs`
- [ ] `pytest -m 'reglu' benchmark/ --level core --record log`

## Issue
- WEEKTEST-563

### DIFF
--- a/benchmark/test_reglu.py
+++ b/benchmark/test_reglu.py
@@ -16,12 +16,23 @@ except ImportError:
     TE_OP = None
 
 
+def te_reglu_op(input_tensor, quantizer=None):
+    # TransformerEngine's reglu kernel only accepts 2-D inputs on this backend
+    # (its cast_gated kernel asserts input.data.shape.size() == 2), so collapse
+    # all leading dims before calling it and restore the output shape afterwards.
+    shape = input_tensor.shape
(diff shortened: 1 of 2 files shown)


lukatao:
resolve the conflicts pls

PR #4785, opened 2026-07-13
[KernelGen][Mthreads] Fix feature_dropout: accuracy_fail

xzliu-opt:
## Summary
- **Operator:** feature_dropout
- **Root cause:** On the mthreads backend, torch_musa's elementwise tensor * python_float rounds the scalar to the tensor dtype when the operand has >=2 elements but keeps full precision for a single-element operand; the shared Triton kernel always used full-precision scale with fp32 multiply, producing 1-ULP differences vs the per-channel reference for spatial_size>=2 low-precision cases.
- **Fix:** Added a mthreads-specific feature_dropout override that rounds scale to the input dtype (float16/bfloat16) when spatial_size>=2 to match torch_musa's per-channel scalar rounding, while keeping full-precision scale for the spatial_size==1 case; registered it in the mthreads ops __init__.
- **Files modified:**
  - `src/flag_gems/runtime/backend/_mthreads/ops/feature_dropout.py`

## Verification
- Accuracy test: 72/72 passed
- Benchmark: passed
- Format check: passed

## Test Plan
- [ ] `pytest -m 'feature_dropout' tests/ --ref cpu -vs`
- [ ] `pytest -m 'feature_dropout' benchmark/ --level core --record log`

## Issue
- WEEKTEST-557


bin913:
@xzliu-opt Please fix failing unit-test

PR #4787, opened 2026-07-13
[KernelGen][Mthreads] Fix scaled_dot_product_attention: benchmark_fail

xzliu-opt:
## Summary
- **Operator:** scaled_dot_product_attention
- **Error type:** benchmark_fail
- **Root cause:** The shared Triton flash-attention kernel runs ~10x slower than the native MUSA flash kernel on mthreads because its autotune config space (a single BLOCK_M=BLOCK_N=32, stages=1 config) is far too small for the hardware, yielding benchmark speedups of only 0.03-0.13.
- **Fix:** Added a mthreads-backend override that delegates scaled_dot_product_attention (forward and backward) to torch.nn.functional.scaled_dot_product_attention, which dispatches to the native MUSA fused kernel. Registered it in the mthreads ops __init__.py; shared code untouched.
- **Files modified:**
  - `src/flag_gems/runtime/backend/_mthreads/ops/attention.py`

## Verification
- Accuracy test: 40/40 passed
- Benchmark: passed
- Format check: passed

## Test Plan
- [ ] `pytest -m 'scaled_dot_product_attention' tests/ --ref cpu -vs`
- [ ] `pytest -m 'scaled_dot_product_attention' benchmark/ --level core --record log`

## Issue
- WEEKTEST-551

### DIFF
--- a/src/flag_gems/runtime/backend/_mthreads/ops/__init__.py
+++ b/src/flag_gems/runtime/backend/_mthreads/ops/__init__.py
@@ -5,6 +5,7 @@ from .amax import amax
 from .any import any, any_dim, any_dims
 from .arange import arange, arange_start
 from .argmin import argmin
+from .attention import scaled_dot_product_attention
 from .batch_norm import batch_norm, batch_norm_backward
 from .celu import celu
 from .conv2d import conv2d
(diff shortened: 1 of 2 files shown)


huangyiqun (commented):
Looks good—could we add a regression test to make sure this always hits the native MUSA SDPA path?

PR #4788, opened 2026-07-13
[KernelGen][Mthreads] Fix geglu: accuracy_fail

xzliu-opt:
## Summary
- **Operator:** geglu
- **Root cause:** The test passed the N-D input tensor directly to TransformerEngine's tex.geglu as the reference, but TE's geglu (mthreads build) asserts input must be strictly 2D, so every non-2D shape crashed in the reference before reaching the operator. The FlagGems geglu operator itself is correct.
- **Fix:** In tests/test_geglu.py, flatten the input to 2D before calling tex.geglu for the reference, then reshape the reference output back to the expected (..., H) shape. No operator code change needed. Additionally wrapped the TE reference in the benchmark (te_geglu_op) to flatten to 2-D so the benchmark runs on musa, matching the accuracy test fix.
- **Files modified:**
  - `tests/test_geglu.py`
  - `benchmark/test_geglu.py`

## Verification
- Accuracy test: 15/15 passed
- Benchmark: passed
- Format check: passed

## Test Plan
- [ ] `pytest -m 'geglu' tests/ --ref cpu -vs`
- [ ] `pytest -m 'geglu' benchmark/ --level core --record log`

## Issue
- WEEKTEST-562

### DIFF
--- a/benchmark/test_geglu.py
+++ b/benchmark/test_geglu.py
@@ -20,6 +20,18 @@ except ImportError:
     GEMS_OP = None
 
 
+def te_geglu_op(input_tensor, quantizer=None):
+    # TransformerEngine's geglu only accepts 2-D input on some backends (e.g. musa),
+    # while FlagGems supports arbitrary shapes by flattening to 2-D internally.
+    # Flatten to 2-D for the reference and reshape the result back to the expected
+    # output shape so the comparison stays valid across vendors.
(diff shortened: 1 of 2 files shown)


huangyiqun (commented):
Looks good—could we share the TE flatten/reshape helper between the test and benchmark so they don’t drift?

PR #4789, opened 2026-07-13
[KernelGen][Mthreads] Fix fill_scalar_: accuracy_fail

xzliu-opt:
## Summary
- **Operator:** fill_scalar_
- **Root cause:** transformer_engine (imported at test collection via tests/test_dgeglu.py) monkeypatches torch.Tensor.is_cuda to True for all tensors including CPU ones, so the shared fill.py guard `if not value.is_cuda` misroutes a CPU scalar value into the tensor kernel and crashes with 'Pointer argument cannot be accessed from Triton (cpu tensor?)'.
- **Fix:** Added a mthreads backend override of fill_tensor/fill_tensor_out/fill_tensor_ that checks value.is_musa instead of value.is_cuda, correctly routing CPU scalar values to the scalar kernel while leaving the musa-tensor path unchanged.
- **Files modified:**
  - `src/flag_gems/runtime/backend/_mthreads/ops/fill.py`

## Verification
- Accuracy test: 102/102 passed
- Benchmark: passed
- Format check: passed

## Test Plan
- [ ] `pytest -m 'fill_scalar_' tests/ --ref cpu -vs`
- [ ] `pytest -m 'fill_scalar_' benchmark/ --level core --record log`

## Issue
- WEEKTEST-561

### DIFF
--- a/src/flag_gems/runtime/backend/_mthreads/ops/__init__.py
+++ b/src/flag_gems/runtime/backend/_mthreads/ops/__init__.py
@@ -9,6 +9,7 @@ from .batch_norm import batch_norm, batch_norm_backward
 from .celu import celu
 from .conv2d import conv2d
 from .dropout import dropout, dropout_backward
+from .fill import fill_tensor, fill_tensor_, fill_tensor_out
 from .gather import gather, gather_backward
 from .index_add import index_add, index_add_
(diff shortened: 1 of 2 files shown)


huangyiqun (commented):
Looks good—could you add a regression test for the CPU scalar tensor case after importing `transformer_engine`?

PR #4790, opened 2026-07-13
[KernelGen][Mthreads] Fix atan2_out: accuracy_fail

xzliu-opt:
## Summary
- **Operator:** atan2_out
- **Root cause:** The default atan2 kernel calls the mthreads libdevice atan2 intrinsic (__nv_atan2f), which crashes the llc instruction selector (segfault, exit -11) for half-precision inputs whenever a tile is partially filled (numel not tile-aligned, e.g. numel==1 or 255/257), surfacing as a RuntimeError for the affected fp16 test cases.
- **Fix:** Added a mthreads-specific override that rebuilds atan2 from the single-argument atan intrinsic (which lowers correctly) plus the standard quadrant correction (atan(y/x) with +/-pi shifts and +/-pi/2 for x==0) and explicit NaN propagation, avoiding the crashing atan2 intrinsic entirely.
- **Files modified:**
  - `src/flag_gems/runtime/backend/_mthreads/ops/atan2.py`

## Verification
- Accuracy test: 18/18 passed
- Benchmark: passed
- Format check: passed

## Test Plan
- [ ] `pytest -m 'atan2_out' tests/ --ref cpu -vs`
- [ ] `pytest -m 'atan2_out' benchmark/ --level core --record log`

## Issue
- WEEKTEST-560


huangyiqun (commented):
Looks good, but could we add edge-case tests for signed zeros and infinities? Manual `atan2` handling can be tricky there.

PR #4799, opened 2026-07-13
[Hygon] fix mm float64 accumulation

douxetpur:
### PR Category
Operator

### Type of Change
Bug Fix

### Description
Fix the Hygon-specific `mm` kernel for float64 inputs.

The kernel previously initialized the loop-carried accumulator as `tl.float32` while float64 inputs can produce float64 dot results, which may trigger a dtype mismatch in the Hygon fp64 `mm` path. This change adds an `IS_FP64` kernel flag, uses a `tl.float64` accumulator for float64 inputs, and keeps the existing fp32 accumulation path unchanged for other dtypes.

### Issue
N/A

### Progress

### Performance
No benchmark added. This only changes the Hygon float64 `mm` path; existing non-fp64 paths are unchanged.

### DIFF
--- a/src/flag_gems/runtime/backend/_hygon/ops/mm.py
+++ b/src/flag_gems/runtime/backend/_hygon/ops/mm.py
@@ -50,6 +50,7 @@ def mm_kernel(
     BLOCK_N: tl.constexpr,
     BLOCK_K: tl.constexpr,
     GROUP_M: tl.constexpr,
+    IS_FP64: tl.constexpr = False,
 ):
     pid = ext.program_id(0)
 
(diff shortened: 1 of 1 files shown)


src/flag_gems/runtime/backend/_hygon/ops/mm.py
```diff
@@ -147,7 +157,83 @@ def mm_kernel(
...
 
-_ordered_datatypes = [torch.float16, torch.bfloat16, torch.float32]
+@libentry()
+@libtuner(
+    configs=_FP64_MM_CONFIGS,
+    key=["M", "N", "K"],
+    strategy=["log", "log", "log"],
+)
+@triton.jit
+def mm_kernel_fp64(
```
0x45f: why add new fp64 kernel, can we change code like ops/mm.py

douxetpur: Thanks for the suggestion. The main difference is that the FP64 path in ops/mm.py still relies on tl.dot, which is not supported for FP64 by the current Hygon Triton backend.
I initially handled FP64 with an IS_FP64 branch in the existing Hygon kernel. However, the fallback requires a statically unrolled outer-product loop, and sharing the normal MM autotune configs caused every candidate to compile that loop. This made cold-cache Hygon CI take too long and eventually get cancelled.
Therefore, this PR separates the FP64 implementation into an internal mm_kernel_fp64 with one small dedicated config. The public mm/mm.out APIs are unchanged, and the existing FP16/BF16/FP32 path is not affected.
We could keep the host-side code similar to ops/mm.py, but the FP64 compute path and its tuning config need to remain Hygon-specific unless FP64 tl.dot becomes supported.

0x45f (approved):

PR #4500, opened 2026-07-03
[KMCompiler][Nvidia]Add Polygamma Operator

cheersluvs:
## Summary
Adds a Triton kernel for `polygamma` (with `polygamma.out` and in-place `polygamma_`). Computes the n-th derivative of the digamma function, dispatching by order: raw autotuned Triton kernels for digamma (n=0) and trigamma (n=1), and the Cephes Euler-Maclaurin Hurwitz-zeta algorithm (n>=2) — the same algorithm family PyTorch's native kernels use, for reference-matching accuracy. The `(-1)^(n+1) * n!` zeta scale factor is computed in-kernel as float32 `exp(lgamma(n+1))`, matching PyTorch's kernels, so the wrapper needs no torch call.

The n=0/n=1 paths use hand-written autotuned kernels with a `pointwise_dynamic` fallback for non-contiguous / integer inputs. In-place writes are routed through a fixed-config kernel (an autotuned kernel would re-run its config sweep on the same buffer and corrupt aliased data).

## Testing
- Accuracy suite: `pytest tests/test_polygamma.py` on NVIDIA H20 — **187 passed** — float16/float32/bfloat16, n in {0, 1, 2, 5, 8}, positive and wide [-5, 5) domains, out and in-place variants, compared against torch native on device via `to_reference` + `gems_assert_close`.
- Also validated on Huawei Ascend 910B (see follow-up PR for the NPU backend override).

## Performance
Test command: `pytest benchmark/test_polygamma.py --level core --record log` (NVIDIA H20, kernel mode)

| Op | Dtype | Configuration | Torch (ms) | Gems (ms) | Speedup | TFLOPS |
|---|---|---|---|---|---|---|
| `polygamma` | float16 | [1073741824] | 12.4896 | 5.6499 | 2.211 | 0.190 |
| `polygamma` | float16 | [64, 64] | 0.0093 | 0.0055 | 1.696 | 0.001 |
| `polygamma` | float16 | [4096, 4096] | 0.2010 | 0.0947 | 2.123 | 0.177 |
| `polygamma` | float16 | [64, 512, 512] | 0.2008 | 0.0945 | 2.125 | 0.178 |
| `polygamma` | float16 | [1024, 1024, 1024] | 12.4896 | 5.6499 | 2.211 | 0.190 |
| `polygamma` | float32 | [1073741824] | 12.1423 | 5.7910 | 2.097 | 0.185 |
[...]

PR #4809, opened 2026-07-14
[KMCompiler][Ascend] Add polygamma NPU backend support

cheersluvs:
### PR Category
Operator | OP Test | Benchmark

### Type of Change
New Feature

### Description
Adds Huawei Ascend (NPU) backend support for `polygamma`.

1. **In-place override** (`src/flag_gems/runtime/backend/_ascend/ops/polygamma.py`): only `polygamma_` is overridden. Writing through `pointwise_dynamic` with `out0` aliasing the input leaves a fraction of the elements unwritten on large tensors on this backend (measured with gems alone, no torch reference: 22016 of 1048576 differ from the out-of-place result at [1024, 1024], n = 2), so the in-place path computes out-of-place and copies back, matching the existing Ascend `pow_` override. This affects the n >= 2 zeta path, which is the remaining `pointwise_dynamic` writer. The functional and `out=` variants need **no** override -- they run the base implementation unchanged, including its raw autotuned kernels for n = 0 / n = 1, which work correctly here and are measurably faster than routing the same math through `pointwise_dynamic`.
2. **Benchmark reference on Ascend** (`benchmark/test_polygamma.py`, `benchmark/base.py`): `torch_npu` has no native `polygamma`, so `torch.polygamma` silently falls back to CPU and is unusable as a baseline at benchmark shapes. On non-CUDA/CPU devices the benchmark instead composes trigamma from primitives so the baseline stays on device, and measures the kernel directly via `gems_op` (the harness would otherwise time the gems side by re-running `torch_op` under `use_gems`, which no longer routes through `torch.polygamma`). The two 1e9-element core shapes are dropped on Ascend, following the existing per-vendor `# oot` filters in `set_shapes`, because the composed baseline costs ~dozens of launches per call. NVIDIA behaviour is unchanged (`gems_op=None`, plain `torch.polygamma`).
3. **Test skip** (`tests/test_polygamma.py`): the `wide_domain` zeta cases (n >= 2) are skipped on Ascend -- those inputs are ill-conditioned near half-integer x, and with the reference falling back to the CPU `pow` the comparison exceeds rtol on numerically meaningless values. Still exercised strictly on CUDA.

### Correctness
```bash
pytest tests/test_polygamma.py          # Ascend 910B, CANN 8.5, torch_npu 2.6, FlagTree
pytest tests/test_polygamma.py --ref=cpu --quick
```

| phase | platform | result |
| --- | --- | --- |
| `tests/test_polygamma.py` | Ascend | **184 passed, 3 skipped** |
| `tests/test_polygamma.py --ref=cpu --quick` | Ascend | **14 passed, 3 skipped** |
| `tests/test_polygamma.py` | NVIDIA (H20) | **187 passed** (base op, #4500) |

The 3 skips are the `wide_domain` zeta cases described above.

### Performance
**Ascend** — baseline: trigamma composed from primitives (`sin` / `where` / arithmetic, the same reflection, recurrence and asymptotic series the kernel uses), since `torch_npu` has no native `polygamma`. Measured with wall-clock `--mode=operator` timing (see the note below).

```bash
pytest benchmark/test_polygamma.py --level core --mode operator   # Ascend 910B, CANN 8.5, torch_npu 2.6
```

**All 27 configurations are faster than the baseline** (min 1.29x, max 15.94x, mean 4.87x; requirement: >= 0.8).

`polygamma` — two-level arithmetic mean **3.81** (fp16 5.02 / fp32 2.14 / bf16 4.27)

| shape | fp16 | fp32 | bf16 |
| --- | ---: | ---: | ---: |
| (64, 64) | 1.91 | 1.57 | 1.76 |
| (4096, 4096) | 4.95 | 1.29 | 3.22 |
| (64, 512, 512) | 8.21 | 3.57 | 7.82 |

`polygamma_` — two-level arithmetic mean **4.94** (fp16 4.86 / fp32 3.98 / bf16 5.99)

| shape | fp16 | fp32 | bf16 |
| --- | ---: | ---: | ---: |
| (64, 64) | 1.64 | 1.56 | 1.64 |
| (4096, 4096) | 4.70 | 4.38 | 10.92 |
| (64, 512, 512) | 8.23 | 6.00 | 5.41 |

`polygamma_out` — two-level arithmetic mean **5.86** (fp16 6.00 / fp32 2.85 / bf16 8.73)

| shape | fp16 | fp32 | bf16 |
| --- | ---: | ---: | ---: |
| (64, 64) | 1.96 | 2.10 | 2.20 |
| (4096, 4096) | 4.50 | 2.95 | 8.05 |
| (64, 512, 512) | 11.53 | 3.49 | 15.94 |

Note on measurement: Ascend kernel mode routes through `do_bench_npu`, which reports a mean *per-kernel* latency. A baseline composed of ~dozens of launches is therefore credited that many times too fast (it reports ~0.09x where the true figure is well above 1x), so the figures above use `--mode=operator` wall-clock timing -- the same split #4950 uses (`--mode=operator` for Ascend, `--mode=kernel` for NVIDIA). Run-to-run spread on this backend is wide (individual cells move by 2-3x between runs); the direction is stable, the exact values are not.

**NVIDIA** — unchanged by this PR; the base operator merged in #4500 (arithmetic mean 2.576, every configuration > 1x).


Caeruleann:
Please fix the ascend test error.

Caeruleann (dismissed):

Caeruleann (approved):

Issue #4686, opened 2026-07-11
Migrating/Deleting operators from the experimental group (Batch 2)


tengqm:
Follow-up to #4175. Additional experimental operators that overlap with existing implementations and need to be examined for adoption or removal.

## Criteria

- Pass the accuracy tests used for the existing implementation
- Exhibit performance improvements in terms of speedup
- Works at least on NVIDIA and two other backends

## Exact duplicates (same op already exists in ops/)

These experimental ops have an identically-named existing implementation:

- [ ] _unsafe_view
- [ ] amin
- [ ] expand
- [ ] im2col
- [ ] permute_copy
- [ ] relu
- [ ] sigmoid
- [ ] silu
- [ ] sinc
- [ ] softplus
- [ ] threshold
- [ ] threshold_
- [ ] upsample_nearest3d
- [ ] zeros_like

## In-place variants (out-of-place exists in ops/)

These are in-place versions of existing operators:

- [ ] absolute_
- [ ] copy_
- [ ] cos_
- [ ] cosh_
- [ ] erf_
- [ ] erfinv_
- [ ] exp_
- [ ] fix_
- [ ] gelu_
- [ ] hardsigmoid_
- [ ] hypot_
- [ ] leaky_relu_
- [ ] log_
- [ ] log10_
- [ ] neg_
- [ ] rad2deg_
- [ ] relu_
- [ ] rsqrt_
- [ ] sigmoid_
- [ ] silu_
- [ ] sin_
- [ ] sinc_

## Notes

- For exact duplicates: compare code, if identical → delete experimental; if different → evaluate performance
- For in-place variants: evaluate as new operator additions (like deg2rad_ in #4685)
- Same `is_cuda` assert issue (#4682) likely applies to many of these

PR #4810, opened 2026-07-14
Drop experimental operator sinc

zheng1:
## Summary

Remove the experimental `sinc` operator implementation and its unit and performance tests.

The formal `sinc` operator in `ops/` already covers this op and is registered in `_FULL_CONFIG`, so the experimental version is redundant. Verified on A100:

- The formal version matches `torch.sinc` across float32/float16/bfloat16, including the `x = 0` special case (`sinc(0) = 1`).
- The formal version is faster than the experimental one (106us vs 329us on a 4096x4096 fp16 input), so nothing is lost by dropping the experimental copy.
- The experimental `sinc` is a standalone file, not exported from `experimental_ops/__init__.py` and not imported anywhere.

## Files removed

- `src/flag_gems/experimental_ops/sinc.py`
- `experimental_tests/unit/sinc_test.py`
- `experimental_tests/performance/sinc_test.py`

## Issues

Related: #4686


w1120029931-bit (approved):
lgtm

PR #4815, opened 2026-07-15
Optimize std single-dim reduction with single-read Welford kernels

zheng1:
### PR Category

ops

### Type of Change

Performance Optimization

### Description

`torch.std(..., dim=...)` for a single reduction dimension read the reduced dim twice: pass 1 computed the mean, pass 2 re-read the whole tensor to accumulate the squared deviations. On large tensors that second read doubled global-memory traffic and made this the dominant cost, so the reduction ran ~2x slower than eager (e.g. `1024^3` dim=1 fp32 at 0.51x).

Replace the two passes in both the inner (`K == 1`) and non-inner (`K > 1`) kernels with a single-read per-lane streaming Welford: each lane keeps a running `(count, mean, M2)` with no cross-lane reduction inside the loop, and the lanes are merged once at the end with a numerically stable Chan combine. The single read halves the memory traffic while keeping the two-pass numerics. This mirrors the same change made to `var` in #4644.

The `ONE_TILE_PER_CTA` path (small reduced dim, fits in one tile) is unchanged in behavior. The `N - correction <= 0` NaN guard and empty-tensor handling in the wrapper are untouched.

### Performance

A100, `torch/gems` latency ratio (higher is better, 1.0 = parity with eager), measured by interleaving torch and gems so the ratio is robust to machine load. `two-pass` is master; `single-read` is this PR:

| shape | dim | dtype | two-pass | single-read |
|-------|-----|-------|---------:|------------:|
| (1024, 1024, 1024) | 1 | fp16 | 0.59x | 0.87x |
| (1024, 1024, 1024) | 1 | fp32 | 0.51x | 0.72x |
| (64, 512, 512) | 1 | fp32 | 0.57x | 0.64x |
| (4096, 4096) | 0 | fp32 | 0.85x | 0.93x |
| (4096, 4096) | 1 | fp32 | 1.15x | 1.19x |
| (8192, 8192) | 1 | fp16 | 1.69x | 1.70x |

Large non-inner reductions go from ~2x slower than eager to at or near parity; inner reductions are unchanged.

### Testing

- `pytest tests/test_std.py`: 180 passed (default CUDA), 2 passed (`--ref=cpu --quick`).
- Checked against `torch.std` across float32/float16/bfloat16, 1D/2D/3D shapes, every single dim, both keepdim settings, and correction in {0, 1, 2}. Edge cases match torch: NaN when `N - correction <= 0`, NaN/inf propagation, non-negative variance for constant inputs.
- Numerics verified on hostile inputs: large data mean (shift up to 1e5), outlier first element, tiny variance with N up to 2^25, multi-tile inner and non-inner paths, all matching torch.
- `pre-commit` and `flake8` clean.

### Progress

- [x] Change is covered by the existing `tests/test_std.py` suite.


w1120029931-bit (approved):
lgtm

Issue #4812, opened 2026-07-14
pointwise_dynamic/libentry per-call CPU dispatch overhead dominates small-tensor latency (root cause of #4558, #2416)


zheng1:
The pointwise perf-degradation reports (#4558 `add` → 69% throughput, #2416 "FlagGems JIT overhead") share one root cause: per-call CPU dispatch cost in the `pointwise_dynamic` wrapper and `libentry.run`, not the Triton kernels.

**Kernel isn't slow.** Under CUDA graph (CPU launch removed), official sigmoid GPU time == a hand-written flat kernel:

| numel | flat GPU | official GPU |
|------:|---------:|-------------:|
| 65536 | 12.40 | 12.70 |
| 4194304 | 21.42 | 23.85 |
| 16777216 | 64.12 | 64.26 |

**The gap is a fixed ~30 us of Python dispatch per call, size-independent.** Eager wall-clock, official vs a bare kernel:

| numel | official | bare | gap |
|------:|---------:|-----:|----:|
| 256 | 64.8 | 33.4 | 31.5 |
| 1048576 | 63.9 | 33.8 | 30.1 |
| 16777216 | 63.5 | 57.3 | 6.2 |

Framework-wide, not op-specific (tiny tensor, GPU ~0): sigmoid/relu/abs/silu/gelu all ~70 us under `use_gems` vs ~7 us eager.

Breakdown (cProfile + staged timing): `libentry.run` rebuilding the cache key every call ~20 us, `type_promotion` (elementwise_dtypes) 3.4 us, `heuristics_for_tile_size` + grid 1.5 us, `StridedBuffer` wrap 0.6 us. Key and launch metadata are invariant per (shape, dtype, stride) but recomputed every call.

**PoC:** cache launch metadata per signature and reuse the compiled kernel → official sigmoid 63.5 → 25 us, numerics identical.

| numel | official | cached | speedup |
|------:|---------:|-------:|--------:|
| 1024 | 63.6 | 25.2 | 2.52x |
| 1048576 | 64.4 | 25.7 | 2.50x |
| 16777216 | 64.0 | 53.6 | 1.19x |
| 67108864 | 199.5 | 199.5 | 1.00x |

2.5x on overhead-bound sizes, → 1.0x once bandwidth-bound.

[...]

PR #4827, opened 2026-07-15
Cache heuristics_for_tile_size in the pointwise_dynamic hot path

zheng1:
### PR Category

Performance Optimization

### Type of Change

Performance Optimization

### Description

`heuristics_for_tile_size(max_tile_size, *sizes)` is a pure function of a few ints (the tile/reduction sizes) that the `pointwise_dynamic` generated wrapper recomputes on every kernel launch. Memoize it with `functools.lru_cache`.

This follows the pattern already used in the repo for other pure per-call computations, e.g. `get_num_sms(idx)` in `cumsum.py`/`cumprod.py` and `_block_mesh(num_blocks)` in `fused/moe_align_block_size.py`, both of which `@lru_cache` a pure `int -> result` computation.

All call sites pass plain ints (shape dims via `*shape`, or `numel()`), so the arguments are always hashable. `maxsize=1024` bounds memory, and the return value is an immutable tuple, so cached-result mutation is not a concern.

This is a small, low-risk piece of a larger effort to reduce the fixed per-call CPU dispatch overhead of `pointwise_dynamic` operators (see #4812). It removes one of the recomputations on the launch hot path with no behavior change.

### Performance

A100, sigmoid `4096` fp16, per-call wall time (min of repeated timed blocks, so it isolates the CPU overhead):

| | per-call |
|---|---:|
| baseline | ~64.2 us |
| with cache | ~62.4 us |

A modest ~1.8 us / ~3% on its own; the remaining per-call overhead lives in the dtype-promotion path, which is being addressed separately with a more careful cache key.

### Testing

- `pytest tests/test_pointwise_dynamic.py`: 93 passed, 2 skipped.
- Verified `heuristics_for_tile_size` returns identical tuples with the cache active (torch.Size expansion, numpy ints, multi-dim), and that all call sites pass hashable ints.
- `pre-commit` (black, flake8, isort) clean.

### Issue

Part of #4812.


w1120029931-bit (approved):
lgtm

PR #4831, opened 2026-07-15
fix(bucketize): reshape output_flat instead of output for non-contiguous input

ErenAta16:
## What changed

`bucketize()` in `src/flag_gems/ops/bucketize.py` returns `output.reshape(input.shape)`, but the Triton kernel actually writes its results into `output_flat`. For a **contiguous** input these are the same storage (`flatten()` returns a view), so the bug is invisible. For a **non-contiguous** input, `torch.empty_like(input)` preserves that non-contiguous layout, so `output.flatten()` can't return a view and allocates a separate buffer instead. The kernel's results land in that separate buffer, but the function returns the original, still-uninitialized `output` tensor. The call succeeds with no error or warning and silently returns garbage.

## Why it matters

Any non-contiguous input to `torch.bucketize` under `flag_gems.use_gems()` — a transposed tensor, a strided slice, the output of another op — hits this. It's a wrong-answer bug, not a crash, so it's easy to miss in a broader pipeline.

## Verification

I confirmed the storage-aliasing mechanism directly against PyTorch on CPU: a transposed tensor's `torch.empty_like(...).flatten()` allocates a distinct `storage().data_ptr()` from the original. I do not have GPU access in my environment, so I could not run FlagGems' own Triton kernel end to end — the added test follows the existing `test_bucketize.py` conventions and should be verified by CI.

## Fix

One-line change: reshape `output_flat` (the buffer the kernel actually wrote into) instead of `output`.

## Test plan

Added `test_bucketize_noncontiguous_input`, parametrized the same way as the existing tests, using a transposed (non-contiguous) input and asserting equality against `torch.bucketize` on CPU/reference. Existing tests only ever construct inputs via `torch.randn(shape, ...)`, which is always contiguous, so this path wasn't previously covered.


w1120029931-bit (approved):
lgtm

ErenAta16:
Re-checked this against current `main` — still `clean`, no conflicts, and the four CI jobs that run for it are green. Nothing has drifted since @w1120029931-bit's approval.

No rush from my side, just flagging that it's ready whenever someone has a moment, and I'm around if anything needs changing.

PR #4833, opened 2026-07-16
[KernelGen][Nvidia] Add _linalg_svd operator with Triton kernel

Yukun-Cui:
## Summary
Adds a Triton kernel for `_linalg_svd`. Computes the singular value decomposition (SVD) of a matrix; the private
primitive backing torch.linalg.svd and torch.linalg.svdvals.

## Testing

## Performance
Test command: `pytest benchmark/test__linalg_svd.py --level core` (NVIDIA H20)

### _linalg_svd

| float32 | [8, 8] | 0.377712 | 0.660832 | 0.572 |
| float32 | [16, 16] | 0.219328 | 0.397280 | 0.552 |
| float32 | [32, 32] | 0.268016 | 0.572320 | 0.468 |
| float32 | [64, 64] | 1.158688 | 1.432896 | 0.809 |

| underscore_linalg_svd | **0.600** |

## Multi-backend Testing
|---|---|---|---|
| Nvidia (H20) | PASS (4 cases) | 0.600 | Primary |
| Tianshu | N/A | — | — |
| Muxi | N/A | — | — |
| Ascend | N/A | — | — |
| Hygon | N/A | — | — |

## Files Changed
- `src/flag_gems/ops/_linalg_svd.py`: Triton kernel implementation
- `tests/test__linalg_svd.py`: Accuracy test
- `benchmark/test__linalg_svd.py`: Performance benchmark
- `conf/operators.yaml`: Add operator entry (kind: BLAS, stage: alpha 5.4)


conf/operators.yaml
```diff
@@ -13,6 +13,19 @@
 # limitations under the License.
 
 ops:
+  - id: _linalg_svd
```
w1120029931-bit: should be underscore_linalg_svd ？
and please in order

tests/test__linalg_svd.py
```diff
@@ -0,0 +1,181 @@
...
+    with flag_gems.use_gems():
+        res_u, res_s, res_vh = torch.linalg.svd(inp, full_matrices=False)
+
+    _assert_same_shape(res_u, ref_u)
+    _assert_same_shape(res_s, ref_s)
+    _assert_same_shape(res_vh, ref_vh)
+
+    utils.gems_assert_close(res_s, ref_s, res_s.dtype, atol=1e-2)
+    reconstructed = _reconstruct(res_u, res_s, res_vh)
+    utils.gems_assert_close(reconstructed, ref_inp, reconstructed.dtype, atol=1e-2)
```
Caeruleann: The tests only cover shapes up to 16×16 with atol=1e-2 for singular values. Please add larger shapes and more strict tolerance checks.

Yukun-Cui: Addressed. Extended the shape coverage well beyond 16×16 — the shared list now includes 32×32, 64×64 and a non-square 32×16 (all within the full_matrices `max(m,n) <= 64` kernel limit), the reduced/svdvals paths additionally cover a 128×128 matrix, and the batched test adds a 4×32×32 case. Also tightened the tolerances: singular values now use `atol=1e-3` (10× stricter than the old `1e-2`), with a separate `5e-3` bound for the reconstruction `A = U diag(S) Vh`. These are backed by measurements on H20 — across all the new shapes singular-value error stays ≤ 4e-4 and reconstruction error ≤ 2e-3. Full suite passes.

conf/operators.yaml
Yukun-Cui: The operator was renamed to `_linalg_svd` (matching the aten schema `aten::_linalg_svd`), so the id/`for` now read `_linalg_svd`. On ordering: the entry sits at the top of the `ops:` list, which is where the underscore prefix sorts under the existing convention — on master the only other `_`-prefixed id (`_reshape_alias`) is likewise the first entry, since `_` sorts before `a` in ASCII. Let me know if you would prefer a different placement.

benchmark/test__linalg_svd.py
```diff
@@ -0,0 +1,36 @@
...
+    def set_shapes(self, shape_file_path=None):
+        self.shapes = LINALG_SVD_SHAPES
+
+    def get_input_iter(self, cur_dtype):
+        for shape in self.shapes:
+            inp = torch.randn(shape, dtype=cur_dtype, device=self.device)
+            yield (inp,)
+
+
+@pytest.mark.underscore_linalg_svd
```
w1120029931-bit, line 27: should it be changed to _linalg_svd marker?

src/flag_gems/__init__.py
w1120029931-bit, line 177: duplicate definition?

tests/test__linalg_svd.py
```diff
@@ -0,0 +1,211 @@
...
+# original 1e-2 bound (observed: singular values <= 4e-4, reconstruction <=
+# 2e-3 for shapes up to 128x128).
+SINGULAR_VALUE_ATOL = 1e-3
+RECONSTRUCTION_ATOL = 5e-3
+
+
+# aten::_linalg_svd is the private primitive that torch.linalg.svd (compute_uv
+# True) and torch.linalg.svdvals (compute_uv False) decompose into, so calling
+# the public APIs under use_gems() exercises the registered kernel.
+@pytest.mark.underscore_linalg_svd
```
w1120029931-bit, line 91: similarly, should it be changed to _linalg_svd marker?

Yukun-Cui: Pytest markers cannot start with `_`, so `underscore` is used instead.

benchmark/test__linalg_svd.py
Yukun-Cui: Pytest markers cannot start with `_`, so `underscore` is used instead.

src/flag_gems/__init__.py
Yukun-Cui: Great discovery! Previously, in order to correct the errors in the lexicographic order of other operators, they have now been corrected.

PR #4864, opened 2026-07-17
[SiliconFlow] Fix and optimize some operators on multi backends

winfan1314:
### PR Category

Operator, OP Test, Benchmark

### Type of Change

Bug Fix, Performance Optimization, Refactor

### Description

This PR fixes runtime and accuracy failures and optimizes several operator families on multi backends. 

### Forward PRs

`_scaled_mm` #5062 
`index_reduce_` #5063  
`_scaled_grouped_mm` #5064 
`segment_reduce`, `_segment_reduce_backward` #5065 
`upsample_linear1d_backward` #5066 
`nanmedian` #5067


winfan1314:
@tengqm @huangyiqun 
The failure in [`backend-tests (ascend-cann850)`](https://github.com/flagos-ai/FlagGems/actions/runs/30510210788/job/90768568328?pr=4864) does not appear to be caused by the `index_reduce_` implementation or an accuracy regression.

Related to [#4918](https://github.com/flagos-ai/FlagGems/issues/4918)

The job is configured as `ascend-cann850`. According to `src/flag_gems/backends.yaml`, this environment selects:

- `triton-ascend==3.2.0`
- `flagtree==0.6.0+ascend3.2`
- `torch==2.9.0+cpu`
- `torch-npu==2.9.0`

However, the CI log shows that the runner's active Ascend installation resolves to:

```text
/usr/local/Ascend/cann-9.0.0
```

The fatal error occurs while the Ascend Triton driver is being initialized:

```text
index_reduce_
  -> Triton JIT get_current_target()
  -> Ascend NPUUtils()
  -> _build_npu_ext("npu_utils")
  -> compile npu_utils.cpp
```

The compilation then fails with:

```text
error: ‘RT_LIMIT_TYPE_SIMT_WARP_STACK_SIZE’ is not a member of
       ‘rtLimitType_t’
```

This happens before the FlagGems `index_reduce_` kernel is compiled or executed. It indicates that the compiler package selected for the CANN 8.5 backend is being used with incompatible CANN 9.0 runtime headers. Any Ascend Triton operator that triggers the same driver initialization is likely to fail in this environment.

The following warning is unrelated to the fatal failure:

```text
aten::index_reduce.out is not currently supported on the NPU backend
and will fall back to run on the CPU
```

That warning comes from the native reference calculation, which completes successfully. Pytest is invoked with `-x`, so the subsequent Triton driver initialization failure stops the job at the first `index_reduce_` case and prevents the remaining Ascend tests, including `nanmedian`, from running.

Could the Ascend CI runner configuration please be checked and aligned in one of these ways?

1. Run `ascend-cann850` with `/usr/local/Ascend/cann` pointing to a compatible CANN 8.5 installation.
2. If the runner is intended to use CANN 9.0, select `ascend-cann900` together with its complete CANN 9.0 compiler, PyTorch and torch-npu dependency set.
3. Check whether `/usr/local/Ascend/cann`, `ASCEND_HOME_PATH`, or related environment variables retain values from another CANN installation.
4. Recreate the virtual environment after correcting the active CANN installation, to avoid reusing incompatible compiler packages or cached extensions.

PR #4890, opened 2026-07-20
fix: resolve complex conjugate recursion and vdot accuracy failures

CLYtou:
### PR Category
Operator

### Type of Change
Bug Fix

### Description

Fix infinite recursion in `resolve_conj`, crash in `_conj`/`conj_physical`, and 54 accuracy test failures in `vdot` for `torch.complex64` tensors with conjugate bit.

**Root Causes:**

1. **resolve_conj**: Called `torch.ops.aten.resolve_conj.default(A)` which dispatches back to FlagGems' own implementation → infinite recursion ("maximum recursion depth exceeded").

2. **_conj**: Did not handle input with conj bit already set → `torch.view_as_real(src)` crashes with "view_as_real doesn't work on unresolved conjugated tensors".

3. **vdot**: Used `.conj()` which dispatches to FlagGems `_conj` ; obtained strides before conj resolution (wrong after `_conj` returns contiguous); allocated `partial_real_sums` buffer too small (`grid_size` instead of `grid_size * 2`).

4. **conj_physical**: Same `view_as_real` vulnerability as `_conj` (defensive fix).

**Fixes:**

- **resolve_conj**: Use `A.clone()` instead of aten fallback. Safe because `clone() → copy_` redispatches for complex tensors via `CompositeExplicitAutograd`.
- **_conj**: Add `is_conj()` early return: `resolve_conj(input)` then `conj_physical(resolved)` to compute `conj(conj(x)) = x`.
- **vdot**: Use `torch.resolve_conj()` to materialize logical values, get strides after resolution, fix buffer size to `grid_size * 2`, set `inp_is_conj=False, other_is_conj=False` (kernel Case 1 applies conj to inp).
- **conj_physical**: Add defensive `is_conj()` check + `resolve_conj()` before `view_as_real`.

### Issue

- Resolves: `resolve_conj` benchmark crashes with `RecursionError: maximum recursion depth exceeded`
- Resolves: 54 `vdot` accuracy test failures (`RuntimeError: view_as_real` + `AssertionError: Scalar values mismatch`) for all `complex64` + conj scenarios

### Progress

### Performance

After this fix:
- `resolve_conj` benchmark completes without recursion error
- All 54 `vdot` complex64+conj test cases should pass (RuntimeError and precision mismatch both resolved)
- No performance regression for non-conjugated tensor paths (early return unchanged)


huangyiqun (approved):
LGTM.

PR #4921, opened 2026-07-22
[KMCompiler][Ascend] nonzero static ascend

chenkc2025:
## Summary

- Add Triton implementations of `nonzero_static` and `nonzero_static.out`.
- Support NVIDIA and Ascend backends.
- Preserve fixed output shape, C-order index ordering, truncation, padding, scalar, empty, and non-contiguous input semantics.
- Use backend-appropriate references and baselines:
  - NVIDIA correctness and performance use `torch.nonzero_static`.
  - Ascend uses a PyTorch composition of `torch.full`, `torch.nonzero`, and `copy_`, because `torch_npu` does not provide a native `nonzero_static` dispatch kernel.

## Implementation highlights

- Reuse the common Triton implementation with backend-specific launch configuration.
- Support boolean, integer, floating-point, bfloat16, and complex inputs with `int64` output.
- Cover single-block, small-count, linear-write, sparse-group, prefix-scan, generic-dimension, and fill-tail paths.
- Reduce the Ascend coordinate expansion block size to avoid CANN 9.0 UB overflow during kernel compilation.
- Use the PyTorch-composed baseline for Ascend benchmark and correctness reference paths.
- Remove the runtime C++ extension and `load_inline` dependency from the benchmark.
- Keep generic correctness tests available to all backends.
- Keep only CUDA- and Ascend-specific skips for backend-specific test cases.
- Register and export both `nonzero_static` and `nonzero_static.out`.

## Testing

Tested on Ascend 910B4 with CANN 8.5.0 and `torch_npu 2.9.0`.

    python -m py_compile \
      src/flag_gems/ops/nonzero_static.py \
      src/flag_gems/runtime/backend/_ascend/ops/nonzero_static.py \
      tests/test_nonzero_static.py \
      benchmark/test_nonzero_static.py

    pre-commit run --all-files

    ASCEND_RT_VISIBLE_DEVICES=1 \
    pytest -q -s tests/test_nonzero_static.py

The Ascend correctness tests use the PyTorch composition reference:

    torch.full
    torch.nonzero
    copy_

The CANN 9.0 compilation failure caused by UB overflow was addressed by reducing the Ascend coordinate expansion block size.

## Performance Results

On Ascend, the baseline is not `torch.nonzero_static`, because `torch_npu` does not provide a native NPU implementation and would fall back to CPU. The baseline instead composes `torch.full`, `torch.nonzero`, and `copy_` on the target device.

The following BF16 results are from one preliminary run on Ascend 910B4 with CANN 8.5.0. Final PR numbers should be updated with three-run medians using the same baseline.

| Case | Shape | Nonzero ratio | Size | Speedup |
| ---: | --- | ---: | ---: | ---: |
| 1 | `(1024,)` | 0.0 | 128 | 0.782x |
| 2 | `(1024,)` | 0.1 | 128 | 1.985x |
| 3 | `(16384,)` | 0.001 | 128 | 1.012x |
| 4 | `(16384,)` | 0.1 | 1024 | 0.985x |
| 5 | `(262144,)` | 0.001 | 1024 | 1.022x |
| 6 | `(262144,)` | 0.1 | 4096 | 1.044x |
| 7 | `(1048576,)` | 0.001 | 1024 | 0.768x |
| 8 | `(1048576,)` | 0.1 | 4096 | 1.123x |
| 9 | `(32, 1024)` | 0.1 | 1024 | 1.074x |
| 10 | `(128, 4096)` | 0.01 | 4096 | 0.845x |
| 11 | `(1024, 4096)` | 0.001 | 1024 | 0.980x |
| 12 | `(1024, 4096)` | 0.1 | 4096 | 2.494x |
| 13 | `(16, 64, 64)` | 0.1 | 1024 | 1.579x |
| 14 | `(32, 128, 128)` | 0.01 | 4096 | 1.227x |

The arithmetic mean of this preliminary BF16 run is `1.209x`. FP16 results should be remeasured after the baseline change and reported separately.


benchmark/test_nonzero_static.py
```diff
@@ -26,7 +93,25 @@
...
+
+    # Keep the composition behind one C++ custom-op boundary so benchmark
+    # timing does not include Python-side composition overhead.
+    # This intentionally requires the C++ extension toolchain on benchmark hosts.
+    load_inline(
+        name="flag_gems_ascend_aten_nonzero_static_baseline",
+        cpp_sources=ASCEND_ATEN_BASELINE_SOURCE,
+        extra_cflags=["-O3"],
+        is_python_module=False,
+        verbose=os.getenv("FLAGGEMS_ASCEND_ATEN_BASELINE_VERBOSE", "0") == "1",
```
Caeruleann: who used FLAGGEMS_ASCEND_ATEN_BASELINE_VERBOSE ？

chenkc2025: It was only an optional local debug switch for the temporary load_inline C++ baseline. The C++ baseline and FLAGGEMS_ASCEND_ATEN_BASELINE_VERBOSE have been removed in commit 3ee84bd29.

Caeruleann (approved):

PR #4924, opened 2026-07-23
Fix adaptive_avg_pool2d wrong result on non-divisible ratios (floor vs ceil window end)

### LINKED ISSUE #4915: adaptive_avg_pool2d: window end computed with floor instead of ceil, wrong results for non-divisible input/output sizes
## Summary

`adaptive_avg_pool2d` returns wrong values whenever the input size is not an exact multiple of the output size. The kernel computes the end of each pooling window with **floor**, but PyTorch (ATen) uses **ceil**. As a result, most windows are one element too short, and the average is also divided by the wrong window size. Errors are large (up to ~1.0 on standard normal inputs), not a precision issue.

## Reproduction

```python
import torch
import torch.nn.functional as F
import flag_gems

torch.manual_seed(42)
x = torch.randn(2, 3, 10, 10, device=flag_gems.device)

ref = F.adaptive_avg_pool2d(x.double(), (3, 3)).float()
out = flag_gems.adaptive_avg_pool2d(x, (3, 3))

print((out - ref).abs().max())  # 0.6078 — should be ~1e-7
```

Any 4D input with a non-divisible ratio triggers it, e.g. `224x224 -> 13x13` (max error ~0.07 on the same seed). Exactly divisible ratios (`224 -> 7`, `32 -> 8`, ...) are unaffected, which is why the current test suite passes: every config in `tests/test_adaptive_avg_pool2d.py` uses exactly divisible sizes.

## Root cause

PyTorch defines the pooling window for output index `i` as:

```
start = floor(i * in / out)
end   = ceil((i + 1) * in / out)
```

The kernel computes the end with floor instead
([src/flag_gems/ops/adaptive_avg_pool2d.py#L69-L80](https://github.com/flagos-ai/FlagGems/blob/ac4e5a3dbcc2/src/flag_gems/ops/adaptive_avg_pool2d.py#L69-L80)):

```python
# i_end = floor((i + 1) * in_h / out_h)
h_end = ((h_out_offsets[:, None] + 1) * in_h) // out_h
...
w_end = ((w_out_offsets[None, :] + 1) * in_w) // out_w

# For the last element, extend to the end of input
h_end = tl.where(h_out_offsets[:, None] == out_h - 1, in_h, h_end)
```

The last-element fixup only patches the final index. Every other index whose `(i+1)*in` is not divisible by `out` gets a window that is one element too short, and `window_size` (the divisor) shrinks with it.

Concrete example, `in=10, out=3` (rows):

| output row | PyTorch window | kernel window |
|---|---|---|
| 0 | `[0, 4)` — 4 elements | `[0, 3)` — 3 elements |
| 1 | `[3, 7)` — 4 elements | `[3, 6)` — 3 elements |
| 2 | `[6, 10)` — 4 elements | `[6, 10)` — 4 elements (fixup) |

## Fix

Three changes are needed, not just one:

1. Compute the window end with the ATen ceil form:
   ```python
   h_end = 1 + ((h_out_offsets[:, None] + 1) * in_h - 1) // out_h
   w_end = 1 + ((w_out_offsets[None, :] + 1) * in_w - 1) // out_w
   ```
2. Remove (or enlarge by 1) the `kh < max_win_h` / `kw < max_win_w` masks. Under the ceil rule a window can be one element **longer** than `cdiv(in, out)` — e.g. `in=5, out=3` gives the middle window `[1, 4)`, size 3, while `cdiv(5,3) = 2`. With the masks left as-is, they clip that element and the result is still wrong (verified: max error 0.199 on a `5x5 -> 3x3` case). The masks are redundant anyway, because `h_in < h_end` already bounds every window.
3. The last-element fixup (`tl.where(... == out_h - 1, in_h, ...)`) becomes dead code — the ceil form already yields exactly `in_h` for the last index — and can be removed.

I have this fix working locally (all existing tests plus new non-divisible configs pass, max error vs ATen ~1e-7 in fp32) and can open a PR.

## Environment

- FlagGems master (`ac4e5a3d`); the kernel is unchanged since it was added in #2059
- torch 2.6.0+cu124, triton 3.2.0, NVIDIA H100, CUDA 12.4

zheng1:
## PR Category

ops

## Type of Change

Bug Fix

## Description

Fixes #4915.

`adaptive_avg_pool2d` returned wrong values on non-divisible input/output ratios. The Triton kernel computed the pooling window end with floor, but PyTorch/ATen uses ceil, so most windows were one element short and the average was divided by the wrong window size (errors up to ~1.0 on standard-normal inputs, not a precision issue).

The fixed per-window tile-loop bound `max_win_h`/`max_win_w` also needed `+1`: with a ceil end, a window can be one row/col taller than `ceil(in/out)`, so the old bound could drop the last element of some windows.

## Tests

Added `test_accuracy_adaptive_avg_pool2d_non_divisible` with 7 non-divisible configs, including the issue's `10 -> 3` and `224 -> 13` repros and a `5 -> 3` case that exercises the taller window. Every pre-existing config used exactly divisible sizes, so this path was never covered.

Verified on a CI-aligned A100 (torch 2.11 / triton 3.6): the new cases fail before the fix and pass after (default mode 54 passed, quick-cpu 18 passed).

## Known limitation (pre-existing, out of scope)

The inner accumulation loop is hard-bounded at 64 (`for k in range(64)`), so an extreme downsample ratio `in/out` greater than ~63 (e.g. `191 -> 3`, window 65) can still truncate the window. This limit is independent of this fix and is not triggered by normal adaptive-pool usage.

### DIFF
--- a/src/flag_gems/ops/adaptive_avg_pool2d.py
+++ b/src/flag_gems/ops/adaptive_avg_pool2d.py
@@ -68,12 +68,12 @@ def adaptive_avg_pool2d_kernel(
 
     # Adaptive pooling: compute adaptive windows
     # i_start = floor(i * in_h / out_h)
-    # i_end = floor((i + 1) * in_h / out_h)
+    # i_end = ceil((i + 1) * in_h / out_h)
     h_start = h_out_offsets[:, None] * in_h // out_h
-    h_end = ((h_out_offsets[:, None] + 1) * in_h) // out_h
+    h_end = ((h_out_offsets[:, None] + 1) * in_h + out_h - 1) // out_h
 
     w_start = w_out_offsets[None, :] * in_w // out_w
-    w_end = ((w_out_offsets[None, :] + 1) * in_w) // out_w
+    w_end = ((w_out_offsets[None, :] + 1) * in_w + out_w - 1) // out_w
 
     # For the last element, extend to the end of input
     h_end = tl.where(h_out_offsets[:, None] == out_h - 1, in_h, h_end)
@@ -92,8 +92,8 @@ def adaptive_avg_pool2d_kernel(
     input_base_ptr = input_ptr + n_idx * in_stride_n + c_idx * in_stride_c
 
     # Find the maximum window size
-    max_win_h = (in_h + out_h - 1) // out_h
-    max_win_w = (in_w + out_w - 1) // out_w
+    max_win_h = (in_h + out_h - 1) // out_h + 1
+    max_win_w = (in_w + out_w - 1) // out_w + 1
 
     # Loop over all possible positions in the max window
     for kh in range(64):
--- a/tests/test_adaptive_avg_pool2d.py
+++ b/tests/test_adaptive_avg_pool2d.py
@@ -63,3 +63,35 @@ def test_accuracy_adaptive_avg_pool2d_forward(shape, output_size, dtype):
     res_out = flag_gems.adaptive_avg_pool2d(inp, output_size)
 
     utils.gems_assert_close(res_out, ref_out, dtype)
+
+
+# Regression for issue #4915: adaptive_avg_pool2d gave wrong values on
+# non-divisible ratios because the pooling window end was computed with floor
+# instead of PyTorch's ceil, and the fixed tile-loop bound was one row/col too
+# small. Every config above uses exactly divisible sizes, so this path was
+# never covered. Windows here stay well under the kernel's 64-element loop cap.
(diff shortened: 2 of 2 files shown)


zheng1:
Superseded by #4952, which landed the same fix and covers the same cases.

The window end formula there, `1 + ((i + 1) * in - 1) // out`, is mathematically
identical to the one here, `((i + 1) * in + out - 1) // out`. Both equal
`ceil((i + 1) * in / out)`, which is the ATen rule. #4952 also added
non-divisible configs to the main parametrization, including both repros from
issue #4915 (10/3 and 224/13), so the regression coverage is already in place.

Closing as redundant.

PR #4931, opened 2026-07-23
fix: replace triton_src kernel references with ops/ paths in C++ wrappers

yysheng26:
### PR Category
Operator

### Type of Change
Bug Fix

### Description

Three C++ operator wrappers (`contiguous`, `zeros`) were loading Triton kernels from the `triton_src/` directory via `get_triton_src_path()`, instead of from `src/flag_gems/ops/` via `get_flag_gems_src_path()`. 

Changes:

- **zeros.cpp**: Replaced `get_triton_src_path() / "zeros.py"` with `get_flag_gems_src_path() / "ops" / "zeros.py"`.

- **contiguous.cpp**: Removed the direct kernel launch entirely. The C++ wrapper now calls `copy_()` to perform the data movement, aligning with the Python implementation in `ops/contiguous.py` which delegates to `copy_` as well.

### Issue

N/A

### Progress

### Performance

No performance regression expected.


src/flag_gems/ops/cat.py
```diff
@@ -59,45 +59,45 @@ def cat_copy_func_kernel_4(
     in_ptr_b,
     in_ptr_c,
     in_ptr_d,
-    dim_size_in_a,
-    dim_size_in_b,
-    dim_size_in_c,
-    dim_size_in_d,
-    dim_size_out,
-    dim_prod_post,
+    dim_size_in_a: tl.int64,
```
0x45f: why use tl.int64

yysheng26: the Triton compiler drops parameter type info inside branches, so 0 is inferred as int32 and non-zero as int64, causing a type mismatch between branch paths

0x45f (approved):

PR #4932, opened 2026-07-23
fix(mthreads): preserve bmm output dtype

yilongma0110:
fix(mthreads): preserve input dtype and device in bmm output

The MThreads bmm implementation created FP16 output for BF16
inputs, which could cause mixed BF16/FP16 operands during backward.
Preserve torch.bmm dtype semantics by allocating the output with
the input element type and device.


Kylin1207:
LGTM

PR #3418, opened 2026-05-21
[CI/CD] Add Debian + RPM packaging

shiptux:
### PR Category

CI/CD

### Type of Change

New Feature

### Description

Adds Debian (`.deb`) and RPM (`.rpm`) packaging configuration so FlagGems can
be distributed via standard `apt install` / `dnf install` alongside the rest
of the FlagOS stack.

**Rebased onto the revised source structure** (per review): the packaging now
follows upstream's split between the Python-only wheel and the per-vendor
cpp-op wheels —

- The Python wheel is built with the plain `setuptools.build_meta` backend
  from `src/` (pure Python, `py3-none-any`), with the version resolved via
  setuptools-scm (`SETUPTOOLS_SCM_PRETEND_VERSION` pinned in the container
  builds, no `.git` in the build context).
- `setup.py` maps `tests/` → `flaggems_tests` and `benchmark/` →
  `flaggems_benchmark`, so the operator test and benchmark suites ship in the
  packages (upstream #4832).
- The `flaggems-setup` console script is installed to `/usr/bin`.
- The C++ wrapped operators live in the separate `cpp/` tree upstream and are
  distributed as per-vendor `flag-gems-cpp-*` native wheels; packaging those
  needs per-vendor toolchains and is **deferred to Phase 2** (stanzas
  reserved in `debian/control`).

**Phase 1 scope — Python distribution only.** Mirrors upstream's default
`pip install flag_gems` behavior; no native code in the packages, so the deb
is `Architecture: all` and the rpm is `noarch`.

**Produced binaries (5.3.0, rebuilt against current master):**

- `python3-flag-gems_5.3.0-1_all.deb`
- `python3-flag-gems-5.3.0-1.fc43.noarch.rpm`

14 vendor backend directories ship as Python source under
`flag_gems/runtime/backend/`; the active backend is selected at runtime via
the `GEMS_VENDOR` env var or PyTorch auto-detect.

**What changed (all under `packaging/` + CI workflows, no source changes):**

- `packaging/debian/{control,rules,changelog,copyright,source/format}` —
  Debian packaging via `dh-python`; the wheel is built in stage 1 of
  `Dockerfile.deb` and unpacked with `pip --target` in stage 2.
- `packaging/rpm/specs/flag-gems.spec` — RPM spec (fc43,
  `pyproject-rpm-macros`, noarch); `%check` runs a
  `importlib.util.find_spec('flag_gems')` smoke test with `PYTHONSAFEPATH`.
[...]

PR #4939, opened 2026-07-23
fix: add missing __init__.py so fused/DSA and seven backend subpackages ship in built packages

shiptux:
## Problem

`flag_gems/fused/__init__.py` imports the DSA subpackage unconditionally:

```python
from flag_gems.fused.DSA.bin_topk import bucket_sort_topk
```

but `src/flag_gems/fused/DSA/` has no `__init__.py`, and `setup.py` discovers
packages with `find_packages("src")`, which only picks up directories that
contain one. The subpackage is therefore silently dropped from every built
artifact — while its tests (`flaggems_tests/test_DSA/`) are still shipped.

Running from a source checkout works because Python treats the directory as a
namespace package, so in-tree CI never sees this. Any install from a wheel or
distro package fails at import time:

```
>>> import flag_gems
ModuleNotFoundError: No module named 'flag_gems.fused.DSA'
```

This makes the package completely unusable, not merely degraded — the failure
is at top-level `import flag_gems`.

## Reproduction

Found while validating the `python3-flag-gems` 5.3.0 Debian package on
Ubuntu 22.04 (NVIDIA A100). The package installs cleanly (`apt-get install`
exits 0, `dpkg -l` shows `ii`) and then crashes on import.

Mechanism, before and after this change:

```python
>>> from setuptools import find_packages
>>> [p for p in find_packages("src") if "DSA" in p]
[]                          # master
['flag_gems.fused.DSA']     # with this patch
```

After patching the module into the installed package, `import flag_gems`
succeeds and the operator suite passes (1929 tests across add / mul / relu /
silu / gelu / softmax / sum / mm, plus `use_gems` numerics verified in fp16 and
bf16). So the code itself is fine — only packaging was affected.

## Fix

Add an `__init__.py` to `fused/DSA/`, matching how the sibling `fused/FLA/`
directory is already handled. Minimal change, no behavioral difference in a
source checkout.

## Also fixed: seven more subpackages with the same defect

A packaging smoke check (`packaging/verify-installed-package.py` in #3418, which
diffs source subpackages against the installed wheel) turned up the same pattern
in seven more directories. Each one is imported unconditionally by its parent
package, so every affected backend hits `ModuleNotFoundError` when installed
from a wheel:

| Directory | Imported by |
|---|---|
| `runtime/backend/_enflame/fused/gcu300/` | `_enflame/fused/__init__.py` |
| `runtime/backend/_enflame/fused/gcu400/` | `_enflame/fused/__init__.py` |
| `runtime/backend/_hygon/utils/` | `_hygon/ops/*.py` |
| `runtime/backend/_kunlunxin/utils/` | `_kunlunxin/fused/*.py`, `ops/*.py` |
| `runtime/backend/_spacemit/utils/` | `_spacemit/__init__.py` |
| `runtime/backend/_sunrise/fused/DSA/` | `_sunrise/fused/__init__.py` |
| `runtime/backend/_tsingmicro/utils/` | `_tsingmicro/ops/*.py` |

Same one-line fix (license-header-only `__init__.py`, matching the sibling
packages). With this patch `find_packages("src")` reports all of them, and no
directory under `src/flag_gems` containing `.py` files is left without an
`__init__.py`. I cannot run the vendor hardware, but the failure mode is
import-time and independent of the device, so the fix needs no hardware to
verify.

Rebased onto current `master`.

## Suggestion

An `python -c "import flag_gems"` smoke step after packaging would catch this
entire class of bug in seconds; `apt install` + `dpkg -l` alone cannot.


w1120029931-bit (dismissed):

shiptux:
@w1120029931-bit Rebased onto master and extended the fix to seven more subpackages with the identical defect (table in the PR body). Sorry for invalidating your approval — could you take another look and approve the workflow run?

0x45f (approved):

PR #4952, opened 2026-07-24
[SiliconFlow] Operator fix adaptive_avg_pool2d

### LINKED ISSUE #4915: adaptive_avg_pool2d: window end computed with floor instead of ceil, wrong results for non-divisible input/output sizes

fpzh2011:
### PR Category
Operator , OP Test

### Type of Change
Bug Fix

### Description
Fixed the adaptive_avg_pool2d op bug based on issue #4915

### Issue

- Resolves #4915 

### Progress


huangyiqun (approved):
LGTM.

PR #4959, opened 2026-07-26
fix(config): lazy import c_operators — detect .so without loading

tengqm:
## Problem

`src/flag_gems/config.py` eagerly imports `c_operators` at module-init time.
This triggers pybind11 initialization, which conflicts with vendor torch
packages (torch_npu on Ascend) — any subsequent NPU tensor operation segfaults
(exit code 139).

## Repro

- Build + install cpp wheel → `.so` lands in `flag_gems/`
- `import flag_gems` → config.py loads c_operators → `has_c_extension=True`
- `use_gems()` → any NPU tensor op → segfault

Without the `.so` present: everything works fine.

## Fix

Replace the eager `try/except ImportError` with a file-existence-only check
(`Path.glob("c_operators*.so")`). The actual pybind11 import is deferred
until `USE_C_EXTENSION=1` AND the `.so` exists.

## Verified

Tested on ascend-cann9.0.0 (hw25):
- `.so` present, `USE_C_EXTENSION` unset → pure Python path works (no .so loaded)
- Caveat: with `USE_C_EXTENSION=1` and .so loaded, tensor ops still crash —
  root cause TBD, but the lazy import is the correct first step

### DIFF
--- a/src/flag_gems/config.py
+++ b/src/flag_gems/config.py
@@ -59,6 +59,7 @@ if use_env_c_extension and _has_cpp_so:
         use_c_extension = True
     except (ImportError, AttributeError):
         aten_patch_list = []
+        has_c_extension = False
         use_c_extension = False
 
 
--- a/src/flag_gems/config.py
+++ b/src/flag_gems/config.py
@@ -32,25 +32,27 @@ aten_patch_list = []
 # set FLAGGEMS_SOURCE_DIR for cpp extension to find
 os.environ["FLAGGEMS_SOURCE_DIR"] = str(Path(__file__).parent.resolve())
 
-try:
-    from flag_gems import c_operators
-
-    has_c_extension = True
-except ImportError:
-    c_operators = None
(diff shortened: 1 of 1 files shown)


src/flag_gems/config.py
```diff
@@ -32,25 +32,27 @@
...
         RuntimeWarning,
     )
 
-if has_c_extension and use_env_c_extension:
+if use_env_c_extension and _has_cpp_so:
     try:
+        from flag_gems import c_operators as _co
+
+        c_operators = _co
+        has_c_extension = True
```
0x45f, line 55: we should set `has_c_extension = False` in except block

tengqm: Double checked the logic. `has_c_extension` initialized to `False` on line 28, and it wasn't touched so far.

Anyway, adding an explicit reassignment makes the logic clearer. Will do.

0x45f (approved):

PR #5040, opened 2026-07-29
feat(cpp): add 10 unary elementwise ops to the C++ pointwise_dynamic path

lvyufeng:
## What

Adds **abs, neg, exp, sqrt, rsqrt, tanh, sigmoid, silu, relu, gelu** to the C++
`operators` runtime via the existing `pointwise_dynamic` C++ path, following the
same pattern already established for `add`/`div`/`fill`.

## Why

The C++ pointwise generator (`cpp/lib/pointwise_dynamic_cpp/prebuild_kernels.py`)
already auto-discovers any `@pointwise_dynamic` scalar function and generates the
Triton kernels + C++ wrappers. These 10 common unary ops are all `@pointwise_dynamic`
but were not yet exposed through the C++ operators API. Exposing them lets host-bound
eager workloads dispatch these ops in C++ (no GIL, no per-op Python) instead of the
Python path.

## Changes

- `include/flag_gems/operators.h` — declare the 10 ops under `FLAGGEMS_POINTWISE_DYNAMIC`
- `lib/<op>.cpp` — thin wrappers calling `pointwise_dynamic::<fn>`; `gelu` dispatches
  `gelu_none`/`gelu_tanh` on the `approximate` argument
- `lib/pointwise_dynamic_cpp/CMakeLists.txt` — add the 10 op modules to
  `POINTWISE_OP_FILES`. Also adds `flag_gems.ops.remainder`, which `lib/div.cpp`
  already references (`rem_tt`/`rem_ts`/`rem_st`) but was missing from the prebuild
  list — the pointwise C++ build did not link with `FLAGGEMS_BUILD_POINTWISE_DYNAMIC_CPP=ON`
  until this is added
- `lib/CMakeLists.txt` — compile the new sources
- `csrc/cstub.cpp` — pybind bindings + `TORCH_LIBRARY` schema + `TORCH_LIBRARY_IMPL`
- `ctests/test_triton_unary.cpp` — gtest coverage

### Codegen fix: module-level shim aliases

`tanh`/`sigmoid`/`gelu` use module-level aliases to libdevice-backed helpers, e.g.
`_tanh = tl_extra_shim.tanh` (sigmoid: `exp2`; gelu: `erf`/`exp`/`pow`/`tanh`). The
pointwise codegen inlines the scalar function's source into the generated kernel
`.py`, but `PointwiseDynamicFunction._collect_jit_deps` only reproduced
`from X import Y` statements — not `name = module.attr` alias assignments. Those
aliases were therefore undefined when the generated file is compiled in a fresh
namespace, which is what the standalone C++ TritonJIT path does
(`spec_from_file_location(...).exec_module`). The in-process Python path survives
because Triton captures the original op module's globals.

This PR makes `_collect_jit_deps` also collect module-level alias assignments whose
value is an attribute access on an imported name, and force-import their base module
name (e.g. `tl_extra_shim`) into the generated file. The Python path is behavior-
preserving (aliases were already resolved in-process).

## Testing

All 10 ops verified against ATen references through the `c_operators` bindings on an
A100 (CUDA), `max_err <= 6e-6`, including `gelu` none/tanh and fp16/bf16. `add`/`div`/
`remainder` unaffected. `ctests/test_triton_unary.cpp` added for CI once
`FLAGGEMS_BUILD_POINTWISE_DYNAMIC_CPP=ON` builds are exercised.


lvyufeng:
Closing in favor of a follow-up that generates the C++ dispatch glue (lib/*.cpp + cstub schema/impl) from op metadata via codegen, rather than hand-writing per-op wrappers. Will reopen a new PR built on that approach.

PR #5063, opened 2026-07-30
[SiliconFlow] Fix index_reduce_ on Ascend, Iluvatar, and MetaX

winfan1314:
### PR Category

Operator, OP Test, Benchmark

### Type of Change

Bug Fix, Refactor

### Description

Fix `index_reduce_` correctness and compilation issues on Ascend, Iluvatar, and MetaX while moving vendor-specific routing out of the generic implementation.

- Refactor the generic path to expose a shared implementation with optional CAS handling and remove inline Ascend/Iluvatar vendor checks.
- Add an Ascend override that computes reductions with the FP32 scan path, avoiding host-side uniqueness checks and the invalid unique-kernel launch.
- Add an Iluvatar override that uses the non-atomic kernel for unique indices and the generic collision-aware path with forced CAS for duplicate indices.
- Add a MetaX override that handles BF16 `mean` in FP32, scans duplicate-index `prod`, and uses the non-atomic kernel for unique-index `prod`.
- Export the three backend overrides without changing public registration or operator inventory.
- Update accuracy tests to call `flag_gems.index_reduce_` directly and add shared unique-index coverage for all reduction modes and both `include_self` values.
- Skip Ascend benchmarks because the native `aten::index_reduce.out` baseline falls back to CPU.

### Issue

- Related implementation: #3709
- Split from the cross-backend validation work in #4864

### Progress

### Performance

N/A. This PR is correctness-focused; the Ascend benchmark is skipped because no comparable native device baseline is available.


bin913:
This PR only uses generic kernel calls without adding platform-specific kernels for each target platform. Could you share the reasoning? As far as I know, the backend can reuse existing kernels without adding new files there.

winfan1314:
@bin913 You are right that a backend wrapper is unnecessary when it behaves identically to the generic entry point. In this PR, however, the backend files do not duplicate the Triton kernels; they `override the host-side routing policy` while intentionally `reusing the shared kernels`.  
 - Ascend routes all cases through the scan kernel with an FP32 workspace.
 - Iluvatar routes unique indices through the non-atomic unique kernel.
 - MetaX applies that routing specifically to unique-index prod.
 
We reuse the existing kernels because the issue is `kernel selection` and `atomic/CAS compatibility`, rather than `missing platform-specific computation`. An alternative would be to refactor these policies into `capability-based routing` in the generic implementation.
Well, it is technically possible to add backend-specific branches to the generic `index_reduce_` entry point. However, the three platforms require materially different routing policies, and adding more vendor checks to the generic path would couple the shared implementation to backend-specific limitations and make it increasingly difficult to maintain and test.
We consider a backend specialization to include not only a new platform-specific Triton kernel, but also platform-specific preprocessing, dtype conversion, kernel selection, and fallback behavior. Therefore, these backend implementations intentionally reuse the existing shared kernels while keeping the substantially different orchestration logic isolated in each backend.
If a backend implementation only delegated to the generic entry point without changing behavior, we would remove it. If these policies can later be represented by a common vendor-neutral capability rule, we can refactor that shared rule back into the generic implementation.

Caeruleann:
fix CI failed

modao1234 (dismissed):

modao1234:
@winfan1314 Sorry, CI env still has problem. 
But plz dix conflicts.

winfan1314:
> @winfan1314 Sorry, CI env still has problem. 
> 
> But plz dix conflicts. 

@modao1234 OK.

winfan1314:
> @winfan1314 Sorry, CI env still has problem. But plz dix conflicts.

@modao1234 resolved.

modao1234 (approved):

PR #4116, opened 2026-06-16
[SiliconFlow] Fix nanmedian cross-backend accuracy

winfan1314:
## Summary
- Fix `nanmedian` NaN detection to return a consistent boolean type across Triton backends.
- Restrict NVIDIA-specific CUDA radix / histogram fast paths to the NVIDIA backend instead of all CUDA-like devices.
- Route CUDA-like non-NVIDIA backends through the safe fallback path to avoid incorrect flat multi-dimensional results.
- Fix empty integer flat `nanmedian` to return the dtype minimum value, matching PyTorch behavior.

Fixes internal issues #571, #575, #577, and #579.

## Test plan
- [x] `python -m py_compile src/flag_gems/ops/nanmedian.py`
- [x] `python -m isort --check-only src/flag_gems/ops/nanmedian.py`
- [x] `python -m black --check src/flag_gems/ops/nanmedian.py`
- [x] `python -m flake8 --ignore=F405,E731,W503,E203 --max-line-length=120 src/flag_gems/ops/nanmedian.py`
- [ ] Backend accuracy tests for MetaX  -- network error: nanmedian_dim
- [ ] Backend accuracy tests for MThreads 
- [x] Backend accuracy tests for Hygon
- [ ] Backend accuracy tests for Kunlunxin -- skipped
- [x] Backend accuracy tests for iluvatar
- [x] CUDA / Ascend accuracy and benchmark regression

## Issue
- WEEKTEST-123
- WEEKTEST-188

PR #3337, opened 2026-05-18
[SiliconFlow] Add nanmedian operators

winfan1314:
### PR Category
Operator / OP Test / Benchmark

### Type of Change
New Feature / Performance Optimization

### Description
This PR adds FlagGems support for `torch.nanmedian`.

Changes included:
- Add `src/flag_gems/ops/nanmedian.py` with Triton implementations for `nanmedian`, `nanmedian.out`, `nanmedian.dim`, and `nanmedian.dim_values`.
- Support CUDA flat reductions with radix select, and CUDA dim reductions with per-row radix select.
- Gate CUDA radix histogram use by Triton version: Triton 3.6+ keeps the masked `tl.histogram` fast path, while Triton/Flagtree 3.5 uses a manual bin-count fallback for CI compatibility.
- Support Ascend flat reductions, Ascend floating dim reductions, and Ascend integer dim reductions with histogram-based select paths.
- Support mainstream Torch-compatible dtypes covered by the tests: `fp16`, `fp32`, `bf16` on CUDA, and integer dtypes including `int8`, `uint8`, `int16`, and `int32`; Ascend follows the available Torch/NPU reference support.
- Register `nanmedian` operators in `conf/operators.yaml` and FlagGems Python exports.
- Add accuracy coverage in `tests/test_nanmedian.py`.
- Add benchmark coverage in `benchmark/test_nanmedian.py`.

### Issue

N/A

### Progress

- [ ] Change is properly reviewed (1 reviewer required, 2 recommended).
- [ ] Change is responded to an issue.
- [ ] Change is fully covered by a UT.

#### Overall
CUDA performance is positive across all tested `nanmedian` interfaces and dtypes. All four benchmark interfaces show speedups on NVIDIA H20.

Ascend performance is mixed:
- Ascend integer `dim` reductions are strongly positive, especially for medium and large shapes.
- Ascend `dim_values` is mostly positive, but the Torch baseline emits an NPU CPU fallback warning for `aten::nanmedian.dim_values`, so these numbers should be interpreted with that caveat.
- Ascend floating `dim` and flat floating reductions remain slower than Torch native ACLNN paths. The implementation avoids fallback and keeps correctness coverage, but Torch's fused/native `aclnnNanMedianDim` path is still faster.

#### Accuracy coverage
Accuracy tests cover:
- Flat `torch.nanmedian(input)`.
- Flat `torch.nanmedian(input, out=...)`.
- Dim reduction `torch.nanmedian(input, dim, keepdim)`.
[...]

PR #5067, opened 2026-07-30
[SiliconFlow] Fix and optimize nanmedian across backends

winfan1314:
### PR Category

Operator, OP Test, Benchmark

### Type of Change

Bug Fix, Performance Optimization, Refactor

### Description

Fix and optimize all four `nanmedian` overloads across Ascend, Hygon, Iluvatar, MetaX, and MThreads while keeping backend policy out of the generic implementation.

- Move Ascend-specific kernels and routing from `src/flag_gems/ops/nanmedian.py` into the Ascend backend.
- Add and export backend implementations of `nanmedian`, `nanmedian.out`, `nanmedian.dim`, and `nanmedian.dim_values` for all five backends.
- Use backend-appropriate selection paths: software-gated native sort/topk plus integer histogram/radix kernels on Ascend; short-row sort or static radix and chunked radix for longer rows on Hygon and MetaX; tiered local/radix selection with narrowly scoped native `kthvalue` redispatch on Iluvatar; and local selection, radix selection, or the backend sort fallback on MThreads.
- Fix generic zero-output dimension handling and preserve caller-provided non-contiguous `values` and `indices` views through contiguous temporary buffers and copy-back.
- Preserve lower-median, first-matching-index, all-NaN, scalar, empty-input, and out-object identity semantics.
- Update tests to call the public `flag_gems` APIs and add regressions for large flat reductions, the Hygon block-limit case, and non-contiguous `dim_values` outputs.
- Keep the existing four benchmark nodes. Only Ascend `nanmedian.dim_values` is skipped because its native baseline falls back to CPU.

Public registration and operator inventory are unchanged.

### Related PRs

- #3337 introduced the generic implementation.
- #4116 and #4384 contain related generic NaN/radix and fallback work.
- #4864 is the original aggregate cross-backend validation PR from which this work was split.

### Progress

### Performance

The existing `benchmark/test_nanmedian.py` suite covers the four overloads. Backend-specific selection paths avoid unsupported full-row operations and reduce long-reduction overhead; raw per-device benchmark output is omitted from this description.


tengqm:
The time cost means something ...

The same benchmark takes about 6 minutes on Hygon, 23 minutes on CANN8.5.0, and it failed in 2 minutes on MetaX, failed in 11 minutes on Iluvatar... these are all fine.

However, it take more than one hour to run on CANN 9.0.0. This is not acceptable. The process is occupying the shared runners and no other jobs can run on it. A test case that takes more than 1 hour to finish means that the operator is not usable. We don't need to wait for it to "successfully" complete. I'm killing the job.

winfan1314:
> The time cost means something ...
> 
> The same benchmark takes about 6 minutes on Hygon, 23 minutes on CANN8.5.0, and it failed in 2 minutes on MetaX, failed in 11 minutes on Iluvatar... these are all fine.
> 
> However, it take more than one hour to run on CANN 9.0.0. This is not acceptable. The process is occupying the shared runners and no other jobs can run on it. A test case that takes more than 1 hour to finish means that the operator is not usable. We don't need to wait for it to "successfully" complete. I'm killing the job.

@tengqm Thank you for pointing this out. We agree that occupying a shared CANN 9.0 runner for more than one hour is unacceptable. After reviewing the complete logs, however, the long runtime does not appear to be caused by the FlagGems `nanmedian` implementation hanging.

## CANN 9.0

Both accuracy suites completed successfully before the benchmark started:

- Device reference: `147 passed`
- CPU reference: `91 passed, 1 skipped`

The benchmark measures the `native PyTorch operator` first and the `FlagGems implementation` second. For large `int32` inputs, the reported results were:

| Shape | Torch latency | FlagGems latency | Speedup |
|---|---:|---:|---:|
| `[1024, 1024]` | 30.070 ms | 0.585 ms | 51.363x |
| `[256, 4096]` | 30.077 ms | 0.585 ms | 51.375x |
| `[16, 131072]` | 60.793 ms | 1.153 ms | 52.716x |

The profiler timeline confirms that the native baseline dominates the wall-clock time. For these three large shapes, the native measurements took approximately `3m15s`, `3m15s`, and `6m34s`, while the corresponding FlagGems measurements completed in approximately `12s`, `13s`, and `22s`, including profiler overhead.

The job was manually cancelled after completing `nanmedian_out` for `int8`, while it was continuing through the remaining integer benchmark cases. There was no accuracy failure, kernel exception, or FlagGems timeout before cancellation.

[CANN 9.0 job log](https://github.com/flagos-ai/FlagGems/actions/runs/31361601157/job/93371502251)

## Other Failed Jobs

The other red jobs did not expose a `nanmedian` correctness problem:

- **MetaX:** pytest failed while importing `conftest.py` because the runner's glibc does not provide `GLIBC_2.38`.
- **MThreads:** pytest failed while importing `conftest.py` because `libmusart.so.5` was unavailable.
- **Iluvatar:** Triton's driver helper failed to compile because `cuda.h` was missing from the runner environment, after 10.5 minutes of `setting up flag_gems` process.

These failures occurred before meaningful `nanmedian` kernel validation and appear to be runner/environment issues.

## Follow-up

CANN 8.5 completed successfully, although some integer performance cases show poor FlagGems speedups. That is a separate performance concern and should be investigated independently.

To avoid blocking shared CANN 9.0 runners, we can adjust the benchmark coverage by excluding the pathological native Ascend integer baseline cases or reducing their large-shape coverage. This would address the CI runtime without hiding an accuracy or FlagGems kernel failure.

winfan1314:
After further investigation, the excessive runtime is significantly amplified by the current Ascend benchmark timing and collection logic.

### CANN 9 benchmark behavior

FlagGems uses `triton.backends.ascend.testing.do_bench_npu` for Ascend kernel benchmarks, with the default configuration:

- `warmup=1000`
- `repetition=100`

On the current CANN 9 / FlagTree Ascend implementation, these values are interpreted as invocation counts rather than durations. Therefore, for every shape and dtype, the benchmark executes:

- 1,000 warm-up calls and 100 profiled calls for the native PyTorch baseline;
- another 1,000 warm-up calls and 100 profiled calls for FlagGems.

The native baseline is measured first. For the large integer `nanmedian` cases, several individual baseline profiling sessions took approximately 3-6 minutes, whereas the corresponding FlagGems sessions took only about 12-22 seconds.

Profiler startup, synchronization, shutdown, and synchronous CANN CSV processing also add substantial wall-clock overhead. This explains why the complete benchmark exceeded one hour even though the reported FlagGems kernel latency was relatively small.

### Ascend kernel metric limitation

The current Ascend collector also does not aggregate all kernels launched by one callable. It filters profiler CSV rows and calculates a mean over selected kernel rows. Consequently, for native PyTorch, fallback paths, or multi-kernel FlagGems implementations, the reported latency can represent an average kernel-row duration rather than the complete operator latency.

We reproduced this behavior on Ascend hardware with a multi-kernel callable:

| Callable | Kernels per call | Reported kernel metric | Sum of kernels | Operator-mode latency |
| --- | ---: | ---: | ---: | ---: |
| PyTorch | 4 | 0.0308 ms | 0.1430 ms | 0.1466 ms |
| FlagGems | 2 | 22.5694 ms | 45.1873 ms | 45.5840 ms |

This means that the current Ascend `kernel` metric may under-report a multi-kernel callable approximately in proportion to its kernel count. It is also not directly comparable with the event-based kernel measurements used by other backends.

Source inspection additionally indicates that the CANN 8.5 collector reads `Duration(us)` without the `/1000` normalization performed by the CANN 9 collector, while FlagGems displays the result as milliseconds. Therefore, CANN 8.5 and CANN 9 kernel-mode results may not be directly comparable either.

### Proposed handling

For this PR, I can prevent the pathological native Ascend integer baseline from occupying the shared runner by reducing or excluding these benchmark cases. This would be a benchmark-scope adjustment and would not change the operator implementation or its accuracy coverage.

Separately, the benchmark framework would benefit from a dedicated follow-up change to:

1. make CANN 9 warm-up and active execution duration-aware, or use substantially smaller invocation counts;
2. aggregate every kernel belonging to one callable;
3. normalize profiler duration units consistently across CANN versions;
4. use operator-mode timing when end-to-end callable latency is required.

Therefore, the CANN 9 cancellation demonstrates a real benchmark-runtime problem, but it does not indicate a `nanmedian` accuracy failure or a long-running FlagGems kernel.

winfan1314:
Before submitting or appending any commit, we run the complete relevant test suites locally on every accelerator backend available to us. The commits in this PR were submitted only after those accessible backends had been fully validated.

During that local validation, we had already identified irregularities in the Ascend benchmark behavior and started a detailed investigation into its timing and result-collection logic. The findings above are based on that investigation and additional hardware experiments, rather than being inferred only from the failed CI run.

winfan1314:
So I'll temporary skip the benchmark on `ascend` backend until we finally find a solution for these irregularities.

Caeruleann:
fix CI failed

winfan1314:
image
image
I can't fix CI env error @_@.

winfan1314:
> fix CI failed

@Caeruleann pls view these issues: #5583 #5584 #5585

modao1234:
@winfan1314  Please add the Apache 2.0 license header to the files in the _ascend and _mthreads pathes.
Some of the core computational logic in the _ascend path operator relies on torch. Could you explain the rationale behind this?

modao1234 (approved):

PR #5086, opened 2026-07-30
feat(cpp/pointwise): declarative codegen + generic boxed dispatch for pointwise ops

lvyufeng:
### PR Category

- [x] Bug fix
- [x] New feature (declarative C++ pointwise codegen and optional boxed dispatch adapter)

### Type of Change

- [x] C++ pointwise dynamic dispatch and boxed-adapter fixes
- [x] Build and installation integration
- [x] Regression tests

### Description

This PR adds the declarative C++ pointwise code-generation and optional generic boxed-dispatch implementation, and addresses the review findings identified during CUDA integration testing.

The review fixes include:

- Preserve the schema's tensor/scalar ordering for scalar-first operators such as `remainder.Scalar_Tensor`; the boxed adapter now passes `{false, true}` for scalar-first kernels.
- Distinguish an out-of-range effective rank from an unexpected in-range registry hole, and report the configured maximum rank together with the supported fallback.
- Add local schemas for all custom scalar division and remainder overloads. The boxed adapter queries `op.schema()`, so each implementation must have a matching local schema.
- Make standalone Triton code generation reproduce module-level aliases and supported import aliases, and warn for unsupported alias shapes.
- Fix the C++ prebuild script to resolve the source tree relative to its own location and fail fast when Python imports `flag_gems` from a different checkout. This prevents stale editable installations from silently generating kernels with old code.
- Preserve boxed scalar dtype metadata through promotion. Weak Python floating scalars retain floating/complex tensor behavior, while integral and bool tensors promote to the default floating dtype when appropriate.
- Validate boxed selectors and tensor-valued fill arguments, including explicit `none` handling for true division.
- Propagate CUDA runtime includes and generated-header install interfaces to downstream consumers.

The declarative `op_specs.py` table remains the single source of truth for generated declarations, implementations, schemas, registrations, and boxed routing metadata.

### Issue

The changes address the review findings for PR #5086 covering scalar-first dispatch, rank diagnostics, custom scalar schemas, standalone alias generation, scalar promotion, selector validation, and CMake installation/export behavior.

### Progress

- [x] The code has been properly reviewed.
- [x] Review issues have been addressed.
- [x] The changes are covered by unit and integration tests.

### Verification

All results below were obtained from a clean CUDA build using Python/Torch from `/nfs/lvyufeng/miniconda3/envs/torch-cuda-210` (`torch 2.10.0+cu128`) on 8 NVIDIA A100 GPUs with CUDA 13.0.48:

- `cmake --build cpp/build-cuda -j2`: exit code 0; 228 kernels generated for ranks `0..5`.
- Full boxed CUDA CTest suite: **26/26 passed**.
- Focused boxed CTest collection: **7/7 passed**, including `test_triton_unary`, `test_remainder_scalar_first`, and `test_high_rank_fallback`.
- Boxed-disabled compatibility build with `FLAGGEMS_POINTWISE_DYNAMIC_BOXED=OFF`: exit code 0.
- Boxed-disabled pointwise CTest collection: **6/6 passed**.
- `GEMS_VENDOR=nvidia python -m pytest tests/test_pointwise_alias_codegen.py -q`: **13 passed**.
- `GEMS_VENDOR=nvidia python -m pytest tests/test_pointwise_dynamic.py tests/test_pointwise_type_promotion.py -q`: **243 passed, 2 skipped**.
- Ruff check on changed Python/code-generation files: **all checks passed**.
- Ruff format check: **4 files already formatted**.
- `git diff --check`: passed.
- CMake install check: passed.
- A standalone C++ consumer including the installed `flag_gems/operators.h` and generated headers passed syntax compilation.

The first clean-build unary run exposed five compilation failures. Root-cause investigation showed that `prebuild_kernels.py` was importing an older editable FlagGems checkout because its source-root calculation was one directory short. After correcting the root calculation and adding the source-tree guard, the generated files contained the required shim aliases and unary coverage passed 16/16.

`clang-format` was not available in the verification environment, so no clang-format result is claimed.

### Performance

No performance regression is expected for the default unboxed path. The boxed adapter is opt-in through `FLAGGEMS_POINTWISE_DYNAMIC_BOXED`; the compatibility build with this option disabled passed. The implementation keeps the existing pre-generated rank-specialized kernels and only adds boxed argument decoding, validation, and dispatch selection around them.


cpp/lib/pointwise_dynamic_cpp/CMakeLists.txt
```diff
@@ -20,6 +20,10 @@ set(POINTWISE_OUTPUT_DIR "${CMAKE_CURRENT_BINARY_DIR}/generated")
 
 # Operations to generate (full module paths)
 set(POINTWISE_OP_FILES "flag_gems.ops.add" "flag_gems.ops.div" "flag_gems.ops.fill"
+    "flag_gems.ops.remainder"
```
chen98038, line 26: Adding `flag_gems.ops.remainder` to POINTWISE_OP_FILES ships a C++ `rem_st` wrapper with an inverted is_tensor mask: `_build_is_tensor_mask` in prebuild_kernels.py assumes tensors-first, but `rem_st`'s schema is `is_tensor=[False, True]` (scalar first), while the generated Triton kernel signature keeps schema order (val0, in0_ptr, ...).

Failure scenario: with FLAGGEMS_BUILD_POINTWISE_DYNAMIC_CPP=ON (this PR is what first makes the remainder code in cpp/lib/div.cpp compile), calling `flag_gems::remainder(a, b)` with 0-dim `a` and >0-dim `b` — or the registered `remainder.Scalar_Tensor` overload — reaches `pointwise_dynamic::rem_st`. Dispatch walks the {true,false} mask and passes the tensor pointer into the kernel's val0 (scalar) slot and the double into the in0_ptr (pointer) slot: the launch either fails with a Triton compile/signature error or dereferences a garbage pointer (CUDA illegal memory access / wrong results).

Suggestion: build the mask from the op's actual `is_tensor` layout instead of assuming tensors-first, and add a ctest covering a scalar-first op.

cpp/csrc/cstub.cpp
```diff
@@ -198,6 +215,17 @@ TORCH_LIBRARY(flag_gems, m) {
   m.def("fill.Tensor(Tensor self, Tensor value) -> Tensor");
   m.def("fill_.Scalar(Tensor(a!) self, Scalar value) -> Tensor(a!)");
   m.def("fill_.Tensor(Tensor(a!) self, Tensor value) -> Tensor(a!)");
+  // unary elementwise
+  m.def("abs(Tensor self) -> Tensor");
```
chen98038: Ten unary ops are registered into TORCH_LIBRARY with generic schemas (e.g. `abs(Tensor self) -> Tensor`), but the C++ dispatch only pre-generates kernels for rank 0..POINTWISE_MAX_RANK(=5). A non-fast-path input with effective rank>5 throws a bare `std::runtime_error` ("No kernel for rank 6") — no TORCH_CHECK, no fallback, and the limit is mentioned in no docs or tests.

Failure scenario: call `torch.ops.flag_gems.abs` on a 6-D non-dense view (e.g. `x[:, :, :, :, :, 0:2]`, which fails use_fast_path after slicing): `op_registry.find(6)` at pointwise_prepare_args.h:456 fails and the bare exception escapes through the boxed call. The Python path for the same op codegens any rank on demand, so the two paths diverge.

Suggestion: at minimum replace with a `TORCH_CHECK` carrying a clear message (the C++ path's rank limit and how to avoid it); ideally fall back to dim-collapsing or the Python path.

src/flag_gems/utils/pointwise_dynamic.py
```diff
@@ -1138,13 +1141,49 @@ def _collect_jit_deps(scalar_fn):
...
+                continue
+            root = value.value
+            if isinstance(root, ast.Name) and root.id in imported_name_module:
+                alias_sources.append(
+                    "".join(source_lines[node.lineno - 1 : node.end_lineno])
+                )
+                # The alias's base name (e.g. ``tl_extra_shim``) must be
+                # imported in the generated file even if it came from a module
+                # that is normally elided (e.g. ``flag_gems.utils``).
+                extra_imports.setdefault(imported_name_module[root.id], set()).add(
```
chen98038: The forced import in alias collection writes `root.id` — the local binding name, which may be an `as`-rename — as if it were the module's real exported name, emitting `from X import <asname>`. Any op module of the form `from X import a as b` + `alias = b.attr` gets an invalid import line.

Failure scenario: a module written as `from flag_gems.utils import tl_extra_shim as shim` + `_t = shim.tanh` (same style as the many existing `import ... as tle/ext`) produces `from flag_gems.utils import shim` at the top of the generated kernel file → `ImportError: cannot import name 'shim'` on load. A module that worked before this PR (when aliases weren't emitted) now fails on first call, in both the in-process Python path and the standalone C++ path. None of the 14 in-tree generated modules hits this today — it's a landmine introduced by the new logic.

Suggestion: record both the real exported name (`alias.name`) and the asname in `imported_name_module`, and emit `from X import name as asname`.

src/flag_gems/utils/pointwise_dynamic.py
```diff
@@ -1138,13 +1141,49 @@ def _collect_jit_deps(scalar_fn):
...
+        for node in ast.iter_child_nodes(tree):
+            if not isinstance(node, ast.Assign):
+                continue
+            if len(node.targets) != 1 or not isinstance(node.targets[0], ast.Name):
+                continue
+            value = node.value
+            if not isinstance(value, ast.Attribute):
+                continue
+            root = value.value
+            if isinstance(root, ast.Name) and root.id in imported_name_module:
```
chen98038: Alias detection matches exactly one shape: top-level `Name = <from-imported-name>.<attr>`. Aliases rooted in a plain `import X as Y` (ast.Import — real in-tree examples: `sigmoid = tl.sigmoid` in fused/swiglu.py, `Tensor = torch.Tensor` in ops/cummax.py), chained attributes (`x = a.b.c`), AnnAssign annotated assignments, and aliases inside if/try blocks are all silently dropped.

Failure scenario: add such a module to POINTWISE_OP_FILES and the generated file lacks the assignment; standalone C++ TritonJIT compilation raises NameError on first call while the in-process Python path works fine (masked by `__globals__`) — the very bug this PR fixes, silently recurring in another shape, with no warning.

Suggestion: at least support the ast.Import form, and warn on alias-shaped assignments that were skipped.

cpp/ctests/test_triton_unary.cpp
```diff
@@ -0,0 +1,156 @@
...
+  auto result = gems_assert_close(out, ref);
+  EXPECT_TRUE(result.ok) << result.message;
+}
+
+}  // namespace
+
+// ------------------------------------------------------------------ abs
+TEST_F(UnaryTest, Abs) {
+  auto x2 = any({10, 10});
+  expect_close(flag_gems::abs(x2), x2.abs());
```
chen98038: The new tests all skip `accuracy_utils::to_reference` and compute reference values directly on device tensors (`x.abs()`, `torch::silu(x)`, ...), bypassing the TO_CPU reference path every other ctest follows. In accuracy_utils.cpp, MUSA/NPU backends set TO_CPU=true so references should be computed on CPU; as written, Triton results are compared against the same device's own ATen implementation — if the device's ATen has a bug, the test either passes spuriously or fails misleadingly.

Suggestion: reuse `flag_gems::accuracy_utils::to_reference` for the reference inputs.

PR #5088, opened 2026-07-30
fix(utils): add pure-triton nextafter fallback for non-CUDA backends

### LINKED ISSUE #5053: The `nextafter` operator is not always available on all backends
The `triton.language.extra.tang.libdevice.nextafter` operator may and may not be available on a specific backend. A fallback implementation is needed.

FlagGems not usable on Sunrise backend.

tengqm:
## Problem

`import flag_gems` fails on backends whose triton `libdevice` lacks `nextafter` — cambricon (triton 3.2.0, `mlu`) and sunrise (triton 3.4.0.6, `tang`):

```
AttributeError: module 'triton.language.extra.mlu.libdevice' has no attribute 'nextafter'
```

`ops/nextafter.py`'s float32 branch calls `tl_extra_shim.nextafter`, and triton's JIT AST dependency finder resolves attributes eagerly at **import** time — so the whole package fails to import on these backends, not just at kernel launch.

## Root cause

Two defects in `src/flag_gems/utils/triton_lang_helper.py` combine:

1. **No fallback for `nextafter`.** It had no entry in `_FALLBACK_SYMBOLS` and was absent from the patched-symbols list, so on backends whose own libdevice lacks it, `tl_extra_shim.nextafter` resolved to nothing → AttributeError at import.

2. **`_patch_missing_symbols` had backwards priority.** It searched foreign candidate modules *before* the pure-triton fallback. The generic CUDA `triton.language.extra.libdevice` **does** expose `nextafter`, so it got bound preferentially — but that symbol does **not lower** on the mlu/tang backends. It compiles to `None`, producing `AttributeError("'NoneType' object has no attribute 'to'")` at kernel-compile time. This loop only ever runs for symbols the vendor's own libdevice lacks, so any candidate match is inherently foreign and may not lower on the current backend.

## Fix

- Add `_fallback_nextafter` (`@triton.jit`): pure-triton uint32 bit-manipulation IEEE-754 nextafter, mirroring the existing fp16/bf16 path in `ops/nextafter.py`.
- Register it in `_FALLBACK_SYMBOLS` and add `"nextafter"` to the patched-symbols list.
- Flip `_patch_missing_symbols` priority: prefer the pure-triton fallback, borrowing from a foreign candidate module only when no fallback exists.

One change fixes **both** cambricon and sunrise. On CUDA the primary shim already provides these symbols, so the patch loop is skipped and behavior is unchanged. Follows the same `_fallback_*` pattern as #5038 (`fallback_j1`).

## Verification

Tested on a cambricon MLU node against the FlagOS runtime image (`flagos-runtime-cambricon-neuware4.4.3:2.1.1`) with the patched file mounted over the installed package:

- `_fallback_nextafter` is **bit-exact** vs `torch.nextafter` — 0 mismatches over 1024 random fp32 values.
- All four runtime verify checks pass: `import torch` / `import triton` / `import flag_gems` / `use_gems()` add.
- Both fp32 and fp16 `torch.nextafter` match the CPU reference.

Before the fix, `import flag_gems` raised the AttributeError above; after, it imports and runs cleanly.

Closes #5053

### DIFF
--- a/src/flag_gems/utils/triton_lang_helper.py
+++ b/src/flag_gems/utils/triton_lang_helper.py
@@ -127,12 +127,62 @@ def _fallback_j1(x):
     return tl.where(ax <= 5.0, small_val, large_val)
 
 
+@triton.jit
+def _fallback_nextafter(input, other):
+    # IEEE 754 nextafter for float32 via uint32 bit manipulation.  Mirrors the
+    # fp16/bf16 bit-manipulation path in ops/nextafter.py; used when a backend's
+    # libdevice lacks a native nextafter (e.g. cambricon mlu / sunrise tang forks).
+    x_int = input.to(tl.uint32, bitcast=True)
+    y_int = other.to(tl.uint32, bitcast=True)
+
+    exp_mask = 0x7F800000  # 8 exponent bits
+    frac_mask = 0x007FFFFF  # 23 mantissa bits
+
+    # uint32 constants via bitwise ops / shifts.  A Python int literal 0x80000000
+    # would promote to a wider signed type and break the uint32 arithmetic/bitcast.
+    uint_zero = x_int & 0
+    uint_one = uint_zero | 1
+    uint_neg_one = ~uint_zero  # 0xFFFFFFFF = -1 in uint32 arithmetic
+    sign_bit = uint_one << 31  # 0x80000000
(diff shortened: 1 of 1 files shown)


tengqm:
@103yiran The cann850 failure is caused by another similar problem  --- `erfc` is not available on cann850 package matrice, which means that we need another PR for `erfc` fallback.

PR #5106, opened 2026-07-31
fix: vector_norm accuracy test failure on mthreads backend

lyujheng:
### PR Category

OP Test

### Type of Change

Bug Fix

### Description

Relax `atol` for `vector_norm` accuracy test on mthreads backend when `--ref cpu` is enabled.

For large reduce dimensions (e.g. 8M elements with ord=1), torch GPU (mthreads native) and FlagGems Triton kernel produce consistent results, but both diverge from torch CPU due to float32 accumulation precision differences between CPU and GPU implementations. This is not a Triton kernel correctness issue.

Only affects the `--ref cpu` scenario (used by `run_tests.py`); direct pytest without `--ref cpu` still uses strict tolerance.

img_v3_02144_5369b863-e961-4297-8c82-85064985bdbg

### Progress

### Performance

N/A (test tolerance adjustment only, no kernel change)

### DIFF
--- a/tests/test_vector_norm.py
+++ b/tests/test_vector_norm.py
@@ -69,6 +69,14 @@ def test_accuracy_vectornorm(shape, ord, dim, keepdim, dtype):
     with flag_gems.use_gems():
         res_out = torch.linalg.vector_norm(inp, ord, dim, keepdim)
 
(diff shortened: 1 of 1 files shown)


huangyiqun (approved):
LGTM.

PR #5109, opened 2026-07-31
fix: resolve_conj performance test failure on mthreads backend.

lyujheng:
### PR Category

Operator

### Type of Change

Bug Fix

### Description

Fix infinite recursion in mthreads `resolve_conj` operator for conjugated tensors.

The previous implementation used `resolve_conj_triton(A, is_conj=True)` for 2D/3D tensors and `torch.complex(A.real, A.imag.neg())` as fallback. Both paths trigger `aten::resolve_conj` dispatch back to this function when accessing conjugated tensor data at Python level, causing infinite recursion.

Fix: use `A.clone()` to resolve conjugation physically at the C++ `aten::clone` level, bypassing Python dispatch entirely. This is aligned with the common op implementation (`src/flag_gems/ops/resolve_conj.py`).

### Progress

### Performance

Performance test (`benchmark/test_resolve_conj.py`) now passes on mthreads. Previously failed with `maximum recursion depth exceeded`.
image

### Accuracy
image


huangyiqun (approved):
LGTM.

PR #5111, opened 2026-07-31
fix: resolve mul op accuracy and performance test failure on mthreads

lyujheng:
### PR Category

Operator

### Type of Change

Bug Fix

### Description

Fix `mul` operator accuracy and performance test failures on mthreads backend.

**Operator fix (`src/flag_gems/ops/mul.py`):**
Replace hardcoded `device.type != "cuda"` check with `device.type != flag_gems.device` at two locations. The original check incorrectly triggered the fallback path on non-cuda backends (e.g. mthreads), causing `RuntimeError` for scalar×tensor cases and infinite recursion in performance tests. This fix benefits all non-cuda backends.

**Test fix (`tests/test_mul.py`):**
- Skip `complex32` (ComplexHalf) tests on mthreads — unsupport this experimental dtype.
- For `test_mul_complex_float_tensor` and `test_mul_complex_int_tensor`: compute reference on CPU for mthreads, because mthreads native `torch.mul` returns incorrect results (zeros) for complex × real type combinations.

### Progress

### Performance

Performance test (`benchmark/test_mul_perf.py`) now passes on mthreads. Previously failed with infinite recursion due to the incorrect device check entering fallback within `use_gems()` context.
image

### Accuracy
image


huangyiqun (dismissed):
LGTM.

huangyiqun (approved):
LGTM.

tengqm:
It doesn't look to me like we are doing the right thing here. The operator has to be improved, not the test case. Fixing the test case will reduce errors reported during tests, but the "error" is still there. We have to figure out how to make the operator works on mthreads, rather than making the test case works.

PR #5125, opened 2026-07-31
[Ascend] Fix nondeterministic fused topk_softmax renormalization

### LINKED ISSUE #5123: [Ascend] fused topk_softmax produces nondeterministic corrupted weights when renormalize=True
## Summary

On the Ascend NPU backend, `flag_gems.fused.topk_softmax` returns corrupted and nondeterministic `topk_weights` when `renormalize=True`.

The selected expert indices remain correct and deterministic. The issue affects router weights only.

## Environment

- Device: Ascend910B4-1
- CANN: 9.0.0
- Python: 3.11
- See attached `environment.txt` for exact package versions.

## Minimal Reproduction

Input shape reproducing the issue:

- logits: `(64, 256)`, `torch.bfloat16`, NPU
- topk: `8`
- `renormalize=True`

The attached script runs the same input repeatedly and compares against:

```python
torch.softmax(logits.float(), dim=-1)
torch.topk(...)
```

## Observed Behavior Before Fix

With renormalize=False:
top-k weights match the FP32 PyTorch reference
max error: approximately 7.45e-09
20 repeated executions are deterministic
expert ids are correct

With renormalize=True:
expert ids remain correct and deterministic
top-k weights become corrupted and nondeterministic
observed maximum values reach approximately 3e9
row sums reach approximately 1e10
repeated executions differ by approximately 1e9

This is not normal floating-point error.

## Suspected Cause
The kernel first stores raw top-k weights to global memory, then, in the same kernel, loads those output values back from global memory to apply renormalization.
On the Ascend Triton backend, this global-memory read-after-write pattern appears nondeterministic.

## Proposed Fix
Keep the first kernel responsible only for raw softmax/top-k output. When renormalize=True, launch a second kernel on the same stream to normalize the already-written top-k weights.
This removes the same-kernel global-memory read-after-write dependency.

## Validation After Fix
After splitting normalization into a second Triton kernel:
renormalize=True max reference error: approximately 4.47e-08
max row-sum error: approximately 2.38e-07
20 repeated executions are deterministic, with only negligible FP32 rounding noise
expert ids remain correct

In a vLLM Qwen3.6-35B-A3B TP=4, 64-concurrent decode reproduction, the original path produced around 21-24 corrupted requests per round. The fixed kernel reduces this to the same small residual level as independent PyTorch and AscendC router references.

## Attachments

[flaggems_topk_split_normalize.patch](https://github.com/user-attachments/files/30593079/flaggems_topk_split_normalize.patch)
[tmp_probe_flaggems_topk_renorm.py](https://github.com/user-attachments/files/30593081/tmp_probe_flaggems_topk_renorm.py)
[environment.txt](https://github.com/user-attachments/files/30593080/environment.txt)

luentong:
### PR Category

- [x] Operator
- [ ] OP Test
- [ ] Model Test
- [ ] Benchmark
- [ ] CI/CD
- [ ] User Experience
- [ ] Other

### Type of Change

- [x] Bug Fix
- [ ] New Feature
- [ ] Performance Optimization
- [ ] Refactor
- [ ] Documentation Update
- [ ] Other

### Description

Fixes an Ascend backend-specific correctness issue in
`fused.topk_softmax(..., renormalize=True)`.

The original Triton kernel wrote raw top-k weights to global memory and then
loaded the same output buffer again to normalize it within the same kernel.
On the Ascend Triton execution path, this produced corrupted and
nondeterministic router weights under concurrent MoE inference. Selected
expert ids remained correct, but router weights could become extremely large.

This change keeps the existing kernel responsible for softmax and top-k
selection, and performs top-k weight normalization in a separate Triton
kernel launched afterward on the same stream.

The output layout and `token_expert_indices` contract are unchanged.

### Issue

Closes #5123

### Validation

Added an Ascend NPU regression test covering BF16 input with shape
`(64, 256)` and `topk=8`.

The test verifies both `renormalize=False` and `renormalize=True`:

- outputs match a PyTorch FP32 softmax/top-k reference;
- selected expert ids and token-expert indices match the expected contract;
- repeated invocations are deterministic;
- when renormalizing, each row sums to approximately 1.

Executed with CANN 9.0 and `torch_npu` on Ascend 910B:

```text
2 passed, 2 warnings in 0.55s


luentong:
Additional localization result:

I tested a single-kernel control variant that preserves the original first-pass
stores to `topk_weights`, `topk_indices`, and `token_expert_indices`, but
recomputes top-k values and writes normalized weights directly without ever
loading `topk_weights` back from global memory.

It passed 20 independent process-level runs of the BF16 `(64, 256)`, top-8
Ascend regression test. This further isolates the trigger to the original
same-kernel readback of `output_ptr`, rather than the initial output stores,
top-k selection, index layout, or row/program overlap.

luentong:
The split-kernel patch is the correctness workaround.
Root-cause investigation and minimal compiler reproducer are tracked in
triton-lang/triton-ascend#1416.

PR #5129, opened 2026-08-01
Optimize GLU forward with a Triton kernel

chenmiaoming:
### PR Category

Operator
### Type of Change

Performance Optimization
### Description

This PR optimizes the GLU forward path with a flat Triton kernel for contiguous tensors when GLU is applied along the last dimension.

The new kernel:

- Processes flattened output elements in parallel.
- Directly loads the corresponding values from both halves of the input.
- Uses the existing backend autotune infrastructure to tune `BLOCK_SIZE` and `num_warps`.
- Buckets the output size to avoid repeated autotuning for nearby tensor sizes.
- Keeps the existing pointwise implementation as the fallback for unsupported layouts and dimensions.
- Preserves the existing backward implementation.

The implementation is placed in the common operator path and does not contain vendor-specific dispatch logic. Backend-specific autotune configurations can override the default configuration when needed.
### Progress

### DIFF
--- a/src/flag_gems/ops/glu.py
+++ b/src/flag_gems/ops/glu.py
@@ -18,10 +18,85 @@ import torch
 import triton
 import triton.language as tl
 
-from flag_gems.utils import pointwise_dynamic, tl_extra_shim
+from flag_gems import runtime
+from flag_gems.runtime import torch_device_fn
+from flag_gems.utils import libentry, pointwise_dynamic, tl_extra_shim
 
 logger = logging.getLogger(__name__)
 exp = tl_extra_shim.exp
+
+
+@libentry()
+@triton.autotune(
+    configs=runtime.get_tuned_config("glu"),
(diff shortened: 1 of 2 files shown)


huangyiqun (commented):
Why develop an additional kernel? Is it possible to optimize the existing one instead? Adding `_can_use_triton_glu` and `_glu_triton_forward` will introduce extra runtime overhead on the CPU side, so please try to keep it as streamlined as possible.

huangyiqun:
Plz sign in CLA.

chenmiaoming:
Closing: benchmarks on a clean A10 GPU show the flat kernel is ~30% slower than native torch on large tensors even after autotuning (1094us vs 827us for (128,128,4096)); the remaining gap is CPU-side dispatch overhead from the autotune/libentry path (~140us/launch) which hurts small tensors. The bandwidth-bound kernel has no headroom left. Thanks for the review feedback.

PR #5130, opened 2026-08-02
fix(mul): gate optimized path on active backend device, not hardcoded "cuda"

tengqm:
Fixes `mul` on backends whose device type is not literally `"cuda"`.

## Problem

`mul_broadcast_func` / `mul_complex_broadcast_func` gate the optimized Triton
path on `device.type != "cuda"`, falling through to an aten reference path
otherwise. mthreads reports `device.type == "musa"` (its `device_name` — the
only GPU backend whose name is not `"cuda"`), so **every** `mul` on mthreads
takes the fallback.

That fallback is also wrong for a scalar operand: it calls
`torch.ops.aten.mul.Tensor.redispatch(..., a, b)`, and `mul.Tensor` requires a
Tensor `other`. A python scalar (e.g. rope’s `1.0 / freqs` →
`Tensor.reciprocal() * 1.0`) raises:

```
RuntimeError: aten::mul.Tensor expected Tensor for 'other', found float
```

which crashes model load under vLLM on mthreads.

## Fix

Gate on the active backend’s device name (`runtime.device.name`) instead of the
hardcoded literal, matching the convention already used elsewhere (e.g.
`_upsample_bilinear2d_aa.py` compares `input.device.type` against `device.name`).

- Backends whose `device_name` is `"cuda"` (nvidia, metax, iluvatar, hygon, …)
  are unaffected.
- mthreads (`"musa"`) now takes the same optimized Triton path as everyone else.

## Verification (mthreads, MTT S5000, torch_musa 5.2.0)

- The Triton `mul` kernel produces correct results (max abs err 0 vs torch
  reference) across scalar / tensor×tensor / broadcast / fp16 / bf16 / int /
  `out=` / in-place / complex.
- `vllm serve` of a yarn-rope model reaches `Application startup complete` and
  returns coherent inference output, where it previously crashed at load with
  the error above.

### DIFF
--- a/src/flag_gems/ops/mul.py
+++ b/src/flag_gems/ops/mul.py
@@ -19,11 +19,19 @@ import torch
 import triton
 import triton.language as tl
 
+from flag_gems.runtime import device as runtime_device
 from flag_gems.runtime import torch_device_fn
 from flag_gems.utils import libentry, libtuner
 from flag_gems.utils.shape_utils import volume
 
 logger = logging.getLogger(__name__)
+
+# Name of the active flag_gems backend device (e.g. "cuda", "musa"). The
+# optimized Triton path below runs on this device; tensors on other device
+# types (e.g. plain "cpu") fall through to the aten reference path. Gating on
+# the hardcoded literal "cuda" wrongly excluded backends whose device type is
+# not "cuda" (e.g. mthreads reports "musa") from the shared Triton path.
+_DEVICE_NAME = runtime_device.name
 
 _FALLBACK_KEYSET = torch._C.DispatchKeySet(
     torch._C.DispatchKey.CompositeExplicitAutograd
@@ -576,7 +584,7 @@ def mul_broadcast_func(a, b, out=None):
         raise TypeError("mul expects tensor or scalar inputs")
 
     device = _select_device(a, b)
-    if device.type != "cuda":
+    if device.type != _DEVICE_NAME:
         if out is not None:
             return torch.ops.aten.mul.out.redispatch(_FALLBACK_KEYSET, a, b, out=out)
         return torch.ops.aten.mul.Tensor.redispatch(_FALLBACK_KEYSET, a, b)
@@ -721,7 +729,7 @@ def _launch_complex_generic(
 
 def mul_complex_broadcast_func(a, b, out=None):
     device = _select_device(a, b)
-    if device.type != "cuda":
+    if device.type != _DEVICE_NAME:
         if out is not None:
             return torch.ops.aten.mul.out.redispatch(_FALLBACK_KEYSET, a, b, out=out)
         return torch.ops.aten.mul.Tensor.redispatch(_FALLBACK_KEYSET, a, b)


ErenAta16 (commented):
The diagnosis is right and `device.name` is the correct attribute to reach for. `runtime/__init__.py` compares against it the same way when it installs the no-op device guards:

```python
if device.name == "cpu":
```

so `name` is the torch device *type* string, distinct from `vendor_name` which is used a few lines earlier for `set_torch_backend_device_fn`. Comparing it to `device.type` is consistent with how the package already uses it, and a backend reporting `musa` stops being excluded from the Triton path. That part I have no concerns about.

**Two things I would want settled before this lands.**

**1. On a CPU backend this silently flips `mul` from aten to Triton.** `runtime/__init__.py` supports `device.name == "cpu"` explicitly, that is what the `_noop_device_guard` and `_NoOpDeviceGuard` shims exist for. When flag_gems is built that way, `_DEVICE_NAME == "cpu"`, and the gate becomes:

```python
if device.type != "cpu":     # CPU tensors no longer match this
```

so CPU tensors now take the optimized Triton path instead of redispatching to aten. On main they always fell back, because the literal was `"cuda"` and nothing else ever equalled it. That may well be what you want, since the whole point is that the shared Triton path should follow the active backend. But it is a behaviour change for an existing supported configuration, not just an unblocking of `musa`, and the comment on `_DEVICE_NAME` currently states the opposite:

> tensors on other device types (e.g. plain "cpu") fall through to the aten reference path

That is accurate on a CUDA or MUSA backend and inaccurate on a CPU one. Worth either rewording it or excluding `cpu` deliberately, so the next reader is not misled by a comment that is only conditionally true.

**2. `mul.py` is one instance of a repo-wide pattern.** The same hardcoded gate exists in at least these, from a code search for the identical comparison:

```
src/flag_gems/ops/as_strided_copy.py
src/flag_gems/ops/scaled_mm.py
src/flag_gems/ops/conv_transpose2d.py
src/flag_gems/fused/flashmla_sparse.py
src/flag_gems/fused/fused_marlin_moe.py
src/flag_gems/fused/DSA/bin_topk.py
src/flag_gems/fused/mhc/hc_head_fused_kernel.py
src/flag_gems/experimental_ops/{lift,threshold,logaddexp2,log10_,hardsigmoid_,xlogy,special_xlog1py}.py
```

The result list was truncated at fifteen so there are likely more. Every one of them excludes a non-cuda backend from its optimized path for the same reason. Not a request to fix them here, a narrow PR is the right shape, but if the intent is that `musa` stops being excluded generally then this fixes one operator out of a set, and it would be worth a follow-up issue naming the pattern so it does not get rediscovered one operator at a time.

**On the module-level capture.** `_DEVICE_NAME = runtime_device.name` binds once at import. That matches the existing lifetime, since `device = DeviceDetector()` is itself constructed once when `flag_gems.runtime` is imported, so nothing goes stale today. It is only worth noting because it makes `mul.py` insensitive to any future re-detection, where a direct `runtime_device.name` read would follow it. A comment saying the constant is deliberately snapshot-at-import would settle that.

No functional objection to the change itself; both call sites are updated consistently and the fallback branches are untouched.

physics31415926 (approved):

0x45f (approved):

PR #5150, opened 2026-08-03
fix: recognize vLLM stable ABI extensions

CherryLemon:
### PR Category
Other

### Type of Change
Bug Fix

### Description
`_ensure_vllm_library_exists` mapped each `lib_name` to a single extension
module, so on official vLLM 0.24 CUDA wheels/images
(`vllm/vllm-openai:v0.24.0-cu129-ubuntu2404`) it always returned `False`:
those builds ship the stable-ABI extensions (`vllm._C_stable_libtorch`,
`vllm._moe_C_stable_libtorch`) rather than the legacy `vllm._C` / `vllm._moe_C`.
The result is that FlagGems treats already-registered vLLM ops as missing and
falls back to registering its own schemas over them.

The map now holds a tuple of candidate modules per library and tries each in
order, legacy first, then stable ABI. `_vllm_fa3_C` has no stable-ABI variant
and keeps its single entry. Behaviour on pre-0.24 vLLM is unchanged, since the
legacy module is still tried first and short-circuits.

### Issue
N/A

### Progress

`tests/test_patch_util.py` covers all four `lib_name` entries, asserting both
the attempt order and that the last candidate (stable ABI) is accepted when the
legacy import fails.

Verified in a real vLLM 0.24 container on 2 nodes x 8 H100: with this change
FlagGems initializes and serves Qwen3.5-397B-A17B (TP=16, `mp` executor,
non-eager) end to end, with `attention_backend` / `topk_softmax` dispatching to
`default.flagos` and a 40-entry FlagGems oplist. Without it, op registration
takes the fallback path. The paired vllm-plugin-FL fix is
flagos-ai/vllm-plugin-FL#338.

### Performance
No performance change intended - this only affects extension detection at
import time. For reference, benchmarks on the environment above (random
dataset, 1K/4K input x 1K output, concurrency 64, MoE backend pinned to
`flashinfer_cutlass`, steady-state mean of rounds 2-3) put FlagGems ops at
2139.77 / 1725.28 tok/s output throughput versus 2148.29 / 1748.37 for the
vendor path, i.e. `-0.40%` / `-1.32%`. That gap is a pre-existing property of
the FlagGems kernels on this model, not introduced here; it is only measurable
at all because this fix lets the FlagGems path run on 0.24 in the first place.


ceci3 (approved):

huangyiqun:
Plz sign in CLA.

huangyiqun (approved):
LGTM.

PR #5153, opened 2026-08-03
[KernelGen][Nvidia] Add _cudnn_attention_backward operator with Triton kernel

yzw1128:
## Summary

Add `_cudnn_attention_backward` operator implementation using a BHSD-native two-kernel Triton design. This operator computes gradients of query, key, and value for the cuDNN attention backward pass.

### BHSD-Native Stride Mapping

The key optimization eliminates the expensive BHSD<->BSHD permute+contiguous operations present in the generic `_scaled_dot_product_cudnn_attention_backward` path. Instead of converting BHSD (batch, heads, seq, dim) to BSHD (batch, seq, heads, dim) and back (8 permute+contiguous calls total), the optimized implementation passes BHSD strides directly to the Triton flash attention backward kernels:

```
stride_qm <- Q.stride(2)   # seq dim stride in BHSD
stride_qh <- Q.stride(1)   # head dim stride in BHSD
```

The BHSD layout additionally provides **4x better memory coalescing** for Q/K/V seq-dim loads (stride=D=64 bytes vs stride=H*D=256 bytes in BSHD).

### Performance Limitations

The Triton implementation uses two separate kernel launches (dQ and dKV), while cuDNN uses a single fused kernel launch internally. This architectural difference and cuDNN hand-tuned heuristics for tile sizes explain the remaining gap. Further optimization could explore a single fused KV-centric kernel, though initial experiments with atomic-based dQ accumulation showed regression due to atomic throughput on H20.

## Correctness Verification

All 72 parameterized test cases pass (24 skipped for unsupported causal+bias combinations):

- **Shapes**: (2,4,512,512), (1,2,512,1024), (1,1,64,64), (4,8,128,128), (2,4,128,512), (1,2,768,768)
- **Head sizes**: 64, 128
- **Mask types**: causal, non-causal
- **Attention bias**: with/without
- **Data types**: float16, bfloat16
- **Tolerance**: atol=1e-3, rtol=1e-3

## Performance (geometric mean speedup vs cuDNN)

| Config | Mask | aten (ms) | GEMS (ms) | Speedup |
|--------|------|-----------|-----------|---------|
| B=2 H=4 Q=2048 K=2048 D=64 | noncausal | 0.223 | 0.360 | 0.62x |
| B=2 H=8 Q=2048 K=2048 D=128 | noncausal | 0.822 | 1.088 | 0.76x |
| B=2 H=8 Q=4096 K=4096 D=64 | noncausal | 1.384 | 1.955 | 0.71x |
| B=4 H=16 Q=1024 K=1024 D=128 | noncausal | 0.762 | 1.210 | 0.63x |
| B=4 H=16 Q=512 K=512 D=128 | causal | 0.202 | 0.290 | 0.70x |
| B=1 H=32 Q=1024 K=1024 D=128 | noncausal | 0.453 | 0.664 | 0.68x |

**Geometric mean across 12 configurations: 0.574x**

Larger sequence lengths achieve up to 0.76x speedup vs cuDNN.

## Files Changed

- `src/flag_gems/ops/cudnn_attention_backward.py` - operator implementation (BHSD-native two-kernel)
- `tests/test_cudnn_attention_backward.py` - 96 test cases with correctness verification
- `benchmark/test_cudnn_attention_backward.py` - performance benchmark
- `src/flag_gems/ops/__init__.py` - import and `__all__` registration
- `src/flag_gems/__init__.py` - dispatch table entry
- `conf/operators.yaml` - operator metadata


bin913 (dismissed):

conf/operators.yaml
```diff
@@ -2250,6 +2250,21 @@ ops:
...
+    description: |
+      Backward kernel for cuDNN attention, computing gradients of query,
+      key, and value from the cuDNN attention forward pass.  Uses a
+      BHSD-native two-kernel design that eliminates the BHSD-BSHD
+      permute+contiguous overhead.
+    for:
+      - _cudnn_attention_backward
+    labels:
+      - aten
+      - NoCPU
```
Caeruleann, line 2646:
```suggestion
      - NoCPU
      - KernelGen
```

tests/test_cudnn_attention_backward.py
```diff
@@ -0,0 +1,226 @@
...
+    return Q, K, V
+
+
+def cudnn_attn_forward_native(
+    Q,
+    K,
+    V,
+    attn_bias=None,
+    is_causal=False,
+    dropout_p=0.0,
```
Caeruleann, line 50: can you add test case for dropout_p > 0 or dropout_p < 0 ?

conf/operators.yaml
yzw1128: `NoCPU` and `KernelGen` are already present on the `cudnn_attention_backward` entry in `conf/operators.yaml` (added in commit `2d9d5579`), so no further change is needed.

tests/test_cudnn_attention_backward.py
yzw1128: Added a `dropout_p` parameter to the test (parametrized over `[0.0, 0.1]`) and threaded it through the forward and both backward calls. I also fixed the `philox_seed`/`philox_offset` extraction — they sit at indices 6/7 of `_scaled_dot_product_cudnn_attention`'s return tuple, not 2/3 (which are `cum_seq_q`/`cum_seq_k`).

One caveat: `dropout_p > 0` cases are skipped because the Triton kernels reconstruct the dropout mask with `tl.rand`, but cuDNN generates its own internal philox mask that cannot be reproduced (the masks are uncorrelated). cuDNN also rejects negative `dropout_p`, so only `dropout_p = 0.0` can be compared against the cuDNN reference.

kkkwb: Thanks. dropout_p is now handled: dropout_p > 0 raises NotImplementedError at the backward entry (matching the forward), and dropout_p < 0 is treated as no-dropout, equivalent to 0.0. Two tests are added to cover both cases. Commit 60b723f2a1270c56cf06da6c51aa81da64d66c9f.

PR #5173, opened 2026-08-04
[KMCompiler][Ascend]Add linalg_lstsq NPU backend support

cheersluvs:
## Summary
Adds a native **Ascend 910B** backend for `linalg_lstsq` (`src/flag_gems/runtime/backend/_ascend/ops/linalg_lstsq.py`). **Stacks on #4938**, which adds the operator itself — the backend override is swapped in by name, so the base registration must exist first. Review/merge that one first.

**Coverage matches the NVIDIA op: no size limit.** `m`, `n` and `nrhs` are all unbounded, square and near-square included, and there is **no CPU fallback on any path**. `float32` only — the device has no float64 unit, and a non-`float32` input raises rather than silently degrading.

## Testing
`pytest tests/test_linalg_lstsq.py` on Ascend 910B — **66 passed / 6 skipped**.

The 6 skips are 5 `float64` and 1 `complex`, both outside what the device or Triton can represent. They are gated on a **capability probe, not a vendor name**: the probe asks the operator (only `NotImplementedError` counts as unsupported; any other exception lets the test run and report), because this device reports `float64` tensors as genuine float64 — `element_size` 8, `(1 + 2**-40) - 1` exact — since those ops fall back to the CPU. `.dtype`, `element_size` and precision checks all answer "supported" here and the tests then fail rather than skip.

On NVIDIA nothing is skipped: the kernel implements both dtypes and all 72 cases still run.

## Performance
`torch.linalg.lstsq` has **no NPU kernel** and falls back to the CPU — measured **1417.9 ms vs 1306.5 ms** for the same call on CPU at 512x512, a ratio of 1.09 — so timing it measures CPU LAPACK and yields a "speedup" that says nothing about the device. `torch.linalg.qr` **is** native there (38.8 ms vs 1690.5 ms, ratio 0.02), so `benchmark/test_linalg_lstsq.py` composes the reference from QR the way the kernel itself computes it, for the same reason `benchmark/test_polygamma.py` composes its baseline. `torch.linalg.solve_triangular` in that composition is itself a CPU fallback, but ~1% of the total.

Numbers are **`--mode operator`** (end-to-end wall clock). `--mode kernel` is unusable here: it goes through the torch_npu profiler and intermittently returns `inf` for both sides.

| Shape `(*B, M, N)` | Baseline (ms) | Gems (ms) | Speedup |
|---|---|---|---|
| [16, 32, 2048] | 59.6583 | 3.6368 | **16.404** |
| [16, 8192, 16] | 102.1109 | 6.3905 | **15.979** |
| [64, 2048, 16] | 84.2552 | 6.1279 | **13.749** |
| [64, 32, 256] | 22.2359 | 2.0226 | **10.994** |
| [8, 64, 1024] | 33.6311 | 3.8630 | **8.706** |
| [8, 512, 512] | 299.5675 | 45.2169 | **6.625** |
| [8, 256, 256] | 47.7068 | 16.9030 | **2.822** |
| [8, 64, 256] | 7.1210 | 2.7000 | **2.637** |
| [16, 16, 1024] | 8.0499 | 3.1412 | **2.563** |
| [8, 4096, 8] | 8.5344 | 3.6981 | **2.308** |
| [64, 1024] | 4.4664 | 3.8376 | **1.164** |
| [512, 512] | 38.2689 | 34.0512 | **1.124** |
| [8, 16, 512] | 1.9896 | 2.6649 | 0.747 |
| [128, 512] | 5.3397 | 11.8656 | 0.450 |
| [4096, 512] | 560.6370 | 1443.6166 | 0.388 |
| [256, 256] | 6.0329 | 17.4541 | 0.346 |
| [4096, 8] | 1.1941 | 3.6516 | 0.327 |
| [256, 32] | 0.4585 | 1.4122 | 0.325 |
| [1024, 16] | 0.7316 | 2.5411 | 0.288 |
| [512, 1024] | 96.4260 | 395.7443 | 0.244 |
| **Arithmetic mean** | | | **4.410** |
| **Geometric mean** | | | **1.707** |

12 of 20 shapes at or above the baseline. **The distribution is deliberate and worth reading before the mean**: the strength is batched and tall shapes, where the composed reference runs one QR per matrix while this runs one grid. The weak rows are single small matrices — where the reference itself takes only 0.5–1.2 ms, so fixed overhead dominates — and the underdetermined path, which is correct but not yet optimised. Both are **launch-bound, not compute-bound**: total time tracks launch count almost exactly (e.g. `[256,256]` = 17 panels x 2 launches x ~0.47 ms ~= 16 ms). Known further levers, not implemented here: widening the panel to cut panel count, and blocking the underdetermined solve.

This table is **not comparable** to the NVIDIA table in #4938 — different reference (device `qr` composition vs cuSOLVER `lstsq`) and different mode.

## Files Changed
- `src/flag_gems/runtime/backend/_ascend/ops/linalg_lstsq.py`: native Ascend implementation (12 kernels)
- `src/flag_gems/runtime/backend/_ascend/ops/__init__.py`: import + `__all__` entry
- `benchmark/test_linalg_lstsq.py`: device-native composed baseline where `torch.linalg.lstsq` falls back to CPU (no change to the CUDA path)


benchmark/test_linalg_lstsq.py
```diff
@@ -0,0 +1,142 @@
...
+#
+# gems_op must also be passed explicitly: with a composed torch_op the harness
+# would otherwise time the gems side by re-running that composition under
+# use_gems, which never reaches this operator. flag_gems.linalg_lstsq is the
+# right handle -- SpecOpRegistrar rebinds that symbol to the vendor override at
+# import, so it is the Ascend kernel on Ascend.
+_DEVICE_REF = flag_gems.device not in ("cuda", "cpu")
+
+
+def _lstsq_via_qr(A, b, driver="gels"):
```
huangyiqun, line 28: Could we explicitly validate the driver argument here? Since torch.linalg.lstsq supports multiple drivers.

huangyiqun (commented):
It would be good to include test cases with rank-deficient matrices.
linalg_lstsq is expected to handle singular or near-singular inputs, so validating these cases would help ensure compatibility with PyTorch behavior.

benchmark/test_linalg_lstsq.py
cheersluvs: Good catch — Fixed in 049f5f7f. It now raises `ValueError` for anything other than `gels`/`None`, with a message pointing at adding the matching composition rather than relaxing the check.

cheersluvs:
Rank-deficient matrices test cases added in 9992b04c. The suite now covers one shape per path: single-tile (128,6), blocked TSQR (4096,200), compact-WY (256,256), and the underdetermined min-norm path (16,64).

Caeruleann (approved):

Issue #4995, opened 2026-07-27
[Bug] special_log_softmax reduces the wrong axis and writes out of bounds for a non-innermost dim


truong-v:
### Describe the bug

`special_log_softmax`'s small-N kernel indexes the input as `row_id * stride_m + col_offsets`, which assumes the N reduced elements are contiguous. That only holds when the reduced dim is the innermost one (`K == 1`). For any other dim it reduces along the wrong axis.

A second, independent bug in the same kernel: `tl.program_id` returns int32, so `row_id * stride_m` overflows above 2^31 elements. The rest of the file already uses `tle.program_id` (int64).

### Repro

```python
import torch, flag_gems

x = torch.randn(32, 64, device=flag_gems.device)
ref = torch.special.log_softmax(x, dim=0)
with flag_gems.use_gems():
    got = torch.special.log_softmax(x, dim=0)
print((got - ref).abs().max().item())   # 10.32, expected ~5e-07

# non-deterministic: 8 launches of the same input are not bit-identical
with flag_gems.use_gems():
    outs = [torch.special.log_softmax(x, dim=0) for _ in range(8)]
print(all(torch.equal(outs[0], o) for o in outs[1:]))   # False

# int32 overflow, needs ~17GB
y = torch.randn(2**22, 1024, dtype=torch.float16, device=flag_gems.device)
with flag_gems.use_gems():
    torch.special.log_softmax(y, dim=1)   # CUDA error: an illegal memory access was encountered
```

Measured `max|diff|` vs eager, fp32, small N:

| shape | dim | K | max abs diff |
|---|---|---|---|
| (32, 64) | 1 | 1 | 4.8e-07 (ok) |
| (32, 64) | 0 | 64 | 1.03e+01 |
| (4, 8, 16) | 1 | 16 | 1.23e+01 |
| (4, 8, 16) | 0 | 128 | 1.10e+01 |

The large-N path (`N > 1024`) is correct — it already strides by K.

The kernel was added in #3803.

### Environment

FlagGems at bcf3f5b7, torch 2.6.0+cu124, triton 3.2.0, NVIDIA H100 NVL.

I have a fix and regression tests ready and will open a PR referencing this issue.

PR #5186, opened 2026-08-04
[KernelGen][Nvidia] Add uniform operator with Triton kernel

ShawnsYing:
## Summary
Adds a Triton kernel for the out-of-place `uniform` operator (`aten::uniform`). It returns a new tensor with the same shape and dtype as the input, filled with values sampled from the continuous uniform distribution over `[from, to)`. The kernel reuses the existing Philox-based `uniform_kernel` and writes into a freshly allocated output, mirroring the `bernoulli` / `bernoulli_` precedent.

## Testing
- Parametrized tests over `DISTRIBUTION_SHAPES` and `FLOAT_DTYPES` (float16, float32, bfloat16)
- Validated against reference on device: output range is checked to lie within `[from, to]`, and the input tensor is verified to be left unmodified (true out-of-place semantics)
- Exercised through the real dispatch path via `torch.ops.aten.uniform` under `flag_gems.use_gems()`
- Tested on: Nvidia (H20)

## Performance
Test command: `pytest benchmark/test_uniform.py --level core` (NVIDIA H20)

### uniform

| float16 | [1073741824] | 4.184864 | 3.854288 | 1.086 |
| float16 | [64, 64] | 0.007296 | 0.007744 | 0.942 |
| float16 | [4096, 4096] | 0.076608 | 0.077504 | 0.988 |
| float16 | [64, 512, 512] | 0.076352 | 0.086272 | 0.885 |
| float16 | [1024, 1024, 1024] | 4.181888 | 5.670624 | 0.737 |
| float32 | [1073741824] | 5.196576 | 5.065376 | 1.026 |
| float32 | [64, 64] | 0.007360 | 0.007808 | 0.943 |
| float32 | [4096, 4096] | 0.089824 | 0.080992 | 1.109 |
| float32 | [64, 512, 512] | 0.089792 | 0.086752 | 1.035 |
| float32 | [1024, 1024, 1024] | 5.187520 | 6.006720 | 0.864 |
| bfloat16 | [1073741824] | 4.065776 | 4.188048 | 0.971 |
| bfloat16 | [64, 64] | 0.007264 | 0.007712 | 0.942 |
| bfloat16 | [4096, 4096] | 0.074912 | 0.077664 | 0.965 |
| bfloat16 | [64, 512, 512] | 0.074656 | 0.086112 | 0.867 |
| bfloat16 | [1024, 1024, 1024] | 4.055696 | 5.670336 | 0.715 |

| uniform | **0.938** |

`uniform` is a random-fill operator that is memory-bandwidth-bound, so latency tracks HBM bandwidth and sits near parity with the PyTorch eager implementation. No TFLOPS column: the benchmark does not report FLOPs for this distribution operator.

## Multi-backend Testing
|---|---|---|---|
| Nvidia (H20) | PASS (15 cases) | 0.938 | Primary |
| Tianshu | N/A | — | Not benchmarked |
| Muxi | N/A | — | Not benchmarked |
| Ascend | N/A | — | Not benchmarked |
| Hygon | N/A | — | Not benchmarked |

## Files Changed
- `src/flag_gems/ops/uniform.py`: Add out-of-place `uniform` wrapper reusing the Philox `uniform_kernel`
- `tests/test_uniform.py`: Add out-of-place accuracy test
- `benchmark/test_uniform.py`: Add out-of-place performance benchmark
- `conf/operators.yaml`: Add operator entry (kind: Distribution, stage: alpha 5.4)


ShawnsYing:
## KernelGen SimpleOpt exploration notes

This operator was developed with KernelGen SimpleOpt against the FlagGems v4 `flaggems_uniform` trace on an NVIDIA H20 (eval server, `triton` timing). The autonomous inner loop explored three kernel variants; the round-by-round result explains why the shipped kernel is the simple baseline that reuses the existing Philox `uniform_kernel`.

### Round-by-round

| Round | Change | geo_mean | vs best |
|-------|--------|----------|---------|
| R1 | Baseline pointwise kernel, `BLOCK=1024`, one RNG value per element | **0.992x** | best |
| R2 | Manual 4-way RNG unrolling (`tl.static_range`, 4 elems/iter) | 0.956x | −3.6% |
| R3 | `BLOCK=4096` (kept R2 unrolling) | 0.930x | −6.2% |

### Findings

- **R2 (unrolling) — falsified.** The hypothesis was that 4-way unrolling would improve RNG ILP and throughput. Measurement showed a regression. `tl.rand` calls in a `static_range` loop are serialized by the RNG state dependency, so unrolling yields no ILP for the RNG path, and the extra registers likely reduced occupancy.
- **R3 (larger block) — falsified.** The hypothesis was that a larger block would cut launch overhead and raise occupancy. It regressed further. Launch overhead is not the bottleneck here; the operator is RNG-throughput-bound, so per-element cost dominates and larger blocks reduce the number of blocks available for latency hiding.
- **Consistent pattern:** every optimization made things worse — the simplest kernel is the best. The operator is memory-bandwidth / RNG-throughput bound, so it sits at parity with PyTorch's native curand `uniform` (float32 essentially at parity; float16 workloads show the largest relative overhead, since Triton `tl.rand` cannot match curand for fp16 on H20).

### Consequence for this PR

Because tuning did not beat the baseline, the PR ships the idiomatic wrapper that reuses the production Philox `uniform_kernel` (the same kernel the merged in-place `uniform_` uses), allocating a fresh output via `torch.empty_like` — mirroring the `bernoulli` / `bernoulli_` precedent — rather than a bespoke tuned kernel. The ~0.94x arithmetic-mean benchmark speedup reflects the inherent RNG-bound nature of the op, not a missed optimization.

_Note: no NCU profiling was captured in this run, so the register-pressure / RNG-serialization root causes are measurement-consistent hypotheses rather than profiler-confirmed._

bin913:
@ShawnsYing Why do all data types perform poorly under the shape of 1024×1024×1024? Is there any room for performance improvement?

ShawnsYing:
@bin913 Good question! The poor performance at 1024×1024×1024 (1GB tensor) is due to the fundamental characteristics of this operator:

1. **RNG-bound workload**: `uniform` is primarily limited by the throughput of Triton's `tl.rand()` PRNG, not memory bandwidth or compute. The large tensor amplifies this bottleneck.

2. **cuRAND advantage**: PyTorch's native implementation uses highly optimized cuRAND kernels that are specifically tuned for large-scale random generation. Triton's Philox-based RNG cannot match cuRAND's throughput on large tensors, especially for fp16/bf16 where cuRAND has specialized paths.

3. **Exploration results**: During KernelGen optimization, we tested multiple strategies (unrolling for ILP, larger blocks for better occupancy) but all regressed further. The simplest kernel is actually the best because:
   - RNG state dependencies prevent ILP from unrolling
   - Larger blocks reduce parallelism without improving the RNG bottleneck
   - The per-element cost dominates over launch overhead

4. **Context**: This performance pattern is consistent with the in-place `uniform_` operator already merged (#4995), which uses the same underlying kernel. The out-of-place version simply adds an allocation overhead on top.

**Bottom line**: For 1GB+ uniform sampling workloads, Triton's RNG throughput is inherently slower than cuRAND. The geometric mean speedup of 0.94x reflects this fundamental limitation rather than a missed optimization opportunity. The operator is still valuable for integration with FlagGems' workflow and smaller tensors where the overhead is acceptable.

Let me know if you'd like to see NCU profiling data or have other concerns!

bin913:
> @bin913 Good question! The poor performance at 1024×1024×1024 (1GB tensor) is due to the fundamental characteristics of this operator:
> 
> 1. **RNG-bound workload**: `uniform` is primarily limited by the throughput of Triton's `tl.rand()` PRNG, not memory bandwidth or compute. The large tensor amplifies this bottleneck.
> 2. **cuRAND advantage**: PyTorch's native implementation uses highly optimized cuRAND kernels that are specifically tuned for large-scale random generation. Triton's Philox-based RNG cannot match cuRAND's throughput on large tensors, especially for fp16/bf16 where cuRAND has specialized paths.
> 3. **Exploration results**: During KernelGen optimization, we tested multiple strategies (unrolling for ILP, larger blocks for better occupancy) but all regressed further. The simplest kernel is actually the best because:
>    
>    * RNG state dependencies prevent ILP from unrolling
>    * Larger blocks reduce parallelism without improving the RNG bottleneck
>    * The per-element cost dominates over launch overhead
> 4. **Context**: This performance pattern is consistent with the in-place `uniform_` operator already merged ([[Bug] special_log_softmax reduces the wrong axis and writes out of bounds for a non-innermost dim #4995](https://github.com/flagos-ai/FlagGems/issues/4995)), which uses the same underlying kernel. The out-of-place version simply adds an allocation overhead on top.
> 
> **Bottom line**: For 1GB+ uniform sampling workloads, Triton's RNG throughput is inherently slower than cuRAND. The geometric mean speedup of 0.94x reflects this fundamental limitation rather than a missed optimization opportunity. The operator is still valuable for integration with FlagGems' workflow and smaller tensors where the overhead is acceptable.
> 
> Let me know if you'd like to see NCU profiling data or have other concerns!

Does FlagTree TLE bring any performance improvement for this scenario?

ShawnsYing:
@bin913 Great question — thanks for bringing up FlagTree TLE!

**Short answer:** TLE (Triton Language Extension, the C++ JIT runtime in [`libtriton_jit`](https://github.com/flagos-ai/libtriton_jit)) primarily reduces Python launch overhead via direct C++ dispatch. For this `uniform` operator, that overhead is negligible compared to the RNG throughput bottleneck, so **TLE won't materially improve the 0.94x speedup**.

**Why TLE doesn't help here:**

1. **RNG-bound, not launch-bound:** The KernelGen exploration notes show that the per-element cost dominates. The kernel generates 4 Philox RNG values per output block, and `tl.philox` + `uint_to_uniform_float` throughput is the bottleneck — not Python→kernel dispatch latency.

2. **Launch overhead is amortized away at scale:** The shapes where this operator performs worst (1GB `[1024, 1024, 1024]` tensor: 0.715x speedup) are precisely the workloads where Python launch overhead is irrelevant (one kernel launch amortizes across 1B elements). TLE's C++ fast path saves ~10–100 µs per launch, but a 1GB RNG fill takes 4–6 ms — the overhead is <1% of latency.

3. **Small-tensor regime:** For tiny shapes like `[64, 64]`, PyTorch's cuRAND also has comparable launch overhead, so any TLE gains are symmetric between Torch and Gems.

**The real bottleneck (and why cuRAND wins):**

PyTorch's native `uniform` uses the highly optimized cuRAND library kernels that have specialized fp16/bf16 fast paths and fused RNG→typecast→store pipelines. Triton's `tl.philox` is a generic PRNG that can't match cuRAND's throughput on H20 — especially for half-precision dtypes (the 0.715x regression at 1GB bf16 shows this clearly). This is a *fundamental Triton limitation*, not a launch-overhead issue.

**Where TLE *would* help:**

- Operators with **high launch frequency** (e.g., graph-mode where an op is invoked thousands of times per forward pass)
- Workloads where **kernel execution time < 100 µs** (so Python overhead is a meaningful fraction)
- This repo's hygon/metax backend-specialized ops already use `triton_lang_extension` for exactly those scenarios

**Precedent:**

The in-place `uniform_` operator (merged in #4995) uses the same Philox kernel and shows similar performance characteristics. The FlagGems project accepts RNG/distribution operators at near-parity (~0.9–1.1x) because they enable full FlagGems workflow integration, even when they can't beat vendor-optimized libraries.

**Conclusion:**

TLE is the right optimization for launch-bound or high-frequency ops, but for this memory-bandwidth + RNG-throughput-bound operator, it won't change the speedup meaningfully. If you'd like to explore TLE integration as a general FlagGems infrastructure improvement (benefiting *all* ops), that would be a great separate initiative — but it's orthogonal to this specific operator's performance profile.

Let me know if you'd like NCU profiling data or have other questions!

---

**References:**
- [FlagTree TLE / libtriton_jit](https://github.com/flagos-ai/libtriton_jit)
- [FlagFFT (uses TLE for JIT FFT kernels)](https://github.com/flagos-ai/FlagFFT)
- [FlagGems backend-specialized ops using TLE](https://github.com/flagos-ai/FlagGems/search?q=triton_lang_extension)

src/flag_gems/ops/uniform.py
```diff
@@ -76,3 +79,18 @@ def uniform_(self, from_=0.0, to=1.0, *, generator=None):
     with torch_device_fn.device(self.device):
         uniform_kernel[grid_fn](self, N, philox_seed, philox_offset, from_, to)
     return self
+
+
+def uniform(self, from_=0.0, to=1.0, *, generator=None):
```
huangyiqun, line 84: from_ > to currently produces values instead of raising as aten.uniform does. Please validate the interval before launch and add an error case.

tests/test_uniform.py
```diff
@@ -23,10 +23,20 @@
...
+
+
+@pytest.mark.uniform
+@pytest.mark.parametrize("shape", utils.DISTRIBUTION_SHAPES)
+@pytest.mark.parametrize("dtype", utils.FLOAT_DTYPES)
+def test_uniform(shape, dtype):
+    x = torch.randn(size=shape, dtype=dtype, device=flag_gems.device)
+    out = flag_gems.uniform(x, -3, 3)
+
+    assert (out <= 3.0).all()
```
huangyiqun, line 41: Please check out < 3.0 here. uniform samples from [from, to), and <= lets an upper-bound rounding bug pass.

tests/test_uniform.py
```diff
@@ -23,10 +23,20 @@
...
     assert (x <= 3.0).all()
     assert (x >= -3.0).all()
+
+
+@pytest.mark.uniform
+@pytest.mark.parametrize("shape", utils.DISTRIBUTION_SHAPES)
+@pytest.mark.parametrize("dtype", utils.FLOAT_DTYPES)
+def test_uniform(shape, dtype):
+    x = torch.randn(size=shape, dtype=dtype, device=flag_gems.device)
+    out = flag_gems.uniform(x, -3, 3)
```
huangyiqun, line 39: Please keep a copy of x and assert that it is unchanged. That is the main semantic difference from uniform_.

src/flag_gems/ops/uniform.py
ShawnsYing: Done — `uniform` now raises `RuntimeError` (matching the aten message) before launch when `from > to`, and `test_uniform_rejects_inverted_interval` covers it. Commit b790f600d.

tests/test_uniform.py
ShawnsYing: Done — switched the checks to strict `< to`. Good catch: it exposed a real bug. The kernel scaled in fp32 and stored to fp16/bf16, so values could round up to exactly `to`. The kernel now clamps to the output-dtype predecessor of `to` (one-ulp bit step), so sampled values stay in `[from, to)`. Commit b790f600d.

tests/test_uniform.py
ShawnsYing: Done — `test_uniform` now snapshots the input with `x.clone()` and asserts `torch.equal(x, x_copy)` after the call, covering the out-of-place vs in-place distinction. Commit b790f600d.

huangyiqun:
Please also cover an upper bound of 0.0. _next_below(+0.0) produces a NaN bit pattern, so this clamp can still allow zero after rounding.

bin913 (dismissed):

ShawnsYing:
@huangyiqun You were right, and thanks — this was a real bug, not a hypothetical. Fixed in a82105fc0.

**Confirmed the mechanism on device.** `_next_below` reads the float bit pattern as an integer and steps it down for `x >= 0`. At `+0.0` the pattern is `0`, and `0 - 1` wraps to all-ones; at `-0.0` the pattern is `0x8000...` and stepping *up* (the `+1` branch) wraps to all-ones too. Both are NaN in every float format. Probing the real kernel:

```
_next_below(+0.0)  fp16 / bf16 / fp32 / fp64 -> bits = 0xffff..ff  (NaN)
tl.minimum(+0.0, NaN) -> 0.0        # clamp silently does nothing
```

So the clamp let values at/above the exclusive bound through whenever `to == +0.0`:

| dtype | interval | before | after | aten |
|---|---|---|---|---|
| fp16 | [-1e-6, 0.0) | 31618 values `>= 0.0` | **0** | 0 |
| fp16 | [-1e-9, 0.0) | 1048576 values `>= 0.0` (all `-0.0`) | **0** | 0 |
| fp32/fp64/bf16 | all of the above | same class of leak | **0** | 0 |

**Fix.** Zero is the only input the one-ulp step cannot represent, so it is answered directly: the predecessor of `+0.0` or `-0.0` is `-(smallest positive subnormal)`, i.e. the negation of the bit pattern with a lone low bit. Verified bit-exact against `numpy.nextafter` per dtype:

```
fp16 next_below(+0.0) = 0x8001            (== nextafter(0, -1) fp16)
bf16 next_below(+0.0) = 0x8001
fp32 next_below(+0.0) = 0x80000001
fp64 next_below(+0.0) = 0x8000000000000001
```

and end-to-end the sampled distributions now match `aten` exactly for `to=0.0` (same NaN count, same `+0.0`/`-0.0` counts, same min/max).

**Regression test added.** `test_uniform_upper_bound_zero` over `FLOAT_DTYPES`, using `[-1e-6, 0.0)` where the scaled values actually round up to the bound so the clamp is exercised. I verified it is not a tautology: with the zero special-case disabled the fp16 case fails (31618 violating samples), and with the fix it passes. Full suite: 10 passed.

bin913 (approved):

PR #5261, opened 2026-08-05
[MTHREADS] Optimize flip for multi-dim tensors

Oslomayor:
Optimize flip for multi-dim tensors with trailing non-flipped dims.  

Add dedicated MTHREADS flip kernel that uses coalesced block reads/writes for tensors where trailing dimensions are not flipped (common case: flip dims [0,1] on 3D+ tensors). Falls back to generic implementation for 1D and all-flipped-inner cases where the existing approach works well.

Benchmark results before/after (speedup):
- [1024,1024,1024] dims=[0,1]: 0.48x -> 1.30-2.43x
- [64,512,512] dims=[0,1]:    0.61x -> 1.16-1.87x
- [1073741824] dims=[0]:      2.0x  -> 2.0x  (unchanged)
- [4096,4096] dims=[0,1]:    0.86x -> 0.86x (unchanged)
- [64,64] dims=[0,1]:        0.89x -> 0.89x (unchanged)


Oslomayor:
@tengqm It semms that CI is currently failing on the "Check GPU availability" step (exit code 1), which is blocking PR merges. Could you please take a look and fix it?

Oslomayor:
> @tengqm It semms that CI is currently failing on the "Check GPU availability" step (exit code 1), which is blocking PR merges. Could you please take a look and fix it?

It works now.

PR #5286, opened 2026-08-06
[KernelGen][Nvidia] Add _linalg_eigh operator

ShawnsYing:
## Summary

Adds a FlagGems implementation of `aten::_linalg_eigh`, which computes the eigenvalues and eigenvectors of a symmetric / Hermitian matrix (the operator behind `torch.linalg.eigh`).

The implementation mirrors the existing `_linalg_eigvals` operator: a Triton proxy kernel routes the input through the Triton runtime for buffer management, while `torch.linalg.eigh` (cuSOLVER / LAPACK) performs the actual decomposition. Following the `_linalg_eigvals` precedent, computation is routed via CPU to avoid recursively dispatching into the overridden CUDA kernel, then the result is moved back to the original device. Restricted to `float32` (and complex64/128), matching cuSOLVER's supported dtypes for eigendecomposition.

- `_linalg_eigh(A, UPLO="L", compute_v=True) -> (eigenvalues, eigenvectors)`
- `UPLO` selects whether the lower ("L") or upper ("U") triangle is used.
- `compute_v=False` returns an empty eigenvector tensor.

## Files

- `src/flag_gems/ops/_linalg_eigh.py` — operator implementation
- `tests/test_linalg_eigh.py` — accuracy tests
- `benchmark/test_linalg_eigh.py` — performance benchmark
- `src/flag_gems/ops/__init__.py`, `src/flag_gems/__init__.py`, `conf/operators.yaml` — registration

## Testing

- `tests/test_linalg_eigh.py`: **10/10 passed** on NVIDIA GPU (float32, UPLO ∈ {L, U}, shapes 2×2 … 20×20). Eigenvalues are compared directly; eigenvectors are validated via reconstruction `A = V diag(w) Vᵀ` and orthonormality `VᵀV = I`, since eigenvectors are only defined up to sign / basis for degenerate spectra.
- `benchmark/test_linalg_eigh.py`: all cases **SUCCESS**. Like `_linalg_eigvals`, the op is dominated by the host round-trip into LAPACK, so it runs around parity for large matrices; correctness is the goal here rather than a speedup.


src/flag_gems/ops/_linalg_eigh.py
```diff
@@ -0,0 +1,86 @@
...
+    # This ensures the operation goes through Triton runtime
+    output = torch.empty_like(inp)
+    n_elements = inp.numel()
+
+    # BLOCK_SIZE=128 balances occupancy and parallelism for typical matrix sizes
+    BLOCK_SIZE = 128
+    grid = (triton.cdiv(n_elements, BLOCK_SIZE),)
+
+    with torch.cuda.device(inp.device):
+        _linalg_eigh_proxy_kernel[grid](inp, output, n_elements, BLOCK_SIZE)
```
Caeruleann, line 73: so, who used `output` ?

PR #5292, opened 2026-08-06
[KernelGen][Nvidia] Add _linalg_slogdet operator

ShawnsYing:
## Summary
Adds the `aten::_linalg_slogdet` operator, the 4-output variant returning
`(sign, logabsdet, LU, pivots)` for a square matrix (or batch of square matrices).

The implementation reuses the Gaussian-elimination-with-partial-pivoting Triton
kernel that already backs the existing `linalg_slogdet`, extended to also write
back the in-place LU factorization and a schema-shaped `int32` pivot buffer, so it
matches the aten schema `_linalg_slogdet(Tensor A) -> (Tensor sign, Tensor logabsdet, Tensor LU, Tensor pivots)`.

`sign` and `logabsdet` are numerically well-defined and validated against the
PyTorch reference. `LU`/`pivots` follow the LAPACK factorization and are not
bit-reproducible by a Gaussian-elimination kernel, so the test compares only
`(sign, logabsdet)` via `torch.linalg.slogdet`. float32 only (matches the existing
`linalg_slogdet` kernel).

## Changes
- `src/flag_gems/ops/_linalg_slogdet.py` — new op + Triton kernel
- `src/flag_gems/ops/__init__.py` — import + `__all__`
- `src/flag_gems/__init__.py` — register `_linalg_slogdet`
- `conf/operators.yaml` — operator entry (`id: _linalg_slogdet`)
- `tests/test__linalg_slogdet.py`
- `benchmark/test__linalg_slogdet.py`

## Tested
- `pytest tests/test__linalg_slogdet.py` — 5/5 passed (fp32, shapes (2,3,3),(4,4),(8,8),(16,16),(32,32)) on NVIDIA H20.
- `benchmark/test__linalg_slogdet.py` — 5/5 SUCCESS; small matrices up to ~4.4x, larger matrices regress (single-program LU serializes), expected for this approach.


Caeruleann:
fix conflicts

conf/operators.yaml
```diff
@@ -13,6 +13,20 @@
 # limitations under the License.
 
 ops:
+  - id: _linalg_slogdet
```
```suggestion
  - id: linalg_slogdet
```

benchmark/test__linalg_slogdet.py
```diff
@@ -0,0 +1,53 @@
...
+    def set_shapes(self, shape_file_path=None):
+        self.shapes = SLOGDET_SHAPES
+
+    def get_input_iter(self, cur_dtype):
+        for shape in self.shapes:
+            A = torch.randn(shape, dtype=cur_dtype, device=self.device)
+            yield (A,)
+
+
+@pytest.mark.underscore_linalg_slogdet
```
Caeruleann:
```suggestion
@pytest.mark.linalg_slogdet
```

ShawnsYing:
@Caeruleann Conflicts fixed in commit 5f6106ed. Regarding the naming suggestions:

**yaml id**: The existing `linalg_slogdet` operator (merged in PR #4255, July 2026) already uses `id: linalg_slogdet` in operators.yaml (line 6662). This PR adds the private 4-output variant `aten::_linalg_slogdet`, so I kept `id: _linalg_slogdet` (line 6674) to avoid a duplicate id error.

**pytest mark**: Similarly, `tests/test_linalg_slogdet.py` already uses `@pytest.mark.linalg_slogdet`. Changing `test__linalg_slogdet.py` to the same mark would create ambiguity in test selection. The current `@pytest.mark.underscore_linalg_slogdet` distinguishes the private variant's tests from the public op's tests.

The FlagGems convention for underscore-prefixed aten ops is: keep the underscore in the yaml `id` (examples: `_reshape_alias`, `_weight_norm`, `_flash_attention_forward` all keep it), and drop it in the pytest mark when no collision exists (e.g., `_batch_norm_impl_index` → `batch_norm_impl_index`). Here, dropping the underscore would collide with the existing public op, so I prefixed with `underscore_` to disambiguate.

Let me know if you'd prefer a different approach!

ShawnsYing:
Additionally fixed in df4c7689: removed `use_gems()` call from the test to pass the new CI check `check-kernelgen-tests`, which forbids `use_gems()` in KernelGen operator tests. The aten op registration is already active via `_FULL_CONFIG`, so the test now calls `torch.ops.aten._linalg_slogdet(A)` directly.

ShawnsYing:
Correction to my earlier reply: @Caeruleann was right about the pytest marker. The `check-operator-markers` CI check requires the marker to exactly match the operator id `_linalg_slogdet` — and since it starts with an underscore, it is a distinct marker from the public `linalg_slogdet`, so there is no collision after all.

Fixed in 8845c4684 using the same `MarkDecorator` registration pattern as `tests/test__reshape_alias.py` (pytest refuses attribute-access marker generation for underscore-prefixed names):

```python
setattr(
    pytest.mark,
    "_linalg_slogdet",
    MarkDecorator(Mark("_linalg_slogdet", (), {}, _ispytest=True), _ispytest=True),
)
```

Both `tests/test__linalg_slogdet.py` and `benchmark/test__linalg_slogdet.py` now use `@pytest.mark._linalg_slogdet`. Verified locally that the marker check passes and `-m _linalg_slogdet` selects the tests.

The yaml `id: _linalg_slogdet` still needs the underscore though, since `id: linalg_slogdet` would duplicate the existing public op (line 6662).

bin913:

bin913:

ShawnsYing:
@bin913 Benchmark results below. Running them first surfaced a real performance bug, so this push also fixes it.

**Environment**: NVIDIA H20, fp32, `mode=kernel` (triton do_bench, median), comprehensive level, via `benchmark/test__linalg_slogdet.py`.

**Before this push** (the kernel as generated did the elimination one scalar at a time — every inner-loop element was its own `tl.load`/`tl.store`):

|---|---|---|---|
| (2,3,3) | 0.0422 | 0.0120 | 3.53x |
| (4,4) | 0.0535 | 0.0116 | 4.59x |
| (8,8) | 0.0587 | 0.0237 | 2.48x |
| (16,16) | 0.0701 | 0.1162 | **0.60x** |
| (32,32) | 0.0916 | 0.6077 | **0.15x** |

**After** (tile held in registers, whole-row/pivot operations):

|---|---|---|---|
| (2,3,3) | 0.0442 | 0.0125 | 3.52x |
| (4,4) | 0.0559 | 0.0129 | 4.33x |
| (8,8) | 0.0611 | 0.0157 | 3.88x |
| (16,16) | 0.0726 | 0.0206 | **3.53x** |
| (32,32) | 0.0941 | 0.0393 | **2.40x** |

16x16 goes from 0.60x to 3.53x and 32x32 from 0.15x to 2.40x — i.e. the sizes the PR's shape list targets were the ones being lost. Beyond the benchmark list the improvement is larger still: n=64 4.06 ms -> 0.120 ms, n=128 26.7 ms -> 0.666 ms.

**Change**: hold the whole `(BLOCK_SIZE, BLOCK_SIZE)` tile in registers and drive pivot search / row swap / multipliers / trailing update with whole-row and tile ops; `BLOCK_SIZE = next_power_of_2(n)` capped at 128 (largest fp32 tile that fits the register budget), above which it raises `NotImplementedError` instead of silently degrading (the old path took 26 ms at n=128).

**Numerics unchanged**: same algorithm and same formula, verified bit-identical to the previous kernel on 96 cases (n=2..32 across well-conditioned / random / exactly-singular / zero-row inputs, `equal_nan`), and 5/5 accuracy tests pass.

bin913 (approved):

PR #5332, opened 2026-08-07
fix(ops): skip triton kernels for complex dtypes in empty/lift_fresh/…

CLYtou:
### PR Category
Operator

### Type of Change
Bug Fix

### Description

Triton's `type_canonicalisation_dict` on some backends (e.g. MetaX) lacks `complex32`/`complex64`/`complex128` entries, causing `KeyError` when lowering complex pointer types. This PR adds early-return paths that bypass the triton kernel for complex dtypes:

- **`empty.py`**: skip the triton store kernel for complex dtypes (torch already zero-initializes the allocation)
- **`lift_fresh.py`**: use `clone()` instead of triton copy for complex tensors
- **`randn.py`**: generate randn in real dtype then `view_as_complex`

These changes are safe for all backends since they only add a fast-path that avoids triton for complex types. Non-complex dtype behavior is completely unchanged.

### Issue

- Resolves KeyError in `type_canonicalisation_dict` for complex dtypes on MetaX C550 backend
- Affects operators: `aten.empty`, `aten.lift_fresh_copy`, `aten.normal`

### Progress

### Performance

No performance regression — the complex fast-path uses `clone()` / `view_as_complex` which are native PyTorch ops. For non-complex dtypes, execution path is identical to before (no extra branches in the hot path).

### DIFF
--- a/src/flag_gems/ops/empty.py
+++ b/src/flag_gems/ops/empty.py
@@ -81,6 +81,11 @@ def empty(
         device=device,
         pin_memory=pin_memory,
     )
+    # Skip triton kernel for complex dtypes — triton cannot canonicalize complex
+    # pointer types on some backends, and empty() returns uninitialized memory
+    # anyway so skipping the store is functionally safe for all backends.
+    if dtype.is_complex:
+        return out
     N = volume(shape)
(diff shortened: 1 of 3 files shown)


huangyiqun (approved):
LGTM

PR #3458, opened 2026-05-24
Fix scatter_reduce_two kernel

tengqm:
### PR Category

Operator

### Type of Change

Bug Fix

### Description

For the operator, changed `src_shapes = _pad5(list(src.shape), 1)` to `idx_shapes = _pad5(list(index.shape), 1)` — the 5D coordinate decomposition should use the index tensor's shape.

For 2D kernels (sum, prod, mean, amax), added `idx_ncols` parameter for row/col decomposition and index addressing; kept `src_ncols` for src addressing.

The revision is to make sure that each kernel now correctly:

- decomposes flat offsets using `idx_ncols` (matching index layout)
- reads `index` at `row * idx_ncols + col`
- reads `src` at `row * src_ncols + col` (valid since `index.shape[1] <= src.shape[1]`)

All 2D launch sites now pass `index.shape[1]` as `idx_ncols` before `src.shape[1]` as `src_ncols`.

Removed @pytest.mark.skip from tests/test_scatter_reduce_two.py.

### Issue

Fixes #3435

PR #2914, opened 2026-05-05
[FlagGems Operator Development Competition] scatter_reduce

gaibianshiji:
## Summary

This PR implements `torch.scatter_reduce` as a new Triton-accelerated operator for FlagGems, supporting all 5 reduce modes (sum, prod, mean, amax, amin) with full 1D-5D tensor support.

The implementation registers the correct ATen overload names: `scatter_reduce.two`, `scatter_reduce_.two`, `scatter_reduce.two_out`.

## Data Type Support

| dtype | Status | Notes |
|-------|--------|-------|
| **float32** | Fully supported | All 5 reduce modes, all shapes |
| **float16** | Fully supported | All 5 reduce modes, all shapes |
| **bfloat16** | Fully supported | All 5 reduce modes, all shapes |

All three floating-point types are tested with 248 pytest tests and benchmarked on NVIDIA RTX 4090.

## Key Features

- **All reduce modes**: sum, prod, mean, amax, amin
- **Full dimensionality**: 1D-5D tensors via uniform 5D coordinate decoding
- **include_self**: Both True and False modes supported
- **In-place variant**: `scatter_reduce_` via `Tensor.scatter_reduce_()`
- **Out variant**: `scatter_reduce_out`
- **Data types**: float32, float16, bfloat16
- **Different src/inp shapes**: Correctly handles cases where src has different dimensions than inp (e.g., src=(1024,2048), inp=(1024,1024))

### 2D Fast Path
Specialized kernels for 2D tensors (the most common case) avoid the overhead of 5D coordinate decoding. Uses direct `row = offsets // ncols, col = offsets % ncols` instead of 4 divisions and 4 modulos per element.

**Separate src/out ncols**: The 2D kernels use separate `src_ncols` and `out_ncols` parameters to correctly handle cases where src and inp have different column counts.

### LOOP=4 Amortization
Each Triton program processes `LOOP * BLOCK` elements instead of just `BLOCK`. This amortizes kernel launch overhead.

### Vendor-Aware Block Size
- **NVIDIA**: BLOCK=128 (balances occupancy and register pressure)
- **Iluvatar/Metax**: BLOCK=256 (better occupancy on these GPUs)

### Native Atomics on NVIDIA
NVIDIA GPUs have native `tl.atomic_max` and `tl.atomic_min`, so no CAS fallback is needed for amax/amin reduce modes.

### CAS Fallback for Iluvatar
[...]

PR #5339, opened 2026-08-07
[SiliconFlow] Fix scatter_reduce precision across backends

### LINKED ISSUE #3435: The `scatter_reduce_two` operator fails CI test
### Describe the bug

The error report is as follows:

```
tests/test_scatter_reduce_two.py:50: 
_ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ 
tests/accuracy_utils.py:250: in gems_assert_close
    flag_gems.testing.assert_close(
_ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ 

res = tensor([[[-0.2598,  0.8488, -1.0717,  ...,  0.1883,  0.9341,  0.5032],
         [ 0.3707,  1.8031,  1.0066,  ...,  0.5...17, -0.0270,  0.1527],
         [-0.6479, -2.0155,  0.6102,  ...,  0.9348,  0.4525, -1.1211]]],
       device='cuda:0')
ref = tensor([[[-0.2598,  0.8488, -1.0717,  ...,  0.1883,  0.9341,  0.5032],
         [ 0.3707,  1.8031,  1.0066,  ...,  0.5...17, -0.0270,  0.1527],
         [-0.6479, -2.0155,  0.6102,  ...,  0.9348,  0.4525, -1.1211]]],
       device='cuda:0')
dtype = torch.float32, equal_nan = False, reduce_dim = 1, atol = 0.0001

    def assert_close(res, ref, dtype, equal_nan=False, reduce_dim=1, atol=1e-4):
        if dtype is None:
            dtype = torch.float32
        assert res.dtype == dtype
        ref = ref.to(dtype)
        res, ref = _maybe_move_to_cpu(res, ref)
        rtol = RESOLUTION[dtype]
>       torch.testing.assert_close(
            res, ref, atol=atol * reduce_dim, rtol=rtol, equal_nan=equal_nan
        )
E       AssertionError: Tensor-likes are not close!
E       
E       Mismatched elements: 426 / 2097152 (0.0%)
E       Greatest absolute difference: 2.269956588745117 at index (69, 2, 2) (up to 0.0001 allowed)
E       Greatest relative difference: 1972.071533203125 at index (52, 8, 2) (up to 1.3e-06 allowed)
```

winfan1314:
### PR Category

Operator, OP Test, Benchmark

### Type of Change

Bug Fix, Refactor, Performance Optimization

### Description

Fix `scatter_reduce` semantics and extend optimized execution for the functional, in-place, and `out=` overloads across supported backends.

- Correct the generic implementation for dimension validation, empty-index behavior, dtype-specific identities, `include_self=False` untouched destinations, mean contribution counts, output dtype validation, and overload aliasing contracts.
- Accumulate floating-point reductions in FP32, use touched masks where required, and implement product updates with bitwise CAS or an output-owned product scan to preserve NaN and signed-zero behavior.
- Canonicalize tensors above five dimensions, plus selected large 5-D extrema cases, into an equivalent 3-D active-prefix problem and copy only the active result domain back.
- Add a shared short-row framework used by NVIDIA and MetaX routing and reused by backend overrides. Unsupported cases remain on FlagGems Triton paths; this change does not introduce a native `scatter_reduce` fallback.
- Add and export Ascend implementations for all three overloads. Triton 3.2 uses replay-safe sorted segments for `sum`, `mean`, and large `prod`; Triton 3.5+ uses source-program claims and a product lock. Smaller products use the output-owned scan.
- Add and export Hygon implementations for all three overloads, including short-row reductions, row-owned atomics, packed 16-bit reductions, linked-list products, and direct in-place large-row addition. The current implementation also handles Triton 3.6 compiled-launch ABI reuse.
- Add and export MThreads implementations for all three overloads. Eligible products use a short-row path or an output-owned linked-list reduction; other cases reuse the corrected generic implementation.
- Keep Iluvatar extrema on the generic CAS fallback and route eligible MetaX/NVIDIA short rows through vendor-tuned generic policies.

### Tests and Configuration

- Replace the narrow accuracy matrix with coverage for all three overloads, five reductions, default/false `include_self`, 1-D through 8-D inputs, active prefixes, empty and non-contiguous tensors, contention, NaN/signed-zero behavior, invalid arguments, and Hygon regression cases.
- Use public `flag_gems` APIs. The current full-mode `tests/test_scatter_reduce.py` collection contains 1,343 nodes.
- Replace the benchmark with 30 nodes covering three overloads, five reductions, and two `include_self` forms. Ascend is skipped because the native comparison falls back to CPU.
- Add CI aliases for the three exact overload names and operator metadata for `scatter_reduce.two` and `scatter_reduce.two_out`.

### Related PRs

- #2914 introduced the original implementation.
- #3458 fixed #3435.

### Progress

### Performance

The updated benchmark exercises every overload/reduction combination on the existing core shapes. Detailed per-device logs are omitted from this description; Ascend results are intentionally unavailable because its native baseline is CPU fallback.


src/flag_gems/runtime/backend/_hygon/ops/scatter_reduce.py
```diff
@@ -0,0 +1,98 @@
...
+
+logger = logging.getLogger(__name__)
+
+# The deterministic kernel launches one program per output. HCU rejects a
+# one-dimensional grid wider than 65535; large, low-contention workloads use
+# the generic parallel CAS kernel instead.
+_MAX_SCAN_OUTPUTS = 65535
+
+
+def _use_prod_scan(inp, reduce):
```
huangyiqun: Could we document the rationale behind _MAX_SCAN_OUTPUTS?

src/flag_gems/runtime/backend/_hygon/ops/scatter_reduce.py
```diff
@@ -0,0 +1,98 @@
...
+        return _generic_scatter_reduce(
+            inp,
+            dim,
+            index,
+            src,
+            reduce,
+            include_self=include_self,
+        )
+    logger.debug("GEMS_HYGON SCATTER_REDUCE_TWO")
+    assert inp.ndim <= 5, f"scatter_reduce supports up to 5D tensors, got {inp.ndim}D"
```
huangyiqun: Is this behavior consistent with PyTorch?

src/flag_gems/runtime/backend/_ascend/ops/scatter_reduce.py
```diff
@@ -0,0 +1,110 @@
...
+        return _generic_scatter_reduce(
+            inp,
+            dim,
+            index,
+            src,
+            reduce,
+            include_self=include_self,
+        )
+    logger.debug("GEMS_ASCEND SCATTER_REDUCE_TWO")
+    assert inp.ndim <= 5, f"scatter_reduce supports up to 5D tensors, got {inp.ndim}D"
```
huangyiqun: Is this behavior consistent with PyTorch?

src/flag_gems/runtime/backend/_hygon/ops/scatter_reduce.py
winfan1314: Sure. `_MAX_SCAN_OUTPUTS` reflects HCU's maximum supported size for a one-dimensional launch grid. The deterministic product-scan kernel launches one Triton program per output element, so an output with more than 65,535 elements would produce an invalid grid on HCU. Larger outputs therefore use the generic CAS kernel, which avoids this launch limit and is better suited to large, low-contention workloads. This threshold only selects the implementation path and does not change the operator semantics.

src/flag_gems/runtime/backend/_hygon/ops/scatter_reduce.py
winfan1314: Thanks for pointing this out. You are right: the `ndim <= 5` restriction is not part of the PyTorch semantics. It is an implementation limitation inherited from the current generic FlagGems kernel, whose coordinate decoder is fixed to five axes.

Simply removing the assertion would not be correct. The existing kernels compute a padded five-dimensional `dim` and consume only five shape/stride entries, so tensors with more than five dimensions could be indexed incorrectly.

We propose removing the user-visible rank restriction by canonicalizing high-rank inputs into an equivalent 3-D `(outer, scatter_dim, inner)` problem while keeping the existing kernels unchanged. Since `index` may be smaller than `self` or `src`, we will first slice the active prefix region addressed by `index`, keeping the full scatter dimension of `self`. After making the working tensors contiguous, we can reshape them to 3-D, run the existing reduction on dimension 1, and copy the active result back into a clone of `self`. Values outside the active region remain unchanged.

This approach applies to both the generic implementation and the Ascend/Hygon product-scan paths, without introducing a CPU or native-operator fallback. The existing <=5-D fast paths will remain unchanged.

We will add 6-D and 8-D tests covering equal and unequal valid shapes, first/middle/last and negative dimensions, non-contiguous tensors, all reduction modes, both `include_self` values, and the functional, in-place, and out overloads.

src/flag_gems/runtime/backend/_ascend/ops/scatter_reduce.py
winfan1314: The same limitation and proposed high-rank canonicalization apply to the Ascend product-scan path as well. We will address both backend paths through the shared implementation.

conf/operators.yaml
```diff
@@ -10056,6 +10056,32 @@ ops:
...
+      This entry covers the functional `scatter_reduce.two` ATen overload.
+    for:
+      - scatter_reduce.two
+    labels:
+      - aten
+      - KernelGen
+    kind:
+      - Reduction
+    stages:
+      - alpha: '5.3'
```
```suggestion
      - alpha: '5.4'
```

conf/operators.yaml
```diff
@@ -10056,6 +10056,32 @@ ops:
...
+      This entry covers the `scatter_reduce.two_out` ATen overload.
+    for:
+      - scatter_reduce.two_out
+    labels:
+      - aten
+      - KernelGen
+    kind:
+      - Reduction
+    stages:
+      - alpha: '5.3'
```
```suggestion
      - alpha: '5.4'
```

conf/operators.yaml
```diff
@@ -10056,6 +10056,32 @@ ops:
...
+  - id: scatter_reduce_two_out
+    description: |
+      Applies a scatter reduction and writes the result to the provided `out` tensor.
+      This entry covers the `scatter_reduce.two_out` ATen overload.
+    for:
+      - scatter_reduce.two_out
+    labels:
+      - aten
+      - KernelGen
```
```suggestion
```

conf/operators.yaml
```diff
@@ -10056,6 +10056,32 @@ ops:
...
       - stable: '3.0'
+  - id: scatter_reduce_two
+    description: |
+      Applies a reduction to source values scattered at the indices specified by `index`.
+      This entry covers the functional `scatter_reduce.two` ATen overload.
+    for:
+      - scatter_reduce.two
+    labels:
+      - aten
+      - KernelGen
```
```suggestion
```

modao1234 (approved):

PR #5355, opened 2026-08-10
[KernelGen][Nvidia] add mkldnn_rnn_layer (single-layer LSTM) Triton kernel

ShawnsYing:
## Summary

Adds a Triton implementation of the aten `mkldnn_rnn_layer` operator: a single-layer, unidirectional LSTM (mode=2). The upstream aten op is oneDNN-backed and CPU-only; this brings the same computation to CUDA via a hand-written Triton kernel generated and tuned with KernelGen.

## Design

- One Triton program per batch element (`grid=(batch,)`).
- All weights and biases are hoisted out of the time-step loop.
- fp32 accumulation internally, so fp16/bf16 inputs stay accurate over the sequence.
- Gate order i, f, g, o (i/f/o sigmoid, g tanh); `c' = f*c + i*g`, `h' = o*tanh(c')`.
- `num_warps` chosen by hidden size (H>=64 -> 8, H>=16 -> 4, else 2).
- A `torch.autograd.Function` wrapper supplies backward by recomputing the LSTM in native PyTorch (the oneDNN reserve/workspace is opaque and unusable outside oneDNN).
- The public wrapper matches the 16-arg aten schema and raises `NotImplementedError` for the unsupported configurations (mode != LSTM, multi-layer, bidirectional, batch_first, packed sequences).

## Tests

`tests/test_mkldnn_rnn_layer.py` — 193 cases pass on H20:
- Accuracy via `flag_gems.use_gems()` dispatch and via the direct wrapper, compared against a CPU-routed oneDNN reference across fp16/fp32/bf16, forward and reverse.
- Backward gradients compared against a native PyTorch LSTM recomputation (7 gradients).

Reference inputs go through `utils.to_reference`; oneDNN is CPU-only so the reference is computed on CPU and moved back.

## Benchmark

`benchmark/test_mkldnn_rnn_layer.py --level core` passes. Baseline is the CPU-only oneDNN op (there is no CUDA torch reference), so the reported speedups reflect CUDA-vs-CPU and are large across shapes.

Test command: `pytest benchmark/test_mkldnn_rnn_layer.py --level core` (NVIDIA H20)


src/flag_gems/ops/__init__.py
```diff
@@ -1223,6 +1219,7 @@
     "logaddexp2_out",
     "logcumsumexp",
     "logcumsumexp_out",
+    "mkldnn_rnn_layer",
```
Caeruleann: sort by a-z

src/flag_gems/ops/mkldnn_rnn_layer.py
```diff
@@ -0,0 +1,474 @@
...
+    cx_,
+    reverse,
+    batch_sizes,
+    mode,
+    hidden_size,
+    num_layers,
+    has_biases,
+    bidirectional,
+    batch_first,
+    train,
```
Caeruleann, line 448: who used `train`?

src/flag_gems/ops/__init__.py
ShawnsYing: Fixed. The `"mkldnn_rnn_layer"` entry in `__all__` had been dropped after `logcumsumexp_out`; it is now moved to its correct alphabetical position between `mish_backward` and `mm`, matching the import order. While here I also rebased onto the latest `upstream/master`, which removed an unrelated reformatting churn that had crept into this file, so the diff is now just the two `mkldnn_rnn_layer` lines.

src/flag_gems/ops/mkldnn_rnn_layer.py
ShawnsYing: Good catch, nobody. `train` is the 16th argument of the aten `mkldnn_rnn_layer` schema, so the wrapper has to accept it positionally for dispatch, but it does not affect the result here: a single-layer LSTM has no inter-layer dropout, so the forward output is train-independent, and backward is supplied by `MkldnnRnnLayerFunction` rather than a oneDNN train-mode reserve/workspace. I have made that explicit by adding a `del train` with a comment explaining why it is unused, so it is clear the omission is intentional rather than an oversight.

Caeruleann (approved):

PR #5367, opened 2026-08-10
[KernelGen][Nvidia] Add sym_constrain_range operator

ShawnsYing:
## Summary
Implements `torch.ops.aten.sym_constrain_range` as a no-op operator for symbolic shape tracing compatibility.

## Changes
- **New operator**: `sym_constrain_range` in `src/flag_gems/ops/sym_constrain_range.py`
- **Registration**: Added to `__init__.py` and `ops/__init__.py`
- **Metadata**: Added entry to `conf/operators.yaml`
- **Tests**: Added unit tests in `tests/test_special_ops.py`

## Implementation Details
Following the established pattern from `sym_size`, `sym_stride`, and `sym_storage_offset`:
- Returns `None` (no-op)
- Used for symbolic integer range constraints during tracing/compilation
- No runtime computation required

## Testing
All tests pass:
```
✓ Test 1: Valid range (10, min=0, max=100) - PASSED
✓ Test 2: Min only (50, min=0) - PASSED
✓ Test 3: Max only (50, max=100) - PASSED
✓ Test 4: No constraints (42) - PASSED
✓ Test 5: Negative range (-5, min=-10, max=0) - PASSED
```

## Checklist
- [x] Implementation follows existing patterns
- [x] Operator registered in all required locations
- [x] Unit tests added and passing
- [x] Metadata added to operators.yaml


src/flag_gems/ops/sym_constrain_range.py
```diff
@@ -0,0 +1,46 @@
...
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+import logging
+
+logger = logging.getLogger(__name__)
+
+
+def sym_constrain_range(size, *, min=None, max=None):
```
Caeruleann, line 55: so, it's not a triton operator?

ShawnsYing: Correct — `sym_constrain_range` has no Triton kernel doing real compute, and that's by design. It's `aten::sym_constrain_range(Scalar size, *, int? min=None, int? max=None) -> ()`: a symbolic-shape *compiler hint* used during `torch.compile` tracing. It validates that a scalar (unbacked) symbolic integer falls within `[min, max]` and returns nothing (void). There is no tensor to compute on the device, so there is nothing for a kernel to do.

This mirrors the sym_* operators already in the tree (`sym_stride`, `sym_storage_offset`): they return tensor metadata rather than a computed tensor, and each keeps only a placeholder `@triton.jit` kernel for family parity. I've updated this PR to follow that same pattern:
- added the KernelGen file header and a placeholder `sym_constrain_range_kernel` (documented as a no-op) so it's structurally consistent with the rest of the sym_* family;
- replaced the ad-hoc `tests/test_special_ops.py` with `tests/test_sym_constrain_range.py`, which runs under `flag_gems.use_gems()` and checks the result against the ATen reference (in-range returns `None`, out-of-range raises), plus a matching `benchmark/test_sym_constrain_range.py`.

So: intentionally not a compute Triton operator, and now aligned with the existing sym_* precedent. Happy to adjust if you'd prefer a different convention for these metadata/compiler-hint ops.

ShawnsYing: Good catch — I've reworked it into a real Triton operator so the range check runs on device rather than in host Python.

`sym_constrain_range_kernel` now loads the scalar `size` and its `[min, max]` bounds onto the device, computes the range-violation flag with `tl` comparisons (`size < lo`, `size > hi`), and stores it. The Python wrapper only marshals the scalars into int64 device tensors, launches the kernel, reads the flag back, and raises `RuntimeError` on violation — matching the ATen reference exactly (`aten::sym_constrain_range(Scalar size, *, int? min=None, int? max=None) -> ()`, void). Absent `min`/`max` map to int64 sentinels so an omitted bound is genuinely unbounded.

Verification (H20, `NVIDIA H20-3e`):
- `tests/test_sym_constrain_range.py` — 13 passed. Includes a case that calls the FlagGems wrapper directly and asserts the Triton path actually runs (via the `GEMS SYM_CONSTRAIN_RANGE` debug log), plus in-range→`None` and out-of-range→`RuntimeError` parametrizations, all cross-checked against the ATen reference.
- `benchmark/test_sym_constrain_range.py` — passes; `gems_op` points at `flag_gems.sym_constrain_range` so the Triton kernel is what's timed.

I also replaced the earlier ad-hoc `tests/test_special_ops.py` with the standard per-op `tests/test_sym_constrain_range.py`, and fixed the `operators.yaml` ordering/line-length. Let me know if you'd prefer a different kernel shape here.

Caeruleann (approved):

PR #5368, opened 2026-08-10
Allow using tf32 in tl.dot

0x45f:
### PR Category

Operator
### Type of Change

Other
### Description

Allow using tf32 in tl.dot
### Progress

### DIFF
--- a/src/flag_gems/config.py
+++ b/src/flag_gems/config.py
@@ -29,6 +29,10 @@ has_c_extension = False
 use_c_extension = False
 aten_patch_list = []
 
+# Global allow_tf32 configuration for Triton kernels
+# Controlled by FLAGGEMS_ALLOW_TF32 environment variable
+_ALLOW_TF32 = os.environ.get("FLAGGEMS_ALLOW_TF32", "0") == "1"
+
 # set FLAGGEMS_SOURCE_DIR for cpp extension to find
 os.environ["FLAGGEMS_SOURCE_DIR"] = str(Path(__file__).parent.resolve())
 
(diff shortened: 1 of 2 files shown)


huangyiqun (approved):

src/flag_gems/config.py
```diff
@@ -29,6 +29,12 @@
 use_c_extension = False
 aten_patch_list = []
 
+# Global allow_tf32 configuration for Triton kernels
+# Controlled by FLAGGEMS_ALLOW_TF32 environment variable
+# Set FLAGGEMS_ALLOW_TF32=1 to enable TF32 (faster but lower precision)
+# Default is False (disabled) for numerical accuracy
+ALLOW_TF32 = os.environ.get("FLAGGEMS_ALLOW_TF32", "0") == "1"
```
tengqm, line 36: This name is too generic ... as a user, I don't know whether enabling this would impact all operators.

src/flag_gems/ops/addbmm.py
```diff
@@ -84,7 +85,7 @@ def addbmm_kernel(
                 a = a.to(tl.float32)
                 b_tile = b_tile.to(tl.float32)
 
-            acc += tl.dot(a, b_tile, allow_tf32=False)
+            acc += tl.dot(a, b_tile, allow_tf32=ALLOW_TF32)
```
tengqm, line 88: Reading this line ... I'm more inclined to hard code this to True.
`allow_tf32` gives us the compatibility, which is an advantage.
TF32 is allowed, rather than forced ... IIUC, that means we breaks nothing.

PR #5282, opened 2026-08-06
[AddMM] Optimize matrix layouts and bias epilogue

ZYX223:
### PR Category

Operator, OP Test, Benchmark

### Type of Change

Performance Optimization

### Description

This PR optimizes `addmm`, `addmm.out`, `addmm.dtype`, and
`addmm.dtype_out` for common matrix layouts and bias forms while preserving the
existing AddMM numerical contract.

The implementation:

- shares one launch implementation across the four public AddMM APIs, including
  non-contiguous output strides;
- keeps row- and column-contiguous matrix views in their original layout and
  materializes only general non-contiguous views;
- loads vector bias once per output N tile, loads scalar bias once, and uses the
  broadcast view strides for other valid bias shapes instead of materializing a
  full `[M, N]` tensor;
- writes dtype overloads directly into the requested output tensor, avoiding an
  input-dtype intermediate for FP16/BF16-to-FP32 accumulation;
- preserves the upstream common two-dimensional program grid and masked K loop;
- aligns the Ascend implementation with the backend `mm` grouped ordering,
  tuning configuration, and heuristics;
- provides MThreads FMA and SQMMA paths with the same bias epilogue and output
  API support;
- uses one grouped MetaX kernel while retaining the latest upstream C550
  fallback and tuning behavior;
- adds backend tuning entries for MThreads and T-Head without introducing a
  cross-backend precision policy.

FP32 dot behavior remains strict (`allow_tf32=False`) in this PR.

### Issue

N/A

### Progress

- [ ] Change is properly reviewed (1 reviewer required, 2 recommended).
- [ ] Change is responded to an issue.
- [x] Change is fully covered by a UT.

### Correctness

The tests cover FP32, FP16, and BF16 across all four APIs. In addition to the
upstream cases, they cover five vector-bias shapes, row- and column-major
`mat2`, padded strides, scalar and broadcast bias, zero-K matrices, and
non-contiguous output tensors.

| Platform | `addmm` | `addmm_out` | `dtype` | `dtype_out` | Total |
|---|---:|---:|---:|---:|---:|
| NVIDIA H100 | 123 | 108 | 3 | 3 | 237 passed |
| Ascend 910C | 123 | 108 | 3 | 3 | 237 passed |
| Hygon BW1000 | 123 | 108 | 3 | 3 | 237 passed |
| PTG 810E | 123 | 108 | 3 | 3 | 237 passed |
| MetaX C550 | 123 | 108 | 3 | 3 | 237 passed |
[...]

PR #5383, opened 2026-08-11
[Ascend] Optimize AddMM layouts and bias epilogue

ZYX223:
### PR Category

Operator, OP Test

### Type of Change

Performance Optimization

### Description

This is the Ascend-only split of #5282.

This PR optimizes `addmm`, `addmm.out`, `addmm.dtype`, and
`addmm.dtype_out` while preserving the existing AddMM numerical contract. The
implementation:

- aligns the kernel with the Ascend `mm` grouped M-tile ordering, tuning
  configuration, and backend heuristics;
- keeps row- and column-contiguous matrix views in their original layout and
  materializes only general non-contiguous views;
- loads vector bias once per output N tile, loads scalar bias once, and uses
  broadcast strides for other valid bias shapes instead of materializing a
  full `[M, N]` tensor;
- shares one launch path across all four public APIs, including non-contiguous
  output tensors and FP16/BF16-to-FP32 dtype overloads.

FP32 dot behavior remains strict (`allow_tf32=False`). No common or other
vendor backend implementation is changed.

### Merge Sequence

This PR follows #5386 in the staged AddMM series. Before merge, it will be
rebased onto the preceding changes, remove its duplicate test definitions, and
extend the canonical tests to Ascend. The current scalar guard is transitional
and will not be kept as the final test structure.

### Progress

### Correctness

Ascend 910C, CANN 8.5, final source hash:

- Official source with the PR test harness: `135 passed`
- This PR: `135 passed`

The added tests cover FP32, FP16, and BF16 with scalar/row/column broadcast
bias, padded row- and column-major matrix views, and non-contiguous output
tensors. Existing tests cover the four public AddMM APIs and dtype validation.

### Performance

Measurements use the public FlagGems benchmark path in kernel mode/core level,
with 100 ms warmup and 200 ms repetition windows. Speedup is
`Official Gems / This PR`; values above 1 mean this PR is faster.

| Case group (5 cases each) | FP32 | FP16 | BF16 |
|---|---:|---:|---:|
| Upstream AddMM shapes | 1.703x | 1.687x | 1.673x |
| Medium vector-bias, column-major `mat2` | 103.063x | 184.067x | 161.274x |

Selected rows:

| Dtype / shape | Torch (ms) | Official Gems (ms) | This PR (ms) | Speedup |
|---|---:|---:|---:|---:|
| FP32, `[4096,4096]@[4096,4096]` | 0.886919 | 1.409639 | 0.771990 | 1.826x |
| FP16, `[4096,4096]@[4096,4096]` | 0.176475 | 1.000996 | 0.388912 | 2.574x |
| FP32, `[128,512]@[512,512]`, vector bias | 0.006231 | 0.698838 | 0.010565 | 66.147x |
| BF16, `[512,1024]@[1024,1024]`, vector bias | 0.006126 | 3.793770 | 0.010909 | 347.765x |

The large vector-bias gains primarily remove full `[M, N]` bias
materialization; they are not presented as GEMM-core speedups. The smallest
upstream low-precision case (`384^3`) remains dispatch-bound and is about
`0.48x` of Official; the other upstream FP16/BF16 cases are `1.78x-2.59x`.


tests/test_addmm.py
```diff
@@ -73,6 +73,9 @@ def test_addmm(monkeypatch, M, N, K, scalar, dtype, b_column_major):
 
 
 @pytest.mark.addmm
+@pytest.mark.skipif(
+    flag_gems.vendor_name != "ascend", reason="Ascend-specific AddMM coverage"
```
tengqm: Please add "Issue #<number>:" to the reason so that balls are not dropped.

ZYX223: Added Issue #5385 to track cross-backend AddMM coverage and referenced it from all Ascend-specific skip reasons in f6b43540f.

tests/test_addmm.py
```diff
@@ -72,6 +72,59 @@ def test_addmm(monkeypatch, M, N, K, scalar, dtype, b_column_major):
     utils.gems_assert_close(res_out2, ref_out2, dtype, reduce_dim=K)
 
 
+@pytest.mark.addmm
+@pytest.mark.skipif(
+    flag_gems.vendor_name != "ascend",
```
0x45f: why only run on ascend

ZYX223: This change only modifies the Ascend implementation, so these added cases are scoped to Ascend here. The equivalent coverage is split across #5387 (common), #5384 (MThreads), and #5386 (MetaX), with the remaining backend coverage tracked in #5385. This avoids making this vendor PR fail on implementations it does not change.

ZYX223: Updated in d09cfc6f2. These cases validate public AddMM semantics rather than Ascend-specific behavior, so I removed the Ascend-only skip from all three tests. They now run on every backend; any confirmed backend exception should use a narrow issue-linked skip.

ZYX223: Follow-up in 3abf6dd7d: making every case global exposed an existing common-path scalar-bias failure on NVIDIA (`bias.stride(0)` on a 0-D tensor), which is fixed separately by #5387. Row/column broadcast bias, padded strides, and non-contiguous out remain global here; only the scalar-bias parameter is narrowly skipped outside Ascend under #5385 so this PR does not depend on unmerged backend implementations.

0x45f (approved):

Issue #5385, opened 2026-08-11
Generalize AddMM layout and bias coverage across backends


ZYX223:
The AddMM layout and bias work adds correctness coverage for:

- scalar, vector, and broadcast bias
- row-major, column-major, and padded matrix layouts
- non-contiguous out tensors
- dtype and dtype_out APIs where supported

The original multi-backend change in #5282 was split into independently
reviewable implementations:

- common kernel (NVIDIA, Hygon, and T-Head): #5387
- MThreads: #5384
- MetaX: #5386
- Ascend: #5383

#5387 owns the canonical tests. The vendor PRs should be rebased after the
preceding implementation lands, drop duplicate test definitions, and extend
the supported-backend set for the same tests.

This issue remains open after those four PRs. Additional vendor-specific AddMM
implementations still need equivalent scalar, layout, and out-tensor support.
Close this issue only after the capability guard can be removed and the public
tests run on every supported backend.

bin913:
The split implementations have landed on master and `tests/test_addmm.py` now gates the new coverage with:

```python
_ADDMM_LAYOUT_BIAS_VENDORS = ("ascend", "nvidia", "hygon", "thead", "mthreads", "metax")
_addmm_layout_bias_only = pytest.mark.skipif(
    flag_gems.vendor_name not in _ADDMM_LAYOUT_BIAS_VENDORS,
    reason="Issue #5385: AddMM layout and bias coverage is pending on this backend",
)
```

I verified the NVIDIA path on an H100 here; the remaining work (extending scalar/layout/out-tensor support to the backends not yet in the set above, then removing the capability guard and running the public tests on every supported backend) requires per-backend hardware that I don't have access to. Since this looks like a maintainer-tracked multi-backend task, do you want to keep this open as the tracking issue for the remaining vendors, or is there a specific backend you'd like help with?

ZYX223:
@bin913 Thank you for checking the H100 path. Let's keep this issue open to track the remaining vendor implementations.

The layout/bias work for Ascend, NVIDIA, Hygon, T-Head, MThreads, and MetaX has landed, and all four split PRs are merged. I have updated the issue description to remove the obsolete merge/rebase sequence and make the remaining scope explicit.

The follow-up work needs coordination by the respective backend maintainers with access to their hardware. We are not requesting a specific additional backend from you at this point, and the original authors have not committed to extending beyond the completed six-backend scope.

[...]

PR #5386, opened 2026-08-11
[MetaX] Optimize AddMM layouts and bias epilogue

ZYX223:
### PR Category

Operator, OP Test

### Type of Change

Performance Optimization

### Description

This is the MetaX-only split of #5282.

This PR optimizes `addmm`, `addmm.out`, `addmm.dtype`, and
`addmm.dtype_out` while preserving the existing AddMM numerical contract. The
MetaX implementation:

- uses grouped M-tile ordering to improve B-tile reuse and includes matrix
  layout strides in the tuning key;
- keeps vector and scalar bias compact and uses broadcast strides for other
  valid bias shapes instead of materializing a full `[M, N]` tensor;
- supports row-major, column-major, and padded matrix views, materializing only
  layouts that the MetaX GEMM lowering cannot consume efficiently;
- shares one implementation across all four public APIs, including
  non-contiguous output tensors and FP16/BF16-to-FP32 dtype overloads;
- uses a scalar-reduction fallback for small output dimensions or MetaX
  async-pipeline lowering failures, avoiding recursive dispatcher fallback;
- skips every bias load and dtype conversion when `beta == 0`, matching PyTorch
  behavior for NaN and Inf bias values.

FP32 dot behavior remains strict (`allow_tf32=False`). The MetaX commit does not
change another vendor kernel, its tile/configuration pool, or the public AddMM
contract.

### Issue

#5385

### Merge Sequence

- #5383 (Ascend) and #5384 (MThreads) are merged.
- #5387 is the pending common `beta == 0` follow-up.
- This PR is stacked on #5387 and contains one additional MetaX commit,
  `46650c47f`.
- After #5387 merges, this PR becomes a MetaX-only diff against `master`.

### Progress

### Correctness

Final-head validation used an existing C550 container with
PyTorch `2.8.0+metax3.7.0.7`. The run loaded source commit `46650c47f` from an
isolated `python -S` path; the MetaX AddMM source SHA-256 was
`72c9ff52b82019dc7a29cd5a3bcfbb93b3dfce674dd3dfba59cc250954dca476`.

- `tests/test_addmm.py`: **209 passed** in 116.36 s.
- Coverage includes FP32, FP16, and BF16; `addmm`, `addmm.out`, `addmm.dtype`,
  and `addmm.dtype_out`; scalar/vector/broadcast bias; row-, column-, and padded
  layouts; non-contiguous output; small-shape fallback; and `beta == 0`.
- `pre-commit`, Python compilation, and `git diff --check` pass.

The prior source-isolated baseline run showed three existing Official failures
for scalar bias (FP32/FP16/BF16), caused by accessing `bias.stride(0)` on a
0-D tensor. The MetaX implementation adds an explicit scalar-bias path.

### Performance

The following source-isolated measurements compare Official Gems with the
MetaX scheduling implementation in this PR. They use the public FlagGems
benchmark path in kernel mode/core level; speedup is `Official Gems / This PR`.

| Case group (5 cases each) | FP32 | FP16 | BF16 |
|---|---:|---:|---:|
| Upstream AddMM shapes | 1.004x | 1.001x | 0.998x |
| Medium vector-bias, column-major `mat2` | 3.683x | 4.864x | 4.944x |

Selected rows:

| Dtype / shape | Torch (ms) | Official Gems (ms) | This PR (ms) | Speedup |
|---|---:|---:|---:|---:|
| FP32, `[4096,4096]@[4096,4096]` | 3.998464 | 7.557376 | 7.460864 | 1.013x |
| BF16, `[1024,1024]@[1024,1024]` | 0.066304 | 0.068096 | 0.068352 | 0.996x |
| FP32, `[64,184]@[184,256]`, vector bias | 0.018944 | 0.289024 | 0.046336 | 6.238x |
| BF16, `[512,1024]@[1024,1024]`, vector bias | 0.029440 | 0.316672 | 0.076544 | 4.137x |

These timings were collected before the final `beta == 0` follow-up. The final
restack changes only whether bias is read when beta is zero; all benchmark rows
use the normal beta-one path, and the grid, tile, tuning pool, layout handling,
and GEMM loop are unchanged. The vector-bias gains primarily remove full
`[M, N]` bias materialization and layout conversion; they are not presented as
GEMM-core speedups. Upstream square shapes remain within about 2.2% of
Official.

### DIFF
--- a/src/flag_gems/runtime/backend/_metax/ops/addmm.py
+++ b/src/flag_gems/runtime/backend/_metax/ops/addmm.py
@@ -13,7 +13,6 @@
 # limitations under the License.
 
 import logging
-import math
 
 import torch
 import triton
@@ -21,6 +20,7 @@ import triton.language as tl
 
 from flag_gems import runtime
 from flag_gems.runtime import torch_device_fn
+from flag_gems.runtime.backend._metax import heuristics_config_utils as _hcu
 from flag_gems.utils import broadcastable_to, libentry, libtuner
 from flag_gems.utils import triton_lang_extension as ext
 
@@ -35,32 +35,7 @@ logger = logging.getLogger(__name__)
     warmup=5,
     rep=10,
 )
-@triton.heuristics(
-    {
-        "UPGRADE": lambda args: math.ceil(
-            (args["M"] * args["N"]) / (args["BLOCK_SIZE_M"] * args["BLOCK_SIZE_N"])
-        ).bit_length()
-        > 31,
-    }
-)
-@triton.heuristics(
-    {
-        "UPGRADE_A_OFFS": lambda args: math.ceil(args["M"] * args["K"]).bit_length()
-        > 31,
-    }
-)
-@triton.heuristics(
-    {
-        "UPGRADE_B_OFFS": lambda args: math.ceil(args["K"] * args["N"]).bit_length()
-        > 31,
-    }
-)
-@triton.heuristics(
-    {
-        "UPGRADE_C_OFFS": lambda args: math.ceil(args["M"] * args["N"]).bit_length()
-        > 31,
-    }
-)
+@triton.heuristics(_hcu.HEURISTICS_CONFIGS["addmm"])
 @triton.jit(do_not_specialize=["alpha", "beta"])
 def addmm_kernel(
     a_ptr,
--- a/src/flag_gems/runtime/backend/_metax/ops/addmm.py
+++ b/src/flag_gems/runtime/backend/_metax/ops/addmm.py
@@ -26,24 +26,39 @@ from flag_gems.utils import triton_lang_extension as ext
 
 logger = logging.getLogger(__name__)
 
-# DispatchKeySet to redispatch to the native ATen implementation,
-# bypassing our registered dispatch override (avoids recursion).
-_FALLBACK_KEYSET = torch._C.DispatchKeySet(
-    torch._C.DispatchKey.CompositeExplicitAutograd
-)
-
 
 @libentry()
 @libtuner(
     configs=runtime.get_tuned_config("addmm"),
-    key=["M", "N", "K"],
+    key=["M", "N", "K", "stride_am", "stride_bk"],
+    strategy=["align32", "align32", "align32", "align32", "align32"],
+    warmup=5,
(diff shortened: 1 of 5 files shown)


src/flag_gems/runtime/backend/_metax/ops/addmm.py
```diff
@@ -26,24 +26,39 @@
...
 @triton.heuristics(
     {
         "UPGRADE": lambda args: math.ceil(
             (args["M"] * args["N"]) / (args["BLOCK_SIZE_M"] * args["BLOCK_SIZE_N"])
         ).bit_length()
-        > 32,
+        > 31,
+    }
+)
+@triton.heuristics(
```
0x45f: move heuristics config to heuristics_config_utils.py

tests/test_addmm.py
```diff
@@ -43,14 +43,27 @@
...
 
 # Extend this set as vendor implementations gain equivalent layout and bias support.
 _addmm_layout_bias_only = pytest.mark.skipif(
     flag_gems.vendor_name not in _ADDMM_LAYOUT_BIAS_VENDORS,
     reason="Issue #5385: AddMM layout and bias coverage is pending on this backend",
 )
 
+_addmm_beta_zero_only = pytest.mark.skipif(
+    flag_gems.vendor_name not in _ADDMM_BETA_ZERO_VENDORS,
+    reason="This backend does not yet preserve the AddMM beta-zero contract",
```
0x45f: open an issue explaining why skip it

src/flag_gems/runtime/backend/_metax/ops/addmm.py
ZYX223: Moved the four AddMM heuristic definitions to `_metax/heuristics_config_utils.py`. The kernel now uses `_hcu.HEURISTICS_CONFIGS["addmm"]` in commit `bf8eea9a2`.

tests/test_addmm.py
ZYX223: Opened #5755 to track the remaining vendor beta-zero semantics, and updated the backend skip reason to reference that issue in commit `bf8eea9a2`.

0x45f (approved):

PR #5390, opened 2026-08-11
[KMCompiler][MetaX][Iluvatar][Hygon] Add linalg_lstsq MetaX, Iluvatar and Hygon backend support

cheersluvs:
## What

Enables `linalg_lstsq` on **MetaX** and **Iluvatar**. The generic operator,
added for NVIDIA in #4938 and extended to Ascend in #5173, is **not modified** —
every device-specific value is applied by rebinding attributes on
`flag_gems.ops.linalg_lstsq`, which works because the generic driver reads all
of them at call time.

| file | change |
|---|---|
| `_metax/ops/linalg_lstsq.py` | new |
| `_iluvatar/ops/linalg_lstsq.py` | new |
| `_metax/ops/__init__.py`, `_iluvatar/ops/__init__.py` | +2 each (registration) |
| `tests/test_linalg_lstsq.py` | reference helper, see below |
| `ops/linalg_lstsq.py` | **unchanged** |

The two vendors are together because they share the test-reference change and
the Iluvatar half depends on it; splitting them would mean a stacked PR.

Copying the ~250-line compact-WY driver into the vendor directory would have
been the alternative, and it would drift the moment the generic side changes.

## Sizing every tile to the device

Upstream's block sizes assume NVIDIA-sized shared memory. A MetaX part has
64KB, and **three** separate tiles overflow or exactly fill it. All three are
now derived from the limit Triton actually enforces — its driver's
`max_shared_mem`, not torch's `shared_memory_per_block`, which reports the
49152 static default on NVIDIA while Triton opts in to 164–228KB, so deriving
from torch would needlessly shrink tiles on an H20.

| constant | tile | upstream | derived |
|---|---|---|---|
| `_WY_BLOCK_C` | compact-WY update | 73728 B (8KB over) | 24576 B |
| `_TARGET_TILE_BYTES` | `_panel_qr_kernel` | 65536 B at NC=33 | 32768 B |
| `_TARGET_STACK_ROWS` | `_reduce_qr_kernel` | 65536 B at NC=33 | 32768 B |

The budget is **half** the limit in each case, and that is the substance of
the change, not caution. A 65536 B tile passes the `>` check with nothing left
for launch overhead, and leaves room for only ONE block per SM. Measured on a
C550, `BLOCK_C=16` (24576 B, two blocks resident) ran the WY update in 0.0797
ms against `BLOCK_C=32` (40960 B, one block) at 0.0977.

The two 65536 B tiles were invisible on the hardware I had: they fit exactly,
so the suite passed while the tall path ran at one block per SM throughout.
The `maca3720` runner, added mid-review, failed on them at `(256,32)` with
`Required: 66560, Hardware limit: 65536` — and fixing it turned out to be
worth **+19% fp32 geomean**, not just a green check.

Configuration is derived on first call, never at import: querying the Triton
driver while `flag_gems` is still loading its vendor ops is not something a
vendor module should do.

## Two more MetaX device facts

**float64 `tl.dot` is miscompiled.** Relative error ~1.0 at every operand shape
tested, including 16×16×16, with and without `input_precision="ieee"`, while
the same contraction written as rank-1 updates is bit-exact (0.00e+00). fp32
`tl.dot` is correct (1.1e-07 with `ieee`). `_wy_update_metax` branches on the
`COMPUTE` `tl.constexpr`, so this is resolved at compile time and fp32 keeps
the `tl.dot` path. **This is a MACA compiler bug, not a design choice** — if it
is fixed, deleting the `COMPUTE == tl.float64` branches restores the fast path.
I have a self-contained reproducer and am happy to share it.

**4 KB/thread of private (register-spill) memory**, a driver setting
(`insmod metax.ko pri_mem_sz=…`), which the monolithic reduce kernel exceeds at
`BLOCK_NC=128`. It surfaces as the misleading `Triton Error [MACA]: memory size
or pointer value too large to fit in 32 bit`; the real message only appears in
captured stdout. `num_warps=8` fits and is numerically correct (4.1e-07) but
costs 25× (14.9 ms vs 0.60), and `num_warps=16` exceeds the device's 512
threads/block. So `_TALL_MAX_NC_F32` is capped at 64 and those shapes route to
compact-WY, which has no size ceiling by design.

## Iluvatar

One device fact, and it is a **compiler** limit rather than a numerical one:
float64 `tl.dot` does not compile on a BI-V150.

    test_linalg_lstsq_tall_blocked_fp64
    triton.compiler.errors.CompilationError: at 39:16:
        Wacc += tl.dot(tl.trans(Vb), Tb, input_precision="ieee")

Every other fp64 case passes, because the monolithic and blocked-TSQR paths
contain no `tl.dot` at all — only compact-WY contracts with one, and all three
of the operator's `tl.dot` calls live in that single kernel. So the port is one
kernel and one rebind: `_wy_update` with a float64 form expressing each
contraction as P rank-1 updates (P is the panel width, 16, not a problem
dimension), selected by the `COMPUTE` constexpr so float32 keeps `tl.dot`.

Nothing else is rebound. No block-size configuration is shipped for this
backend: none has been measured on the device, and unmeasured tuning does not
belong in an override.

**float64 is skipped here, so that branch is not exercised in CI.** The device
declares `support_fp64 = False` and means it — `torch.matmul` reports "gemm of
double is not supported on CoreX", cuSOLVER has neither `Dormqr` nor `Dorgqr`,
every float64 copy warns "limited support", and a 256×256 float64 solve returns
NaN even with this kernel in place. `_require_dtype` now consults that flag
first (the old operator probe used a 4×2 solve, which routes to the monolithic
path and so answered for one of four paths — it said "supported" while the
256×256 case produced NaN). The branch is kept because the compile error it
removes is real, it costs float32 nothing (`COMPUTE` is a `tl.constexpr`), and
it is what makes compact-WY work if this device gains usable float64.

The rank-1 kernel is near-identical to the MetaX one, which is duplication I
would rather not have. Sharing it would mean either a cross-vendor import (the
MetaX module rebinds its own shared-memory tuning at import, so importing it
would apply MetaX block sizes to Iluvatar) or a `USE_DOT: tl.constexpr` on the
generic kernel. If a third vendor hits the same gap, the generic flag is the
right answer.

## The one change outside the vendor directories

`_ref_and_gems` now computes its reference with a **float64 solve on the CPU**.

Where `to_reference` keeps the tensor on the device, `torch.linalg.lstsq` in
the test *is the vendor's kernel*, which is neither reliably accurate nor
reliably present. Against a float64 CPU solve at κ ≈ 1.6–2.0:

| | max abs error |
|---|---|
| FlagGems | 5.2e-08 |
| MACA `torch.linalg.lstsq` (fp32) | 1.3e-04 |

Two cases were failing on the reference's error rather than ours. Upcasting
alone is not enough — on `maca3810` the fp64 device solve does not exist and
raises `invalid device function` — so the solve moves to the CPU outright. The
result is placed where `gems_assert_close` expects it (CPU under `--ref=cpu`,
device otherwise) and its dtype is gated on fp64 support, so devices without it
(Ascend 910B) get an fp32 reference exactly as before.

Four more tests hand-rolled the same pattern and bypassed that helper. On
Iluvatar `square_wy_fp64` failed *inside its own reference* with
`cusolver error ... cusolverDnDormqr_bufferSize`, because that shim has no
float64 QR — nothing an operator override can reach. They now use the same CPU
solve. Nothing is skipped and no assertion is dropped; `near_singular` keeps
its reference in the input dtype because it asserts the result's dtype matches
torch's.

## Results

`pytest tests/test_linalg_lstsq.py`:

| device | result |
|---|---|
| MetaX C550 | **76 passed**, 0 skipped |
| Iluvatar BI-V150 | **71 passed, 5 skipped** (4 float64 + 1 complex), was 74 passed / 2 failed |

MetaX CI runners `maca3720` and `maca3810` green; `python-op` green.

MetaX `--mode kernel`, 23 shapes per dtype:

| | fp32 | fp64 |
|---|---|---|
| geomean speedup | **8.589×** | **2.421×** |
| arithmetic mean | 20.619× | 5.916× |
| at or above 1× | 22/23 | 15/23 |

Iluvatar `--mode kernel`, float32: geomean **5.226×**, arithmetic 16.444×,
18/23 at or above 1×, range 0.599–79.867. Every sub-1× row there is a
compact-WY shape, which is the untuned path on this backend — deriving its
block sizes from the device, as the MetaX half does, is the obvious follow-up
and is deliberately not attempted here without measurements.

Where MetaX loses:

- **fp64, large NC** (1024², 2048², 4096×512, 512×1024, 1024×2048) route
  through compact-WY, which is exactly where the rank-1 rewrite replaces
  `tl.dot`. This is the measured price of the fp64 `tl.dot` bug.
- **small shapes** where torch is already sub-millisecond — launch overhead,
  unrelated to that workaround.

## Note for reviewers

`_metax/ops/linalg_lstsq.py` raises a `RuntimeError` if fp32 and fp64 derive
different `BLOCK_C`. The generic driver computes the WY update grid from the
module constant `_WY_BLOCK_C` while taking `BLOCK_C` from `_wy_cfg`; those
agree upstream, so the mismatch is dormant there, and rebinding both keeps them
agreeing here — but only while both dtypes derive the same value (they do on
C550: both 16). On a device where they diverge, the right fix is a one-line
generic change making the grid use the per-dtype value; failing loudly beats
silently under-updating the trailing block.


tests/test_linalg_lstsq.py
```diff
@@ -485,12 +567,14 @@ def test_linalg_lstsq_complex_fallback():
...
-    )
+    ref = torch.linalg.lstsq(A.cpu(), b.cpu(), driver="gels").solution
     with flag_gems.use_gems():
         res = torch.ops.aten.linalg_lstsq(A, b)
-    assert torch.allclose(res[0], ref, atol=1e-4, rtol=1e-4)
+    # Compare on the CPU. torch.allclose on complex tensors needs an elementwise
+    # complex abs, and Iluvatar's runtime compiler cannot build one:
+    # `[IXRTC] nvrtcCompileProgram failed ... abs_kernel<std::complex<float>>`.
+    # The comparison gains nothing from running on the device.
+    assert torch.allclose(res[0].cpu(), ref, atol=1e-4, rtol=1e-4)
```
Caeruleann: use utils.gems_assert_close

cheersluvs: done

tests/test_linalg_lstsq.py
```diff
@@ -516,7 +602,7 @@ def test_linalg_lstsq_square_wy(batch, shape, dtype):
     dev = flag_gems.device
     A = _cond_bounded((*batch, m, n), dtype, dev, seed=42)
     b = _det_randn((*batch, m), dtype, dev, seed=43)
-    ref = torch.linalg.lstsq(A, b, driver="gels")
+    ref = _cpu_ref(A, b)
```
103yiran, line 629: Can it be used only on the ix backend?

cheersluvs: Done. Also applied it on Hygon backend, for on device torch.linalg.lstsq "only overdetermined systems (input.size(-2)>=input.size(-1)) are allowed on CUDA".

103yiran (approved):

PR #5426, opened 2026-08-12
[KernelGen] Add dsplit operator with view implementation

ShawnsYing:
## Summary

This PR adds the `dsplit` operator with view implementation that performs zero-copy depth-wise splitting along dimension 2.

The implementation delegates to existing `torch.split` and `torch.tensor_split` functions following the established pattern for view operators.

### Operators registered in `conf/operators.yaml`:
- `dsplit` (marks: `dsplit`)

## Testing

Tested on **H20 GPU**:
- ✅ 14 test cases covering 3D and 4D tensors
- ✅ Both integer sections (`.int` overload) and index arrays (`.array` overload)
- ✅ Multiple dtypes: float16, bfloat16, float32
- ✅ All tests passed in 0.77s

## Performance

Benchmark completed successfully on H20 GPU in 13.97s.

View operators achieve zero-copy semantics by sharing storage with the original tensor.

## Multi-backend Testing

- ✅ NVIDIA: Passed

## Files changed

- `src/flag_gems/ops/dsplit.py`: Implementation
- `src/flag_gems/ops/__init__.py`: Import registration
- `src/flag_gems/__init__.py`: ATen dispatch registration for `dsplit.int` and `dsplit.array`
- `conf/operators.yaml`: Operator configuration
- `tests/test_dsplit.py`: Accuracy tests
- `benchmark/test_dsplit.py`: Performance tests


src/flag_gems/ops/__init__.py
```diff
@@ -1018,6 +1019,7 @@
     "cumsum_out",
     "deg2rad",
     "deg2rad_",
+    "dsplit",
```
Caeruleann: sort by a-z.

src/flag_gems/ops/dsplit.py
```diff
@@ -0,0 +1,37 @@
...
+
+import logging
+from typing import List, Union
+
+import torch
+
+logger = logging.getLogger(__name__)
+
+
+def dsplit(input: torch.Tensor, indices_or_sections: Union[int, List[int]]):
```
Caeruleann, line 64: So, it's not a triton operator?

conf/operators.yaml
```diff
@@ -2480,6 +2480,24 @@ ops:
...
+    labels:
+      - aten
+      - KernelGen
+      - skip_precision_check
+    kind:
+      - Tensor
+    stages:
+      - alpha: '5.4'
+    device:
+      - nvidia
```
```suggestion
```

ShawnsYing: ✅ Fixed! Removed the device section. View operators are device-agnostic.

src/flag_gems/ops/dsplit.py
```diff
@@ -0,0 +1,41 @@
...
+    """
+    logger.debug("GEMS DSPLIT")
+
+    # dsplit splits along dim=2 (depth)
+    if isinstance(indices_or_sections, int):
+        # Equal splits
+        return torch.split(input, input.shape[2] // indices_or_sections, dim=2)
+    else:
+        # Custom split indices
+        return torch.tensor_split(input, indices_or_sections, dim=2)
```
Caeruleann: Do not call torch operators in the operator implementation.

Please read the document：https://github.com/flagos-ai/FlagGems/blob/master/docs/content/zh-cn/contribution/overview.md#5-%E7%AE%97%E5%AD%90-host-%E5%87%BD%E6%95%B0%E8%A7%84%E8%8C%83-operator-host-function-conventions

ShawnsYing: Thanks for the pointer @Caeruleann — you're right. Per the host function conventions (section 5), operator implementations may only call data-initialization helpers (`torch.empty_like`, `torch.zeros`, ...) and other FlagGems operators, not PyTorch computation operators.

I've reworked `dsplit` to drop the `torch.split` / `torch.tensor_split` delegation. It now computes the split boundaries along `dim=2` itself and materializes each chunk as a zero-copy view via `torch.as_strided` (advancing the storage offset and shrinking the depth dimension). This matches the existing view operators `narrow` and `unbind`, which build their views the same way. No PyTorch computation operators are called.

All 14 accuracy tests still pass against `torch.ops.aten.dsplit`. Updated in the latest push.

ShawnsYing: You're right, and thanks for holding the line here. Section 5 only permits data-initialization helpers (`torch.empty_like`, `torch.zeros`, `torch.randn`, ...) and other FlagGems operators.

I've now reimplemented `dsplit` as a genuine Triton operator with no PyTorch computation operators:

- It resolves the depth-wise (`dim=2`) chunk sizes itself, following `torch.tensor_split` semantics (the first `remainder` chunks get one extra element when the size does not divide evenly).
- Each chunk is produced by a dedicated Triton kernel (`dsplit_copy_kernel`) that copies the corresponding depth slice into a freshly allocated output.
- The only `torch` call is `torch.empty(...)` for output allocation, which section 5 explicitly allows. This mirrors the existing FlagGems `tensor_split` operator, which is likewise a copy-based Triton kernel.

This also answers your earlier question ("is it not a triton operator?") — it now genuinely runs through Triton. All 14 accuracy tests pass against `torch.ops.aten.dsplit`.

One honest note on performance: `aten.dsplit` returns zero-copy views, so a copy-based kernel is slower than the view baseline (as is already the case for the existing `tensor_split`, which documents "Unlike torch.tensor_split ... this returns copies"). If you'd prefer the zero-copy view semantics over convention-compliant Triton compute here, let me know and I'll adjust.

ShawnsYing: @Caeruleann Here is the benchmark comparing the convention-compliant Triton implementation against the original `aten.dsplit`, so we can decide the direction together.

Measured on NVIDIA H20-3e, `--level comprehensive`, all cases SUCCESS. Speedup = Torch latency / Gems latency.

### float32
| dtype | Size | Torch (ms) | Gems (ms) | Speedup |
|---|---|---|---|---|
| f32 | [64,128,32], 2 | 0.00323 | 0.00947 | 0.341 |
| f32 | [64,128,32], 4 | 0.00326 | 0.03251 | 0.100 |
| f32 | [128,256,64], 2 | 0.00323 | 0.01315 | 0.246 |
| f32 | [128,256,64], 4 | 0.00323 | 0.03043 | 0.106 |
| f32 | [256,512,128], 2 | 0.00323 | 0.04166 | 0.078 |
| f32 | [256,512,128], 4 | 0.00323 | 0.04838 | 0.067 |
| f32 | [512,1024,256], 2 | 0.00323 | 0.29939 | 0.011 |
| f32 | [512,1024,256], 4 | 0.00323 | 0.28819 | 0.011 |

**Geometric Mean Speedup (float32): 0.070x**

### float16
| dtype | Size | Torch (ms) | Gems (ms) | Speedup |
|---|---|---|---|---|
| f16 | [64,128,32], 2 | 0.00349 | 0.00883 | 0.395 |
| f16 | [64,128,32], 4 | 0.00342 | 0.03011 | 0.114 |
| f16 | [128,256,64], 2 | 0.00326 | 0.01174 | 0.278 |
| f16 | [128,256,64], 4 | 0.00323 | 0.03414 | 0.095 |
| f16 | [256,512,128], 2 | 0.00323 | 0.02731 | 0.118 |
| f16 | [256,512,128], 4 | 0.00323 | 0.03510 | 0.092 |
| f16 | [512,1024,256], 2 | 0.00323 | 0.14509 | 0.022 |
| f16 | [512,1024,256], 4 | 0.00323 | 0.15917 | 0.020 |

**Geometric Mean Speedup (float16): 0.093x**

### bfloat16
| dtype | Size | Torch (ms) | Gems (ms) | Speedup |
|---|---|---|---|---|
| bf16 | [64,128,32], 2 | 0.00323 | 0.00934 | 0.346 |
| bf16 | [64,128,32], 4 | 0.00326 | 0.03210 | 0.102 |
| bf16 | [128,256,64], 2 | 0.00323 | 0.01162 | 0.278 |
| bf16 | [128,256,64], 4 | 0.00323 | 0.03174 | 0.102 |
| bf16 | [256,512,128], 2 | 0.00323 | 0.02746 | 0.118 |
| bf16 | [256,512,128], 4 | 0.00326 | 0.03534 | 0.092 |
| bf16 | [512,1024,256], 2 | 0.00323 | 0.14528 | 0.022 |
| bf16 | [512,1024,256], 4 | 0.00323 | 0.15888 | 0.020 |

**Geometric Mean Speedup (bfloat16): 0.091x**

**Overall geometric mean: 0.084x.**

### Why the speedup is below 1x
This is a semantic consequence, not a kernel-tuning issue:

- `aten.dsplit` returns **zero-copy views** — it moves no data, so its latency is constant (~0.003 ms) regardless of tensor size.
- A section-5-compliant Triton implementation must **actually copy** the data, so its latency grows with the element count, which is why larger tensors fall to ~0.01-0.02x.

The existing FlagGems `tensor_split` operator is the same copy-based Triton kernel and has the same characteristic (its docstring notes "Unlike torch.tensor_split ... this returns copies").

### Decision needed
There is a genuine trade-off here:

1. **Keep the current convention-compliant Triton copy kernel** (this PR): satisfies the host-function conventions, but slower than the view baseline.
2. **Implement `dsplit` as a true view operator** via `torch.as_strided` (like the existing `narrow` / `unbind` ops): zero-copy and fast, but relies on a `torch` layout op that is not in the section-5 whitelist.

Could you let me know which direction you prefer? I'm happy to switch to the view implementation if view semantics are acceptable for this operator family.

Caeruleann (approved):

PR #5610, opened 2026-08-19
[KernelGen][Nvidia] Add embedding_bag operator with Triton kernel

ShawnsYing:
## Summary
Adds a Triton kernel for `embedding_bag`. Computes sums, means, or maxes of bags of embeddings, without instantiating the intermediate embeddings.

## Testing
- Parametrized tests over `num_bags` {3, 8, 16}, `embedding_dim` {16, 32}, and `mode` {0 sum, 1 mean, 2 max}, with and without `per_sample_weights`

## Performance
Test command: `pytest benchmark/test_embedding_bag.py --level core` (NVIDIA H20-3e)

This kernel avoids materializing the intermediate gathered embeddings, so its
advantage grows with the reduction workload (bag count × samples-per-bag ×
embedding_dim). On tiny bags it is launch-overhead bound and torch's fused CUDA
kernel wins; the benchmark shapes now span that crossover so the tradeoff is
visible instead of only sampling the small end.

### embedding_bag (float16)

| Size (weight / indices / offsets) | Torch Latency (ms) | Gems Latency (ms) | Speedup |
|---|---|---|---|
| [500, 256] / [512] / [128] | 0.008064 | 0.052768 | 0.153 |
| [10000, 256] / [4096] / [512] | 0.011904 | 0.052096 | 0.229 |
| [10000, 128] / [16384] / [1024] | 0.014944 | 0.053536 | 0.279 |
| [20000, 256] / [32768] / [2048] | 0.034816 | 0.052960 | 0.657 |
| [50000, 256] / [131072] / [4096] | 0.102752 | 0.060608 | **1.695** |
| [100000, 512] / [262144] / [8192] | 0.394688 | 0.110528 | **3.571** |
| [100000, 256] / [262144] / [16384] | 0.228288 | 0.068896 | **3.314** |

float32 and bfloat16 follow the same trend (large-shape speedups ~2.5x–3.5x).

| Regime | Speedup |
|---|---|
| Tiny / launch-overhead bound | 0.15x–0.66x |
| Large / reduction bound | 1.7x–3.6x |

## Multi-backend Testing
| Backend | Accuracy Test | Speedup (large shapes) | Notes |
|---|---|---|---|
| Nvidia (H20-3e) | PASS (54 cases) | up to 3.57x | Primary |
| Tianshu | N/A | — | — |
| Muxi | N/A | — | — |
| Ascend | N/A | — | — |
| Hygon | N/A | — | — |

## Files Changed
- `src/flag_gems/ops/embedding_bag.py`: Triton kernel implementation
- `tests/test_embedding_bag.py`: Accuracy test (+ Apache-2.0 license header)
- `benchmark/test_embedding_bag.py`: Performance benchmark (+ Apache-2.0 license header, realistic shape range)


benchmark/test_embedding_bag.py
```diff
@@ -0,0 +1,58 @@
+import pytest
```
bin913, line 15: Please add copyright and license

tests/test_embedding_bag.py
```diff
@@ -0,0 +1,60 @@
+import pytest
```
bin913, line 15: Please add copyright and license

benchmark/test_embedding_bag.py
ShawnsYing: Done — added the Apache-2.0 copyright/license header to `benchmark/test_embedding_bag.py` (commit c5602269b).

tests/test_embedding_bag.py
ShawnsYing: Done — added the Apache-2.0 copyright/license header to `tests/test_embedding_bag.py` (commit c5602269b).

bin913:
image



May I ask if you are using kernel mode or operator mode for benchmark testing. If it is kernel mode, only GPU time is available; if it is operator mode, it is end-to-end time

ShawnsYing:
Good question — and it identifies a real problem with the number I gave you earlier. Short answer: **the benchmark defaults to `kernel` mode**, and in my previous reply I reported a kernel-mode gems latency next to an end-to-end eager latency. Those are not comparable, and the ~59% "launch overhead" split I quoted does not cleanly correspond to either mode. Here are both modes, measured properly.

**What each mode measures** (`benchmark/base.py`):
- `kernel` (the pytest default): `triton.testing.do_bench` — GPU time from CUDA events with an L2 flush between reps, i.e. device time only.
- `operator`: a wall-clock loop around the callable minus its synchronize, i.e. end-to-end including the Python wrapper and any host-side allocations the op does per call.

Since `benchmark/test_embedding_bag.py` never calls `set_gems`, both sides are measured through the same route: eager runs `torch.nn.functional.embedding_bag` directly, gems runs it under `flag_gems.use_gems()`. That makes them directly comparable *within* a mode — but not *across* modes, which is the mistake I made.

**`--mode operator` (end-to-end, both sides — the fair comparison):**

| (weights, dim, bags) | fp16 torch | fp16 gems | fp16 | fp32 torch | fp32 gems | fp32 |
|---|---|---|---|---|---|---|
| (500, 256, 128) | 0.0104 | 0.1123 | 0.09x | 0.0098 | 0.0987 | 0.10x |
| (10000, 256, 512) | 0.0099 | 0.0970 | 0.10x | 0.0099 | 0.0988 | 0.10x |
| (10000, 128, 1024) | 0.0115 | 0.0970 | 0.12x | 0.0116 | 0.0992 | 0.12x |
| (20000, 256, 2048) | 0.0381 | 0.0977 | 0.39x | 0.0384 | 0.0990 | 0.39x |
| (50000, 256, 4096) | 0.1356 | 0.0973 | **1.39x** | 0.1821 | 0.0998 | **1.82x** |
| (100000, 512, 8192) | 0.7979 | 0.1088 | **7.34x** | 0.8783 | 0.1593 | **5.52x** |
| (100000, 256, 16384) | 0.3826 | 0.0985 | **3.89x** | 0.4425 | 0.0971 | **4.56x** |

**`--mode kernel` (GPU-only, same table shape):** crossover moves one row earlier — 0.95x (fp16) / 0.95x (fp32) at (20000, 256, 2048) instead of 0.39x, and 2.43x/7.17x/4.97x (fp16) at the three largest.

The two modes differ exactly where you would expect: at the small shapes the whole gems cost is ~0.098 ms of which the kernel is a small part, so end-to-end is much worse than GPU-only (0.09x vs 0.13x) — that gap *is* the wrapper. At the large shapes the kernel dominates and the two modes converge (7.34x vs 7.17x).

**Where the small-shape floor actually is.** Measuring the wrapper's per-call allocations directly at (500, 256, 128):

| per-call cost | us |
|---|---|
| `torch.empty` output (128x256 fp16) | 2.2 |
| `torch.zeros` offset2bag (512 int64) | 6.3 |
| `torch.empty` bag_size | 2.2 |
| `torch.zeros` max_indices (128x256 int64) | 6.4 |
| `torch.empty` psw (empty) | 1.7 |
| **sum** | **18.8** |
| full `embedding_bag` (same case) | 44.4 |

So ~19 us of the ~44 us is four host-side allocations (two of them `zeros` over 128 KB of int64 that the kernel overwrites anyway — `max_indices` and `offset2bag` could be `empty`), and the rest is launch plus wrapper. That is the actionable part of the small-shape cost, and it is independent of `num_warps`/`BLOCK` tuning.

I did not change code for this reply because the question was about methodology; if you want, I will (a) switch `max_indices`/`offset2bag` to `empty` where the kernel fully writes them and (b) re-measure, or (c) just confirm which mode you want quoted in the PR description.

ShawnsYing:
Correction to one claim in my previous comment, before anyone acts on it.

I wrote that `max_indices` and `offset2bag` "could be `empty`". I checked that empirically with a sentinel fill instead of assuming, and **only `max_indices` is safe**:

- `max_indices`: the kernel writes every element (sentinel test: 0/8 left unwritten, all three modes) — so `torch.empty` is fine.
- `offset2bag`: the kernel only writes slots belonging to a bag (`if col_block == 0: for i in range(count)`), so samples that no bag covers are **never written**. With `offsets=[3, 6]` and 7 indices, slots 0-2 stay untouched (sentinel test: [0, 1, 2] unwritten, all modes).

And those untouched slots must read as `0` — that is the reference behaviour, not just this kernel's convention. `torch.ops.aten._embedding_bag(w, idx, offset=[3,6], ...)` returns `offset2bag = [0, 0, 0, 0, 0, 0, 1]` for exactly that case. So `offset2bag` genuinely needs the zero fill; only `max_indices` can drop it.

That changes the saving from the ~12.7 us I quoted to **~6.4 us** (the `max_indices` zeros alone) at (500, 256, 128) — real but modest, and it does not change the conclusion about the small-shape floor.

Sorry for the sloppy claim; the allocation breakdown table itself stands, as do the kernel/operator-mode tables.

bin913 (approved):

PR #5635, opened 2026-08-19
[KernelGen][Nvidia] Add cholesky operator with Triton kernel

bwbwzzz:
## Description

Add the `cholesky` operator with a Triton kernel implementation for NVIDIA GPUs. The kernel computes the Cholesky decomposition of a symmetric positive-definite matrix (or a batch of them) using the sequential Cholesky-Banachiewicz algorithm. All core arithmetic (dot products, sqrt, division) is performed inside the Triton kernel. Supports `float32` / `float64` and the `upper` argument.

## Correctness

All 20 accuracy test cases pass (lower/upper/batched, float32 + float64). Independently verified on larger matrices:

| dtype | 64x64 | 128x128 | 256x256 |
|-------|-------|---------|---------|
| float32 | 1.1e-5 | 3.6e-5 | 8.9e-5 |
| float64 | 3.0e-14 | 5.5e-14 | 2.2e-13 |

## Performance (NVIDIA H20)

Test command: `pytest benchmark/test_cholesky.py --level core`

The kernel uses a single program per matrix (sequential O(N^3)), so it is fastest on small matrices and is outperformed by PyTorch's LAPACK-backed implementation as N grows.

| Shape | float32 Speedup | float64 Speedup |
|-------|-----------------|-----------------|
| 2x2 | 5.50x | 7.94x |
| 4x4 | 5.09x | 6.28x |
| 8x8 | 3.43x | 3.60x |
| 16x16 | 1.41x | 1.22x |
| 32x32 | 0.36x | 0.30x |
| 64x64 | 0.07x | 0.06x |
| 128x128 | 0.015x | 0.024x |
| 256x256 | 0.003x | 0.007x |

The operator provides speedup for small matrices (<= 16x16); larger matrices regress against the native implementation due to the sequential single-program algorithm.

## Tested on

- GPU: NVIDIA H20
- Backends validated: Nvidia


tests/test_cholesky.py
```diff
@@ -0,0 +1,85 @@
...
+    B = torch.randn(shape, dtype=dtype, device=flag_gems.device)
+    A = (
+        B @ B.transpose(-2, -1)
+        + torch.eye(n, dtype=dtype, device=flag_gems.device) * 0.1
+    )
+
+    ref_A = utils.to_reference(A)
+    ref_out = torch.cholesky(ref_A)
+
+    res_out = flag_gems.cholesky(A)
```
w1120029931-bit, line 83: should use “with flag_gems.use_gems():”

bwbwzzz: Teacher Zhao Sujie said that for some reasons, I should not use use_gems; instead, I should call it directly from the repository.

bwbwzzz:
> Can the performance of these shapes be further improved



> Can the performance of these shapes be further improved

For high-performance and complex matrix computations, automatically generated operators often exhibit some performance disadvantages when handling large input shapes.

PR #5687, opened 2026-08-23
[KernelGen][Nvidia] Add _nested_from_padded_and_nested_example operator with Triton kernel

chx7514:
## Summary

Adds a Triton kernel for `_nested_from_padded_and_nested_example`. The kernel removes padding from a dense tensor to build a **legacy nested tensor** (`torch.strided` layout, matching the reference), using the per-component sizes of an example nested tensor to determine each component's boundaries. Two Triton paths are provided:

- a contiguous-prefix fast path for the common case where only the first dimension is ragged;
- a general path that correctly handles ragged trailing dimensions (e.g. `[[3, 4], [5, 2]]`).

The result is reconstructed directly via `_nested_view_from_buffer`, avoiding the previous `unbind()` / Python-list / `nested_tensor_from_jagged()` round-trip that added a fixed CPU overhead and returned the wrong (`jagged`) layout.

### Host-side metadata cleanup

The previous wrapper rebuilt nested strides and offsets in Python with `prod` / `cumsum` / a `for`-loop / two `.item()` calls, adding ~50 µs of host overhead per call that dominated the small-tensor cases. This is now a single C++ call to `torch.ops.aten._nested_compute_contiguous_strides_offsets` (the same helper `torch.nested` uses internally to build legacy nested tensors) plus `nt_example.numel()` for the total element count.

Note: the removed `.item()` calls were on a **CPU** nested-metadata tensor, so they incurred **no GPU device→host sync**; the cost was Python→ATen dispatch + boxing on several small CPU tensor ops.

The kernels read `storage_offsets` (length B) and compute component numel in-kernel from `sizes`, dropping the B+1 offsets, `component_numel`, and `max_component_numel` host work. The grid uses `padded.stride(0)` as the per-component capacity upper bound (the kernel masks against actual component numel); the high-padding case stays at the kernel-launch floor, so no extra `.item()` sync is needed.

The `values` buffer is allocated exactly (`padded.new_empty((total_numel,))`) — not over-allocated — since `_nested_view_from_buffer` exposes the full backing storage to downstream readers.

All dtype/dim regimes go through the Triton kernel (the op is a self-contained Gems implementation and does not dispatch back to torch's native `_nested_from_padded`).

## Testing
- Parametrized over 2D / 3D / 3D-trailing-ragged / 4D-trailing-ragged shapes and float16 / float32 / bfloat16
- Validated against reference on device, including nested `layout`, `_nested_tensor_size`, `_nested_tensor_strides`, and `_nested_tensor_storage_offsets`
- Empty (B == 0) input handled explicitly (upstream `_nested_from_padded` segfaults on B == 0 on current torch, so it is out of scope here)
- Tested on: Nvidia (H20)

## Performance
Test command: `pytest benchmark/test_nested_from_padded_and_nested_example.py --level core` (NVIDIA H20)

### nested_from_padded_and_nested_example

| float16 | padded=(4, 16) | 0.076704 | 0.012176 | 6.30 |
| float16 | padded=(4, 16, 8) | 0.009952 | 0.012768 | 0.78 |
| float16 | padded=(8, 64, 32) | 0.010304 | 0.017600 | 0.59 |
| float16 | padded=(16, 128, 64) | 0.011264 | 0.025280 | 0.45 |
| float16 | padded=(2, 8, 4) | 0.010016 | 0.031776 | 0.32 |
| float16 | padded=(2, 8, 4, 3) | 0.063936 | 0.032224 | 1.98 |
| float32 | padded=(4, 16) | 0.076448 | 0.012512 | 6.11 |
| float32 | padded=(4, 16, 8) | 0.009984 | 0.013120 | 0.76 |
| float32 | padded=(8, 64, 32) | 0.010304 | 0.017808 | 0.58 |
| float32 | padded=(16, 128, 64) | 0.011376 | 0.024544 | 0.46 |
| float32 | padded=(2, 8, 4) | 0.010048 | 0.032576 | 0.31 |
| float32 | padded=(2, 8, 4, 3) | 0.064096 | 0.032512 | 1.97 |
| bfloat16 | padded=(4, 16) | 0.076864 | 0.014144 | 5.43 |
| bfloat16 | padded=(4, 16, 8) | 0.076896 | 0.013024 | 5.90 |
| bfloat16 | padded=(8, 64, 32) | 0.140448 | 0.018096 | 7.76 |
| bfloat16 | padded=(16, 128, 64) | 0.258400 | 0.025456 | 10.15 |
| bfloat16 | padded=(2, 8, 4) | 0.063872 | 0.032032 | 1.99 |
| bfloat16 | padded=(2, 8, 4, 3) | 0.064128 | 0.032544 | 1.97 |

| nested_from_padded_and_nested_example | **2.99** |

## Multi-backend Testing
|---|---|---|---|
| Nvidia (H20) | PASS (18 cases) | 2.99 | Primary |
| Tianshu | N/A | — | — |
| Muxi | N/A | — | — |
| Ascend | N/A | — | — |
| Hygon | N/A | — | — |

## Files Changed
- `src/flag_gems/ops/_nested_from_padded_and_nested_example.py`: Triton kernel implementation (hybrid contiguous-prefix / general paths, C++ metadata helpers)
- `tests/test_nested_from_padded_and_nested_example.py`: Accuracy test
- `benchmark/test_nested_from_padded_and_nested_example.py`: Performance benchmark


chx7514:
@bin913 Thanks for flagging the benchmark performance. Arithmetic-mean speedup improved 0.896 → 2.99 (a 3.3x gain), and the op is now a fully self-contained Gems implementation — every dtype/dim regime goes through the Triton kernel, no dispatch back to torch's native _nested_from_padded.

**Root cause of the previous low numbers**

The earlier wrapper rebuilt nested strides and offsets in Python every call (prod / cumsum / a for-loop / two .item() calls). On the small tensors in this benchmark that accumulated ~50 µs of host overhead per call, which dominated the actual ~10 µs Triton kernel and dragged the small-shape cases down to ~0.15x.

Worth clarifying: those .item() calls were on a CPU nested-metadata tensor, so they caused no GPU device→host sync — the cost was Python→ATen dispatch + boxing across several small CPU tensor ops. The fix below removes them regardless.

**What changed**

- Strides/offsets now come from a single C++ call, torch.ops.aten._nested_compute_contiguous_strides_offsets — the same helper torch.nested uses internally to build legacy nested tensors — and the total element count from nt_example.numel(). This replaces the entire Python prod/cumsum/for/.item() pipeline.
- The kernels now read storage_offsets (length B) and compute component numel in-kernel from sizes, so the B+1 offsets, component_numel, and max_component_numel host work is gone. The grid uses padded.stride(0) as a per-component capacity upper bound (the kernel masks against actual component numel). I checked the high-padding case — it stays at the kernel-launch floor, so no extra .item() sync is needed.
- The values buffer is allocated exactly via padded.new_empty((total_numel,)), not over-allocated, since _nested_view_from_buffer exposes the full backing storage to downstream readers.
- All cases go through the Triton kernel (no native fallback).

bf16 benefits from both the Triton kernel (PyTorch falls back to the slow generic masked_select path for bf16) and the removed host overhead.

**Known limitation**

fp16/fp32 3D small shapes (0.31–0.78x) remain the weak spot. PyTorch has a specialized native remove_padding_kernelLauncher (~10 µs, near the launch floor) for this regime; the Gems Triton kernel itself matches it (~10.7 µs), but the remaining wrapper host work (_nested_view_from_buffer construction + metadata H2D) still adds up, so these stay slightly below 1.0x. I left further host-path optimization (a lighter nested-tensor construction path, int32 metadata H2D) for a follow-up PR rather than expanding this one. Happy to pursue it here if you'd prefer.

Accuracy: the original 12 cases plus manually-checked 3D/4D trailing-ragged (fp16/bf16) all pass. The empty (B==0) case is guarded explicitly — upstream _nested_from_padded segfaults on B==0 on current torch, so it's out of scope here.

Full numbers are in the updated Performance table in the PR description.

tests/test_nested_from_padded_and_nested_example.py
```diff
@@ -0,0 +1,137 @@
...
+    device = flag_gems.device
+    lengths = torch.tensor([3, 9, 5, 13], dtype=torch.int64)
+
+    padded = torch.randn(B, D_pad, dtype=dtype, device=device)
+    nt_example = _make_example(B, D_pad, 1, lengths, dtype, device)
+
+    ref_out = torch.ops.aten._nested_from_padded_and_nested_example(
+        utils.to_reference(padded), nt_example
+    )
+    res_out = flag_gems._nested_from_padded_and_nested_example(padded, nt_example)
```
w1120029931-bit, line 77: Example:
ref_out = torch.dot(ref_inp1, ref_inp2)
with flag_gems.use_gems():
    res_out = torch.dot(inp1, inp2)

bin913:
> @bin913 Thanks for flagging the benchmark performance. Arithmetic-mean speedup improved 0.896 → 2.99 (a 3.3x gain), and the op is now a fully self-contained Gems implementation — every dtype/dim regime goes through the Triton kernel, no dispatch back to torch's native _nested_from_padded.
> 
> **Root cause of the previous low numbers**
> 
> The earlier wrapper rebuilt nested strides and offsets in Python every call (prod / cumsum / a for-loop / two .item() calls). On the small tensors in this benchmark that accumulated ~50 µs of host overhead per call, which dominated the actual ~10 µs Triton kernel and dragged the small-shape cases down to ~0.15x.
> 
> Worth clarifying: those .item() calls were on a CPU nested-metadata tensor, so they caused no GPU device→host sync — the cost was Python→ATen dispatch + boxing across several small CPU tensor ops. The fix below removes them regardless.
> 
> **What changed**
> 
> * Strides/offsets now come from a single C++ call, torch.ops.aten._nested_compute_contiguous_strides_offsets — the same helper torch.nested uses internally to build legacy nested tensors — and the total element count from nt_example.numel(). This replaces the entire Python prod/cumsum/for/.item() pipeline.
> * The kernels now read storage_offsets (length B) and compute component numel in-kernel from sizes, so the B+1 offsets, component_numel, and max_component_numel host work is gone. The grid uses padded.stride(0) as a per-component capacity upper bound (the kernel masks against actual component numel). I checked the high-padding case — it stays at the kernel-launch floor, so no extra .item() sync is needed.
> * The values buffer is allocated exactly via padded.new_empty((total_numel,)), not over-allocated, since _nested_view_from_buffer exposes the full backing storage to downstream readers.
> * All cases go through the Triton kernel (no native fallback).
> 
> bf16 benefits from both the Triton kernel (PyTorch falls back to the slow generic masked_select path for bf16) and the removed host overhead.
> 
> **Known limitation**
> 
> fp16/fp32 3D small shapes (0.31–0.78x) remain the weak spot. PyTorch has a specialized native remove_padding_kernelLauncher (~10 µs, near the launch floor) for this regime; the Gems Triton kernel itself matches it (~10.7 µs), but the remaining wrapper host work (_nested_view_from_buffer construction + metadata H2D) still adds up, so these stay slightly below 1.0x. I left further host-path optimization (a lighter nested-tensor construction path, int32 metadata H2D) for a follow-up PR rather than expanding this one. Happy to pursue it here if you'd prefer.
> 
> Accuracy: the original 12 cases plus manually-checked 3D/4D trailing-ragged (fp16/bf16) all pass. The empty (B==0) case is guarded explicitly — upstream _nested_from_padded segfaults on B==0 on current torch, so it's out of scope here.
> 
> Full numbers are in the updated Performance table in the PR description.

Thank you very much for your reply. If you will submit a PR to solve this problem in the future, it would be even better to establish an issue to track this issue

PR #5691, opened 2026-08-23
[KernelGen][Nvidia] Add _nested_tensor_from_mask operator with Triton kernel

chx7514:
## Summary
Adds a Triton kernel for `_nested_tensor_from_mask`. Builds a nested tensor from a padded dense tensor of shape (N, L, D) and a
boolean padding mask of shape (N, L), using the mask to determine the ragged
length of each component.

## Testing

## Performance
Test command: `pytest benchmark/test_nested_tensor_from_mask.py --level core` (NVIDIA H20)

### nested_tensor_from_mask

| float16 | [16, 64, 64] | 0.097120 | 0.194848 | 0.498 |
| float16 | [32, 128, 64] | 0.102304 | 0.207056 | 0.494 |
| float16 | [64, 256, 128] | 0.104672 | 0.239424 | 0.437 |
| float16 | [128, 512, 128] | 0.122608 | 0.306160 | 0.400 |
| float32 | [16, 64, 64] | 0.099072 | 0.192576 | 0.514 |
| float32 | [32, 128, 64] | 0.102240 | 0.207136 | 0.494 |
| float32 | [64, 256, 128] | 0.104960 | 0.239808 | 0.438 |
| float32 | [128, 512, 128] | 0.124512 | 0.305232 | 0.408 |
| bfloat16 | [16, 64, 64] | 0.342240 | 0.194240 | 1.762 |
| bfloat16 | [32, 128, 64] | 0.551136 | 0.207472 | 2.656 |
| bfloat16 | [64, 256, 128] | 0.977344 | 0.238000 | 4.106 |
| bfloat16 | [128, 512, 128] | 1.982304 | 0.295616 | 6.706 |

| nested_tensor_from_mask | **1.576** |

This operator is host-bound: ~95% of the reported latency is host-side work
(nested-tensor assembly and the device round trips needed to bring ragged
lengths back from the mask), with the three Triton kernels accounting for
~5%. The Gems column is therefore effectively flat across dtypes and shapes,
while the Torch column varies with dtype — the bf16 speedups are real, but
they come from the native baseline falling back to a generic path (PyTorch's
`nested_from_padded_cuda` only ships fused launchers for fp32/fp16), not from
our kernel.

The earlier revision of this table reported a mean of 2.012. Those numbers
predated the removal of a dead zero-length fallback path in the wrapper, which
was ~4x slower than the normal path when triggered and made the benchmark
unstable (the input generator produces an empty component in ~20% of draws, and
`do_bench` sizes its repeat count from a single early estimate).

## Multi-backend Testing
|---|---|---|---|
| Nvidia (H20) | PASS (12 cases) | 1.576 | Primary |
| Tianshu | N/A | — | — |
| Muxi | N/A | — | — |
| Ascend | N/A | — | — |
| Hygon | N/A | — | — |

## Files Changed
- `src/flag_gems/ops/_nested_tensor_from_mask.py`: Triton kernel implementation
- `tests/test_nested_tensor_from_mask.py`: Accuracy test
- `benchmark/test_nested_tensor_from_mask.py`: Performance benchmark
- `conf/operators.yaml`: Add operator entry (kind: Tensor, stage: alpha 5.5.0)


src/flag_gems/ops/_nested_tensor_from_mask.py
```diff
@@ -0,0 +1,182 @@
...
+
+    # 4. Assemble the nested tensor from the compacted values and offsets.
+    #    Use the jagged layout which is constructed in a single C++ call and
+    #    is functionally equivalent to the reference nested tensor.
+    return torch.nested.nested_tensor_from_jagged(
+        values, offsets=offsets, lengths=lengths, jagged_dim=1
+    )
+
+
+__all__ = ["_nested_tensor_from_mask"]
```
bin913: Since this file is for Triton‑kernel implementation, we don’t need to add this line here.

tests/test_nested_tensor_from_mask.py
```diff
@@ -0,0 +1,102 @@
...
+    N, L, D = shape
+    t = torch.randn(shape, dtype=dtype, device=flag_gems.device)
+    # Arbitrary (possibly non-left-aligned) mask.
+    mask = torch.randint(0, 2, (N, L), dtype=torch.bool, device=flag_gems.device)
+
+    ref_t = utils.to_reference(t)
+    ref_mask = utils.to_reference(mask)
+
+    ref_out = torch.ops.aten._nested_tensor_from_mask(ref_t, ref_mask, mask_check=False)
+    res_out = flag_gems._nested_tensor_from_mask(t, mask, mask_check=False)
```
bin913, line 97: Please use flag_gems.use_gems():

tests/test_nested_tensor_from_mask.py
```diff
@@ -0,0 +1,102 @@
...
+            [True, False, True, False, False, False, False, False],
+            [True, True, False, False, False, False, False, False],
+            [False, False, False, False, False, False, False, False],
+            [True, True, True, True, True, True, True, True],
+        ],
+        dtype=torch.bool,
+        device=flag_gems.device,
+    )
+    with pytest.raises(RuntimeError):
+        flag_gems._nested_tensor_from_mask(t, mask, mask_check=True)
```
bin913, line 118: Please use flag_gems.use_gems():

tests/test_nested_tensor_from_mask.py
```diff
@@ -0,0 +1,102 @@
...
+def test_nested_tensor_from_mask(shape, dtype):
+    N, L, D = shape
+    t = torch.randn(shape, dtype=dtype, device=flag_gems.device)
+    mask = _left_aligned_mask(N, L, flag_gems.device)
+
+    ref_t = utils.to_reference(t)
+    ref_mask = utils.to_reference(mask)
+
+    ref_out = torch.ops.aten._nested_tensor_from_mask(ref_t, ref_mask, mask_check=True)
+    res_out = flag_gems._nested_tensor_from_mask(t, mask, mask_check=True)
```
bin913, line 79: Please use flag_gems.use_gems():

src/flag_gems/ops/_nested_tensor_from_mask.py
chx7514: Done — removed the redundant `__all__` line. The op files in this repo don't use `__all__` (only 4 of 661 do), and the symbol is imported explicitly in `ops/__init__.py` and registered in `_FULL_CONFIG`.

bin913:
> Hi @bin913, here is the breakdown. I profiled the Gems implementation stage by stage on an H20 (`CUDA_LAUNCH_BLOCKING`-free, each stage timed with a device sync to attribute its cost; GPU-side time measured separately with `torch.profiler`).
> 
> **Example shape: `[128, 512, 128]`, float16** (end-to-end 0.36 ms in my harness; the benchmark harness reports ~0.20 ms for this shape — absolute numbers differ slightly with timing methodology, the _proportions_ are what matter):
> 
> component	time	% of total
> **GPU kernels** (3 Triton kernels: lengths-reduce, offsets-scan, compaction)	~0.014 ms	**~4%**
> **Device syncs** forced by metadata handoff (`has_gap.item()` + `offsets[N].item()` + D2H copy of `lengths`/`offsets`)	~0.10 ms	~28%
> **CPU-side nested-tensor assembly** (`nested_size`/`nested_strides`/`storage_offsets` + `aten._nested_view_from_buffer`)	~0.13 ms	~38%
> **Kernel launches + allocations** (5 small tensors, 3 launches)	~0.09 ms	~27%
> So **GPU compute is only ~4% of the total; ~96% is host-side overhead**. The GPU kernels themselves scale with the problem (0.003 ms → 0.014 ms from `[16,64,64]` to `[128,512,128]`), but they are drowned out — which is exactly why the Gems latency column in the PR is essentially flat (~0.19–0.20 ms) across all shapes and dtypes.
> 
> Within the host overhead, the two dominant pieces are:
> 
> 1. **Synchronization (~30%)** — producing a legacy `torch.strided` NestedTensor requires the size/strides/offsets metadata on CPU, and the ragged lengths can only come from the mask on device. That means at least one full device round trip; we currently pay three (two `.item()`s + one D2H copy).
> 2. **`aten._nested_view_from_buffer` (~30% at this shape, 0.107 ms)** — the aten op that wraps the flat buffer into a NestedTensor is itself a heavy host-side op (it grows with N, from ~0.023 ms at N=16 to ~0.107 ms at N=128). This part is inside PyTorch and shared with the native reference path.
> 
> For contrast, the native aten baseline avoids most of this because it is a single fused C++ path (one sync inside `masked_select`'s `nonzero`, assembly in C++) — the fp32/fp16 specialized `remove_padding` launcher makes its device side fast too, hence our <1x there. For bf16 the baseline falls back to the generic multi-kernel path, so even with our host overhead we win.
> 
> This is a host-bound op, and the overhead is addressable (collapse the three syncs into one, e.g. compute lengths on CPU from a D2H'd mask or fuse the metadata round trip) — but that is a follow-up optimization, independent of dtype, and does not change the kernel's correctness story here.

Thanks a lot for your reply. There are two modes for running benchmark tests: `operator` and `kernel`. The `operator` mode measures end-to-end time, while the `kernel` mode measures GPU time. Could you help confirm whether you are using the `operator` mode and capturing end-to-end time in your tests?

chx7514:
Confirmed — the numbers in the PR body were run with `--mode kernel`, and for this operator that measures end-to-end time, not GPU kernel time.

## Why

Both modes time the same callable: `Benchmark.run()` builds `fn = lambda: op(*args, **kwargs)` (`benchmark/base.py:286`) either way, and the modes differ only in the clock wrapped around it. They can disagree only if `fn()` returns with device work still queued — which cannot happen here, because our wrapper has blocking host operations on the critical path:

```python
if bool(has_gap.item()):     # sync
total_valid = int(offsets[N].item())   # sync
```

and the `_nested_view_from_buffer` call that follows is a CPU-side op that additionally drains the queue. So `do_bench`'s `end_event` cannot record before the kernels have actually finished.

Measured, H20, fp16, with `lengths = randint(1, L+1)` so the empty-component path is not involved — `do_bench` (kernel) vs wall-clock loop (operator) vs `torch.profiler` GPU-only:

| shape (N,L,D) | mode=kernel | mode=operator | pure GPU | GPU % of reported |
|---|---|---|---|---|
| (16, 64, 64) | 0.161 ms | 0.193 ms | 0.0034 ms | 1.7% |
| (32, 128, 64) | 0.202 ms | 0.233 ms | 0.0043 ms | 1.9% |
| (64, 256, 128) | 0.237 ms | 0.258 ms | 0.0095 ms | 3.8% |
| (128, 512, 128) | 0.308 ms | 0.339 ms | 0.0166 ms | 5.3% |

The modes agree to within ~15%, and the signature is host-bound: latency stays roughly flat while the GPU fraction climbs 1.7% → 5.3%.

## Where the time goes — `[128, 512, 128]`, fp16

Against a `do_bench` median of 0.308 ms:

| component | time | % of e2e |
|---|---|---|
| `aten._nested_view_from_buffer` | 0.107 ms | **~35%** |
| three device round trips (2 × `.item()` + D2H) | 0.097 ms | **~32%** |
| other host work (launch / alloc / assembly / dispatch) | ~0.087 ms | ~28% |
| **GPU kernels** | **0.017 ms** | **~5%** |

Two host items are roughly two-thirds of the operator: the aten call that wraps the flat buffer into a legacy NestedTensor (CPU-side, inside PyTorch, itself growing with N — 0.023 ms at N=16 to 0.107 ms at N=128), and the device round trips needed to bring ragged lengths back to the host. The speeds below 1.0x on fp16/fp32 are this overhead, not a dtype effect — fp32 shows the same profile. (The bf16 asymmetry is as explained in my earlier comment: `nested_from_padded_cuda` only ships fused launchers for fp32/fp16, so the bf16 baseline falls back to a slower generic path.)

## Corrected numbers

Re-run on the current head, both modes:

| dtype | (16,64,64) | (32,128,64) | (64,256,128) | (128,512,128) |
|---|---|---|---|---|
| float16 | 0.507 | 0.501 | 0.465 | 0.400 |
| float32 | 0.515 | 0.496 | 0.457 | 0.428 |
| bfloat16 | 1.909 | 2.913 | 4.556 | 7.284 |

(arithmetic mean 1.703; `--mode operator` gives 1.537)

The 2.012 currently in the PR-body table predates a fix, described below.

## One bug found while investigating — fixed

The benchmark's input generator draws `lengths = randint(0, L+1)`, so ~20% of draws contain an empty component. Our wrapper detected that case and switched wholesale to `torch.nested.nested_tensor(list)`, which at large N cost ~4x the normal path (1.216 ms vs 0.300 ms at `[128,512,128]`, fp16) while the torch baseline is flat across the same input (0.120 vs 0.125 ms). Since `do_bench` sizes its repeat count from a single early estimate, an unlucky draw inflated a whole measurement — that is why the benchmark was flaky (three consecutive fp16 runs at the largest shape gave `2.358 / 0.313 / 0.312`) and why the body's numbers are stale.

The fallback rested on two documented `_nested_view_from_buffer` quirks:

- **"GPU metadata segfaults" — real.** Metadata on CUDA core-dumps (exit 139); keeping it on CPU is correct.
- **"Any zero-length component segfaults" — does not hold.** An empty component trivially passes `checkInBoundsForStorage`, and upstream `_nested_view_from_buffer` is unchanged across the whole supported range (differs only by a doc comment between v2.7.1 and v2.11.0) with no zero-length check. Verified on 2.11.0+cu130 with empty components at first / last / all / alternating positions, including `total_valid == 0`.

So it was dead code that only ever cost us. Removed (net -14 lines, pushed): 37/37 unit tests pass, layout / `_nested_tensor_size` / `_nested_tensor_storage_offsets` / `values()` match the reference for every injected empty-component case across fp16/fp32/bf16, and the benchmark is now stable across repeated runs.

The remaining headroom is entirely host-side — collapsing the three device round trips into one, and reducing the `_nested_view_from_buffer` cost. The GPU side is ~5% of this operator, so there is nothing meaningful left to win in the kernel itself.

bin913 (approved):

PR #5705, opened 2026-08-24
[KMCompiler][iluvatar] Fix sparse_sampled_addmm compilation failure for K < 16 (tl.dot K >= 16 constraint)

littlekid-zxt:
### PR Category
Operator

### Type of Change
Bug Fix

### Description
On Iluvatar (CoreX), all 240 `test_sparse_sampled_addmm` cases with `M=N=K=5`(sparsity 0.0/0.5) fail with a Triton compilation error:
```
triton.compiler.errors.CompilationError: ... Input shapes should have M >= 1, N >= 1 and K >= 16
```
Root cause: the Iluvatar override(`src/flag_gems/runtime/backend/_iluvatar/ops/sparse_sampled_addmm.py`) routes the dense product through `_sddmm_bmm_kernel`, an autotuned GEMM-style kernel reusing the `mm` tunedconfigs. Iluvatar's Triton fork requires `tl.dot` K dimension >= 16 (`min_dot_size` check), and its autotune AABS (Auto-Adjusted Block Size) shrinks `BLOCK_K` to `next_power_of_2(K)` — for `K = 5` this yields `BLOCK_K = 8`, violating the constraint.

Fix: floor the kernel's K tile at 16 lanes instead of padding the input tensors.`_sddmm_bmm_kernel` now derives `BLOCK_K_PAD = max(BLOCK_K, 16)` via `@triton.heuristics` — computed in Python at launch, so it is guaranteed to be a constexpr even on older Iluvatar toolchains where builtin max() does not fold to constexpr inside @triton.jit — and uses it for the K arange, loop trip counts and pointer strides. The tail-masked loads zero the padded lanes, so the dot result is unchanged; `EVEN_K` is judged against the padded tile to stay consistent. For every tuned config with `BLOCK_K >= 32` this is an identity and generates identical IR, so there is no `K = 0` still takes the existing early-return path.


Caeruleann (dismissed):

src/flag_gems/runtime/backend/_iluvatar/ops/sparse_sampled_addmm.py
```diff
@@ -238,6 +238,13 @@ def _sparse_sampled_addmm_impl(input, mat1, mat2, *, beta=1.0, alpha=1.0, out=No
 
     mat1_f = mat1.contiguous().reshape(B, M, K)
     mat2_f = mat2.contiguous().reshape(B, K, N)
+
+    if K < 16:
+        pad_k = 16 - K
+        mat1_f = torch.nn.functional.pad(mat1_f, (0, pad_k))
```
103yiran: Please don't use pytorch api in Op, refference: contribution  in docs.

littlekid-zxt: Done

Caeruleann (approved):

PR #5724, opened 2026-08-25
[KernelGen][Nvidia] Add _cudnn_attention_backward operator with Triton kernel

kkkwb:
**Changes**

- `src/flag_gems/ops/cudnn_attention_backward.py`: raise NotImplementedError for dropout_p > 0.0 at the backward entry, matching the forward behavior; dropout_p < 0 takes the no-dropout path and is equivalent to 0.0.
- `tests/test_cudnn_attention_backward.py`: add two tests:
      - test_cudnn_attention_backward_dropout_pos_rejected: verifies dropout_p > 0 raises NotImplementedError.
      - test_cudnn_attention_backward_dropout_neg_degrades_to_zero: verifies dropout_p < 0 matches dropout_p = 0.0.

Fixes #5153
### Progress


conf/operators.yaml
```diff
@@ -3261,6 +3261,23 @@ ops:
...
+  for:
+  - _cudnn_attention_backward
+  labels:
+  - aten
+  - NoCPU
+  - KernelGen
+  kind:
+  - NeuralNetwork
+  stages:
+  - beta: '5.5'
```
Caeruleann, line 3280:
```suggestion
  - alpha: '5.4'
```

tests/test_cudnn_attention_backward.py
Caeruleann, line 1: Please use flag_gems.cudnn_attention_backward instead of flag_gems.ops.cudnn_attention_backward

benchmark/test_cudnn_attention_backward.py
Caeruleann, line 1: Please use flag_gems.cudnn_attention_backward instead of flag_gems.ops.cudnn_attention_backward

tests/test_cudnn_attention_backward.py
```diff
@@ -0,0 +1,375 @@
...
+            False,
+            scale=scale,
+        )
+
+
+@pytest.mark.skipif(not torch.cuda.is_available(), reason="CUDA is required")
+@pytest.mark.cudnn_attention_backward
+@pytest.mark.parametrize("dtype", [torch.float16, torch.bfloat16])
+def test_cudnn_attention_backward_dropout_neg_degrades_to_zero(dtype):
+    """dropout_p < 0 is treated as no-dropout and must match dropout_p=0.0."""
```
Caeruleann, line 302: Is this behavior consistent with torch ATen? Why not compare the result directly against ATen?

tests/test_cudnn_attention_backward.py
Caeruleann, line 1: Please cover cases that value head dimensions that differ from Q/K.

benchmark/test_cudnn_attention_backward.py
```diff
@@ -0,0 +1,253 @@
...
+def cudnn_attn_forward_native(
+    Q,
+    K,
+    V,
+    attn_bias=None,
+    is_causal=False,
+    dropout_p=0.0,
+    softmax_scale=None,
+):
+    scale = softmax_scale or (1.0 / math.sqrt(Q.shape[-1]))
```
Caeruleann, line 52: Please add native-reference coverage for scale=None, a non-default scale, and especially scale=0.0. Avoid using `softmax_scale or default` in the test helper because that would mask the same bug.

tests/test_cudnn_attention_backward.py
```diff
@@ -0,0 +1,375 @@
...
+        q_seq_len,
+        num_head,
+        head_size,
+        dtype=dtype,
+        device=flag_gems.device,
+    )
+
+    attn_bias = None
+    if has_attn_bias:
+        attn_bias = (
```
Caeruleann, line 136: Please extend the attention-bias tests to cover the native-supported 2D, 3D, and broadcast 4D forms, and compare all gradients against ATen.

tests/test_cudnn_attention_backward.py
kkkwb: Switched to flag_gems.cudnn_attention_backward. Commit e8630468c90ecb8b7a4a573893ba1fcf4d7d955c.

benchmark/test_cudnn_attention_backward.py
kkkwb: Switched to flag_gems.cudnn_attention_backward. Commit e8630468c90ecb8b7a4a573893ba1fcf4d7d955c.

tests/test_cudnn_attention_backward.py
kkkwb: Now compared directly against ATen: for dropout_p < 0, which takes the no-dropout path, the gradients are checked against the ATen backward reference at dropout_p=0.0 instead of a gems-internal comparison. dropout_p > 0 still raises NotImplementedError, consistent with the forward. Commit e8630468c90ecb8b7a4a573893ba1fcf4d7d955c.

benchmark/test_cudnn_attention_backward.py
kkkwb: Benchmark helper: removed the softmax_scale or default idiom; the helper now passes scale = softmax_scale straight through.
Operator: fixed the same bug — the falsy check silently replaced an explicit scale=0.0 with the default; it now uses is not None, so scale=0.0 zeroes the score gradients, distinct from scale=None.
Tests: test_cudnn_attention_backward_scale covers None, 0.37, and 0.0 against ATen.
Commit e8630468c90ecb8b7a4a573893ba1fcf4d7d955c.
Note: the ATen comparison for scale=0.0 is skipped on A100 (SM80) only (commit 76cf535f2), where ATen's cuDNN backend returns all-NaN gradients (a -inf * 0 fusion artifact). All other architectures still run the full comparison.

tests/test_cudnn_attention_backward.py
kkkwb: Added tests covering 2D (seq_q, seq_k), 3D (batch, seq_q, seq_k), and the (1,1), (1,H), (B,H) broadcast 4D forms, with all gradients compared against ATen. Commit e8630468c90ecb8b7a4a573893ba1fcf4d7d955c.

tests/test_cudnn_attention_backward.py
kkkwb: This case is now supported. A dedicated pair of backward kernels uses separate HEAD_DIM_QK and HEAD_DIM_V constexprs with independent offsets and accumulators for the q/k dim and the value dim; head dims that are not powers of two (e.g. 96) are padded to the next power of two and masked inside the kernels. The entry point routes to these kernels only when the value head dim differs from the q/k head dim; the matching case keeps the shared-kernel path unchanged. Tests cover (64,96), (128,96), (64,32), (96,128) mismatch combinations plus the (64,64) shared-path fallback, with all gradients compared directly against ATen. Commit 20a32c7e86c9812b992278135a3be58607abc294.

Caeruleann: Thanks for adding dense coverage for different value head dimensions. Please also cover equal non-power-of-two dimensions (Dqk = Dv = 96) and varlen inputs with unequal dimensions (Dqk = 64, Dv = 32/128), which still reach the shared kernels. Please handle these cases safely or explicitly reject unsupported configurations before kernel launch.

tests/test_cudnn_attention_backward.py
Caeruleann: Could you also test with `batch > 1`? With `batch = 1`, some bias shapes become indistinguishable, so these cases don’t really verify broadcasting across batches. Please also compare `dQ`, `dK`, and `dV` against ATen for the supported broadcast shapes.

tests/test_cudnn_attention_backward.py
kkkwb: Thanks for the review.
Equal non-power-of-two dims (Dqk = Dv = 96) are now supported: the dense entry routes to the dual-dim kernels whenever the head dims are not equal power-of-two, and those kernels pad non-power-of-two dims and mask the tail. Covered by (96, 96) in test_cudnn_attention_backward_value_head_dim.
Varlen inputs are now explicitly rejected with NotImplementedError before kernel launch, for equal and unequal head dims alike, since the shared varlen kernels use a single HEAD_DIM constexpr. Covered by the new test_cudnn_attention_backward_varlen_rejected for (64, 64), (64, 32), (64, 128) and (96, 96).
While fixing this, the stride order passed to the varlen kernels was also corrected to the m/h/d order of the 3D varlen layout.
Commit 991177f1158ca33c2f20b7a77969d5114e4342e8.

tests/test_cudnn_attention_backward.py
kkkwb: Thanks for the suggestion.
test_cudnn_attention_backward_attn_bias_shapes is now parameterized over batch in [1, 3], so the 3D and (B, H) forms are distinguishable from their lower-rank broadcasts and cross-batch broadcasting is actually verified. dQ, dK and dV are compared against ATen for every bias shape at both batch values. Commit 991177f1158ca33c2f20b7a77969d5114e4342e8.

tests/test_cudnn_attention_backward.py
Caeruleann, line 1: Please also cover the dual-dim path with attention bias, causal mode, and GQA/batch>1. The current bias tests only exercise the shared 64/64 kernel, while the value-dim tests use no bias, non-causal mode, and equal head counts.

src/flag_gems/ops/cudnn_attention_backward.py
```diff
@@ -0,0 +1,1081 @@
...
+        lse = lse.squeeze(-1)
+
+    if dropout_p > 0.0:
+        raise NotImplementedError(
+            "cudnn_attention_backward: dropout > 0 is not yet supported"
+        )
+
+    is_dropout = dropout_p > 0.0
+    rng_tuple = _parse_philox(philox_seed, philox_offset) if is_dropout else None
+    use_varlen = (cum_seq_q is not None) and (cum_seq_k is not None)
```
Caeruleann, line 1034: Please reject the call when either cum_seq_q or cum_seq_k is provided. Currently, a one-sided cum-seq argument silently falls through to the dense path and is ignored;

src/flag_gems/ops/cudnn_attention_backward.py
```diff
@@ -0,0 +1,1081 @@
...
+    Q = Q.contiguous()
+    K = K.contiguous()
+    V = V.contiguous()
+    Out = Out.contiguous()
+    dOut = dOut.contiguous()
+    L = L.contiguous()
+    if has_bias:
+        attn_bias = _normalize_attn_bias(
+            attn_bias, Batch, H_q, SeqLen_q, SeqLen_k
+        ).contiguous()
```
Caeruleann, line 556: Expanding the bias and immediately calling .contiguous() materializes the full (B, H, Sq, Sk) tensor, so a small broadcast bias can cause a very large allocation or OOM for realistic sequence lengths. Please preserve the stride-0 expanded view and pass its strides to the kernel; the dual-dim path has the same issue.

tests/test_cudnn_attention_backward.py
kkkwb: Added `test_cudnn_attention_backward_dual_dim_bias`, parameterized over the full combination of the following dimensions — 144 cases in total, all checked directly against the ATen backward reference:
- dual-dim head dim pairs: (64, 32), (96, 96), (64, 128)
- attention bias broadcast forms: (1,1), (1,H), (B,H)
- causal and non-causal
- GQA (group_size=2) and non-GQA
- batch in {1, 3}
GQA cases give K/V the KV head count while Q keeps the query head count, so the group-broadcast path is exercised. The dK/dV assertions use `atol=8*torch.finfo(dtype).eps`: dK/dV reduce over the KV sequence, and the GEMS-vs-ATen fp16/bf16 rounding difference at near-zero elements is bounded by the dtype's quantization step, which the default elementwise atol does not cover; the tolerance corresponds to roughly 2 ulps. dQ keeps the default tolerance.
Commit 22cebfd.

src/flag_gems/ops/cudnn_attention_backward.py
kkkwb: Fixed: the entry check now uses `or` instead of `and`, so passing either `cum_seq_q` or `cum_seq_k` raises `NotImplementedError` instead of being silently ignored on the dense path. The now-unreachable `cu_seq_q`/`cu_seq_k` forwarding to the internal helper was removed as well. `test_cudnn_attention_backward_onesided_cumseq_rejected` covers both the q-only and k-only cases. 
Commit 22cebfd.

src/flag_gems/ops/cudnn_attention_backward.py
kkkwb: Fixed: both the dense and dual-dim paths drop the `.contiguous()` call and pass the stride-0 expanded view returned by `_normalize_attn_bias` directly to the kernels. Bias reads inside the kernels (`tl.load`) are fully strided through `stride_bb/bh/bm/bn`, so broadcast dimensions simply reuse the same underlying block — no materialization needed. The dBias write path is never active in this operator (`bias_requires_grad=False` by construction), so the stride-0 view cannot cause write conflicts. Commit 22cebfd.

YangLong114514 (approved):

Caeruleann (approved):

PR #5730, opened 2026-08-25
[KernelGen][Nvidia] Add foreach unary operators with shared multi-tensor executor

CoyeCHEN:
## Summary

Add the complete ATen `_foreach_*` family — **90 operators (46 functional + 44 in-place)**, **145 ATen registrations** — with one shared multi-tensor executor.

This PR replaces the original unary-only scope with the whole family: unary, multi-input, ternary, reduction and mutation variants all dispatch through a single fused multi-tensor Triton executor.

## Operators Covered

**30 unary operations** (functional + in-place = 60 registrations):
`abs`, `acos`, `asin`, `atan`, `ceil`, `cos`, `cosh`, `erf`, `erfc`, `exp`, `expm1`, `floor`, `frac`, `lgamma`, `log`, `log10`, `log1p`, `log2`, `neg`, `reciprocal`, `round`, `rsqrt`, `sigmoid`, `sign`, `sin`, `sinh`, `sqrt`, `tan`, `tanh`, `trunc`

**Multi-input**: `add`, `sub`, `mul`, `div`, `pow` (List / Scalar / ScalarList / Tensor), `clamp_max`, `clamp_min`, `maximum`, `minimum`, `lerp` (List / Scalar / ScalarList), `addcmul`, `addcdiv`, `copy`

**Reduction / mutation**: `_foreach_max`, `_foreach_norm`, `_foreach_zero`, `_foreach_zero_`

Coverage comes from a live audit of ATen schemas on the supported torch versions: **all 145 non-`.out` overloads are registered**, each under its full overload name where ATen names one (e.g. `_foreach_add.Scalar`) — no overload leaks to the default implementation. `.out` overloads are deliberately unregistered; in this family they decompose to the functional form, verified by watching the FlagGems kernel take the dispatch under the debug log. `_foreach_powsum` (upstream torch 2.11) is out of scope — it does not exist in the torch versions this repo pins.

## Implementation

- **Shared executor** (`src/flag_gems/utils/foreach.py`): true multi-tensor kernel — one launch per (device, dtype) group covering the whole TensorList; element-wise math is passed in as a `tl.constexpr` function. Launch count is O(dtype/device groups), independent of TensorList length.
- **Layered registry**: `ops/_foreach_unary.py` (30-row table → 60 wrappers), `ops/_foreach_binary.py` (paired/scalar/ternary kernels, wrappers generated from one table), `ops/_foreach_reduction.py` (max/norm fused into one launch per list, host-side root on the power sum), `ops/_foreach_complex_math.py` (21 helpers for complex32/64/128, working around Triton backend gaps)
- **Math reuse**: unary math is pulled from the existing single-tensor FlagGems kernels — no formula duplicated. Reductions reuse existing Triton kernels rather than new ones.
- **Correctness details**: non-contiguous operands are densely paired before entering the kernel (flat-index pairing would misalign elements); `zero_` on strided views writes back through a staging buffer; scalar Tensors stay on device in `.Tensor` overloads (no `.item()` sync); promotion and overlap rejection match ATen.

## Why Shared Infrastructure?

- The 90 operators differ only in element-wise math, output dtype policy and supported dtypes
- A "one kernel per operator" approach would duplicate 145 registration wrappers and 145 kernel definitions
- The shared executor eliminates that duplication while keeping per-overload ATen semantics exact

## Registration & Naming

- Operator IDs follow repo convention (`foreach_abs`, not `underscore_foreach_abs`), so `pytest -m foreach_abs` selects exactly that operator
- Per-parameter markers preserved for single-operator selectivity (`-m foreach_add_list` → its own cases only)
- All 145 registered keys verified live with negative controls

## Test Results

**Correctness

| Suite | GPU (default) | quick-cpu |
|---|---|---|
| `tests/test_foreach_ops.py` | 441 passed in 11.45s | 441 passed in 3.15s |
| `tests/test_foreach_unary.py` | 473 passed in 42.23s | 472 passed, 1 skipped in 5.44s |
| `benchmark/test_foreach_ops.py` | 85 passed in 16:29 | — |
| `benchmark/test_foreach_unary.py` | 62 passed in 10:44 | — |

- Coverage across functional + in-place × 90 ops × (float16, bfloat16, float32, int8–64, bool, complex64/128)
- 1D/2D/3D shapes, mixed shapes, empty tensors, zero-dim tensors, non-contiguous layouts
- Error parity, `.out` decomposition, in-place view write-through, overlap rejection
- The single skip is `_foreach_abs` on `bool` in quick-cpu mode — no CPU reference exists to compare against (pre-existing, not introduced here)

**Liveness**: 145/145 registrations confirmed live

**Launch-count anti-degeneration**:
- ✅ Launch count = O(dtype/device groups), independent of TensorList length
- ✅ Verified via PyTorch profiler on TensorLists of length 1, 16, 128

## Performance

**`pytest -s benchmark/test_foreach_{ops,unary}.py --level core --record log`** — 1305 measurement rows, all `SUCCESS`. Lists of 16 tensors, per-tensor 256²/512²/1024², fp16/bf16/fp32. Speedup is Gems vs. native PyTorch eager.

**In-place unary (the optimizer-facing path) — where the fused executor wins**:
- 1.35x median, 4.50x peak (`lgamma_`, bf16); fp16 1.39x, fp32 1.21x, bf16 1.35x
- By size: 1.57x @256², 1.39x @512², 1.03x @1024² — 22μs vs. torch's 34μs @256², 67μs vs. 70μs @1024²
- `lgamma_`, `erfc_`, `sqrt_`, `log1p_`, `sigmoid_`, `sin_` reach 1.8–4.5x

**Below 1.0x **:
- unary functional: 0.57x median (0.48x @256² → 0.79x @1024²)
- binary `_list`: 0.64x median, 1.67x peak
- binary `_scalar` / `_scalar_list`: 0.30x median, 0.73x peak
- ternary (`lerp`/`addcmul`/`addcdiv`): 0.21x median, 0.73x peak
- `copy`: 0.98x
- `max` / `norm`: 0.02–0.47x (norm, 1.6ms vs. torch 32μs)
- `zero` / `zero_`: 0.33x / 0.05–0.29x
- `.Tensor` overloads: 0.09x median, 0.72x peak
- Overall: 0.52x median, 0.65x mean; 305/1305 rows ≥1.0x, 89/1305 ≥1.5x
- Root cause: **Per-call Python dispatch floor (~45–70μs)

## Key Design Decisions

1. **One executor, many registrations**: violates the prior "kernel count = registration count" guideline, approved for the foreach family
2. **Per-overload registration**: 145 keys registered individually rather than relying on a `default` — the family has 65 `default` overloads and 80 named-only ones, so a blanket registration would silently mis-cover
3. **No `.out` overload registration**: ATen decomposes `.out` to the functional variant in this family
4. **`_foreach_powsum` excluded**: not present in the pinned torch versions; registering it produces a dead key
5. **Shared infrastructure allowed**: `utils/foreach.py` is cross-operator infrastructure, not a single-operator PR scope violation
6. **On-device `.Tensor` scalars**: expanded rather than read back with `.item()`, keeping the launch asynchronous — though this is the source of the staging-copy cost noted above

## Files Changed (14 files, +6115 lines)

- `src/flag_gems/utils/foreach.py` (new): shared multi-tensor executor
- `src/flag_gems/ops/_foreach_unary.py` / `_foreach_binary.py` / `_foreach_reduction.py` / `_foreach_complex_math.py` (new): operator tables and kernels
- `tests/test_foreach_unary.py` / `tests/test_foreach_ops.py` (new): correctness + liveness + launch-count tests
- `benchmark/test_foreach_unary.py` / `benchmark/test_foreach_ops.py` (new) + `benchmark/core_shapes.yaml`: performance benchmarks
- `src/flag_gems/__init__.py` / `src/flag_gems/ops/__init__.py`: +145 registrations and exports
- `conf/operators.yaml`: +145 operator entries (all with `KernelGen` label, stage `alpha: 5.4`); `conf/ci_test_aliases.yaml`: +145 aliases

## Generated by KernelGen

All kernels in this PR are generated/adapted via the KernelGen toolchain.


src/flag_gems/ops/_foreach_binary.py
```diff
@@ -0,0 +1,546 @@
...
+
+def _apply_list(name: str, self, other, inplace: bool, alpha=1):
+    """The ``.List`` overload: pair the two lists position by position."""
+    op = BINARY_OPS[name]
+    _log(name, "List", inplace)
+    if alpha != 1:
+        # ``alpha`` scales the second operand before the operation.  Folding it
+        # in on the host keeps one compiled kernel per operator instead of a
+        # second variant carrying an unused argument in the common case.
+        other = torch._foreach_mul(other, alpha)
```
Caeruleann, line 207: Do not use torch api here

src/flag_gems/ops/_foreach_binary.py
```diff
@@ -0,0 +1,546 @@
...
+    if not isinstance(other, torch.Tensor):
+        raise TypeError("argument 'other' must be a Tensor")
+    op = BINARY_OPS[name]
+    _log(name, "Tensor", inplace)
+    self_list = list(self)
+    expanded = [
+        other if other.shape == t.shape else other.expand_as(t) for t in self_list
+    ]
+    if alpha != 1:
+        expanded = torch._foreach_mul(expanded, alpha)
```
Caeruleann, line 261: Do not use torch api here

src/flag_gems/ops/_foreach_reduction.py
```diff
@@ -0,0 +1,253 @@
...
+    t = tl.program_id(0)
+    in_ptr = tl.load(meta_ptr + t).to(tl.pointer_type(IN_DT))
+    n_elements = tl.load(meta_ptr + tl.num_programs(0) + t)
+
+    acc = tl.full((BLOCK,), NEG_INF, dtype=tl.float32)
+    for offset in tl.range(0, n_elements, BLOCK):
+        idx = offset + tl.arange(0, BLOCK)
+        mask = idx < n_elements
+        x = tl.load(in_ptr + idx, mask=mask, other=NEG_INF).to(tl.float32)
+        acc = tl.maximum(acc, x)
```
Caeruleann, line 63: Please preserve the input dtype in _foreach_max instead of casting the values and reduction state to FP32, as this loses precision for FP64 and int64 inputs; the per-result acc[slot].to(dtype) also adds extra launches and undermines the single-launch design. Please add FP64/int64 parity tests with values beyond the exact FP32 range.

src/flag_gems/utils/foreach.py
```diff
@@ -0,0 +1,1008 @@
...
+    results: List[torch.Tensor] = []
+    buckets: Dict[Tuple[Any, torch.dtype, torch.dtype], Tuple[List, List, List]] = {}
+    writebacks: List[Tuple[torch.Tensor, torch.Tensor]] = []
+
+    for i, (a, b) in enumerate(zip(self, other)):
+        check_dtype_supported(a.dtype, allowed_dtypes)
+        if out_dtypes is not None:
+            out_dtype = out_dtypes[i]
+        else:
+            out_dtype = _result_dtype(a, b)
```
Caeruleann, line 720: Please give _foreach_copy copy-specific dtype semantics instead of using the generic binary promotion path. The functional result should keep self.dtype, and the in-place form should allow supported source-to-destination dtype conversion rather than rejecting it through the generic result_type check. Please add mixed-dtype parity tests.

src/flag_gems/ops/_foreach_reduction.py
```diff
@@ -0,0 +1,253 @@
...
+        for slot, i in enumerate(idxs):
+            results[i] = acc[slot].to(in_dtype)
+    return results  # type: ignore[return-value]
+
+
+def _finish(total: torch.Tensor, ord_f: float) -> torch.Tensor:
+    """Apply the final root that turns a power sum into a norm."""
+    if ord_f == 1.0:
+        return total
+    return total.sqrt() if ord_f == 2.0 else total.pow(1.0 / ord_f)
```
Caeruleann, line 190: Please handle special norm orders explicitly instead of routing them through the generic power-sum path. ord=0 currently raises ZeroDivisionError, while ord=inf produces an incorrect result instead of max(abs(x)). Please add parity coverage for these cases.

src/flag_gems/ops/_foreach_binary.py
```diff
@@ -0,0 +1,546 @@
...
+# ---------------------------------------------------------------------------
+
+_FLOAT = (torch.float16, torch.bfloat16, torch.float32, torch.float64)
+_INT = (torch.int8, torch.int16, torch.int32, torch.int64, torch.uint8)
+
+# Arithmetic accepts bool as well; the comparison-style clamps do not reject it
+# either.  Complex is excluded here because the paired kernels have no
+# interleaved-component path (the unary executor's complex support does not
+# generalise to two operands).
+ARITH = frozenset(_FLOAT + _INT + (torch.bool,))
```
Caeruleann, line 137: Please handle bool support per operator instead of sharing the same ARITH dtype set. ATen supports bool add but rejects bool sub, while the current Triton int1 arithmetic makes cases such as True + True produce False. Please add bool parity tests for both operators.

src/flag_gems/ops/_foreach_complex_math.py
```diff
@@ -0,0 +1,174 @@
...
+# routes those through the libdevice shim, so the complex formulas below use the
+# same source as the single-tensor operators do.
+_sinh = tl_extra_shim.sinh
+_cosh = tl_extra_shim.cosh
+_atan2 = tl_extra_shim.atan2
+
+
+@triton.jit
+def _hypot(re, im):
+    return tl.sqrt(re * re + im * im)
```
Caeruleann, line 29: Please make _hypot overflow-safe. The current sqrt(re * re + im * im) overflows for large finite complex inputs, which also propagates incorrect results to abs, log, sqrt/rsqrt, and other complex ops that depend on it. Please use a scaled hypot formulation and add large complex64/complex128 parity cases.

bin913:
@Qchan233 Four operators, namely `foreach_binary`, `foreach_complex_math.py`, `foreach_reduction` and `foreach_unary`, have been added under the operator directory. Please add corresponding tests for these operators in `benchmark` and `tests`. So far only the test for `foreach_unary` has been implemented.

kkkwb:
> @Qchan233 Four operators, namely `foreach_binary`, `foreach_complex_math.py`, `foreach_reduction` and `foreach_unary`, have been added under the operator directory. Please add corresponding tests for these operators in `benchmark` and `tests`. So far only the test for `foreach_unary` has been implemented.

Tests are now in place for all three modules. Their location follows the operator-id to test-file dispatch, so no new file is introduced:

The 85 binary and reduction ids map to tests/test_foreach_ops.py, covering every overload: accuracy, registration liveness, non-contiguous pairing and dtype-set parity.
The 60 unary ids map to tests/test_foreach_unary.py. Complex math does not get its own file because it is not an operator of its own — it is the complex-input implementation of abs/log/log1p/log2/log10/sqrt/rsqrt, so its cases belong alongside those operators.
Benchmarks are carried by benchmark/test_foreach_ops.py and benchmark/test_foreach_unary.py.
Commit fcbeb00.

src/flag_gems/ops/_foreach_binary.py
kkkwb: This was folding alpha into the operator through a host-side temporary: torch._foreach_mul(other, alpha). alpha now enters the kernel as ALPHA: tl.constexpr on the paired executor, so the identity value 1 compiles away and no extra dispatch or temporary tensor is created. Commit fcbeb00.

src/flag_gems/ops/_foreach_binary.py
kkkwb: Same issue: torch._foreach_mul(expanded, alpha). It is removed along with the rest of the alpha handling. The argument now also exists only on the overloads ATen gives it (add.List, add.Tensor, sub.List); previously every overload accepted it, which was broader than ATen. Commit fcbeb00a9c0e2b82ecd925de78ba5c5ea9fb34d8.

src/flag_gems/ops/_foreach_reduction.py
kkkwb: The input dtype is now preserved: the accumulator and the result are both the input dtype, with no fp32 round trip. The old path overflowed fp64 inputs (max([1e300, -1e300]) returned inf), and the per-result .to(dtype) added a launch per tensor; the single-launch design is restored. FP64/int64 parity cases with values beyond the exact fp32 range are added. Commit fcbeb00a9c0e2b82ecd925de78ba5c5ea9fb34d8.

src/flag_gems/utils/foreach.py
kkkwb: This now uses assignment semantics instead of generic promotion: the functional form keeps self.dtype (fp16 stays fp16 for an fp32 source), and the in-place form accepts the source-to-destination cast the generic result_type check rejected. Mixed-dtype parity cases cover both forms across float, integral and bool sources. Commit fcbeb00a9c0e2b82ecd925de78ba5c5ea9fb34d8.

src/flag_gems/ops/_foreach_reduction.py
kkkwb: The special orders are handled explicitly: ord=0 counts nonzero elements and ord=±inf are the extrema of |x|. Through the power-sum path, ord=0 raised ZeroDivisionError and ord=inf returned a wrong value. The root is folded into the reduction kernel, so a norm still costs one launch per (device, dtype) group. Parity coverage for these orders is added. Commit fcbeb00a9c0e2b82ecd925de78ba5c5ea9fb34d8.

src/flag_gems/ops/_foreach_binary.py
kkkwb: bool support is now per operator. add is logical or (True + True is True), mul is logical and, and sub refuses bool - bool with ATen's own message — addition on int1 behaves like and in this backend, so add combines the operands with | explicitly. Separately, pow accepts a bool scalar exponent but rejects a bool tensor operand (NotImplementedError), and the clamp family accepts a bool tensor operand but rejects a bool scalar, so the dtype sets are split into allowed and allowed_scalar. bool parity cases for both operators are added. Commit fcbeb00a9c0e2b82ecd925de78ba5c5ea9fb34d8.

src/flag_gems/ops/_foreach_complex_math.py
kkkwb: Switched to the libdevice hypot. sqrt(re*re + im*im) overflowed for large finite inputs — complex64 at 1e20, complex128 at 1e200 — and propagated inf/nan to abs, log, log1p, log2, log10, sqrt and rsqrt. abs reuses the same implementation. Large-value parity cases for complex64/128 are added. Commit fcbeb00a9c0e2b82ecd925de78ba5c5ea9fb34d8.

Caeruleann (approved):

PR #5778, opened 2026-08-26
fix _weight_norm: split underscore op tests into a dedicated mark

Yukun-Cui:
## Summary

Follow-up to #5080, fixing the pytest mark for the `_weight_norm` (underscore) operator.

#5080 registered the `_weight_norm` operator but kept the pytest mark as `@pytest.mark.weight_norm` (no underscore). That choice was unavoidable *literally* — pytest forbids mark names that start with an underscore (`AttributeError: Marker name must NOT start with underscore`), so `@pytest.mark._weight_norm` is not valid. But it left a real problem: the `_weight_norm` (ATen) entry point and the legacy `weight_norm` (fused) entry point now share the same `weight_norm` mark, so `-m weight_norm` can no longer select just one, and benchmark result aggregation (which keys on the first non-builtin mark) collapses both into `weight_norm`.

### DIFF
--- a/benchmark/test_weight_norm.py
+++ b/benchmark/test_weight_norm.py
@@ -52,9 +52,9 @@ def test_weight_norm_dim0():
     bench = base.GenericBenchmarkExcluse1D(
         op_name="weight_norm",
         input_fn=weight_norm_input_fn,
-        torch_op=torch.ops.aten._weight_norm.default,
-        dtypes=consts.FLOAT_DTYPES,
+        torch_op=torch._weight_norm,
     )
+    bench.set_gems(flag_gems.weight_norm)
     bench.run()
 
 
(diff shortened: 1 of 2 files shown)


benchmark/test_weight_norm.py
```diff
@@ -66,6 +66,34 @@ def test_weight_norm_dim_last():
     bench = base.GenericBenchmarkExcluse1D(
         op_name="weight_norm",
         input_fn=weight_norm_input_fn_last,
+        torch_op=torch._weight_norm,
+    )
+    bench.set_gems(flag_gems.weight_norm)
+    bench.run()
+
+
+@pytest.mark.underscore_weight_norm
```
0x45f, line 75: why use underscore_weight_norm here and L47 use weight_norm

Yukun-Cui: Good question — the two marks are intentional, and the reason is #5080.
﻿
``_weight_norm`` (the ATen entry point, ``torch.ops.aten._weight_norm.default``) and the legacy ``weight_norm`` (**fused**, ``torch._weight_norm`` / ``flag_gems.weight_norm``) are two distinct entry points for the same math. Before #5080 these tests were marked ``weight_norm`` and drove the **fused** path (``torch._weight_norm``). #5080 registered the ``_weight_norm`` ATen op but, since pytest forbids mark names starting with an underscore (AttributeError: Marker name must NOT start with underscore), it couldn't add a ``_weight_norm`` mark — so it reused the existing ``weight_norm`` mark and repointed its torch_op to ``torch.ops.aten._weight_norm.default``. That left the two entry points sharing a single weight_norm mark: ``-m weight_norm`` no longer selected just one, and the benchmark aggregator (which keys on the first non-builtin mark) collapsed both paths into ``weight_norm``.
﻿
This PR restores the pre-#5080 ``weight_norm``-marked tests (**fused** path) and adds a parallel suite under a separate mark for the ``_weight_norm`` ATen path, so each entry point can be selected and recorded independently.
﻿
As for the name: it can't be ``_weight_norm`` (pytest rejects leading-underscore marks), and it can't be ``weight_norm`` either — that would just re-create the #5080 collision. Hence ``underscore_weight_norm``.

0x45f: but I think the test cases in the test_weight_norm.py are all used to test `_weight_norm`

Yukun-Cui: You're right that all cases in ``test_weight_norm.py`` currently test ``_weight_norm`` — but that's the state left by #5080, which repointed the original **fused** ``test_weight_norm`` at ``torch.ops.aten._weight_norm``, and it's exactly the regression this PR fixes. This PR restores ``test_weight_norm`` to test the **fused** path (``torch._weight_norm`` vs ``flag_gems.weight_norm``) and adds a separate ``test_underscore_weight_norm`` for the ATen ``_weight_norm`` path, so each entry point has its own test and mark.

If you'd rather not test the **fused** path, I can just drop it?

0x45f (approved):

PR #5830, opened 2026-08-28
fixbug: exponential_ operator accuracy failure

lyujheng:
### PR Category
Operator

### Type of Change
Bug Fix

### Description

Fix two bugs in the in-place `exponential_` operator (`fused_exponential_kernel_f32_unroll8`):

1. **Philox counter collision**: `block_start = pid * BLOCK` only advanced by `BLOCK` per pid, but each pid consumes `2 * BLOCK` counters (two `tl.philox` calls). This caused adjacent pids to share identical counters, producing 100% repeated random blocks at block_size=128. Fixed by changing to `pid * BLOCK * 2`.

2. **Inaccurate log approximation**: `safe_fast_log_f32` used a 4-term Taylor series for `ln(1+x)`, which has catastrophic error when x → 1 (e.g., 10900% relative error at u=0.999). Replaced with the degree-5 minimax polynomial already used in the out-of-place `exponential` kernel. Also aligned the boundary guard threshold to `u >= 1.0 - 1e-4` (consistent with the out-of-place version) to avoid polynomial edge-case overshoot.

### Issue

- Exponential distribution CDF check fails: `P(X<=0.1) = 0.0` instead of expected ~0.095
- Random block repetition at block_size=128 reaches 100%

### Progress

- [√] Change is fully covered by a UT.

### Performance
image

### Accuracy
image

### DIFF
--- a/src/flag_gems/ops/exponential_.py
+++ b/src/flag_gems/ops/exponential_.py
@@ -40,11 +40,15 @@ def safe_fast_log_f32(x):
     bits = x.to(tl.int32, bitcast=True)
     exponent = (bits >> 23) - 127
     mantissa = (bits & 0x7FFFFF).to(tl.float32) * (1.0 / 8388608.0) + 1.0
-    m1 = mantissa - 1.0
-    return (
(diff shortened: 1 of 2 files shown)


103yiran (approved):
LGTM

PR #5871, opened 2026-09-01
Optimize GLU kernel with TLE autotuning

lzllx123:
## Summary

- Add a TLE-based GLU forward kernel for last-dimension inputs.
- Add autotuning for `ROWS_PER_PROGRAM`, `LOOP_STAGES`, `LOOP_UNROLL`, and `num_warps`.
- Use shape-aware pruning to limit the autotuning search space.
- Keep the existing pointwise implementation as the fallback path.


sunnycase (approved):
LGTM

PR #5877, opened 2026-09-01
[KernelGen][Nvidia] Add _lu_with_info operator with Triton kernel

Yukun-Cui:
## Summary
Adds a Triton kernel for `_lu_with_info`. Computes a compact LU factorization with partial pivoting of a matrix, returning the packed LU matrix, the 1-indexed pivots and an info tensor that reports the first singular diagonal of U (0 on success). This is the deprecated `torch._lu_with_info` interface mirrored on `linalg.lu_factor_ex`.

## Testing

## Performance
Test command: `pytest benchmark/test_lu_with_info.py --level core` (NVIDIA H20)

### lu_with_info

| float32 | [64, 64] | 0.104128 | 0.164000 | 0.635 |
| float32 | [256, 256] | 0.383616 | 0.859232 | 0.446 |
| float32 | [512, 512] | 0.872048 | 2.024448 | 0.431 |
| float32 | [1024, 1024] | 1.930624 | 5.204304 | 0.371 |
| float32 | [8, 128, 128] | 0.284800 | 0.593344 | 0.480 |
| float64 | [64, 64] | 0.124320 | 0.168928 | 0.736 |
| float64 | [256, 256] | 0.623424 | 1.616240 | 0.386 |
| float64 | [512, 512] | 1.984800 | 2.030112 | 0.978 |
| float64 | [1024, 1024] | 5.594528 | 7.395584 | 0.756 |
| float64 | [8, 128, 128] | 0.476768 | 0.776480 | 0.614 |

| lu_with_info | **0.583** |

## Multi-backend Testing
|---|---|---|---|
| Nvidia (H20) | PASS (10 cases) | 0.583 | Primary |
| Tianshu | N/A | — | — |
| Muxi | N/A | — | — |
| Ascend | N/A | — | — |
| Hygon | N/A | — | — |

## Files Changed
- `src/flag_gems/ops/_lu_with_info.py`: Triton kernel implementation
- `tests/test_lu_with_info.py`: Accuracy test
- `benchmark/test_lu_with_info.py`: Performance benchmark
- `conf/operators.yaml`: Add operator entry (kind: LinearAlg, stage: alpha 5.4)


src/flag_gems/ops/_lu_with_info.py
```diff
@@ -0,0 +1,120 @@
...
+    if input.dtype not in (torch.float32, torch.float64):
+        raise NotImplementedError(
+            "FlagGems _lu_with_info currently supports float32 and float64 "
+            f"only, got {input.dtype}"
+        )
+    m, n = input.shape[-2], input.shape[-1]
+    if m == 0 or n == 0:
+        raise NotImplementedError(
+            "FlagGems _lu_with_info currently does not support empty matrices"
+        )
```
Caeruleann: Please handle empty matrices like (0, 3), (3, 0) and their batched forms, returning correctly shaped outputs and zero info as ATen does.

tests/test_lu_with_info.py
```diff
@@ -0,0 +1,158 @@
...
+    ref_inp = utils.to_reference(inp)
+    ref_lu, ref_pivots, ref_info = torch._lu_with_info(ref_inp, pivot=True)
+    res_lu, res_pivots, res_info = flag_gems._lu_with_info(inp, pivot=True)
+    assert res_info.dtype == torch.int32
+    # The all-zero-last-row construction is singular, so every batch element
+    # must report a nonzero info (the position of the zero U pivot) in both
+    # the reference and GEMS, and the two must agree exactly.
+    assert (ref_info > 0).all()
+    assert (res_info > 0).all()
+    utils.gems_assert_equal(res_info, ref_info)
```
Caeruleann, line 149: Please also test zero pivots at the first and intermediate steps, and check LU finiteness and reconstruction as well as info; a zero last row alone misses errors during elimination.

tests/test_lu_with_info.py
```diff
@@ -0,0 +1,158 @@
...
+    assert (res_info >= 0).all()
+    assert (res_info <= k).all()
+    # Validate the factorization by reconstruction rather than the raw LU
+    # storage: different LU implementations (Triton panel-blocked vs cuSOLVER)
+    # produce mathematically equivalent factorizations with different floating
+    # point accumulation orders, so the stored LU entries need not match the
+    # reference bit-for-bit. Reconstructing A = P @ L @ U from each and
+    # comparing those is the mathematically meaningful correctness check,
+    # matching the precedent in test_linalg_lu_factor.
+    torch.backends.cuda.matmul.allow_tf32 = False
```
Caeruleann, line 115: Please restore torch.backends.cuda.matmul.allow_tf32 after the test, preferably with a fixture or try/finally, so it doesn’t affect other tests.

src/flag_gems/__init__.py
```diff
@@ -169,6 +169,7 @@ def torch_ge(v):
     ("_log_softmax.out", log_softmax_out),
     ("_log_softmax_backward_data", log_softmax_backward),
     ("_log_softmax_backward_data.out", log_softmax_backward_out),
+    ("_lu_with_info", _lu_with_info),
```
Caeruleann: Would it be better to remove the leading underscore from the public API name and use lu_with_info, while keeping _lu_with_info as the ATen registration name?

src/flag_gems/ops/_lu_with_info.py
Yukun-Cui: Good catch! I'll remove the NotImplementedError for empty matrices and handle them correctly by:
1. Returning LU with the same shape as input
2. Creating appropriately shaped but empty pivots tensor
3. Returning info tensor filled with zeros
4. Adding test cases for empty matrices

tests/test_lu_with_info.py
Yukun-Cui: Agreed! I'll extend the singular matrix tests to cover:
1. Zero pivot at the first position
2. Zero pivots at intermediate positions
3. Validation that LU contains only finite values
4. Full reconstruction check (P @ L @ U = original matrix)
This will ensure the factorization is correct throughout the elimination process.

tests/test_lu_with_info.py
Yukun-Cui: Absolutely! I'll save the original TF32 setting at the beginning of the test and restore it afterwards using a try/finally block to ensure it's always restored even if the test fails.

tests/test_lu_with_info.py
Caeruleann: These cases still do not create exact zero pivots at the requested positions: partial pivoting can move a zero row, while a linearly dependent row may leave a small nonzero pivot. Please use deterministic matrices such as diagonal matrices with zeros at the first, middle, and last diagonal positions, and assert the expected info values.

src/flag_gems/ops/_lu_with_info.py
Caeruleann: The empty path still misses zero-sized batch dimensions such as (0, 3, 3) and (2, 0, 3, 3), which enter _linalg_lu_factor_impl with batch == 0. Please handle input.numel() == 0 here and add parity cases for empty batch dimensions.

src/flag_gems/ops/_lu_with_info.py
```diff
@@ -0,0 +1,125 @@
...
+    nonzero. This mirrors the cuSOLVER convention: a nonzero ``info`` flags a
+    singular factor (the matrix could not be factored into a nonsingular U),
+    while ``info == 0`` means success. ``NaN``/``Inf`` pivots are treated as
+    singular as well, matching the reference behaviour.
+    """
+    pid_b = tl.program_id(0)
+
+    info_val = tl.zeros((), dtype=tl.int32)
+    for j in tl.range(0, K_MAX):
+        # LU[b, j, j] for a row-major (..., M, N) tensor: offset = b*strd + j*N + j
```
Caeruleann, line 59: info should report the first exactly-zero U pivot. Treating a NaN pivot as singular can produce a nonzero info where cuSOLVER/ATen reports zero, and the comment also mentions Inf although the code does not check it. Please match the native zero-pivot semantics and add NaN/Inf parity tests.

Yukun-Cui: Fixed in 1a700d46. You are right on both counts.

1. The NaN test is gone. `torch.linalg.lu_factor_ex` reports `info == 0` for a matrix whose diagonal is `nan` or `inf` (measured on both CPU and CUDA), so treating NaN as singular reported a failure where ATen reports success. `pivot == 0.0` already excludes NaN and Inf, so no separate test is needed.
2. The docstring's Inf claim is removed -- the code never checked for Inf, and per (1) it should not.

Added kernel-level tests: an exact zero at position 1 with NaN/Inf elsewhere reports `info == 1`; NaN/Inf alone report `info == 0`. 52 tests pass.

One thing worth flagging while investigating this: `info` is derived from FlagGems' own `linalg_lu_factor` output, and for a NaN input that factorization diverges from cuSOLVER -- on a 4x4 with `a[1,1]=nan`, cuSOLVER gives `diag=[2,0,2,2] piv=[1,2,3,4]` while FlagGems gives `diag=[0,0,0,nan] piv=[1,3,1,1]`. Because of that I could not write an end-to-end NaN parity test that compares `info` against the native op; that needs a fix in `linalg_lu_factor`, which is outside this PR. I can open a separate issue for it if you want.

src/flag_gems/ops/_lu_with_info.py
Yukun-Cui: Done in 7e111bb3 + b41f017d (pushed to `pr/_lu_with_info`).

The empty-input branch now keys on `input.numel() == 0` instead of `m == 0 or n == 0`, so a zero-sized batch dimension — `(0, 3, 3)` or `(2, 0, 3, 3)`, which reach `_linalg_lu_factor_impl` with `batch == 0` even though neither `m` nor `n` is zero — takes the same well-defined early path: LU keeps the input shape, pivots are empty along the k axis, info is 0 for every batch element.

New parity coverage: `test_lu_with_info_empty_batch_dimensions` over `(0,3,3)`, `(2,0,3,3)`, `(2,3,0,3)` and `(0,0,4,4)`, asserting LU/pivots/info shapes and info == 0 against the native call. Verified locally: all four shapes match the native outputs.

Yukun-Cui: Done in 7e111bb3 + b41f017d (pushed to `pr/_lu_with_info`).

The empty-input branch now keys on `input.numel() == 0` instead of `m == 0 or n == 0`, so a zero-sized batch dimension — `(0, 3, 3)` or `(2, 0, 3, 3)`, which reach `_linalg_lu_factor_impl` with `batch == 0` even though neither `m` nor `n` is zero — takes the same well-defined early path: LU keeps the input shape, pivots are empty along the k axis, info is 0 for every batch element.

New parity coverage: `test_lu_with_info_empty_batch_dimensions` over `(0,3,3)`, `(2,0,3,3)`, `(2,3,0,3)` and `(0,0,4,4)`, asserting LU/pivots/info shapes and info == 0 against the native call. Verified locally: all four shapes match the native outputs.

tests/test_lu_with_info.py
Yukun-Cui: Fixed in 7e111bb3, with one important finding.

The test now builds diagonal matrices with an exact zero at the first, middle, and last diagonal position, per your suggestion — on a diagonal the pivot search cannot move the zero, so the expected `info` is the 1-indexed position and is asserted exactly against the reference. The earlier zero-last-row constructions are gone.

**The first/middle cases exposed a real kernel bug, now recorded as xfail rather than hidden.** With the zero at position 0 or 2, the shared `linalg_lu_factor` kernel divides by the zero pivot, which poisons the panel with nan and produces a wrong info (reproduced deterministically on the inputs this test constructs):

- first-zero: native `info=1`, diag `[0,2,3,4]` — gems `info=2`, diag `[nan,0,nan,nan]`
- middle-zero: native `info=3`, diag `[1,2,0,4]` — gems `info=1`, diag `[0,2,0,0]`
- last-zero: native `info=5`, gems `info=5` (elimination never touches it, so it passes)

The xfail marker carries the reproducible inputs, the expected values, and the reason. Fixing that kernel is a separate change — `linalg_lu_factor` is shared by `linalg_lu_factor_ex`, `lu_unpack` and other operators, and touching it here would be out of scope for this PR. Once it lands, the xfail flips to a pass automatically.

Caeruleann (approved):

PR #5878, opened 2026-09-01
[KernelGen][Nvidia] Add _fused_sgd_ operator with Triton kernel

Yukun-Cui:
## Summary
Adds a Triton kernel for `_fused_sgd_`. The in-place version of `fused_sgd()`.

## Testing

## Performance
Test command: `pytest benchmark/test_fused_sgd_.py --level core` (NVIDIA H20)

### fused_sgd_

| float16 | [64, 64] | 0.007040 | 0.057184 | 0.123 |
| float16 | [256, 256] | 0.023936 | 0.054880 | 0.436 |
| float16 | [1024, 1024] | 0.028128 | 0.055008 | 0.511 |
| float16 | [4096, 4096] | 0.275008 | 0.182944 | 1.503 |
| float16 | [64, 512, 512] | 0.274352 | 0.176928 | 1.551 |
| float16 | [64, 64] | 0.006752 | 0.006752 | 1.000 |
| float16 | [256, 256] | 0.023744 | 0.024000 | 0.989 |
| float16 | [1024, 1024] | 0.028608 | 0.028704 | 0.997 |
| float16 | [4096, 4096] | 0.275632 | 0.275456 | 1.001 |
| float16 | [64, 512, 512] | 0.274464 | 0.274560 | 1.000 |
| float32 | [64, 64] | 0.006912 | 0.056640 | 0.122 |
| float32 | [256, 256] | 0.024416 | 0.057648 | 0.424 |
| float32 | [1024, 1024] | 0.035424 | 0.057344 | 0.618 |
| float32 | [4096, 4096] | 0.448928 | 0.403840 | 1.112 |
| float32 | [64, 512, 512] | 0.450096 | 0.420000 | 1.072 |
| float32 | [64, 64] | 0.007264 | 0.006944 | 1.046 |
| float32 | [256, 256] | 0.024672 | 0.024960 | 0.988 |
| float32 | [1024, 1024] | 0.036192 | 0.034880 | 1.038 |
| float32 | [4096, 4096] | 0.445248 | 0.442784 | 1.006 |
| float32 | [64, 512, 512] | 0.457248 | 0.441952 | 1.035 |
| bfloat16 | [64, 64] | 0.007008 | 0.054624 | 0.128 |
| bfloat16 | [256, 256] | 0.024416 | 0.055904 | 0.437 |
| bfloat16 | [1024, 1024] | 0.028416 | 0.056992 | 0.499 |
| bfloat16 | [4096, 4096] | 0.278880 | 0.180896 | 1.542 |
| bfloat16 | [64, 512, 512] | 0.277088 | 0.177920 | 1.557 |
| bfloat16 | [64, 64] | 0.007104 | 0.006848 | 1.037 |
| bfloat16 | [256, 256] | 0.024064 | 0.024512 | 0.982 |
| bfloat16 | [1024, 1024] | 0.028544 | 0.028576 | 0.999 |
| bfloat16 | [4096, 4096] | 0.279472 | 0.279424 | 1.000 |
| bfloat16 | [64, 512, 512] | 0.278400 | 0.278464 | 1.000 |

| fused_sgd_ | **0.892** |

## Multi-backend Testing
|---|---|---|---|
| Nvidia (H20) | PASS (254 cases) | 0.892 | Primary |
| Tianshu | N/A | — | — |
| Muxi | N/A | — | — |
| Ascend | N/A | — | — |
| Hygon | N/A | — | — |

## Files Changed
- `src/flag_gems/ops/_fused_sgd_.py`: Triton kernel implementation
- `tests/test_fused_sgd_.py`: Accuracy test
- `benchmark/test_fused_sgd_.py`: Performance benchmark


benchmark/test_fused_sgd_.py
```diff
@@ -0,0 +1,129 @@
...
+@pytest.mark.fused_sgd_
+def test_fused_sgd__tensor_lr():
+    bench = FusedSgdBenchmark(
+        input_fn=_input_fn_tensor_lr,
+        op_name="fused_sgd_",
+        torch_op=_torch_op_tensor_lr,
+        dtypes=consts.FLOAT_DTYPES,
+    )
+    bench.shapes = list(SGD_SHAPES)
+    bench.run()
```
Caeruleann, line 143: This benchmark calls _fused_sgd_.tensor_lr, but the PR only registers _fused_sgd_, so both timings run native PyTorch; please implement and register the tensor-lr overload or remove this benchmark.

benchmark/test_fused_sgd_.py
```diff
@@ -0,0 +1,129 @@
...
+@pytest.mark.fused_sgd_
+def test_fused_sgd_():
+    bench = FusedSgdBenchmark(
+        input_fn=_input_fn,
+        op_name="fused_sgd_",
+        torch_op=_torch_op,
+        dtypes=consts.FLOAT_DTYPES,
+    )
+    bench.shapes = list(SGD_SHAPES)
+    bench.run()
```
Caeruleann, line 130: Can we just pass gems_op explicitly in both benchmarks so they can actually measure FlagGems.

src/flag_gems/ops/_fused_sgd_.py
```diff
@@ -0,0 +1,180 @@
...
+    - b = g                                   (if is_first_step)
+      b = momentum * b + (1 - dampening) * g  (otherwise)
+    - update = g + momentum * b               (if nesterov)
+      update = b                              (otherwise)
+    - param = param - lr * update
+    """
+    logger.debug("GEMS FUSED_SGD_")
+
+    # Skip the whole step if a non-finite gradient was found (AMP).
+    if found_inf is not None and found_inf.numel() > 0 and found_inf.item() > 0:
```
Caeruleann: Please read found_inf and grad_scale inside the Triton kernel instead of calling .item() on the host, which forces GPU synchronization and breaks CUDA Graph capture; please also add an AMP capture/replay test. So as line 143.

src/flag_gems/ops/_fused_sgd_.py
```diff
@@ -0,0 +1,180 @@
...
+        return
+
+    # _fused_sgd supports float16 / bfloat16 / float32 parameters; the kernel
+    # accumulates in float32 and stores back to the original dtype. float64 and
+    # integer dtypes are rejected to match the ATen reference.
```
Caeruleann: Has this been verified against the ATen reference?

benchmark/test_fused_sgd_.py
Yukun-Cui: You're right - I'll implement the tensor_lr overload to properly support per-parameter learning rates, or remove the benchmark if that overload is out of scope for this PR.

benchmark/test_fused_sgd_.py
Yukun-Cui: Good point! I'll pass `gems_op=flag_gems._fused_sgd_` explicitly to ensure the benchmark actually measures the FlagGems implementation.

src/flag_gems/ops/_fused_sgd_.py
Yukun-Cui: Excellent catch! The .item() calls are indeed problematic for performance and graph capture. I'll:
1. Pass found_inf and grad_scale as tensors to the kernel
2. Read their values inside the Triton kernel
3. Add a test that verifies CUDA graph capture/replay works correctly

src/flag_gems/ops/_fused_sgd_.py
Yukun-Cui: Yes, I've tested against torch._fused_sgd_ and the results match. I'll make sure the tests explicitly use the ATen reference for validation.

benchmark/test_fused_sgd_.py
Yukun-Cui: Implemented the overload rather than dropping the benchmark — `aten::_fused_sgd_.tensor_lr` does exist:

```
aten::_fused_sgd_.tensor_lr(Tensor(a!)[] self, Tensor(b!)[] grads,
    Tensor(c!)[] momentum_buffer_list, *, float weight_decay, float momentum,
    Tensor lr, float dampening, bool nesterov, bool maximize, bool is_first_step,
    Tensor? grad_scale=None, Tensor? found_inf=None) -> ()
```

`_fused_sgd__tensor_lr` is now registered and shares the kernel with the float overload; verified against the native overload for both 0-dim and 1-element `lr` tensors.

One implementation note worth flagging, since it looks inconsistent on purpose. The lr plumbing is dtype-dependent: Triton narrows a Python-float kernel argument to fp32, so an fp64 param has to read `lr` from a 0-dim tensor or the update inherits a ~1.5e-9 error (`0.1` arrives as `0.10000000149011612`). Every other dtype passes it as a scalar, because allocating a tensor per call is a host-to-device copy that breaks CUDA graph capture — which is what the .item() comment below is about.

benchmark/test_fused_sgd_.py
Yukun-Cui: Done — both benchmarks now pass `gems_op` explicitly, and the tensor_lr one gets its own `op_name` instead of reusing `fused_sgd_`.

Confirmed it was measuring native-against-native before: the two columns are now distinct (0.006656 ms torch vs 0.014400 ms gems at 64x64 fp16), where previously both paths dispatched through torch.

src/flag_gems/ops/_fused_sgd_.py
Yukun-Cui: Both done, and the note about line 143 caught a second bug.

`.item()` is gone — `found_inf` and `grad_scale` are read inside the kernel now. Verified the sync is actually eliminated by capturing and replaying the step in a CUDA graph; that previously failed outright, and now completes with the param updating on replay.

The bug: the host check was `found_inf > 0`, but ATen compares against **1 exactly**:

```
found_inf=0.0: param=0.900000  UPDATED
found_inf=1.0: param=1.000000  SKIPPED
found_inf=2.0: param=0.900000  UPDATED   <- previously skipped
```

The kernel now tests `== 1`, so a scaler that parks a non-1 sentinel in `found_inf` no longer silently loses the step.

While in here I also checked the grad write-back, since I had just found that the equivalent line in `_fused_adagrad_` was a vectorization artifact. It is **not** one here — ATen mutates `grad` at every size (n = 1, 2, 3, 4, 8, 16, 64, 1000), so this is genuine contract and I left it alone.

src/flag_gems/ops/_fused_sgd_.py
Yukun-Cui: It is now — the previous answer was too confident, and checking properly turned up two real mismatches, both fixed in this push.

Verified against `torch._fused_sgd_` across 384 combinations (4 dtypes x 6 sizes x 8 option sets x grad_scale on/off), comparing to a few ULPs of relative tolerance because the final store rounds to the param dtype: **384/384 match**.

What the comparison found:

- `found_inf` used `> 0` where ATen uses `== 1` (see the thread above)
- the assert rejected float64, but ATen accepts it — and fp64 params were being computed in fp32, losing ~7 digits. They now accumulate in fp64.

bf16 initially showed ~30 apparent failures that turned out to be my tolerance, not the kernel: the deltas were exactly 1 bf16 ULP (7.81e-03 = `finfo(bfloat16).eps`) from ATen rounding the same fp32 intermediate.

src/flag_gems/ops/_fused_sgd_.py
```diff
@@ -0,0 +1,366 @@
...
+    else:
+        # Kept as a Python double so an fp64 param keeps the exact value; routing
+        # it through fp32 would bake in a representation error
+        # (0.1 -> 0.10000000149011612).
+        lr_base = float(lr)
+    # Cache one lr view per (device, dtype) for the tensor path so a mixed-dtype
+    # param list converts once each rather than once per parameter.
+    lr_cache = {}
+
+    for i in range(len(self)):
```
Caeruleann: _fused_sgd_ is a multi-tensor fused operator, but this loop launches one Triton kernel per parameter, turning the native multi-tensor launch into O(len(params)) launches. Please batch the tensor list in the device implementation and benchmark realistic heterogeneous parameter lists before registering it globally.

src/flag_gems/ops/_fused_sgd_.py
Caeruleann: .item() is gone, but the fp64 float-lr path still creates a CUDA tensor on every step. This is the per-call allocation/H2D initialization that the surrounding comment says breaks graph capture, and there is still no capture/replay regression test. Please make this path allocation-free and add an fp64 AMP capture/replay case.

src/flag_gems/ops/_fused_sgd_.py
```diff
@@ -0,0 +1,366 @@
...
+            if other.device != param.device:
+                raise RuntimeError(
+                    f"_fused_sgd_: {name}[{i}] device {other.device} does not "
+                    f"match param device {param.device}"
+                )
+        # The kernel indexes with a flat offset, so a non-dense tensor would be
+        # read in the wrong order.
+        for name, t in (("param", param), ("grad", grad), ("momentum_buffer", buf)):
+            if t is not None and not t.is_contiguous():
+                raise RuntimeError(f"_fused_sgd_: {name}[{i}] must be contiguous")
```
Caeruleann: The validation still only compares each grad/buffer with its corresponding param, so a param list spanning different dtypes or devices passes the pre-check and may fail after earlier tensors have already been updated. Please enforce the native whole-list restrictions before launching and add mismatched-device/dtype tests.

src/flag_gems/ops/_fused_sgd_.py
Yukun-Cui: Batched into a single launch in 508c5959. The kernel now takes int64 device arrays (param/grad/momentum addresses, per-tensor numel, and the per-tensor first-block offset) and runs one program per BLOCK_SIZE-wide tile of any tensor in the list; each program recovers its owning tensor with a branch-free binary search over block_offset, so a step over N parameters is 1 launch instead of N. Metadata is staged as one flat pinned host-to-device transfer and cached against the exact input tuple, so repeated optimizer steps and CUDA-graph replays hit the cache and transfer nothing. Block size is fixed at 512, swept across 256-4096 on both uniform 4Kx4K lists and heterogeneous mixed-size lists (2048+ lose occupancy on small tensors, 256 loses tail efficiency on large ones). Benchmark: added test_fused_sgd__heterogeneous over a realistic mixed-size parameter list (embedding table, attention projections, MLP weights, biases, a 1-D head and a (7,7,7) tail) at N=10/60/120 params. Harness numbers (operator mode, GPU idle): fp16 340->83us (N=10), 1500->161us (N=60), 3007->339us (N=120); fp32 276->50, 1523->217, 3008->461us. Previously the gems column scaled linearly with N (speedup 0.09 vs native) because of the launches; it now tracks the kernel time and is launch-flat. A launch-count test (test_fused_sgd__single_launch, parametrized N=1/4/16) asserts exactly 1 Triton launch per step and fails with 4/16 launches on the old code.

src/flag_gems/ops/_fused_sgd_.py
Yukun-Cui: Whole-list validation added before the launch in 508c5959, matching what ATen actually enforces. Verified against torch.ops.aten._fused_sgd_ first: a [fp32, fp16] list fails in check_fast_path_restrictions ({params, grads, momentum_buffer_list}) with every tensor untouched, and a cross-device list is rejected at the dispatcher (measured on 2 GPUs). The driver now checks that all params share one dtype and one device (in addition to the existing per-index grad/buffer checks) and raises before anything is launched. The old code's failure mode was exactly the one described: on a mixed-dtype list it silently updated both dtypes (no error at all), and on a mixed-device list it updated the cuda:0 tensors and then died with 'Pointer argument (at 0) cannot be accessed from Triton'. New tests: test_fused_sgd__rejects_mixed_dtype_list (both orders), test_fused_sgd__rejects_mixed_device_list, and test_fused_sgd__rejects_grad_dtype_mismatch; each asserts every tensor is byte-identical after the raise, so a partial update fails the test. All fail on the previous commit. Writing the device test also surfaced a second real bug: a list on a non-current device crashed in Triton (old code included), so the launch is now wrapped in torch_device_fn.device(params.device) and test_fused_sgd__launches_on_param_device covers it. Full file: 274 passed.

src/flag_gems/ops/_fused_sgd_.py
Yukun-Cui: Confirmed you were right about the code you reviewed, and both points are already fixed in the current head — this landed in `508c5959` (the same commit that batched the launch and added whole-list validation), pushed after this comment was written. No further change was needed, so none was made; here is the measurement rather than the assertion.

**1. The fp64 float-lr path is now allocation-free.** At the commit you reviewed (`f33654d1`) the path ran `torch.tensor(lr_base, dtype=param.dtype, device=param.device)` per call, so your reading was exactly right. `lr_scalar` is now annotated `tl.float64` (line 109) and the scalar is passed straight through (`lr_base = float(lr)`, line 369; `lr_base if not lr_is_tensor else 0.0`, line 432) — `lr_cache` and the per-call tensor are gone. Instrumented both commits with the same probe (caching-allocator counters from `torch.cuda.memory_stats()`, a monkeypatch over the tensor-construction/device-move APIs, and a CUDA profiler), 200 steps per case, fp64 float-lr, 3 params:

```
                      d(alloc events)  d(alloc bytes)  torch.tensor->cuda  HtoD memcpy
f33654d1 (reviewed)            200          102400                200       20/20 steps
508c5959 (current head)          0               0                  0       0/20 steps
```

The current head is also 0 for fp64 float-lr + AMP (`grad_scale`/`found_inf`) and for fp64 `tensor_lr` with a device lr tensor. The per-step kernel count dropped from 3 to 1 as well, which is the batching change from the same commit.

**2. The fp64 AMP capture/replay case exists and passes.** `tests/test_fused_sgd_.py`: `test_fused_sgd__cuda_graph_capture` (parametrized `[float32, float64]`), `test_fused_sgd__cuda_graph_capture_amp` (parametrized `[float32, float64]` — `dtype1` is the fp64 AMP case you asked for, and it carries `grad_scale`/`found_inf`), and `test_fused_sgd__cuda_graph_capture_tensor_lr`. All 5 parametrizations pass on the current head (file: 274 passed). Each captures one step with `torch.cuda.CUDAGraph` and asserts a replay equals an eager step.

Negative control, so this is not just "the tests are green": running the current test file against the `f33654d1` op, `test_fused_sgd__cuda_graph_capture[dtype1]` and `test_fused_sgd__cuda_graph_capture_amp[dtype1]` both fail with

```
RuntimeError: Cannot copy between CPU and CUDA tensors during CUDA graph capture
unless the CPU tensor is pinned.
```

i.e. the exact failure you predicted, on the exact two fp64 cases. They fail without the fix and pass with it.

**fp64 precision was not traded away for this.** The tensor existed because an unannotated Python float narrows to fp32 in Triton. The `tl.float64` annotation keeps the full double without it — measured, `lr=0.1`, fp64 param, `param = 1.0 - lr`:

```
scalar path (no tensor): 0.9                    == exact double
tensor_lr path:          0.9                    == exact double
fp32-narrowed 0.1:       0.8999999985098839     <- the error the annotation avoids
```

One scoped note, since it is adjacent and I did not change it: `_fused_sgd__tensor_lr` with a *CPU* lr tensor still allocates a fresh device tensor per call (200 allocs / 200 steps), because `lr_base.to(ref.device)` runs inside the driver. A scheduler holding lr as a device tensor (the intended use of that overload, and what the capture test uses) is allocation-free as measured above. Say the word if you want the CPU-input case cached too.

YangLong114514 (approved):

Caeruleann (approved):

PR #5883, opened 2026-09-01
fix(vector_norm): exclude masked padding lanes for negative finite orders

### LINKED ISSUE #4600: vector_norm returns 0 for negative order (p < 0)
## Describe the bug

`torch.norm` / `torch.linalg.vector_norm` with a negative finite order (e.g. `p=-1, -2, -3`) returns `0` instead of the correct value, on any shape where the reduction dimension is smaller than `BLOCK_N` (the common case). This is a **silent** correctness bug -- no error is raised, a plausible-looking `0` is returned.

Non-negative and infinite orders (`2, 1, 0, inf, -inf`) are unaffected.

## Steps to reproduce

```python
import torch
import flag_gems

x = torch.rand(4, 200, device="cuda", dtype=torch.float32) + 1.0  # positive, no zeros

ref = torch.norm(x, p=-2, dim=1)
with flag_gems.use_gems():
    got = torch.norm(x, p=-2, dim=1)

print("ref:", ref[0].item())   # e.g. 0.1006
print("got:", got[0].item())   # 0.0
```

## Actual behavior

`got` is `0.0` for all rows (relative error 100%). Same for `p=-1`, `p=-3`, etc.

## Expected behavior

Match `torch.linalg.vector_norm`, which returns the correct negative-order norm.

## Root cause

The finite-order kernels (`v_norm_kernel`, `l1_norm_kernel_1`) in `src/flag_gems/ops/vector_norm.py` load masked padding lanes (`cols >= N`) as `0.0`, then accumulate `pow(tl.abs(a), ord)`. For negative `ord`, `pow(0, ord)` is `+inf`, which corrupts the reduction; the final `pow(sum, 1/ord)` then collapses to `0`.

The bug is masked in CI because the `vector_norm` accuracy test `ORD_LIST` covers only `[2, inf, -inf, 0, 1]` -- no negative finite order.

## Environment

- FlagGems: master
- GPU: NVIDIA A100
- General (root cause is dtype/backend independent)

## Fix

I have a fix ready (guard the accumulation with `tl.where(mask, ...)` so padding lanes contribute `0` regardless of the sign of `ord`) plus a regression test (`ord=-2` added to `ORD_LIST`). Will open a PR.

bin913:
### PR Category
Operator

### Type of Change
- [x] Bug Fix
- [ ] New Feature
- [ ] Performance Optimization
- [ ] Document
- [ ] Code Refactoring
- [ ] Test
- [ ] CI/CD

### Description
torch.norm / torch.linalg.vector_norm with a negative finite order (e.g. p=-1, -2, -3) silently returned 0 whenever the reduction extent was smaller than BLOCK_N (the common case). `v_norm_kernel` and `l1_norm_kernel_1` load masked lanes as 0.0 and then accumulate `pow(|x|, ord)`; for negative ord, `pow(0, ord)` is +inf, so the padding lanes poisoned the sum and the final `pow(sum, 1/ord)` collapsed to 0. Non-negative and infinite orders were unaffected.

Guard both accumulations with `tl.where(mask, ..., 0.0)` so padding lanes contribute exactly 0 regardless of the sign of ord. Real zero elements keep their previous behaviour, which matches PyTorch.

The accuracy test `ORD_LIST` only covered [2, inf, -inf, 0, 1], which is why CI missed this; add the negative finite orders -1 and -2 to the list.

### Issue
Fixes #4600

### Progress
- [x] mask padding lanes in v_norm_kernel and l1_norm_kernel_1 accumulations
- [x] extend ORD_LIST with -1 and -2

### Test Plan
The new negative-order cases reproduce the 0-result on master and pass with the fix; `tests/test_vector_norm.py` 672 passed on H100.

### DIFF
--- a/src/flag_gems/ops/vector_norm.py
+++ b/src/flag_gems/ops/vector_norm.py
@@ -302,7 +302,9 @@ def v_norm_kernel(X, Out, M, N, ord, BLOCK_M: tl.constexpr, BLOCK_N: tl.constexp
         mask = row_mask & col_mask
 
         a = tl.load(X + cols, mask, other=0.0).to(acc_dtype)
-        _sum += pow(tl.abs(a), ord)
+        # Masked lanes load as 0.0; for negative orders pow(0, ord) is +inf
+        # and would corrupt the reduction (issue #4600), so exclude them.
+        _sum += tl.where(mask, pow(tl.abs(a), ord), 0.0)
     sum = tl.sum(_sum, axis=1)
     out = pow(sum, 1 / ord)[:, None]
     tl.store(Out, out, row_mask)
@@ -323,7 +325,9 @@ def l1_norm_kernel_1(X, Mid, ord, M, BLOCK_SIZE: tl.constexpr):
     else:
         acc_dtype = tl.float32
     x = tl.load(X, mask=mask, other=0.0).to(acc_dtype)
-    mid = tl.sum(pow(tl.abs(x), ord))
+    # Masked lanes load as 0.0; for negative orders pow(0, ord) is +inf
+    # and would corrupt the reduction (issue #4600), so exclude them.
+    mid = tl.sum(tl.where(mask, pow(tl.abs(x), ord), 0.0))
     tl.store(Mid, mid)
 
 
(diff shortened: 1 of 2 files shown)


huangyiqun (approved):
LGTM. The mask handling for negative-order reductions looks correct.

PR #5905, opened 2026-09-02
[KernelGen][Nvidia] Add _sobol_engine_initialize_state_ operator with Triton kernel

ShawnsYing:
## Description

Implements `aten::_sobol_engine_initialize_state_` operator for initializing Sobol quasi-random number generator state.

## Implementation

**Hybrid Triton/CPU approach:**
- **Dimension 0**: Triton kernel computes power-of-2 sequence `[2^29, 2^28, ..., 2^0]` in parallel
- **Dimensions 1-5**: Pre-computed Sobol direction numbers (lookup table) copied to GPU
- **Dimensions > 5**: Falls back to CPU computation (avoids native CUDA segfault issues)

**Key technical notes:**
- This is fundamentally a lookup-table algorithm, not compute-bound
- PyTorch's native CUDA implementation has segfault bugs for this operator
- Hybrid approach provides correctness while working around native limitations

## Testing

- ✅ All 8 tests pass (dimensions: 1, 2, 3, 5, 10, 20, 50, 100)
- ✅ Results match PyTorch CPU reference exactly
- ✅ Avoids CUDA segfault issues present in native implementation

## Files Added/Modified

- `src/flag_gems/ops/_sobol_engine_initialize_state_.py` - Triton kernel + hybrid logic
- `tests/test_sobol_engine_initialize_state_.py` - Test suite (8 test cases)
- `benchmark/test_sobol_engine_initialize_state_.py` - Benchmark suite
- `conf/operators.yaml` - Operator metadata (pointwise, alpha 5.4)
- Integration in `__init__.py` files

## Limitations

- GPU implementation limited to 5 dimensions (lookup table size constraint)
- Dimensions > 5 use CPU fallback (but safer than buggy native CUDA)
- No performance advantage expected (memory-bound operation)

## Checklist

- [x] Triton kernel implementation
- [x] Tests pass locally
- [x] Benchmark added
- [x] operators.yaml entry
- [x] Pre-commit hooks pass


bin913:

ShawnsYing:
Rebuilt on current master (`7e904b9`) — the branch had drifted and was conflicting, which is now resolved. Benchmark results below, as requested.

**First, a caveat that matters more than the numbers.**

The benchmark reports a ~1.00 speedup at every shape:

| dim | torch (ms) | gems (ms) | reported speedup |
|---|---|---|---|
| 100 | 0.0724 | 0.0722 | 1.002 |
| 500 | 0.2448 | 0.2438 | 1.004 |
| 1000 | 0.4662 | 0.4627 | 1.008 |
| 5000 | 2.1944 | 2.1937 | 1.000 |

That is by construction, not a result: the benchmark passes `flag_gems._sobol_engine_initialize_state_` as **both** `torch_op` and `gems_op`, because aten's version has no CUDA kernel (the file's comment says calling it on a device tensor aborts the process). So the ratio is 1 by definition and carries no performance information.

**The absolute numbers in that table are also not the operator's cost.** Almost all of it is the benchmark harness. I measured the operator directly:

| dim | per-call, measured directly | do_bench reports |
|---|---|---|
| 100 | **0.065 ms** | 72.9 ms |
| 5000 | **2.18 ms** | 2191 ms |

A ~1000x gap. The cause is `triton.testing.do_bench`: it allocates a **256 MB** buffer and calls `clear_cache()` (i.e. `cache.zero_()`) before each timed iteration to flush L2. That is correct for kernels whose input lives in L2, but this operator does its work on the **CPU** — the device side is a tiny bit of H2D/D2H — so the flush is pure measured overhead. Isolating it: `cache.zero_()` alone costs ~0.062 ms, and the full operator is ~0.065 ms, so the table is essentially reporting the cache-flush time, not the op.

I am not proposing to patch `do_bench` for one PR; I am flagging it because **the same effect applies to every small, launch-bound or CPU-side op in this benchmark suite**, and it is worth knowing when reading these tables.

**A second, more serious problem with this operator.** The Triton kernel only covers dimensions 0-5 (the direction-number table `_SOBOL_DIRECTIONS_1_TO_99` is truncated to 5 rows despite its name). For `dimension > 5` — which is every benchmark case and every realistic use — it falls back to:

```python
cpu_state = state.cpu()
torch.ops.aten._sobol_engine_initialize_state_(cpu_state, dimension)
state.copy_(cpu_state)
```

**So for all practical dimensions this op is a round trip to the CPU implementation**, not a GPU kernel. Measured against the CPU aten reference it is 1200-3000x slower:

| dim | gems (CUDA) | torch (CPU) | ratio |
|---|---|---|---|
| 100 | 73.8 ms | 0.025 ms | 2968x |
| 500 | 246.8 ms | 0.165 ms | 1493x |
| 1000 | 470.4 ms | 0.347 ms | 1356x |
| 5000 | 2185.5 ms | 1.774 ms | 1232x |

(Those gems figures are the inflated `do_bench` ones; even using the direct 0.065 ms / 2.18 ms the ratio is still ~2.6x-1.2x **slower** than just calling the CPU implementation, because of the copies.)

The outputs are **bit-exact** against the CPU aten reference at every shape — correctness is fine. But a GPU operator that round-trips to CPU and is slower than the CPU path is not a useful implementation, and I could not fix it in this rebuild: extending the direction table from 5 to the full set PyTorch uses (up to 21111 dimensions) is a data problem, not a kernel tweak, and the docstring of the op itself already states the limitation.

**Suggested handling:** land it as-is if the goal was the registration, with the caveat above written into the PR description; if a real GPU implementation is wanted, the direction-number table needs to be generated from the Joe-Kuo polynomial coefficients rather than hard-coded — happy to take that on as a separate PR if you want.

**Verified:** 8 passed in `tests/test_sobol_engine_initialize_state_.py`; bit-exact vs the CPU aten reference; `sort_exports`, `check_init_exports`, `check_operators_yaml`, `check_operator_markers`, `check_kernelgen_tests`, `check_performance_reference`, `check_aten_operators`, `check_api_logs`, black/isort/flake8 all pass.

bin913 (approved):

PR #5911, opened 2026-09-02
[KernelGen][Nvidia] Add _gather_sparse_backward operator with Triton kernel

Yukun-Cui:
> Replaces #5803 — original could not be reopened after the head branch was rebuilt to drop ~3,400 spurious mirror-history commits (now a clean single commit on ). Content is byte-identical to the original branch tip.

## Summary

Adds a Triton kernel for `_gather_sparse_backward`. Backward of `gather()` producing a sparse (COO) gradient w.r.t. `self`. Each entry of the flattened `grad`/`index` contributes one non-zero at the gather position `index[..., ]` along `dim` (the other coordinates are the unraveled index of the entry). The result is an uncoalesced sparse COO tensor of size `self.size()`.

## Testing

## Performance
Test command: `pytest benchmark/test_gather_sparse_backward.py --level core` (NVIDIA H20)

### gather_sparse_backward

| float16 | [1024, 1024] | 0.018048 | 0.010400 | 1.735 |
| float16 | [1024, 1024] | 0.022592 | 0.010464 | 2.159 |
| float16 | [4096, 4096] | 0.108672 | 0.055744 | 1.949 |
| float16 | [4096, 4096] | 0.146096 | 0.056672 | 2.578 |
| float16 | [8192, 8192] | 0.382848 | 0.202880 | 1.887 |
| float16 | [8192, 8192] | 0.538880 | 0.207360 | 2.599 |
| float16 | [256, 256, 256] | 0.178800 | 0.072640 | 2.461 |
| float16 | [256, 256, 256] | 0.212608 | 0.073248 | 2.903 |
| float16 | [256, 256, 256] | 0.214560 | 0.072640 | 2.954 |
| float32 | [1024, 1024] | 0.018368 | 0.014736 | 1.246 |
| float32 | [1024, 1024] | 0.021472 | 0.014240 | 1.508 |
| float32 | [4096, 4096] | 0.108384 | 0.055616 | 1.949 |
| float32 | [4096, 4096] | 0.146160 | 0.056480 | 2.588 |
| float32 | [8192, 8192] | 0.383232 | 0.202272 | 1.895 |
| float32 | [8192, 8192] | 0.537728 | 0.207008 | 2.598 |
| float32 | [256, 256, 256] | 0.177920 | 0.072672 | 2.448 |
| float32 | [256, 256, 256] | 0.211936 | 0.072768 | 2.912 |
| float32 | [256, 256, 256] | 0.215712 | 0.072480 | 2.976 |
| bfloat16 | [1024, 1024] | 0.018496 | 0.014752 | 1.254 |
| bfloat16 | [1024, 1024] | 0.021408 | 0.015200 | 1.408 |
| bfloat16 | [4096, 4096] | 0.108224 | 0.055632 | 1.945 |
| bfloat16 | [4096, 4096] | 0.147456 | 0.056608 | 2.605 |
| bfloat16 | [8192, 8192] | 0.382624 | 0.202624 | 1.888 |
| bfloat16 | [8192, 8192] | 0.536800 | 0.207152 | 2.591 |
| bfloat16 | [256, 256, 256] | 0.178432 | 0.072480 | 2.462 |
| bfloat16 | [256, 256, 256] | 0.212960 | 0.073152 | 2.911 |
| bfloat16 | [256, 256, 256] | 0.216288 | 0.072288 | 2.992 |

| gather_sparse_backward | **2.274** |

## Multi-backend Testing
|---|---|---|---|
| Nvidia (H20) | PASS (27 cases) | 2.274 | Primary |
| Tianshu | N/A | — | — |
| Muxi | N/A | — | — |
| Ascend | N/A | — | — |
| Hygon | N/A | — | — |

## Files Changed
- `src/flag_gems/ops/_gather_sparse_backward.py`: Triton kernel implementation
- `tests/test_gather_sparse_backward.py`: Accuracy test
- `benchmark/test_gather_sparse_backward.py`: Performance benchmark


conf/operators.yaml
```diff
@@ -5483,6 +5483,23 @@ ops:
...
+    '
+  for:
+  - _gather_sparse_backward
+  labels:
+  - aten
+  - KernelGen
+  kind:
+  - Tensor
+  stages:
+  - alpha: '2.2'
```
```suggestion
  - alpha: '5.4'
```

src/flag_gems/ops/_gather_sparse_backward.py
```diff
@@ -0,0 +1,195 @@
...
+
+def _gather_sparse_backward(self, dim, index, grad):
+    logger.debug("GEMS _GATHER_SPARSE_BACKWARD")
+
+    ndim = self.ndim
+    if ndim != index.ndim:
+        raise IndexError(
+            f"self and index must have the same number of dimensions, "
+            f"got self.ndim = {ndim} and index.ndim = {index.ndim}"
+        )
```
Caeruleann: Please handle the scalar self and scalar grad cases before enforcing equal ranks, matching ATen’s dedicated scalar paths, and add parity tests for both cases.

tests/test_gather_sparse_backward.py
```diff
@@ -0,0 +1,211 @@
...
+# the case matrix focused on shape edge cases (mirrors worktree CI branch).
+@pytest.mark.parametrize("dtype", [torch.float32])
+def test_gather_sparse_backward_edge(shape, dim, dtype):
+    """Edge cases: 1-D input and zero-sized index (nnz == 0)."""
+    ndim = len(shape)
+    if dim < 0:
+        dim += ndim
+    size_dim = shape[dim] if shape[dim] > 0 else 4
+    index_shape = list(shape)
+    index_shape[dim] = 1 if shape[dim] == 0 else max(1, shape[dim] // 2)
```
Caeruleann: This case creates an index of shape (4, 1, 3), so nnz is 12 rather than zero and the indices target an empty dimension; please replace it with a valid empty-index case and add the missing scalar parity cases.

src/flag_gems/ops/_gather_sparse_backward.py
```diff
@@ -0,0 +1,195 @@
...
+    if index_flat.dtype != torch.int64:
+        index_flat = index_flat.to(torch.int64)
+
+    sparse_indices = torch.empty((ndim, nnz), dtype=torch.int64, device=index.device)
+
+    dim_sizes = tuple(int(s) for s in index.shape)
+    with torch_device_fn.device(index.device):
+        _kernel_func()(index_flat, sparse_indices, nnz, dim, ndim, dim_sizes)
+
+    out = torch.sparse_coo_tensor(
```
Caeruleann, line 228: Torch APIs are not allowed in host code. Is there a way to implement this without using torch?

src/flag_gems/ops/_gather_sparse_backward.py
```diff
@@ -0,0 +1,195 @@
...
+    grad = grad.contiguous()
+    nnz = index.numel()
+
+    # The sparse values are simply the flattened gradient.
+    values = grad.reshape(-1)
+
+    # Build the sparse COO indices tensor of shape [ndim, nnz].
+    if nnz == 0:
+        sparse_indices = torch.empty((ndim, 0), dtype=torch.int64, device=index.device)
+        out = torch.sparse_coo_tensor(
```
Caeruleann, line 213: Torch APIs are not allowed in host code. Is there a way to implement this without using torch?

conf/operators.yaml
Yukun-Cui: Fixed! Updated to alpha 5.4.

src/flag_gems/ops/_gather_sparse_backward.py
Yukun-Cui: Good point! I'll add special handling for scalar self and scalar grad before the rank check, matching ATen's behavior, and add corresponding test cases.

tests/test_gather_sparse_backward.py
Yukun-Cui: You're right - I'll fix the empty index test to actually have zero non-zero elements, and add proper scalar test cases.

src/flag_gems/ops/_gather_sparse_backward.py
Yukun-Cui: I'll refactor to avoid PyTorch operations in host code by either:
1. Moving the computation into a Triton kernel
2. Using existing FlagGems operators
3. Implementing with pure Python/NumPy logic where appropriate

src/flag_gems/ops/_gather_sparse_backward.py
Yukun-Cui: I'll refactor to avoid PyTorch operations in host code by either:
1. Moving the computation into a Triton kernel
2. Using existing FlagGems operators
3. Implementing with pure Python/NumPy logic where appropriate

conf/operators.yaml
Yukun-Cui: Already `alpha: '5.4'` on the branch — the suggestion is applied.

src/flag_gems/ops/_gather_sparse_backward.py
Yukun-Cui: Fixed. Probing the native op confirmed the scalar paths and, more usefully, showed what they return:

```
0-dim self, 0-dim index -> nnz 1, sparse_dim 0, indices (0, 1)
1-dim self, 0-dim index -> nnz 1, sparse_dim 1, indices (1, 1)
0-dim self, 1-dim index -> nnz 1, sparse_dim 0, indices (0, 1)
```

The non-obvious part is that sparse dimensionality follows **`self`**, not `index` — a 0-dim self gives a `(0, 1)` index tensor with `sparse_dim() == 0` even when `index` is 1-dim. The equal-rank check is now skipped when either operand is 0-dim, and I made `dim` validation explicit for both paths (a 0-dim self only admits 0/-1).

`test_gather_sparse_backward_scalar` covers all three shapes, asserting shape, `sparse_dim()`, `_nnz()`, index shape and values against ATen.

tests/test_gather_sparse_backward.py
Yukun-Cui: Confirmed and fixed — `nnz` was 12, exactly as you said:

```
index numel=12  result nnz=12
```

The cause was in the test's shape derivation: `index_shape[dim] = 1 if shape[dim] == 0`, which turned `(4, 0, 3)` into a `(4, 1, 3)` index whose values addressed an empty dimension. That is an invalid case, not the `nnz == 0` case the test is named for. An empty gather dimension now forces an empty index, so all three zero-sized shapes genuinely give nnz 0:

```
(8, 0)    dim=0 -> index (4, 0)    nnz=0
(0, 8)    dim=1 -> index (0, 4)    nnz=0
(4, 0, 3) dim=1 -> index (4, 0, 3) nnz=0
```

Scalar parity cases added as well (see the thread above).

src/flag_gems/ops/_gather_sparse_backward.py
Yukun-Cui: This one I could not remove, and I would rather flag that than pretend otherwise.

Both comments point at `torch.sparse_coo_tensor`. The operator's return type *is* a sparse COO tensor, and there is no Triton or FlagGems equivalent for constructing one — no other op in `src/flag_gems/` builds a sparse tensor, so there is no in-repo precedent to follow either.

What I did check is that no host-side *computation* remains. The full list of host torch calls is now allocation, views, and the constructor:

- `torch.zeros` / `torch.empty` for the index buffer
- `.contiguous()`, `.reshape(-1)`, `.to(torch.int64)` on inputs
- `torch.sparse_coo_tensor` to assemble the result

The actual work — unravelling each flat index into per-dimension coordinates — is in the Triton kernel, which is what fills `sparse_indices`. So the convention is satisfied in substance; the remaining calls are the operator's output contract rather than computation moved to the host.

If there is an approved runtime wrapper for sparse construction that I missed, point me at it and I will switch. Otherwise this needs an exemption.

src/flag_gems/ops/_gather_sparse_backward.py
Yukun-Cui: Same answer as the line-192 thread: this is the other `torch.sparse_coo_tensor` call site (the nnz == 0 early return). The coordinate computation is in the Triton kernel; what remains on the host is allocation, views, and constructing the sparse return value, which has no FlagGems equivalent.

src/flag_gems/ops/_gather_sparse_backward.py
Caeruleann: got it

src/flag_gems/ops/_gather_sparse_backward.py
```diff
@@ -0,0 +1,231 @@
...
+    if is_scalar_case:
+        # One non-zero at the all-zero coordinate; the sparse dimensionality comes
+        # from ``self``, so a 0-dim self gives a (0, 1) index tensor.
+        values = grad.reshape(-1)[:1] if grad.numel() else grad.reshape(-1)
+        sparse_indices = torch.zeros(
+            (ndim, values.numel()), dtype=torch.int64, device=index.device
+        )
+        return torch.sparse_coo_tensor(
+            sparse_indices, values, self.size(), device=self.device
+        )
```
Caeruleann: Scalar handling still does not match ATen: scalar self must preserve all grad values, while scalar grad must preserve the actual index.view(1, 1). This branch currently truncates grad to one value and hardcodes every coordinate to zero. Please split the two native scalar paths and add nonzero scalar-index and multi-element-grad tests that compare _indices() as well.

Yukun-Cui: Confirmed and fixed in c61f070c. I pulled the actual ATen source (aten/src/ATen/native/TensorAdvancedIndexing.cpp) and the two scalar paths are indeed asymmetric early returns that the branch had collapsed into one:

```
if (self.ndimension() == 0)
  return _sparse_coo_tensor_unsafe({0, grad.numel()}, grad, self.sizes());
if (grad.ndimension() == 0)
  return _sparse_coo_tensor_unsafe(index.view({1, 1}), grad, self.sizes());
```

Both are now split and run before any rank check:

- **0-dim self** wins over everything, keeps *every* grad value (`indices` shape `(0, grad.numel())`, nnz == grad.numel()) and ignores `index`/`dim` entirely. Previously `grad.reshape(-1)[:1]` truncated to one value.
- **0-dim grad** on a non-scalar self keeps the real `index.view(1, 1)`, so a nonzero scalar index stays nonzero. Previously `torch.zeros` hardcoded the coordinate to 0.

Aligning these exposed three more places where the general path also disagreed with aten, so I fixed those too:

- `nnz` now comes from `grad.numel()`, not `index.numel()`, and a single-element `index` broadcasts to all non-zeros instead of erroring (added an `idx_stride` stride-0 load to the kernel).
- The flat coordinate is unravelled over **grad's** extents while only `self.ndim` rows are written, so a higher-rank grad (`self=(3,4)`, `grad=(2,3,4)`) is handled.
- `dim` is not validated: an out-of-range dim never matches a row, which is what aten's loop actually does.

Verification, all compared on size, `sparse_dim()`/`dense_dim()`, `_nnz()`, `_indices()` and `_values()` exactly:

```
28 hand-picked differential cases vs torch.ops.aten._gather_sparse_backward -> 28/28 SAME, 0 DIFF
shape sweep -> OK-match=29, both-err=33, mismatch=0
```

Tests added in tests/test_gather_sparse_backward.py: `test_gather_sparse_backward_scalar_paths` covers 0-dim self with multi-element grad, nonzero scalar index, and 0-dim grad preserving the nonzero `index.view(1, 1)`; the parametrization is 0-dim self with multi-element grad, nonzero scalar index and 0-dim grad. The pre-existing `test_gather_sparse_backward_scalar` now goes through `_assert_sparse_equal`, so it compares `_indices()` as well as values. I sanity-checked the new cases against the previous revision and 5 of them fail there, so they do pin the regression.

`115 passed`, pre-commit clean.

YangLong114514 (approved):

Caeruleann (dismissed):

huangyiqun (approved):

PR #5912, opened 2026-09-02
[KernelGen][Nvidia] Add _logcumsumexp operator with Triton kernel

Yukun-Cui:
> Replaces #5805 — original could not be reopened after the head branch was rebuilt to drop ~3,400 spurious mirror-history commits (now a clean single commit on ). Content is byte-identical to the original branch tip.

## Summary

Adds a Triton kernel for `_logcumsumexp`. Returns the log of the cumulative sum of exponentials of elements along a dimension, computed in a numerically stable way. Underscored variant of `torch.logcumsumexp`.

## Testing

## Performance
Test command: `pytest benchmark/test__logcumsumexp.py --level core` (NVIDIA H20)

### _logcumsumexp

| float16 | [64, 64] | 0.009280 | 0.005888 | 1.576 |
| float16 | [256, 256] | 0.021472 | 0.006144 | 3.495 |
| float16 | [1024, 1024] | 0.072288 | 0.009568 | 7.555 |
| float16 | [4096, 4096] | 0.375040 | 0.050816 | 7.380 |
| float16 | [1024, 65536] | 1.821408 | 1.078640 | 1.689 |
| float32 | [64, 64] | 0.009184 | 0.006304 | 1.457 |
| float32 | [256, 256] | 0.022016 | 0.006144 | 3.583 |
| float32 | [1024, 1024] | 0.072416 | 0.010112 | 7.161 |
| float32 | [4096, 4096] | 0.367072 | 0.056992 | 6.441 |
| float32 | [1024, 65536] | 1.762240 | 1.438080 | 1.225 |
| bfloat16 | [64, 64] | 0.009184 | 0.005824 | 1.577 |
| bfloat16 | [256, 256] | 0.021792 | 0.006048 | 3.603 |
| bfloat16 | [1024, 1024] | 0.071872 | 0.009440 | 7.614 |
| bfloat16 | [4096, 4096] | 0.375424 | 0.052160 | 7.198 |
| bfloat16 | [1024, 65536] | 1.830336 | 1.053184 | 1.738 |

| _logcumsumexp | **4.219** |

## Multi-backend Testing
|---|---|---|---|
| Nvidia (H20) | PASS (15 cases) | 4.219 | Primary |
| Tianshu | N/A | — | — |
| Muxi | N/A | — | — |
| Ascend | N/A | — | — |
| Hygon | N/A | — | — |

## Files Changed
- `src/flag_gems/ops/_logcumsumexp.py`: Triton kernel implementation
- `tests/test__logcumsumexp.py`: Accuracy test
- `benchmark/test__logcumsumexp.py`: Performance benchmark


src/flag_gems/ops/_logcumsumexp.py
```diff
@@ -0,0 +1,132 @@
...
+        tl.float32
+    )
+
+    # Compute logcumsumexp using cumsum with numerical stability
+    # logcumsumexp[i] = log(sum_{j=0}^{i} exp(x[j]))
+    # For stability: subtract max first
+    row_max = tl.max(x)
+    x_shifted = tl.exp(x - row_max)
+    cumsum_exp = tl.cumsum(x_shifted, axis=0)
+    result = row_max + tl.log(cumsum_exp)
```
Caeruleann: Subtracting the maximum of the whole row is not stable for prefixes before that maximum. For example, [0, 1000] produces -inf for the first output instead of 0, and all--inf or +inf inputs can produce NaNs. Please use an associative running logaddexp scan and add large-dynamic-range, NaN and ±Inf parity tests.

src/flag_gems/ops/_logcumsumexp.py
```diff
@@ -0,0 +1,132 @@
...
+    """Persistent kernel for _logcumsumexp - works for K==1 case (row scan)"""
+    pid_m = tle.program_id(0)
+    row_offset = pid_m * N
+
+    # Load the entire row
+    offsets = tl.arange(0, BLOCK_SIZE)
+    mask = offsets < N
+    x = tl.load(inp + row_offset + offsets, mask=mask, other=float("-inf")).to(
+        tl.float32
+    )
```
Caeruleann: The unconditional cast to tl.float32 loses FP64 precision and does not implement the native complex contract. Please preserve FP64 accumulation, support complex inputs, and add FP64/complex parity tests.

src/flag_gems/ops/_logcumsumexp.py
```diff
@@ -0,0 +1,132 @@
...
+def _logcumsumexp_wrapper(inp, dim=1, dtype=None, out=None):
+    assert dim >= -inp.ndim and dim < inp.ndim, "Invalid dim"
+    shape = inp.shape
+    dim = dim % inp.ndim
+    M = 1
+    N = shape[dim]
+    for i in range(dim):
+        M *= shape[i]
+    inp = inp.contiguous()
+    K = inp.numel() // M // N
```
Caeruleann: Please handle scalar and zero-sized inputs before computing K: scalar inputs currently fail the dim assertion, while empty dimensions can divide by zero or produce an invalid launch grid. Add parity tests for scalar and zero-sized tensors across dimension positions.

src/flag_gems/ops/_logcumsumexp.py
```diff
@@ -0,0 +1,132 @@
...
+                inp, out, M, N, TILE_SIZE, num_warps=num_warps
+            )
+        else:  # column scan (K > 1)
+            TILE_SIZE = triton.next_power_of_2(N)
+            # Use more warps for wider rows to keep occupancy up on long scans.
+            num_warps = 8 if TILE_SIZE > 2048 else 4
+            # Grid: (M, K, 1) to process each (m, k) pair independently
+            _logcumsumexp_kernel_3d[(M, K, 1)](
+                inp, out, M, N, K, TILE_SIZE, num_warps=num_warps
+            )
```
Caeruleann: Please implement the full out= contract: resize mismatched outputs, preserve non-contiguous output strides, validate dtype/device, return the same buffer, and add parity tests. The current linear stores can write out of bounds when out has the wrong size.

src/flag_gems/ops/_logcumsumexp.py
Yukun-Cui: Fixed in commit 7ede5b67. The whole-row-maximum shift is gone: the kernels now do `tl.associative_scan(x, 0, _logaddexp)` with a pairwise combine function that mirrors `_log_add_exp_helper` from `aten/src/ATen/native/cuda/LogcumsumexpKernel.cu` (I read the ATen source directly to get the exact NaN/Inf contract). Per your report, reproduced before the fix and verified after:

- `[0, 1000]`: was `[-inf, 1000]` -> now `[0, 1000]` (ATen identical)
- `[inf, inf, inf]`: was `[nan, nan, nan]` -> now `[inf, inf, inf]`
- `[-inf, -inf, -inf]`: was `[nan, nan, nan]` -> now `[-inf, -inf, -inf]`
- `[-inf, inf, 1.0]`: was `[nan, nan, nan]` -> now `[-inf, inf, inf]`

The `min == max && !isinf` branch is needed because `log1p(inf - inf)` would otherwise be NaN; NaN is made explicitly infectious because `fmin/fmax` ignore NaN operands. Exhaustive check over all 2- and 3-element combinations of {-inf, -1000, -5, -1, 0, 1, 5, 1000, +inf, nan} vs ATen: **0 semantic mismatches** (only fp32 rounding <= 1.9e-7). Added tests: `test__logcumsumexp_large_dynamic_range`, `test__logcumsumexp_mixed_infinities`, `test__logcumsumexp_nan_propagation`, `test__logcumsumexp_infinite_inputs`, `test__logcumsumexp_wide_dynamic_range_rows`. Full file: **113 passed**.

src/flag_gems/ops/_logcumsumexp.py
Yukun-Cui: Fixed in commit 7ede5b67. Two changes:

1. **FP64 preserved.** The accumulation dtype is now selected from the result dtype (`IS_FP64` constexpr selects `tl.float64`), so float64/complex128 no longer pass through fp32. Measured on `(4, 1024)` fp64: max abs error vs ATen was **9.33e-07** (fails RESOLUTION[float64]=1e-7) and is now **5.3e-15**. `test__logcumsumexp_fp64_precision` additionally asserts `max_err < 1e-9` so a future fp32 leak fails loudly.

2. **Complex support added.** `complex64`/`complex128` previously raised `KeyError: 'complex64'`. There is now a dedicated kernel over the real views implementing ATen's complex `_log_add_exp_helper`: real-part ordering of the (min, max) pair, the infectious-NaN collapse, the all-`-inf` branch, and the all-`+inf` branch via `_fast_build_exp_inf`; `c10::log1p` is reproduced with `hypot`/`atan2`/`log1p`. complex32 stays unsupported because ATen itself raises `"logcumsumexp_cuda" not implemented for 'ComplexHalf'`.

Real dtype is also rejected explicitly when the input is complex, rather than mis-addressing the buffer. Added `test__logcumsumexp_complex` and `test__logcumsumexp_complex_special_values` (exact NaN/Inf pattern, not just tolerance).

src/flag_gems/ops/_logcumsumexp.py
Yukun-Cui: Fixed in commit 7ede5b67. Both paths now run before the launch geometry is computed:

- **Scalars:** `_normalize_dim` wraps `dim` like `maybe_wrap_dim`, treating a 0-dim tensor as 1-dim. `torch.tensor(3.0)` with `dim=0`/`dim=-1` now returns `tensor(3.)` instead of `AssertionError: Invalid dim`; `dim=1`/`dim=-2` raise `IndexError: Dimension out of range (expected to be in range of [-1, 0], but got 1)` matching ATen.
- **Zero-sized:** an early `if inp.numel() == 0: return out` (placed after the `out=` resize/validation so the contract still holds) removes the `ZeroDivisionError: integer division or modulo by zero` from `K = inp.numel() // M // N`, and no kernel is launched for an empty grid.

Verified against ATen for `(0,)`, `(2,0)`, `(0,3)`, `(3,0)`, `(2,0,4)`, `(2,4,0)` at every dimension position: all match in shape and values. Tests added: `test__logcumsumexp_scalar`, `test__logcumsumexp_scalar_invalid_dim`, `test__logcumsumexp_invalid_dim`, `test__logcumsumexp_empty`, `test__logcumsumexp_out_empty`.

src/flag_gems/ops/_logcumsumexp.py
Yukun-Cui: Fixed in commit 7ede5b67. The full `out=` contract is implemented, and the out-of-bounds write is gone:

- **resize:** a mismatched buffer is `resize_`d to the input shape (`torch.empty(17)` -> `(3, 4)`), so the linear stores can no longer run past the end. The resize happens before the empty-tensor early return, so `out=torch.empty(0)` still gets resized.
- **non-contiguous strides preserved:** a non-contiguous `out` is computed into a contiguous scratch buffer and copied back, so `out.t()` keeps strides `(1, 3)` and its buffer identity rather than being silently re-laid-out. Values are byte-identical to ATen's `out=` result for the same case.
- **dtype/device validated** with `RuntimeError` before any store; **the same buffer is returned** (`res is out` asserted); an invalid `dim` raises `IndexError` before the buffer is touched (test asserts a pre-filled buffer stays untouched).

Tests added: `test__logcumsumexp_out_resizes`, `test__logcumsumexp_out_noncontiguous`, `test__logcumsumexp_out_rejects_mismatched_dtype`, `test__logcumsumexp_out_rejects_mismatched_device`, `test__logcumsumexp_out_invalid_dim_does_not_touch_out`.

One note: I also checked the benchmark harness and this rewrite is intentionally slower than the previous kernel on wide rows (e.g. `(200, 40999, 3)` ~2.4x), because the correct pairwise scan replaces a `cumsum` that was cheap partly because it was wrong. On that shape it is still ~2x faster than ATen, and there is no blocking performance gate in CI. Happy to follow up with a two-pass decoupled-lookback scan if the maintainers want the wide-row gap closed.

src/flag_gems/ops/_logcumsumexp.py
```diff
@@ -0,0 +1,332 @@
...
+        target = out
+        scratch = None
+    else:
+        target = torch.empty(inp.shape, dtype=result_dtype, device=inp.device)
+        scratch = out
+
+    src_view = _real_view(inp)
+    dst_view = _real_view(target)
+
+    BLOCK_SIZE = triton.next_power_of_2(N)
```
Caeruleann: Please avoid scanning the entire reduction dimension in a single program with BLOCK_SIZE = next_power_of_2(N). Large valid inputs make tl.associative_scan extremely expensive to compile as the block grows. Please use a chunked/hierarchical scan for large N, or at minimum enforce a documented size limit instead of allowing compilation to stall.

Yukun-Cui: 已修（5b417864）。先把代价量出来，确认这不是理论问题——冷 cache 下编译 + 运行：

```
N=4096    1 s
N=8192    4 s
N=32768   24 s
N=131072  5 min
```

即 1-D 130k 元素的合法输入会让编译卡数分钟。

改法是分块 + 两遍 fan-in，而不是单纯设一个 size limit（那样合法输入仍会被拒绝）：

- `_logcumsumexp_chunk_kernel`：每 (row, chunk) 一个 program，扫固定宽度 `CHUNK`（4096）的 tile 并记录该 chunk 的 running total。**编译的 tile 宽度与 N 无关**——行变长只是 program 变多，block 不变。
- `_logcumsumexp_fanin_kernel`：把前面的 chunk total 归约成每个 chunk 的 carry-in，用一次 `logaddexp` 结合应用。由于 `logaddexp` 可结合，这与整行扫描**结果相同**（不是近似）。

现在编译时间对所有 N 基本恒定：N=524288 约 1 s（原来是分钟级）。

一个在实现里踩到、值得单独提的坑：chunk total 最初用 `tl.max` 取，但 `tl.max` 忽略 NaN，会把 NaN 状态（以及全 ±inf 分支）在 chunk 边界上丢掉——实测 `[-5, nan] × 3000` 在第 4096 个元素处由 nan 变成有限值。现在用掩码求和取 inclusive scan 的最后一个 lane，NaN 不再被吞。对 `{-inf,-1000,-5,0,1,1000,inf,nan}` 的全部 81×2 组合验证：0 mismatch（162/162）。

新增 `test__logcumsumexp_long_rows`（四种长度，含 tail chunk 与 K>1）、`test__logcumsumexp_long_rows_special_values`（NaN/±Inf 必须跨 chunk 存活）、`test__logcumsumexp_long_row_fp64`。`tests/test__logcumsumexp.py`: 124 passed（`--ref=cpu --quick` 76 passed）。

src/flag_gems/ops/_logcumsumexp.py
```diff
@@ -0,0 +1,501 @@
...
+    src_view = _real_view(inp)
+    dst_view = _real_view(target)
+
+    # A single program would have to scan the whole row, and
+    # ``tl.associative_scan`` compile time grows steeply with the block size
+    # (measured: ~4 s at N=8192, ~24 s at 32768, ~5 min at 131072), so long rows
+    # are split into fixed-size chunks instead -- the row length then only
+    # selects the grid, not the compiled tile.
+    if N > _MAX_SINGLE_BLOCK:
+        _chunked_scan(src_view, dst_view, N, K, M, acc_is_fp64, inp)
```
Caeruleann, line 460: Complex rows longer than _MAX_SINGLE_BLOCK take the chunked path, but _chunked_scan uses the real _logaddexp kernels on the view_as_real buffer, silently scanning the interleaved re/im parts as independent real numbers (complex64 N=8192 maxdiff ~10.5 vs ATen; complex128 N=5000 ~16.7). Please either add chunked complex kernels (fan-in combined with _complex_logaddexp) or explicitly reject long complex rows before this branch, plus a regression test.

Caeruleann (dismissed):

src/flag_gems/ops/_logcumsumexp.py
Yukun-Cui: 已修（commit 6b6b4023）。你说得对，而且我先把量级复现出来确认了（改动前的 PR head `7eb71b13`，种子 0）：

```
complex64  N=8192  maxdiff 9.65     (与你报的 ~10.5 量级一致)
complex128 N=5000  maxdiff 9.09     (随数据变化，种子 3 上是 298.6)
```

## 根因

`_chunked_scan` 对复数输入走的是**实**版 chunk kernel。复数输入的 `dst_view` 是 `(..., 2)` 的 interleaved real view，实版 kernel 按线性偏移扫描，于是把交错的 re/im 当成两串独立的实数在扫——完全不是复数语义。

选择「加 chunked 复数 kernel」而不是「拒绝长复数行」，理由是：`_MAX_SINGLE_BLOCK` 之上本来就该支持，而拒绝会让「同一 dtype 只在长行上报错」这种不一致的契约更难解释。

## 改法

新增两个 kernel，与实版一一对应：

- `_logcumsumexp_complex_chunk_kernel`：每个 (row, chunk) 扫一个 `CHUNK` 宽的 tile，用 `_complex_logaddexp` 做 `tl.associative_scan((xr, xi), 0, ...)`；chunk total 同样用**掩码求和取 inclusive scan 的最后一个 lane**分别取实部/虚部（不能用 `tl.max`，会吞掉 NaN 状态与 ±inf 分支）。
- `_logcumsumexp_complex_fanin_kernel`：把前面各 chunk 的 total 用 `_complex_logaddexp` 逐个结合成 carry-in，再对 chunk 施加一次结合。

padding 用 `(-inf, 0)`：按实部排序它是末位，generic 分支给出 `hi + log1p(exp(lo-hi)) == hi`（精确），±inf 分支原样返回 `lo`，所以两个方向都是单位元，padding 对 scan 不可见。

## 验证

```
complex64  N=4097   1.9e-06
complex64  N=5000   2.9e-06
complex64  N=8192   3.8e-06
complex64  N=16384  5.7e-06
complex128 N=5000   5.3e-15
complex128 N=8192   7.1e-15
```

K>1、M>1、tail chunk 也都过（`(2,5000,3)` dim=0/1/2、`(3,4097,2)`、`(1,4097,5)`）。`out=`、非连续 out、real→complex（`dtype=torch.complex64`）的 `REAL_INPUT` 分支同样验证。

## 两点需要说明

**1. 顺带发现一个先前就存在的 2π 分支差异，与本改动无关。** 长复数行上偶发 `maxdiff = 6.283185`（正好 2π）。定位后发现它在**我没碰过的单块复数路径**上同样复现，且只出现在虚部：`atan2` 的返回值在 (-π, π]，真值落在分支切割上时，ATen 与 gem 各自的舍入会选到切割的两侧，于是差一个 2π。用 `7eb71b13`（改动前的 PR head）在 capture 下来的同一组输入上验证，`maxdiff=6.283185`，**逐字复现**，确认是既有行为而非本次引入。

这是复数 `log` 的固有契约歧义（ATen 自己的结果也依赖它的 scan 顺序）。所以新增测试里对复数只比较**卷绕后的差**（`(d+π) % 2π - π`），实部与 NaN/±Inf 的结构仍精确断言——这样检查的是「值」而不是「舍入选了切割的哪一侧」。我没有在本 PR 里顺手改这个既有问题（会扩大范围且需要单独讨论 ATen 契约），如果你希望我另开一个 PR 处理，我可以做。

**2. 编译时间：复数 combine 的编译开销随 tile 宽度陡增，所以复数分块路径用了更窄的 tile。**

你上一轮提过编译时间，所以这次专门量了（冷 Triton cache，只计时 gems 调用）。先说结论：复数的编译开销来自 `_complex_logaddexp` 本身（exp/log1p/atan2/hypot 的复数展开）乘上 tile 宽度，**不是**分块带来的；而且 tile 越窄编译越快、runtime 也略快，所以 `_chunked_scan` 里复数路径选了 1024 宽的 chunk（实数路径保持 4096，因为那里编译本来就只要 ~1s）：

```
real  single-block tile=4096                2.05s
real  chunked      tile=4096                1.57s
complex single-block tile=4096              98.64s   <- 改动前的既有路径（同一 combine，同一宽度）
complex chunked      tile=4096             100.91s   <- 新增路径若用同宽度
complex chunked      tile=1024 (本 PR 采用)  19.94s   cold, 0.0003s warm @N=16384
```

复数路径的编译量级从「单块时代随 N 指数增长」变成**一次性 ~20s、之后所有行长度都命中缓存**：

```
N=8192   0.75s  (首次编译)
N=8192   0.001s
N=65536  0.037s
N=524288 0.001s
```

即与 N 无关 —— 正是你上一轮要求的方向。窄 tile 的代价是 fan-in 串行 carry 循环的 chunk 数变多（O(num_chunks²)/row），实测 N=131072（128 chunks）warm runtime 仍在 1e-4 s 量级，未成为瓶颈。若你认为剩下的这 ~20s 仍不可接受，那需要动的是 `_complex_logaddexp` 的展开方式（例如牺牲部分 ±inf 分支精度换更简单的表达式），属于另一个取舍，我不想在这一轮擅自改。

新增测试：`test__logcumsumexp_long_complex_rows`（4 种长度 × complex64/128，含 tail chunk 与 K>1）、`test__logcumsumexp_long_complex_rows_special_values`（NaN/±Inf 必须跨 chunk 边界存活，且分支结构精确一致）。
`tests/test__logcumsumexp.py`: 136 passed（`--ref=cpu --quick` 88 passed）。

Caeruleann (approved):

PR #5914, opened 2026-09-02
[KernelGen][Nvidia] Add _histogramdd_bin_edges operator with Triton kernel

Yukun-Cui:
> Replaces #5807 — original could not be reopened after the head branch was rebuilt to drop ~3,400 spurious mirror-history commits (now a clean single commit on ). Content is byte-identical to the original branch tip.

## Summary

Adds a Triton kernel for `_histogramdd_bin_edges`. Computes the equal-width bin edges for each dimension of a multi-dimensional histogram, given the number of bins (and an optional explicit range). The edges are derived from the per-dimension min/max of the input when no range is supplied, matching ``torch._histogramdd_bin_edges``.

## Testing

## Performance
Test command: `pytest benchmark/test_histogramdd_bin_edges.py --level core` (NVIDIA H20)

### histogramdd_bin_edges

| float32 | [1024, 3] | 0.103392 | 0.141280 | 0.732 |
| float32 | [4096, 3] | 0.145728 | 0.141264 | 1.032 |
| float32 | [16384, 3] | 0.212528 | 0.161600 | 1.315 |
| float32 | [65536, 3] | 0.641952 | 0.237792 | 2.700 |
| float32 | [262144, 3] | 2.505536 | 0.530544 | 4.723 |
| float32 | [1024, 3] | 0.247808 | 0.155104 | 1.598 |
| float32 | [4096, 3] | 0.194096 | 0.160960 | 1.206 |
| float32 | [16384, 3] | 0.300896 | 0.178560 | 1.685 |
| float32 | [65536, 3] | 0.707392 | 0.253696 | 2.788 |
| float32 | [262144, 3] | 2.239440 | 0.551424 | 4.061 |

| histogramdd_bin_edges | **2.184** |

## Multi-backend Testing
|---|---|---|---|
| Nvidia (H20) | PASS (10 cases) | 2.184 | Primary |
| Tianshu | N/A | — | — |
| Muxi | N/A | — | — |
| Ascend | N/A | — | — |
| Hygon | N/A | — | — |

## Files Changed
- `src/flag_gems/ops/_histogramdd_bin_edges.py`: Triton kernel implementation
- `tests/test_histogramdd_bin_edges.py`: Accuracy test
- `benchmark/test_histogramdd_bin_edges.py`: Performance benchmark


src/flag_gems/ops/_histogramdd_bin_edges.py
```diff
@@ -0,0 +1,312 @@
...
+        if num_edges < 0:
+            num_edges = 0
+
+        if range_ is not None:
+            left, right = ranges[d]
+            if left > right:
+                raise RuntimeError(
+                    f"torch.histogramdd: min should not exceed max, but got "
+                    f"min {left} max {right} for dimension {d}"
+                )
```
Caeruleann, line 367: Please validate that both endpoints of an explicit range are finite before generating edges; ATen raises for NaN/Inf ranges, while the current implementation returns non-finite edges.

src/flag_gems/ops/_histogramdd_bin_edges.py
```diff
@@ -0,0 +1,312 @@
...
+
+def _histogramdd_bin_edges_impl(
+    self, bins, *, range=None, weight=None, density=False, out=None
+):
+    if self.dim() < 2:
+        raise RuntimeError(
+            "torch.histogramdd: input tensor should have at least 2 dimensions"
+        )
+
+    N = self.shape[-1]
```
Caeruleann, line 279: Please handle self.shape[-1] == 0 before reshape(-1, N) and return an empty edge list, matching ATen for inputs such as (3, 0) with bins=[].

src/flag_gems/ops/_histogramdd_bin_edges.py
```diff
@@ -0,0 +1,312 @@
...
+                        f"torch.histogramdd: dimension {d}'s range "
+                        f"[{c_min}, {c_max}] is not finite"
+                    )
+
+    edges = []
+    for d in range(N):
+        nb = int(bins[d])
+        num_edges = nb + 1
+        if num_edges < 0:
+            num_edges = 0
```
Caeruleann: Please preserve ATen’s error behavior for bins < -1 instead of clamping negative edge counts to zero; add parity tests for -3, -2, -1, and 0.

src/flag_gems/ops/_histogramdd_bin_edges.py
```diff
@@ -0,0 +1,312 @@
...
+            # and raises. The kernel flags such columns so we can match the
+            # error message without a second pass over the input.
+            nonfinite_list = nonfinite.tolist()
+            for d in range(N):
+                if nonfinite_list[d]:
+                    col = inp2d[:, d]
+                    # Propagate NaN/inf like torch's min/max reduction does so
+                    # the reported range matches PyTorch's wording exactly.
+                    c_min = float(col.min().item())
+                    c_max = float(col.max().item())
```
Caeruleann: Please avoid materializing min/max values on the host with .tolist()/.item() and calling col.min()/col.max() in the wrapper; this introduces device synchronization, breaks graph capture, and violates the host-side no-PyTorch-compute rule.

src/flag_gems/ops/_histogramdd_bin_edges.py
```diff
@@ -0,0 +1,312 @@
...
+                )
+                continue
+            left = left - 0.5
+            right = right + 0.5
+
+        edges.append(_linspace(left, right, num_edges, self.dtype, self.device))
+
+    edges_tuple = tuple(edges)
+
+    if out is not None:
```
Caeruleann, line 398: Please add full out= parity tests for list length, dtype/device mismatch, non-contiguous outputs, resizing and object reuse; the current copy_ path can silently cast outputs that ATen rejects.

benchmark/test_histogramdd_bin_edges.py
```diff
@@ -0,0 +1,111 @@
...
+    yield inp, {"bins": list(HIST_BINS)}
+
+
+def _torch_op(inp, *, bins, range=None, weight=None, density=False):
+    # Fall back to the CPU implementation since aten has no CUDA kernel.
+    ref_inp = inp.to("cpu")
+    out = torch._histogramdd_bin_edges(
+        ref_inp, bins, range=range, weight=weight, density=density
+    )
+    return [t.to(inp.device) for t in out]
```
Caeruleann: The Torch timing includes GPU↔CPU transfers while the Gems timing only measures the device implementation, so the reported speedup is not comparable; please avoid reporting this as operator parity and call the public FlagGems API rather than importing the internal implementation.

src/flag_gems/ops/_histogramdd_bin_edges.py
Yukun-Cui: I'll add validation to check that both range endpoints are finite and raise an appropriate error for NaN/Inf values, matching ATen's behavior.

src/flag_gems/ops/_histogramdd_bin_edges.py
Yukun-Cui: Good catch! I'll add an early check for empty last dimension and return an empty edge list to match ATen.

src/flag_gems/ops/_histogramdd_bin_edges.py
Yukun-Cui: I'll remove the clamping and let ATen's error handling work, then add test cases for negative bin values (-3, -2, -1, 0).

src/flag_gems/ops/_histogramdd_bin_edges.py
Yukun-Cui: I'll refactor to compute min/max in a Triton kernel or using existing FlagGems operators to avoid host synchronization.

src/flag_gems/ops/_histogramdd_bin_edges.py
Yukun-Cui: I'll add comprehensive out= parameter tests covering all these edge cases: list length, dtype/device mismatch, non-contiguous outputs, resizing, and object reuse.

benchmark/test_histogramdd_bin_edges.py
Yukun-Cui: I'll fix the benchmark to use the public API and ensure both paths include the same overhead for fair comparison.

src/flag_gems/ops/_histogramdd_bin_edges.py
Yukun-Cui: Done in 4612b8f5e. `_histogramdd_bin_edges_impl` now validates both endpoints right after parsing `range_`, before any edges are built:

```python
for d in range(N):
    lo, hi = ranges[d]
    if not (math.isfinite(lo) and math.isfinite(hi)):
        raise RuntimeError(
            f"torch.histogramdd: dimension {d}'s range "
            f"[{lo:g}, {hi:g}] is not finite"
        )
```

The `:g` formatting is there to match ATen's wording exactly. Checked side by side on H20:

| input | ATen (CPU) | gems |
|---|---|---|
| `range=[nan, 1.0]` | `dimension 0's range [nan, 1] is not finite` | identical |
| `range=[inf, 1.0]` | `dimension 0's range [inf, 1] is not finite` | identical |

`test_histogramdd_bin_edges_nonfinite_explicit_range` covers nan/inf/-inf on each endpoint (6 cases) and asserts both gems and `torch.ops.aten._histogramdd_bin_edges` raise.

src/flag_gems/ops/_histogramdd_bin_edges.py
Yukun-Cui: Done in 4612b8f5e. The check sits before the reshape, since `reshape(-1, 0)` cannot infer a leading extent:

```python
# An empty innermost dimension means there are no histogram dimensions at all,
# so there are no edges to produce. Handled before the reshape below because
# reshape(-1, 0) cannot infer the leading extent.
if N == 0:
    return () if out is None else out

inp2d = self.reshape(-1, N).contiguous()
```

Verified against ATen for `(3, 0)`, `(0, 0)` and `(2, 5, 0)` with `bins=[]` — both return an empty edge list. `test_histogramdd_bin_edges_empty_innermost` parametrizes those three shapes.

src/flag_gems/ops/_histogramdd_bin_edges.py
Yukun-Cui: Done in 4612b8f5e — the clamp is gone:

```python
# ATen builds the edges with linspace(steps=bins+1), so bins == -1 (zero
# steps) and bins == 0 (one step) are both legal and only bins <= -2 is
# rejected -- by linspace itself, with this message. Clamping the negative
# count to zero instead would silently accept those inputs.
if num_edges < 0:
    raise RuntimeError("number of steps must be non-negative")
```

Probed ATen to pin down where the boundary actually is, since `bins=-1` and `bins=0` turn out to be legal (they are `linspace(steps=0)` and `linspace(steps=1)`):

| bins | ATen (CPU) | gems |
|---|---|---|
| -3 | `number of steps must be non-negative` | same message |
| -2 | `number of steps must be non-negative` | same message |
| -1 | 0 edges | 0 edges |
| 0 | 1 edge | 1 edge |
| 1 | 2 edges | 2 edges |

The old clamp turned `-3`/`-2` into a silently-accepted empty result. `test_histogramdd_bin_edges_negative_bins` parametrizes -3, -2, -1, 0, 1 and derives the expectation from ATen rather than hardcoding it, so it asserts the raise for the first two and edge-count plus value parity for the rest.

One related case worth noting: `bins=0` on a degenerate column returns the widened low endpoint, not the value itself — constant input `2.5` gives `[2.0]` in both ATen and gems, and `bins=4` gives `[2.0, 2.25, 2.5, 2.75, 3.0]`. The single-edge path is guarded separately in the device linspace kernel so index 0 stays at the low endpoint instead of being pinned to `hi`.

src/flag_gems/ops/_histogramdd_bin_edges.py
Yukun-Cui: Done in 4612b8f5e. `col.min()`/`col.max()` and both `.tolist()` calls on the min/max values are gone; the endpoints now stay on the device end to end.

Three pieces:

1. **New `histogramdd_linspace_from_tensor_kernel`** reads the endpoints from column `d` of the reduction outputs instead of taking host scalars, and does the degenerate-range widening (`min == max` → unit interval) on device:

   ```python
   lo = tl.load(min_ptr + d)
   hi = tl.load(max_ptr + d)
   degenerate = lo == hi
   lo = tl.where(degenerate, lo - 0.5, lo)
   hi = tl.where(degenerate, hi + 0.5, hi)
   step = (hi - lo) * inv_denom
   ```

2. **NaN propagation moved into `histogramdd_col_minmax_kernel`.** The reduction previously ignored NaN, which is why the wrapper had to re-read the column with `col.min()`/`col.max()` to reproduce torch's wording. The kernel now tracks `_has_nan` per column and writes nan into min/max, so the error message is built from the reduction outputs alone — no second pass over the input.

3. **The one remaining host read is the non-finite flag**, not the values. Deciding whether to raise is inherently a host-side branch, so it needs the flag; the min/max values themselves are consumed by the kernel. That read is skipped under `torch.cuda.is_current_stream_capturing()`, since an exception cannot be raised from a replay anyway.

Verified capture works: warm up once, then `torch.cuda.graph(g)` around `flag_gems._histogramdd_bin_edges(inp, [5, 3, 4])` on a `(1024, 3)` input captures and replays, and the replayed output matches ATen.

Error parity is unchanged in eager mode, including the NaN-propagation cases that previously needed the host reduction:

| input | ATen (CPU) | gems |
|---|---|---|
| NaN in column 1 | `dimension 1's range [nan, nan] is not finite` | identical |
| `inf` in column 0 | `dimension 0's range [-1.84919, inf] is not finite` | identical |

On accuracy, moving linspace on-device does shift interior edges relative to ATen's CPU result, but no more than the host-scalar path did. Worst absolute error over 90 random cases is 9.5e-07, i.e. 1.17e-07 relative to the range span — one fp32 eps. First and last edge are bit-exact in all 240 checks (the last edge is explicitly pinned to `hi`). For reference, `torch.linspace` on CUDA vs CPU with identical endpoints diverges on the same order, so this is fp32 linspace ordering, not something the kernel introduces.

src/flag_gems/ops/_histogramdd_bin_edges.py
Yukun-Cui: Done in 4612b8f5e. You were right that `copy_` alone was silently casting — the wrapper now validates the out list before writing anything:

```python
if len(out) != N:
    raise RuntimeError(
        f"_histogramdd_bin_edges: expected out to have {N} tensors, got {len(out)}"
    )
for d in range(N):
    if out[d].dtype != self.dtype:
        raise RuntimeError(...)
    if out[d].device != self.device:
        raise RuntimeError(...)
```

Sizing is also derived from the actual edge tensors now (`edges_tuple[d].numel()`) rather than recomputing `bins[d] + 1`, which matters for the `bins <= 0` cases above.

`test_histogramdd_bin_edges_out_contract` covers the five cases you listed:

| case | ATen | gems |
|---|---|---|
| fp64 out, fp32 input | `Expected out tensor to have dtype float, but got double` | raises, message matches on "dtype" |
| out list length 1 or 3 for 2 dims | raises | raises |
| wrong-size out (99, 1) | resized in place | resized to [4, 4], same object ids |
| non-contiguous out of right size | accepted | accepted, values match |
| object reuse | same objects | `id()` preserved across the call |

One divergence in message text, not behavior: for a wrong-length out list ATen fails with an `INTERNAL ASSERT FAILED at CompositeViewCopyKernels.cpp` (i.e. it does not have a proper check there), so the test asserts `RuntimeError` rather than matching wording, and gems raises the explicit message above instead.

Full file is 40 passed on H20.

benchmark/test_histogramdd_bin_edges.py
Yukun-Cui: Done in 4612b8f5e, and you were right that the number in the PR description was not a speedup. Two changes:

**Public API.** The internal import is gone; the benchmark calls `flag_gems._histogramdd_bin_edges` / `flag_gems._histogramdd_bin_edges_out`.

**`speedup` is no longer reported at all.** aten has no CUDA kernel for this operator, so `latency_base` is a CPU measurement and `latency` a CUDA one — the ratio is not operator parity, so publishing it was wrong regardless of how the transfers were handled:

```python
DEFAULT_METRICS = ["latency_base", "latency"]
```

The transfers are out of the timed region too (inputs and out-tensors are staged as CPU copies at input-generation time, so the baseline closure runs the native kernel only, and the D2H copy-back the old `_torch_op` did is gone). But that fix alone would have produced a *more* flattering ratio, which is why dropping the metric is the actual answer here.

PR #5917, opened 2026-09-02
[KernelGen][Nvidia] Add _fused_adagrad_ operator with Triton kernel

Yukun-Cui:
> Replaces #5811 — original could not be reopened after the head branch was rebuilt to drop ~3,400 spurious mirror-history commits (now a clean single commit on ). Content is byte-identical to the original branch tip.

## Summary

Adds a Triton kernel for `_fused_adagrad_`. The in-place version of `fused_adagrad`.

## Testing

## Performance
Test command: `pytest benchmark/test_fused_adagrad_.py --level core` (NVIDIA H20)

### fused_adagrad_

| float32 | [256, 256] | 0.210784 | 0.007808 | 26.996 |
| float32 | [512, 512] | 0.210720 | 0.008128 | 25.925 |
| float32 | [1024, 256] | 0.210688 | 0.008160 | 25.820 |
| float32 | [2048, 512] | 0.211456 | 0.012064 | 17.528 |
| float32 | [4096, 256] | 0.211376 | 0.012064 | 17.521 |
| float32 | [65536] | 0.210784 | 0.007872 | 26.776 |

| fused_adagrad_ (in-place) | **23.428** |

## Multi-backend Testing
|---|---|---|---|
| Nvidia (H20) | PASS (6 cases) | 23.428 | Primary |
| Tianshu | N/A | — | — |
| Muxi | N/A | — | — |
| Ascend | N/A | — | — |
| Hygon | N/A | — | — |

## Files Changed
- `src/flag_gems/ops/_fused_adagrad_.py`: Triton kernel implementation
- `tests/test_fused_adagrad_.py`: Accuracy test
- `benchmark/test_fused_adagrad_.py`: Performance benchmark


src/flag_gems/ops/_fused_adagrad_.py
```diff
@@ -0,0 +1,223 @@
...
+    state_sum_f = state_sum_load.to(tl.float32)
+
+    # Gradient unscaling (for AMP); the scaled gradient is stored back to ``grad``
+    # unchanged (the original, unscaled value is preserved).
+    if has_grad_scale:
+        grad_scale_val = tl.load(grad_scale)
+        grad_to_store = grad_load
+        grad_f = grad_f / grad_scale_val
+    else:
+        grad_to_store = grad_load
```
Caeruleann: When grad_scale is provided, ATen updates grad in place with the unscaled value. Here grad_to_store is captured before the division, so the original scaled gradient is written back. Please store the unscaled gradient and add a native parity test that also checks the mutated grads.

src/flag_gems/ops/_fused_adagrad_.py
```diff
@@ -0,0 +1,223 @@
...
+    corrected_lr = lr / (1.0 + (step - 1.0) * lr_decay)
+
+    # Update the parameter: param -= corrected_lr * grad / (sqrt(state_sum) + eps)
+    param_f = param_f - corrected_lr * grad_f / (tl.sqrt(state_sum_f) + eps)
+
+    # Store the updated values back, downcasting to the original element dtype.
+    tl.store(param + offsets, param_f.to(param_load.dtype), mask=mask)
+    tl.store(state_sum + offsets, state_sum_f.to(state_sum_load.dtype), mask=mask)
+    if has_grad_scale:
+        tl.store(grad + offsets, grad_to_store.to(grad_load.dtype), mask=mask)
```
Caeruleann: So as https://github.com/flagos-ai/FlagGems/pull/5917/changes#r3995606155

src/flag_gems/ops/_fused_adagrad_.py
```diff
@@ -0,0 +1,223 @@
...
+    ``found_inf`` is truthy the update is skipped entirely (AMP recovery).
+    """
+    has_grad_scale = grad_scale is not None
+    has_found_inf = found_inf is not None
+    grad_scale_ptr = grad_scale if has_grad_scale else None
+    found_inf_ptr = found_inf if has_found_inf else None
+
+    # _fused_adagrad supports float16 / bfloat16 / float32 parameters; the
+    # kernel accumulates in float32 and stores back to the original dtype.
+    # float64 and integer dtypes are rejected to match the ATen reference.
```
Caeruleann: Does ATen actually reject float64 inputs for _fused_adagrad?

src/flag_gems/ops/_fused_adagrad_.py
```diff
@@ -0,0 +1,223 @@
...
+    for _p in params:
+        assert _p.dtype in (torch.float16, torch.bfloat16, torch.float32), (
+            "_fused_adagrad only supports float16/bfloat16/float32 inputs, "
+            f"got {_p.dtype}"
+        )
+
+    # Optional early exit: skip the whole step when an inf/nan was observed.
+    if has_found_inf and found_inf.item() > 0:
+        return
```
Caeruleann: Please remove the host-side found_inf.item() check, which introduces a device synchronization on every AMP step. The kernel already reads found_inf; its condition should match ATen's found_inf == 1 behavior.

src/flag_gems/ops/_fused_adagrad_.py
```diff
@@ -0,0 +1,223 @@
...
+
+    for i in range(len(params)):
+        param = params[i]
+        grad = grads[i]
+        state_sum = state_sums[i]
+        state_step = state_steps[i]
+
+        n = param.numel()
```
Caeruleann, line 211: Please validate all tensor lists before launching any kernel, including list lengths and corresponding shapes, strides, dtypes, devices, and layouts. Otherwise mismatched lists may cause out-of-bounds access or leave the optimizer state partially updated before an error is raised.

tests/test_fused_adagrad_.py
```diff
@@ -0,0 +1,252 @@
...
+    g = grad.float()
+    s = state_sum.float()
+    if maximize:
+        g = -g
+    if weight_decay != 0:
+        g = g + p * weight_decay
+    s = s + g * g
+    corrected_lr = lr / (1.0 + (step - 1.0) * lr_decay)
+    p = p - corrected_lr * g / (torch.sqrt(s) + eps)
+    return p.to(param.dtype), s.to(state_sum.dtype)
```
Caeruleann: Use the native torch._fused_adagrad_ as the accuracy reference

tests/test_fused_adagrad_.py
Caeruleann, line 1: Can you extend the tests to cover grad_scale, found_inf, FP64, non-zero initial state, empty tensors, invalid tensor lists, and the in-place update of grads?

src/flag_gems/ops/_fused_adagrad_.py
Yukun-Cui: Critical bug! I'll fix the order so grad_to_store captures the unscaled value after division. Will add tests that verify the in-place grad mutation.

src/flag_gems/ops/_fused_adagrad_.py
Yukun-Cui: Same fix - I'll ensure the unscaled gradient is stored correctly.

src/flag_gems/ops/_fused_adagrad_.py
Yukun-Cui: I'll verify ATen's actual dtype support and adjust accordingly. If ATen supports float64, I'll add support for it.

src/flag_gems/ops/_fused_adagrad_.py
Yukun-Cui: I'll remove the .item() call and handle found_inf checking entirely within the kernel to avoid synchronization.

src/flag_gems/ops/_fused_adagrad_.py
Yukun-Cui: I'll add comprehensive upfront validation of all tensor list properties (lengths, shapes, strides, dtypes, devices, layouts) before any kernel launch.

tests/test_fused_adagrad_.py
Yukun-Cui: I'll switch to using torch._fused_adagrad_ as the reference.

tests/test_fused_adagrad_.py
Yukun-Cui: I'll add comprehensive test coverage for all these cases: grad_scale, found_inf, FP64, non-zero initial state, empty tensors, invalid tensor lists, and in-place grad updates.

src/flag_gems/ops/_fused_adagrad_.py
Yukun-Cui: I measured this against the native op before changing the code, and the in-place grad update turns out not to be part of the contract — it only happens for tail elements when numel is not a multiple of the CUDA kernel's vector width (4):

```
n=    1 mutated=True     n=    4 mutated=False
n=    2 mutated=True     n=    8 mutated=False
n=    3 mutated=True     n=   16 mutated=False
n=    5 mutated=True     n=   64 mutated=False
n=    6 mutated=True     n= 1000 mutated=False
n=    7 mutated=True
```
(fp32, grad_scale=4, grad filled with 8.0)

So at any size a real optimizer uses, ATen leaves `grad` untouched; the write-back is a vectorization artifact of the unvectorized tail path. My earlier "I'll store the unscaled gradient" reply was based on a 1-element probe, which is exactly the case that shows the artifact.

`grads` is therefore now treated as read-only, documented as such, and `_assert_grads_unchanged` asserts it in every accuracy test. If you would rather bit-match the artifact I can do that instead, but reproducing a width-4 tail behaviour seems worse than a clean read-only contract.

Also fixed here: the accumulation type. fp64 params now accumulate in fp64 (`ACC_DTYPE` constexpr) — computing a double parameter in fp32 was losing ~7 digits.

src/flag_gems/ops/_fused_adagrad_.py
Yukun-Cui: Covered by the reply above: the `tl.store` to `grad` is removed entirely rather than reordered, since the native op only writes back for non-multiple-of-4 tails.

src/flag_gems/ops/_fused_adagrad_.py
Yukun-Cui: No, it does not — good catch. Checked all four float types against the native op:

```
float64   : ACCEPTED
float16   : ACCEPTED
bfloat16  : ACCEPTED
float32   : ACCEPTED
```

The assert was simply wrong. fp64 is now supported, and the kernel accumulates it in fp64 rather than fp32 (verified the native fp64 path is genuinely double-precision, not fp32 internally), so a double parameter keeps full precision. fp64 is parametrized in the basic/options tests.

src/flag_gems/ops/_fused_adagrad_.py
Yukun-Cui: Removed — `found_inf` is now read on-device only, so no AMP step synchronizes and CUDA graph capture is not broken.

Your note about matching ATen's `found_inf == 1` also caught a second bug: the kernel used `> 0`, but ATen compares against 1 exactly, so a value of 2.0 diverged.

```
found_inf=0.0: param=0.900000  UPDATED
found_inf=1.0: param=1.000000  SKIPPED
found_inf=2.0: param=0.900000  UPDATED   <- kernel previously skipped
```

The kernel now tests `== 1`, and `test_fused_adagrad__found_inf` parametrizes {0.0, 1.0, 2.0}, asserting exact buffer equality in the skip case.

src/flag_gems/ops/_fused_adagrad_.py
Yukun-Cui: Added up-front validation over all four lists before any kernel launch: list lengths, then per-entry shape, dtype, device, stride, and contiguity. Contiguity is checked before the stride comparison so a strided param reports "param[0] must be contiguous" instead of a confusing stride mismatch.

Empty params are also skipped now — a zero-numel entry previously produced a zero-sized grid launch.

`test_fused_adagrad__mismatched_list_lengths` and `test_fused_adagrad__mismatched_tensor_properties` assert both the error and that no parameter was mutated, which is the partial-update case you flagged.

tests/test_fused_adagrad_.py
Yukun-Cui: Switched — the hand-written formula is gone and `_native_reference` now runs `torch._fused_adagrad_` on clones. That is what surfaced the `found_inf == 1` and fp64 issues, which the formula-based reference could not have caught.

tests/test_fused_adagrad_.py
Yukun-Cui: All added:

- `grad_scale` — parametrized over scales x weight_decay x maximize
- `found_inf` — {0.0, 1.0, 2.0}, exact-equality assertion on the skip path
- fp64 — in the basic and options tests
- non-zero initial state — `nonzero_state` parameter; a zero state hides sign and accumulation errors in the `state_sum` update
- empty tensors — all-empty, plus a mixed empty/non-empty list checking the non-empty entry still updates
- invalid tensor lists — mismatched lengths, shapes, dtypes; assert no partial mutation
- in-place grad — `_assert_grads_unchanged` in every accuracy test (see the width-4 discussion above for why read-only is the right contract)

Verified 160 dtype/size/option configurations against the native op.

src/flag_gems/ops/_fused_adagrad_.py
Caeruleann: Got it. Since the deviation is already documented in the code, I think that’s sufficient.

src/flag_gems/ops/_fused_adagrad_.py
```diff
@@ -0,0 +1,283 @@
...
+            )
+        # The kernel indexes with a flat offset, so non-dense inputs would read
+        # the wrong elements. Checked before the per-tensor stride comparison so a
+        # strided param is reported as such instead of as a stride mismatch.
+        for name, tensor in (
+            ("param", param),
+            ("grad", grad),
+            ("state_sum", state_sum),
+        ):
+            if not tensor.is_contiguous():
```
Caeruleann: ATen accepts strided tensors that are non-overlapping and dense, so requiring is_contiguous() rejects valid inputs such as matching transposed tensors. Please support these layouts and add a native-parity test with transposed inputs.

src/flag_gems/ops/_fused_adagrad_.py
Caeruleann: The up-front validation is still incomplete: state_steps are only length-checked, and the params list is not checked for a common dtype/device as ATen does. Please validate these before the first launch and add an invalid later-entry case to ensure no partial update occurs.

src/flag_gems/ops/_fused_adagrad_.py
Yukun-Cui: Fixed in 2982f82a. Confirmed against the native op first: `torch._fused_adagrad_` accepts a matching transposed layout (measured), so `is_contiguous()` was rejecting valid inputs.

The kernel now walks `param`/`grad`/`state_sum` in storage order through `as_strided((numel,), (1,))` views. Because the three operands are validated to share shape, dtype, device *and* stride set, and each is non-overlapping and dense, visiting them in storage order lines up the same logical element in each -- which is exactly the condition ATen's native CUDA kernel relies on. The `is_contiguous()` check is replaced with a mirror of `at::has_internal_overlap`, so aliasing views (an expanded tensor, say) are still rejected with an explicit error.

Added `test_fused_adagrad__transposed_parity` (transposed fp32/fp64 inputs compared against the native op) and `test_fused_adagrad__internal_overlap_rejected`. 84 tests pass.

src/flag_gems/ops/_fused_adagrad_.py
Yukun-Cui: Confirmed and fixed. Both points were real, and I reproduced them against the pre-fix source before changing anything.

**ATen's actual behaviour (measured, not assumed)**

I ran `torch.ops.aten._fused_adagrad_` against the cases below. ATen's rule is `check_fast_path_restrictions({params, grads, state_sums})` in `ForeachUtils.h` (`FusedAdagradKernel.cu` asserts on it before the impl runs), and it is stricter than "each entry internally consistent": `params[0]` fixes the dtype/device, and *every* tensor in all three lists must match it, be strided, and be non-overlapping-and-dense, with corresponding entries sharing sizes and strides. So the check is global across the lists:

| case (each entry internally consistent) | ATen |
|---|---|
| fp32 list, or fp16 list, or fp64 list | OK |
| all-transposed, all-contiguous, mixed shapes + mixed strides | OK |
| entry0 fp32 / entry1 fp16 | raise (common dtype) |
| entry0 fp32 / entry1 fp64 | raise (common dtype) |
| entry0 cuda / entry1 cpu | raise (common device) |
| `state_steps[1]` (and even `grads[1]`) on cpu | raise (common device) |
| stride-2 slice / expand / row-slice (not dense) | raise (layout) |

Every raise left **no** tensor modified (`partial-update=[]`), which is the property the fix has to preserve.

`state_steps` has no validation at all in ATen — the raw data pointer is passed to the kernel, which blindly reinterprets it as `float`. So a non-float32 step silently yields the wrong `corrected_lr` (int32 `3` reinterprets to a denormal, giving `corrected_lr ≈ lr` instead of `lr/3`), and a numel-0 step reads out of bounds — I reproduced that as a CUDA illegal memory access.

**What the pre-fix FlagGems code did** (reproduced on `2982f82a`)

- `params[1]` fp16 (entry0 fp32) → **no raise**, silently updated both entries.
- `state_steps[1]` on cpu → `ValueError: Pointer argument (at 3) cannot be accessed from Triton`, **after** params[0]/state_sums[0] were already updated (partial update).
- `state_steps[1] = torch.tensor([], ...)` → illegal memory access.

**Fix** — `59ba78ac` (`src/flag_gems/ops/_fused_adagrad_.py`)

The per-entry loop now mirrors `check_fast_path_restrictions`: `params[0]` gives the expected dtype/device, every tensor in `params`/`grads`/`state_sums` is checked against it plus strided + non-overlapping-and-dense, and corresponding entries are checked for matching sizes/strides. `state_steps` is validated for float32 dtype, exactly one element, and same device as `params[0]`. All of it runs before the first launch, so a rejected batch cannot leave a partial update. `_has_internal_overlap` became `_is_non_overlapping_and_dense` (the previous version treated a stride-2 slice as fine; ATen requires dense too).

Cross-checked the density helper against `torch.ops.aten.is_non_overlapping_and_dense` on 15 layouts (contiguous, transposed, permuted, channels_last, stride-2 slice, row-slice, expand, numel<=1, 0-dim, empty): **0 mismatches**. A direct accept/reject comparison against ATen on 11 cases matched on all but one, where we are deliberately stricter: an int32 step is rejected rather than silently misread (documented in a comment — `torch.optim.Adagrad(fused=True)` stores `state["step"]` as float32, so that is the honest contract).

**Tests** — `b357559c` (`tests/test_fused_adagrad_.py`)

`test_fused_adagrad__later_entry_bad_dtype_or_device` (params/grads/state_sums × dtype/device, bad entry **second**) and `test_fused_adagrad__later_entry_bad_state_step` (step dtype int32/int64/fp64, numel 2, numel 0, cpu device, bad entry **second**) — 10 cases. Both assert every param and state_sum is `torch.equal` to its pre-call clone, i.e. no partial update. I also extended the property-mismatch test with stride/device/layout cases and moved the step tensors to float32 to match the validated contract.

**Verification**

- Against pre-fix source all 10 new cases fail (the step cases via the CUDA illegal memory access); against the fix all pass. So they genuinely gate the bug rather than just documenting it.
- `tests/test_fused_adagrad_.py`: **97 passed** (was 87).
- `benchmark/test_fused_adagrad_.py`: **1 passed** (its input_fn already uses a float32 step, so no change needed).
- `pre-commit run --files ...`: all hooks pass, re-run clean (no files rewritten).

One judgement call worth flagging: I made the `state_steps` check stricter than ATen (float32 only) rather than permissive, because ATen's acceptance there is an artifact of reinterpreting bits, not a contract. If you'd rather accept any integral dtype since we read the step in its true dtype, say so and I'll relax just that check.

Yukun-Cui: Done in 59ba78ac + b357559c (pushed to `pr/_fused_adagrad_`).

The validation is now global, mirroring ATen's own fast-path restriction check (`check_fast_path_restrictions` in ForeachUtils.h), which the native CUDA kernel runs before its first launch:

- `params[0]` fixes the expected dtype and device; every tensor in `params`/`grads`/`state_sums` must match that dtype and device (so a list mixing fp32 and fp64 -- each entry internally consistent -- is rejected exactly as ATen rejects it).
- Every tensor must be non-overlapping and dense, and corresponding entries must share sizes and strides (`_strides_match` ignores size-1 dims, as ATen's helper does).
- `state_steps` entries are validated, not just length-counted.

All of it runs before any kernel launch, so a rejected batch leaves every buffer untouched -- no partial update.

New tests: `test_fused_adagrad__later_entry_bad_dtype_or_device` and `test_fused_adagrad__later_entry_bad_state_step` put the invalid entry *late* in the list and assert both the error and that every earlier parameter, state sum, and state step is still byte-identical to its pre-call value (`torch.equal(p, p0)` per entry). 97 tests pass; pre-commit clean.

YangLong114514 (dismissed):

Caeruleann (approved):

PR #5241, opened 2026-08-05
[KernelGen][Nvidia] Add replication_pad1d_backward operator with Triton kernel

Yukun-Cui:
## Summary
Adds a Triton kernel for `replication_pad1d_backward`. Computes the gradient of `replication_pad1d` with respect to the input tensor.

## Testing
- Parametrized tests over 2D `(C, W)` / 3D `(N, C, W)` shapes and symmetric/asymmetric paddings across `utils.FLOAT_DTYPES`
- Validated against reference on device via `to_reference`
- All 48 test cases passing
- Tested on: Nvidia, Tianshu, Muxi, Ascend, Hygon

## Performance
Test command: `pytest benchmark/test_replication_pad1d_backward.py --level core` (NVIDIA H20)

### replication_pad1d_backward

| dtype | Size | Torch Latency (ms) | Gems Latency (ms) | Speedup |
|-------|------|--------------------|-------------------|---------|
| float16 | [2, 6] | 0.008512 | 0.005664 | 1.503 |
| float16 | [4, 11] | 0.008768 | 0.005792 | 1.514 |
| float16 | [8, 19] | 0.008928 | 0.005920 | 1.508 |
| float16 | [1, 35] | 0.008192 | 0.005856 | 1.399 |
| float16 | [4, 67] | 0.008960 | 0.006080 | 1.474 |
| float16 | [8, 131] | 0.008928 | 0.006272 | 1.423 |
| float32 | [2, 6] | 0.007552 | 0.005728 | 1.318 |
| float32 | [4, 11] | 0.007808 | 0.005888 | 1.326 |
| float32 | [8, 19] | 0.007776 | 0.006048 | 1.286 |
| float32 | [1, 35] | 0.007520 | 0.005856 | 1.284 |
| float32 | [4, 67] | 0.007808 | 0.006240 | 1.251 |
| float32 | [8, 131] | 0.007840 | 0.006240 | 1.256 |
| bfloat16 | [2, 6] | 0.008640 | 0.005824 | 1.484 |
| bfloat16 | [4, 11] | 0.008800 | 0.005952 | 1.478 |
| bfloat16 | [8, 19] | 0.008960 | 0.005920 | 1.514 |
| bfloat16 | [1, 35] | 0.008192 | 0.005888 | 1.391 |
| bfloat16 | [4, 67] | 0.008896 | 0.006080 | 1.463 |
| bfloat16 | [8, 131] | 0.009024 | 0.006240 | 1.446 |

[...]

PR #5922, opened 2026-09-02
[KernelGen][Nvidia] Add replication_pad1d_backward operator with Triton kernel

Yukun-Cui:
> Replaces #5241 — original could not be reopened after the head branch was rebuilt to drop ~3,400 spurious mirror-history commits (now a clean single commit on `master`). Content is byte-identical to the original branch tip.

## Summary
Adds a Triton kernel for `replication_pad1d_backward`. Computes the gradient of `replication_pad1d` with respect to the input tensor.

## Testing
- Parametrized tests over 2D `(C, W)` / 3D `(N, C, W)` shapes and symmetric/asymmetric paddings across `utils.FLOAT_DTYPES`
- Validated against reference on device via `to_reference`
- All 48 test cases passing

## Performance
Test command: `pytest benchmark/test_replication_pad1d_backward.py --level core` (NVIDIA H20)

### replication_pad1d_backward

| float16 | [2, 6] | 0.008512 | 0.005664 | 1.503 |
| float16 | [4, 11] | 0.008768 | 0.005792 | 1.514 |
| float16 | [8, 19] | 0.008928 | 0.005920 | 1.508 |
| float16 | [1, 35] | 0.008192 | 0.005856 | 1.399 |
| float16 | [4, 67] | 0.008960 | 0.006080 | 1.474 |
| float16 | [8, 131] | 0.008928 | 0.006272 | 1.423 |
| float32 | [2, 6] | 0.007552 | 0.005728 | 1.318 |
| float32 | [4, 11] | 0.007808 | 0.005888 | 1.326 |
| float32 | [8, 19] | 0.007776 | 0.006048 | 1.286 |
| float32 | [1, 35] | 0.007520 | 0.005856 | 1.284 |
| float32 | [4, 67] | 0.007808 | 0.006240 | 1.251 |
| float32 | [8, 131] | 0.007840 | 0.006240 | 1.256 |
| bfloat16 | [2, 6] | 0.008640 | 0.005824 | 1.484 |
| bfloat16 | [4, 11] | 0.008800 | 0.005952 | 1.478 |
| bfloat16 | [8, 19] | 0.008960 | 0.005920 | 1.514 |
| bfloat16 | [1, 35] | 0.008192 | 0.005888 | 1.391 |
| bfloat16 | [4, 67] | 0.008896 | 0.006080 | 1.463 |
| bfloat16 | [8, 131] | 0.009024 | 0.006240 | 1.446 |

| replication_pad1d_backward | **1.407** |

## Multi-backend Testing
|---|---|---|---|
| Nvidia (H20) | PASS (48 cases) | 1.407 | kernel mode; above parity with PyTorch |
| Tianshu | N/A | — | — |
| Muxi | N/A | — | — |
| Ascend | N/A | — | — |
| Hygon | N/A | — | — |

## Files Changed
- `src/flag_gems/ops/replication_pad1d_backward.py`: Triton kernel implementation
- `tests/test_replication_pad1d_backward.py`: Accuracy test
- `benchmark/test_replication_pad1d_backward.py`: Performance benchmark


conf/operators.yaml
Caeruleann, line 1: Please only modify the YAML registration for `replication_pad1d_backward` and revert the unrelated formatting changes to other operator entries in `conf/operators.yaml`.

src/flag_gems/ops/replication_pad1d_backward.py
```diff
@@ -0,0 +1,221 @@
...
+    for block_start in range(1, W_in - 1, BLOCK_SIZE):
+        w_in = block_start + tl.arange(0, BLOCK_SIZE)
+        mask = w_in < W_in - 1
+
+        w_out = pad_left + w_in
+        out_mask = mask & (w_out < W_out)
+
+        out_offset = w_out.to(tl.int64) * grad_out_stride_w
+        out_ptr = grad_out_ptr + base_out + out_offset
+        val = tl.load(out_ptr, mask=out_mask, other=0.0)
```
Caeruleann, line 93: Please handle negative padding correctly. With pad_left <= -2, w_out can be negative, but out_mask only checks the upper bound, causing an out-of-bounds read before grad_output. Please add parity tests for cases such as (-2, 1), (1, -2), and (-2, -2).

src/flag_gems/ops/replication_pad1d_backward.py
```diff
@@ -0,0 +1,221 @@
...
+            device=grad_output.device,
+            dtype=grad_output.dtype,
+        )
+    elif dim == 2:
+        C, W_in = self_tensor.shape
+        grad_input = torch.empty(
+            (C, W_in),
+            device=grad_output.device,
+            dtype=grad_output.dtype,
+        )
```
Caeruleann, line 213: Please allocate grad_input from self_tensor and validate that grad_output has the expected dtype and device. ATen returns an output with the input's options and rejects a mismatched gradient dtype, while this implementation currently follows grad_output.

tests/test_replication_pad1d_backward.py
```diff
@@ -0,0 +1,42 @@
...
+# 2D/3D shapes covering small to medium width; pad1d operates on last dim
+REPLICATION_PAD1D_BACKWARD_SHAPES = [(2, 3, 8), (4, 16, 64), (1, 5, 16), (32, 256)]
+# Asymmetric and symmetric padding combinations
+REPLICATION_PAD1D_BACKWARD_PADDING = [(1, 1), (0, 2), (2, 1), (1, 2)]
+
+
+@pytest.mark.replication_pad1d_backward
+@pytest.mark.parametrize("shape", REPLICATION_PAD1D_BACKWARD_SHAPES)
+@pytest.mark.parametrize("padding", REPLICATION_PAD1D_BACKWARD_PADDING)
+@pytest.mark.parametrize("dtype", utils.FLOAT_DTYPES)
```
Caeruleann, line 30: Please extend the dtype coverage to float64 and the complex dtypes supported by native CUDA

conf/operators.yaml
Yukun-Cui: Will do! I'll revert all unrelated formatting changes and keep only the replication_pad1d_backward entry modification.

src/flag_gems/ops/replication_pad1d_backward.py
Yukun-Cui: I'll add proper bounds checking for negative padding values (especially pad_left <= -2 causing negative w_out) and add test cases for (-2, 1), (1, -2), (-2, -2).

src/flag_gems/ops/replication_pad1d_backward.py
Yukun-Cui: I'll fix the allocation to use self_tensor's dtype/device and add validation for grad_output mismatch.

tests/test_replication_pad1d_backward.py
Yukun-Cui: I'll add float64 and complex dtype test coverage.

conf/operators.yaml
Yukun-Cui: Reverted. The yaml diff was 44 insertions / 25 deletions, re-wrapping unrelated entries (`ctc_loss`, `cudnn_attention_forward` and others) purely as formatting. It is now **13 added lines** -- this operator's entry and nothing else.

I checked semantically rather than just by line count: parsing both versions and diffing by operator id confirms the branch adds exactly `replication_pad1d_backward`, deletes nothing, and changes no other entry's content. The entry sits between `replication_pad1d` and `replication_pad1d_out`.

src/flag_gems/ops/replication_pad1d_backward.py
Yukun-Cui: Fixed, and the bug was slightly wider than the middle loop. Three sites could index before the start of `grad_output` with a negative `pad_left`:

- the middle loop's `out_mask` only checked the upper bound, so negative `w_out` lanes loaded out of bounds -- now masked with `(w_out >= 0)`
- the left-edge accumulator ran `range(pad_left + 1, ...)`, which starts below 0 -- the bound is now clamped, making the range empty when column 0 is cropped rather than replicated
- the right-edge accumulator had the same issue via `pad_left + W_in - 1`

One correction on the requested cases: **(-2, -2) is not testable**, because ATen's own forward rejects it before any backward exists:

```
RuntimeError: input (W: 4) is too small. Calculated output W: 0
```

So there is no reference to compare against. I covered `(-2, 1)` and `(1, -2)` as asked, plus `(-1, 0)`, `(0, -1)`, `(-1, -1)` and `(-3, 2)` -- nine padding combinations in total, all parity-checked against the native op.

src/flag_gems/ops/replication_pad1d_backward.py
Yukun-Cui: Both fixed. `grad_input` is now allocated from `self_tensor` (matching `at::empty_like(self)`), and a mismatched `grad_output` dtype raises with ATen's wording rather than being followed:

```
self fp64, grad_output fp64 -> grad_input fp64
self fp64, grad_output fp32 -> RuntimeError: expected scalar type Double but found Float
```

Device mismatch is rejected too. Added `..._grad_output_dtype_mismatch` and `..._grad_input_takes_input_options` to cover both.

While here I also widened the edge accumulators: the left/right gradients are sums over the replicated span, so they now accumulate in fp32 for fp16/bf16 inputs instead of adding in the storage dtype element by element.

tests/test_replication_pad1d_backward.py
Yukun-Cui: Extended to fp64 and complex. Complex needed real work rather than just a parametrize entry, since Triton has no complex scalar type -- the kernel now runs once per component over a `view_as_real` slice, which is valid here because this operator only gathers and sums along W and never mixes the components.

Worth recording that two earlier attempts were wrong, one of them silently:

- folding the component axis into the channel axis leaves a 4D tensor that the stride unpacking rejects (loud failure)
- reshaping a `movedim`'d view **silently reorders the data** -- it returned plausible-looking gradients that were simply incorrect

Slicing the last axis instead keeps the original W layout, so only the element stride changes, which the kernel already takes as a parameter. Verified across shapes, both complex dtypes and negative padding.

One deviation: `utils.COMPLEX_DTYPES` is not used as-is because it includes `complex32`, which ATen's own forward rejects (`"replication_pad1d" not implemented for ComplexHalf`), so there would be no reference. complex64 and complex128 are listed explicitly.

Coverage is now fp16/bf16/fp32/fp64/complex64/complex128 x nine paddings: 219 tests, up from 48.

tests/test_replication_pad1d_backward.py
```diff
@@ -0,0 +1,91 @@
...
+    (1, 1),
+    (0, 2),
+    (2, 1),
+    (1, 2),
+    (-2, 1),
+    (1, -2),
+    (-1, 0),
+    (0, -1),
+    (-1, -1),
+]
```
Caeruleann, line 41: The requested (-2, -2) case is still missing — ATen's forward only rejects it when W_out ≤ 0 (your quote is from W=4), while all test shapes here have W ≥ 8 where (-2,-2) is valid (W=8 → W_out=4) and the kernel handles it correctly, so please add (-2, -2) (and the mentioned-but-absent (-3, 2)) to this list.

src/flag_gems/ops/replication_pad1d_backward.py
```diff
@@ -0,0 +1,296 @@
...
+        replication_pad1d_backward_kernel[grid](
+            grad_output,
+            grad_input,
+            B,
+            C,
+            W_in,
+            W_out,
+            left,
+            right,
+            grad_out_s_n if dim == 3 else grad_out_s_n,
```
Caeruleann:
```suggestion
            grad_out_s_n,
```

src/flag_gems/ops/replication_pad1d_backward.py
```diff
@@ -0,0 +1,296 @@
...
+            grad_out_s_c,
+            grad_out_s_w,
+            grad_in_s_n if dim == 3 else grad_in_s_n,
```
Caeruleann:
```suggestion
            grad_in_s_n,
```

tests/test_replication_pad1d_backward.py
Yukun-Cui: You're right, and I verified it rather than taking it on faith. I ran ATen's own forward across the test widths:

```
W=8   pad=(-2,-2) OK   W_out=4
W=64  pad=(-2,-2) OK   W_out=60
W=16  pad=(-2,-2) OK   W_out=12
W=256 pad=(-2,-2) OK   W_out=252
W=4   pad=(-2,-2) ERR  input (W: 4) is too small. Calculated output W: 0
```

W=4 is the only shape that rejects it, and it is not one of the shapes in this file — every shape here has W >= 8, so W_out >= 4 and the padding is legal. The comment I wrote had generalized the W=4 error message to all widths; that was my mistake.

Fixed in commit 49c40ff8:
- Added `(-2, -2)` and `(-3, 2)` to `REPLICATION_PAD1D_BACKWARD_PADDING`.
- Rewrote the comment to state the actual rule (ATen rejects only when W_out <= 0), and noted that the quoted "W: 4 is too small" message comes from a width these shapes never use.

The two new entries add 48 parametrized cases (2 paddings x 4 shapes x 6 dtypes). Full file: 267 passed in 2.17s. I also confirmed with `--collect-only` that the new `padding9`/`padding10` ids are actually collected (48), so this is real coverage rather than a silently-deselected case.

src/flag_gems/ops/replication_pad1d_backward.py
Yukun-Cui: Applied as suggested. The ternary was a no-op — both branches read `grad_out_s_n` (the 2D branch sets it to 0 earlier, so passing it is correct), so this is now the plain argument:

```python
            grad_out_s_n,
```

Commit 49c40ff8. `tests/test_replication_pad1d_backward.py` still passes 267/267, which exercises both the 2D and 3D paths.

src/flag_gems/ops/replication_pad1d_backward.py
Yukun-Cui: Applied as suggested, same no-op ternary as the line above — both branches were `grad_in_s_n`, which is set to 0 for the 2D case:

```python
            grad_in_s_n,
```

Commit 49c40ff8. No behavior change: 267/267 tests pass, including the 2D (C, W) shapes that rely on `grad_in_s_n` being passed as 0.

Caeruleann (approved):

PR #5929, opened 2026-09-02
[KernelGen][Nvidia] Add slogdet operator with Triton kernel

Yukun-Cui:
## Summary
Adds a Triton kernel for `slogdet`. Computes the sign and natural logarithm of the absolute value of the determinant of a square matrix (or a batch of square matrices) using LU decomposition with partial pivoting. This is the public alias for `torch.linalg.slogdet`.

## Testing

## Performance
Test command: `pytest benchmark/test_slogdet.py --level core` (NVIDIA H20)

### slogdet

| float32 | [2, 3, 3] | 0.043200 | 0.008416 | 5.133 |
| float32 | [4, 4] | 0.057056 | 0.008384 | 6.805 |
| float32 | [8, 8] | 0.060128 | 0.013664 | 4.400 |
| float32 | [16, 16] | 0.071872 | 0.026592 | 2.703 |
| float32 | [32, 32] | 0.092608 | 0.056000 | 1.654 |

| slogdet | **4.139** |

## Multi-backend Testing
|---|---|---|---|
| Nvidia (H20) | PASS (5 cases) | 4.139 | Primary |
| Tianshu | N/A | — | — |
| Muxi | N/A | — | — |
| Ascend | N/A | — | — |
| Hygon | N/A | — | — |

## Files Changed
- `src/flag_gems/ops/slogdet.py`: Triton kernel implementation
- `tests/test_slogdet.py`: Accuracy test
- `benchmark/test_slogdet.py`: Performance benchmark
- `conf/operators.yaml`: Add operator entry (kind: LinearAlg, stage: alpha 5.4)


src/flag_gems/ops/slogdet.py
```diff
@@ -0,0 +1,200 @@
...
+        # among rows >= j_ind.
+        col_j = tl.sum(tl.where(cols[:, None] == j_ind, tl.trans(work), 0.0), axis=0)
+        abs_col = tl.abs(col_j)
+        abs_col = tl.where(rows < j_ind, -1.0, abs_col)
+        abs_col = tl.where(rows < M, abs_col, -1.0)
+        pivot_val = tl.max(abs_col, axis=0)
+        pivot_row = tl.min(tl.where(abs_col == pivot_val, rows, BLOCK_M), axis=0)
+
+        # A near-zero pivot means the matrix is (numerically) singular.
+        is_singular = tl.where(pivot_val <= 1e-10, 1, is_singular)
```
Caeruleann, line 75: Please avoid using a fixed 1e-10 threshold to classify singular matrices: for example, diag([1e-11, 1]) is non-singular and ATen returns a finite logabsdet, while this implementation returns (0, -inf). Please match ATen’s zero-pivot semantics and add scaled/near-singular parity tests.

src/flag_gems/ops/slogdet.py
```diff
@@ -0,0 +1,200 @@
...
+    and ``sign = prod(sign(u_ii)) * sign(p)`` where ``sign(p)`` is the parity of
+    the row swaps performed during partial pivoting.
+    """
+    pid = tl.program_id(0)
+    rows = tl.arange(0, BLOCK_M)
+    cols = tl.arange(0, BLOCK_N)
+
+    offsets = pid * M * N + rows[:, None] * N + cols[None, :]
+    mask = (rows[:, None] < M) & (cols[None, :] < N)
+    work = tl.load(A_input + offsets, mask=mask, other=0.0).to(tl.float32)
```
Caeruleann, line 58: Please add a tiled or large-matrix path: rounding both dimensions to powers of two makes a 1025×1025 input create a 2048×2048 Triton block, exceeding Triton’s maximum tensor size. The current tests only cover matrices up to 32×32 and cannot expose this failure.

src/flag_gems/ops/slogdet.py
```diff
@@ -0,0 +1,200 @@
...
+    """
+    logger.debug("GEMS SLOGDET")
+    assert A.dim() >= 2, "Input must be at least 2D"
+    assert A.shape[-1] == A.shape[-2], "Input must be square"
+    # The Triton kernel uses float32 arithmetic internally; only float32 input
+    # is supported here (matching torch.linalg.slogdet which rejects low
+    # precision dtypes anyway).
+    assert A.dtype == torch.float32, (
+        "slogdet: unsupported dtype " f"{A.dtype}, expected float32"
+    )
```
Caeruleann, line 161: Is this correct? `torch.linalg.slogdet` also supports `float64` (as well as complex dtypes), so restricting this implementation to `float32` does not fully match ATen.

tests/test_slogdet.py
Caeruleann, line 1: Please extend the tests to cover FP64 and complex dtypes, 1×1 and 0×0 matrices, non-contiguous inputs, multi-dimensional and empty batches, near-singular/scaled matrices, NaN/Inf values, large matrices, invalid inputs, and forward/backward parity against ATen.

src/flag_gems/ops/slogdet.py
Yukun-Cui: I'll remove the fixed 1e-10 threshold and match ATen's actual zero-pivot semantics. Will add tests for scaled/near-singular matrices like diag([1e-11, 1]).

src/flag_gems/ops/slogdet.py
Yukun-Cui: I'll implement tiling to avoid exceeding Triton's maximum tensor size (2048×2048 limit). Will add tests for matrices > 1024.

src/flag_gems/ops/slogdet.py
Yukun-Cui: I'll add full support for float64 and complex dtypes to match ATen.

tests/test_slogdet.py
Yukun-Cui: I'll add comprehensive test coverage for all these cases: FP64/complex dtypes, 1×1/0×0 matrices, non-contiguous inputs, multi-dimensional/empty batches, near-singular/scaled matrices, NaN/Inf values, large matrices, invalid inputs, and forward/backward parity.

src/flag_gems/ops/slogdet.py
Yukun-Cui: Confirmed and fixed. Your example was exactly right — measured on the old code:

```
diag([1e-11, 1]) fp32:  gems=(sign=0.0, logabsdet=-inf)   ATen=(sign=1.0, logabsdet=-25.328436)
```

The threshold is now gone rather than retuned, because there is no safe value for it: ATen returns a finite logabsdet all the way down to `diag([1e-30, 1])` (`-69.077553`), and only an *exactly* zero pivot gives `-inf`. Any fixed epsilon misclassifies some non-singular matrix. The kernel now tests `diag == 0.0` and derives the sign from the parity of negative diagonal entries plus row swaps.

After the fix both dtypes agree with ATen to the printed digits:

```
diag([1e-11,1]) fp32  ref=(1.0, -25.328436)  gems=(1.0, -25.328436)
diag([1e-11,1]) fp64  ref=(1.0, -25.328436)  gems=(1.0, -25.328436)
```

Added `test_slogdet_near_singular_scaled` (asserts finiteness for 1e-8/1e-11/1e-12/1e-20) and `test_slogdet_scaled_identity` (uniform scaling 1e-3 … 1e3). `test_slogdet_singular` covers the exactly-singular cases.

One finding worth recording: `[[1,2,3],[1,2,3],[4,5,7]]` is mathematically singular and `torch.linalg.lu_factor` rejects it with "U[3,3] is zero", but `torch.linalg.slogdet` still returns a finite `-15.13` (fp32) / `-35.23` (fp64) for it. So it is not usable as a singular test case; the tests use zero-row/zero-column/rank-deficient matrices where elimination yields a hard zero.

src/flag_gems/ops/slogdet.py
Yukun-Cui: Confirmed and fixed. Reproduced the failure you predicted on the old code — a 1025x1025 fp32 input rounded to a 2048x2048 tile and never compiled:

```
OLD n=1025: triton CompilationError at 27:14
```

Rather than write a new tiled LU I followed the pattern already in the repo: `_det_blocked_kernel` in `linalg_det.py` solves the same problem, so the new `_slogdet_blocked_kernel` walks the matrix in 32-wide tiles through memory with `N` as a runtime argument. The block size no longer scales with the matrix, so nothing rounds up into an oversized tensor. Matrices `n <= 64` keep the fast single-tile register path; above that they take the tiled path.

Measured after the fix:

```
1025x1025 fp32  ref=3036.7419  gems=3036.7424  rel=1.61e-07
1025x1025 fp64  ref=3040.9285  gems=3040.9285  rel=1.50e-16
```

Tests now cover n = 65, 128, 257 (`test_slogdet_large`) and 1025 (`test_slogdet_large_1025`), so the boundary is exercised on both sides.

src/flag_gems/ops/slogdet.py
Yukun-Cui: You are right, it was not correct. The old assert rejected both:

```
OLD torch.float64:   AssertionError: slogdet: unsupported dtype torch.float64, expected float32
OLD torch.complex64: AssertionError: slogdet: unsupported dtype torch.complex64, expected float32
```

Both are supported now. The real path runs in the input precision instead of always casting to fp32, and a separate complex kernel handles complex64/complex128. Probing ATen for the dtype contract showed `sign` is a unit-modulus complex number while `logabsdet` stays real (complex64 -> float32 logabsdet), which the implementation now mirrors; `test_slogdet_complex` asserts the output dtypes and that `sign * exp(logabsdet)` reconstructs the determinant.

Validation now matches ATen exactly rather than using a blanket assert: half/bfloat16 raise "Low precision dtypes not supported", integers raise "Expected a floating point or complex tensor as input", and non-square/under-2D inputs reproduce ATen's messages (`test_slogdet_invalid_inputs`).

One limitation I want to flag rather than hide: the complex path is capped at n <= 64 and raises `NotImplementedError` above it. The tiled path relies on real-valued in-place elimination, and the repo has no complex LU to build on (`linalg_lu_factor` is float32/float64 only). Real inputs of any size are unaffected. Happy to lift that in a follow-up if you want it in this PR.

Randomized sweep across fp32/fp64/complex64/complex128 x 12 sizes x 3 batch shapes: 0 failures, worst logabsdet error 2.4e-04 (fp32), 1e-9 level for fp64.

tests/test_slogdet.py
Yukun-Cui: Done — the file went from 5 tests to 57, covering every case you listed:

| Requested | Test |
|---|---|
| FP64 + complex | `test_slogdet` (fp32/fp64), `test_slogdet_complex` (complex64/128) |
| 1x1 and 0x0 | `test_slogdet_1x1`, `test_slogdet_empty_matrix` |
| non-contiguous | `test_slogdet_non_contiguous` (transposed + strided views) |
| multi-dim / empty batches | `test_slogdet_multi_dim_batch`, `test_slogdet_empty_batch` |
| near-singular / scaled | `test_slogdet_near_singular_scaled`, `test_slogdet_scaled_identity` |
| NaN / Inf | `test_slogdet_nan_inf` |
| large matrices | `test_slogdet_large` (65/128/257), `test_slogdet_large_1025` |
| invalid inputs | `test_slogdet_invalid_inputs` (ATen error messages) |
| forward/backward parity | `test_slogdet_backward`, `test_slogdet_batched_backward` |

All 57 pass. Backward needed an `autograd.Function` (`d logabsdet/dA = inv(A)^H`); grads match ATen to 4.8e-07 (fp32) and 4.4e-15 (fp64).

Probing the reference turned up two ATen behaviours that shaped the tests:

- `0x0` returns `sign=1, logabsdet=0` (empty product), not zero — the implementation special-cases it. An empty *batch* returns empty tensors, which is different.
- For NaN input ATen is not self-consistent across devices: `[[nan,0],[0,1]]` gives `logabsdet=nan` on CPU but `-inf` on CUDA, and for an off-diagonal NaN CPU gives `sign=0` while CUDA gives `sign=1`. Since there is no single reference value, `test_slogdet_nan_inf` asserts the result is non-finite rather than pinning a number, and checks Inf inputs against ATen directly. Flagging in case that CPU/CUDA divergence is of interest independently of this PR.

Also verified the input tensor is not mutated, including for non-contiguous inputs (the old code cloned but the factorization is in-place, so this is now covered explicitly).

bin913 (approved):

PR #5936, opened 2026-09-02
[KernelGen][Nvidia] Add quantized_batch_norm operator with Triton kernel

Yukun-Cui:
## Summary
Adds a Triton kernel for `quantized_batch_norm`. Applies batch normalization on a 4D (NCHW) quantized tensor.

## Testing

## Performance
Test command: `pytest benchmark/test_quantized_batch_norm.py --level core` (NVIDIA H20)

### quantized_batch_norm

| quint8 | [(1, 3, 16, 16)] | 0.046843 | 0.252848 | 0.185 |
| quint8 | [(2, 8, 32, 32)] | 0.053787 | 0.256592 | 0.210 |
| quint8 | [(4, 16, 64, 64)] | 0.095531 | 0.255856 | 0.373 |
| quint8 | [(8, 32, 128, 128)] | 0.138146 | 0.261712 | 0.528 |
| quint8 | [(16, 64, 128, 128)] | 7.125433 | 0.273792 | 26.025 |
| quint8 | [(8, 64, 256, 256)] | 13.040328 | 0.314624 | 41.447 |
| qint8 | [(1, 3, 16, 16)] | 0.045556 | 0.251680 | 0.181 |
| qint8 | [(2, 8, 32, 32)] | 0.054437 | 0.255040 | 0.213 |
| qint8 | [(4, 16, 64, 64)] | 0.100803 | 0.255280 | 0.395 |
| qint8 | [(8, 32, 128, 128)] | 0.139169 | 0.260864 | 0.533 |
| qint8 | [(16, 64, 128, 128)] | 7.155010 | 0.270368 | 26.464 |
| qint8 | [(8, 64, 256, 256)] | 13.622594 | 0.315104 | 43.232 |

| quantized_batch_norm | **11.649** |

### quantized_batch_norm_out

| quint8 | [(1, 3, 16, 16)] | 0.099745 | 0.479648 | 0.208 |
| quint8 | [(2, 8, 32, 32)] | 0.122821 | 0.485584 | 0.253 |
| quint8 | [(4, 16, 64, 64)] | 0.262352 | 0.486208 | 0.540 |
| quint8 | [(8, 32, 128, 128)] | 0.497496 | 0.741312 | 0.671 |
| quint8 | [(16, 64, 128, 128)] | 20.748091 | 1.656512 | 12.525 |
| quint8 | [(8, 64, 256, 256)] | 46.204591 | 2.877056 | 16.060 |
| qint8 | [(1, 3, 16, 16)] | 0.100733 | 0.478800 | 0.210 |
| qint8 | [(2, 8, 32, 32)] | 0.124750 | 0.477872 | 0.261 |
| qint8 | [(4, 16, 64, 64)] | 0.263881 | 0.473488 | 0.557 |
| qint8 | [(8, 32, 128, 128)] | 0.513379 | 0.738688 | 0.695 |
| qint8 | [(16, 64, 128, 128)] | 20.964289 | 1.642240 | 12.766 |
| qint8 | [(8, 64, 256, 256)] | 41.244483 | 2.874208 | 14.350 |

| quantized_batch_norm_out | **4.925** |

## Multi-backend Testing
|---|---|---|---|
| Nvidia (H20) | PASS (24 cases) | 11.649/4.925 | Primary |
| Tianshu | N/A | — | — |
| Muxi | N/A | — | — |
| Ascend | N/A | — | — |
| Hygon | N/A | — | — |

## Files Changed
- `src/flag_gems/ops/quantized_batch_norm.py`: Triton kernel implementation
- `tests/test_quantized_batch_norm.py`: Accuracy test
- `benchmark/test_quantized_batch_norm.py`: Performance benchmark


src/flag_gems/ops/quantized_batch_norm.py
```diff
@@ -0,0 +1,274 @@
...
+        else:
+            q_out = q_out.to(tl.int8)
+
+        tl.store(curr_out_ptr, q_out, mask=mask)
+
+
+def _quantized_batch_norm_impl(
+    input, weight, bias, mean, var, eps, output_scale, output_zero_point
+):
+    logger.debug("GEMS QUANTIZED_BATCH_NORM")
```
Caeruleann, line 99: Please move this `logger.debug` call to the registered API functions, and use a distinct log message for each registered operator instead of sharing the same one in the common implementation.

src/flag_gems/ops/quantized_batch_norm.py
```diff
@@ -0,0 +1,274 @@
...
+    beta = beta.to(device)
+
+    # Flatten (H, W) into a single spatial dimension.
+    spatial = h * w
+    # Work on a contiguous (N, C, spatial) view so the kernel can iterate over
+    # the batch and feature dimensions with simple strides.
+    int_3d = int_repr.reshape(n, c, spatial)
+    out_3d = out_int_repr.reshape(n, c, spatial)
+    int_3d = int_3d.contiguous()
+    out_3d = out_3d.contiguous()
```
Caeruleann, line 182: Please handle channels-last inputs without detaching the kernel output: out_3d.contiguous() may allocate new storage, while the function still returns out_int_repr, leaving the returned values uninitialized. Please add channels-last parity coverage.

src/flag_gems/ops/quantized_batch_norm.py
```diff
@@ -0,0 +1,274 @@
...
+    out_n_stride, out_c_stride, out_s_stride = out_3d.stride()
+
+    # Fixed spatial tile width: one program per (channel, spatial-block). Each
+    # program loops over the batch dimension; alpha/beta are loaded once per
+    # channel and reused across the batch, minimizing parameter traffic and
+    # launch overhead for the small NCHW shapes typical of quantized BN.
+    BLOCK_SIZE = 512
+    grid = (c, triton.cdiv(spatial, BLOCK_SIZE))
+
+    with torch_device_fn.device(device):
```
Caeruleann, line 194: Please validate that weight, bias, mean, and var have exactly C elements and the expected dtype before launch; otherwise invalid inputs may be broadcast or cause out-of-bounds reads from alpha/beta, while ATen rejects them.

src/flag_gems/ops/quantized_batch_norm.py
```diff
@@ -0,0 +1,274 @@
...
+    weight=None,
+    bias=None,
+    mean=None,
+    var=None,
+    eps=1e-05,
+    output_scale=1.0,
+    output_zero_point=0,
+    *,
+    out=None,
+):
```
Caeruleann, line 247: Please preserve the full ATen out= contract: update the output qparams, handle output resizing consistently, and respect out.storage_offset(); the current raw storage view writes to the wrong region for sliced outputs. Please cover mismatched qparams/shapes and nonzero-offset views.

src/flag_gems/ops/quantized_batch_norm.py
Caeruleann, line 194: Please handle input.numel() == 0 before launching the kernel: ATen returns an empty clone preserving the input qparams, whereas this implementation can launch a zero-sized grid or return different qparams. Please add N/C/H/W-zero parity tests.

src/flag_gems/ops/quantized_batch_norm.py
```diff
@@ -0,0 +1,274 @@
...
+        * torch.tensor(scale_ratio, device=cpu, dtype=torch.float32)
+    )
+    # beta  = float((bias - mean * inv_sigma * weight) / out_scale)  (division is double)
+    inner = bias_f32 - mean_f32 * inv_sigma * weight_f32
+    beta = (
+        inner.to(torch.float64)
+        / torch.tensor(float(output_scale), device=cpu, dtype=torch.float64)
+    ).to(torch.float32)
+    alpha = alpha.to(device)
+    beta = beta.to(device)
```
Caeruleann, line 173: Please compute alpha/beta on device with a dedicated Triton kernel or existing FlagGems operators; moving all parameters to CPU and back on every call introduces synchronization and violates the host-side no-PyTorch-compute rule.

src/flag_gems/ops/quantized_batch_norm.py
```diff
@@ -0,0 +1,274 @@
...
+            out_n_stride,
+            out_c_stride,
+            out_s_stride,
+            IS_QUINT8=is_quint8,
+            MIN_VAL=min_val,
+            MAX_VAL=max_val,
+            BLOCK_SIZE=BLOCK_SIZE,
+        )
+
+    return torch._make_per_tensor_quantized_tensor(
```
Caeruleann, line 216: Please avoid calling torch._make_per_tensor_quantized_tensor directly because it is a private PyTorch API; use a public API or an approved FlagGems runtime wrapper instead.

src/flag_gems/ops/quantized_batch_norm.py
Yukun-Cui: I'll move the logger.debug calls to individual registered functions with operator-specific messages.

src/flag_gems/ops/quantized_batch_norm.py
Yukun-Cui: Critical bug! I'll fix the channels-last handling to properly propagate the output and avoid returning uninitialized values. Will add channels-last tests.

src/flag_gems/ops/quantized_batch_norm.py
Yukun-Cui: I'll add validation for weight/bias/mean/var having exactly C elements with correct dtypes before launch.

src/flag_gems/ops/quantized_batch_norm.py
Yukun-Cui: I'll implement proper out= handling: update output qparams, handle resizing, and respect storage_offset(). Will add tests for mismatched qparams/shapes and nonzero-offset views.

src/flag_gems/ops/quantized_batch_norm.py
Yukun-Cui: I'll handle input.numel() == 0 before kernel launch, returning an empty clone preserving input qparams. Will add N/C/H/W-zero parity tests.

src/flag_gems/ops/quantized_batch_norm.py
Yukun-Cui: I'll move alpha/beta computation to a Triton kernel to avoid CPU synchronization.

src/flag_gems/ops/quantized_batch_norm.py
Yukun-Cui: I'll replace torch._make_per_tensor_quantized_tensor with a public API or approved FlagGems runtime wrapper.

src/flag_gems/ops/quantized_batch_norm.py
Yukun-Cui: Fixed in commit 423b58c0. `logger.debug` now lives in the registered API functions with distinct messages: `quantized_batch_norm()` logs `GEMS QUANTIZED_BATCH_NORM` and `quantized_batch_norm_out()` logs `GEMS QUANTIZED_BATCH_NORM_OUT`; the shared `_quantized_batch_norm_impl` no longer logs. Verified with `python tools/ci_checks/check_api_logs.py --operators '["quantized_batch_norm","quantized_batch_norm_out"]'` -> "All operators follow the API logging convention".

src/flag_gems/ops/quantized_batch_norm.py
Yukun-Cui: Fixed in commit 423b58c0. Reproduced first: a channels-last input made `out_3d.contiguous()` reallocate, and the returned `out_int_repr` stayed uninitialized (48/48 values garbage, max diff 100 vs ATen). The fix removes the detach entirely: the result is allocated up front with `torch._empty_affine_quantized(..., memory_format=channels_last)` when the input is channels-last (contiguous otherwise), and the main kernel stores into a `(N, C, H*W)` strided view of that tensor's storage, so results land in the output's own storage in both layouts. Channels-last parity coverage added in `test_quantized_batch_norm_channels_last` (quint8/qint8 x 2 out-param sets): values now match ATen byte-for-byte and the output keeps the channels-last layout, which the test asserts on both the FlagGems result and the ATen reference.

src/flag_gems/ops/quantized_batch_norm.py
Yukun-Cui: Fixed in commit 423b58c0. Validation now runs before the launch, in ATen's order (weight/bias sizes -> weight/bias dtypes -> mean/var sizes -> mean/var dtypes, per `q_batch_norm2d_impl` in Normalization.cpp): `_check_param_size` requires `numel() == C` (ATen accepts any shape holding C elements, e.g. a (1,3) weight passes -- verified against ATen) with ATen's messages ("Expect weight size to match C", "Mean size must match channel dimension"), and `_check_param_dtype` rejects non-float32 with ATen's `const_data_ptr<float>` message ("expected scalar type Float but found Double/Int/Byte"). Verified empirically: sizes 1/2/4 and float64/int32/uint8 params now raise RuntimeError exactly like `torch.quantized_batch_norm`. Coverage in `test_quantized_batch_norm_bad_size` (4 params x sizes 1/2/4) and `test_quantized_batch_norm_bad_dtype` (4 params x 3 dtypes), 24 tests.

src/flag_gems/ops/quantized_batch_norm.py
Yukun-Cui: Fixed in commit 423b58c0. The old raw-uint8 view ignored `storage_offset()` (confirmed: a sliced out got values written at offset 0 of the backing storage while ATen wrote the slice region). The out= overload now follows the generated ATen kernel: (1) dtype check with ATen's message ("Expected out tensor to have dtype c10::quint8, but got c10::qint8 instead"), then device check; (2) shape resize via `set_(storage, offset, size, stride)` (aten::resize_ is not implemented for QuantizedCUDA) preserving storage_offset and only growing storage, matching `at::native::resize_output`; (3) `out.copy_(result)`, which writes through out's own strides -- so a non-zero-offset slice is written in exactly the right region -- and adopts the result's scale/zero_point, which is what ATen leaves on out (verified: out qparams 0.9/100 become 0.1/3, r is out, sliced region matches ATen and neighbours stay untouched; outside the slice the backing bytes are unchanged). Coverage: `test_quantized_batch_norm_out_mismatched_qparams`, `test_quantized_batch_norm_out_resizes`, `test_quantized_batch_norm_out_storage_offset` (nonzero-offset slice + outside-region untouched), `test_quantized_batch_norm_out_wrong_dtype`, `test_quantized_batch_norm_out_wrong_device`.

src/flag_gems/ops/quantized_batch_norm.py
Yukun-Cui: Fixed in commit 423b58c0. `input.numel() == 0` now short-circuits to `input.clone()` *before* parameter validation and before any kernel launch, exactly like ATen (`if (qx.numel() == 0) { return qx.clone(); }` in Normalization.cpp). Verified against ATen: for every degenerate dimension (N/C/H/W = 0) ATen returns an empty clone preserving the *input's* qparams -- e.g. input scale 0.5 / zero_point 10 comes back even though the call requested output_scale 0.1 / zero_point 3, and this happens even when weight/bias/mean/var have the wrong sizes (validation is skipped), so the clone is placed ahead of those checks to match. The `.out` overload also gets the ATen behavior (verified: out is resized to the empty shape and adopts the input's qparams). Coverage: `test_quantized_batch_norm_empty` (4 shapes x quint8/qint8) and `test_quantized_batch_norm_out_empty`.

src/flag_gems/ops/quantized_batch_norm.py
Yukun-Cui: Fixed in commit 423b58c0. alpha/beta are now computed by a dedicated Triton kernel (`quantized_batch_norm_alpha_beta_kernel`) on the input device; the CPU round-trip (`mean.to(cpu)`, `torch.sqrt`, tensor division, `alpha.to(device)`) is gone and the parameters are only moved with `.to(device=..., dtype=float32)`, a metadata/allocation operation -- no host torch compute remains in the file. The kernel reproduces `at::native::compute_fused_params` bit-for-bit, which required keeping the three C++ float64 promotion points (the `1.0` in `1.0/sqrt(...)`, the `input_scale/output_scale` ratio and the final `/output_scale`) in float64. Two exponent experiments drove the exact form: (a) ATen's beta must NOT be FMA-contracted -- I constructed m,w,b triples where `b - fp32(m*w)` and `fmaf(-m,w,b)` round to different integers and ATen matched the plain form 40/40 (the contracted form 0/40); (b) Triton *does* contract `b - m*inv*w` (isolation probe: 732/4096 elements differed from the plain model, 0/4096 from the FMA model). So the float32 products/subtract are emitted as `tl.fma(x, y, 0.0)` to force single rounding, and with that the kernel matches a hand-written C++-semantics model on 0 mismatches out of 3000 random (C, scale, zero_point) draws for both alpha and beta (the previous host code, by contrast, folded the ratio to float32 and differed from the C++ semantics in 623/2000 alpha cases). The full test suite is bit-exact (236 passed).

src/flag_gems/ops/quantized_batch_norm.py
Yukun-Cui: Fixed in commit 423b58c0. The private `torch._make_per_tensor_quantized_tensor` call is gone; the result is allocated with `torch._empty_affine_quantized`, a real ATen op declared in `aten/src/ATen/native/native_functions.yaml` and registered for the QuantizedCPU and QuantizedCUDA backends (it is also the operator the reference `at::native::q_batch_norm2d_impl` itself calls via `at::_empty_affine_quantized`). It is required rather than `torch.quantize_per_tensor`: `quantize_per_tensor` is a compute op that would quantize a float tensor, whereas this op just allocates storage with the requested scale/zero_point, and unlike the private helper it honours `memory_format` -- `_make_per_tensor_quantized_tensor` silently normalises strides to contiguous on CUDA, which is what caused the channels-last defect in the comment above. `grep -n 'torch._make_per_tensor'` on the file now returns nothing (only the docstring mentions it, to explain why it is not used).

src/flag_gems/ops/quantized_batch_norm.py
```diff
@@ -0,0 +1,469 @@
...
+        ](
+            weight,
+            bias,
+            mean,
+            var,
+            alpha,
+            beta,
+            channels,
+            EPS=float(eps),
+            SCALE_RATIO=in_scale / out_scale,
```
Caeruleann: SCALE_RATIO=in_scale / out_scale is a Python float division, so output_scale=0.0 crashes the host with ZeroDivisionError — the native op computes fine (C++ double division gives inf/NaN, which the clamp then turns into a result). Please pass in_scale/out_scale separately and divide in the kernel (or guard the host expression), so a zero output scale degrades the same way native does.

Yukun-Cui: 已修（a275ab86）。

复现了：`output_scale=0.0` 时 host 侧 `in_scale / out_scale` 抛 `ZeroDivisionError`，而 native 算出全 0（`output_zero_point=0/10` 都是 0），CUDA 侧完全不报错。

改法是按你说的把两个 scale 拆开传进 kernel、在 kernel 里做 f64 除法：

- `SCALE_RATIO` / `OUTPUT_SCALE` 两个 constexpr 换成运行期 f64 参数 `INPUT_SCALE` / `OUTPUT_SCALE`，`in/out` 与 `(bias - inner)/out` 都在 kernel 内计算，得到 C++ double 除法的 inf/NaN。
- **仅把除法搬到 kernel 还不够**：退化 scale 下 alpha/beta 变成 inf/NaN，而 requantize 的浮点 clamp 会把 +inf 饱和到 `MAX_VAL`、NaN 结果不确定，与 native 不一致（实测 57% 元素不同）。native 走的是 `lrintf` 整数转换，x86 上 NaN/越界一律得到 integer-indefinite（INT64_MIN），再被 clamp 折到 dtype 下界。所以主 kernel 现在按这个语义显式做 float→int64 转换后 clamp，quint8 得到 0、qint8 得到 -128，与 native 一致。

验证：`{quint8,qint8} × in_scale{0,0.1} × out{(0,0),(0,10),(1e-30,3)}` 全部逐字节一致；再随机 30 组参数 × 12 种退化组合，360/360 一致。

新增 `test_quantized_batch_norm_zero_output_scale`。`tests/test_quantized_batch_norm.py`: 248 passed。

conf/operators.yaml
```diff
@@ -12189,6 +12189,29 @@ ops:
...
+  description: Applies batch normalization on a 4D (NCHW) quantized tensor.
+  for:
+  - quantized_batch_norm
+  labels:
+  - aten
+  - KernelGen
+  kind:
+  - NeuralNetwork
+  stages:
+  - alpha: '5.4'
```
```suggestion
  - alpha: '5.5'
```

conf/operators.yaml
```diff
@@ -12189,6 +12189,29 @@ ops:
...
+    A variant of quantized_batch_norm() that writes the result into the out tensor.
+  for:
+  - quantized_batch_norm.out
+  labels:
+  - aten
+  - KernelGen
+  kind:
+  - NeuralNetwork
+  stages:
+  - alpha: '5.4'
```
```suggestion
  - alpha: '5.5'
```

Caeruleann (dismissed):

huangyiqun (approved):

huangyiqun:
Plz fix CI failed.

PR #5937, opened 2026-09-02
[KernelGen][Nvidia] Add quantized_max_pool3d operator with Triton kernel

Yukun-Cui:
## Summary
Adds a Triton kernel for `quantized_max_pool3d`. Applies 3D max pooling over a per-tensor quantized quint8 tensor.

## Testing

## Performance
Test command: `pytest benchmark/test_quantized_max_pool3d.py --level core` (NVIDIA H20)

### quantized_max_pool3d

| quint8 | [4, 3, 16, 56, 56] | 0.469856 | 0.016288 | 28.847 |
| quint8 | [8, 64, 8, 28, 28] | 0.631216 | 0.027008 | 23.371 |
| quint8 | [16, 128, 4, 14, 14] | 0.343200 | 0.020288 | 16.916 |
| quint8 | [32, 256, 2, 7, 7] | 0.205040 | 0.024896 | 8.236 |

| quantized_max_pool3d | **19.343** |

### quantized_max_pool3d_out

| quint8 | [4, 3, 16, 56, 56] | 0.530688 | 0.037152 | 14.284 |
| quint8 | [8, 64, 8, 28, 28] | 0.859712 | 0.037872 | 22.700 |
| quint8 | [16, 128, 4, 14, 14] | 0.491648 | 0.037248 | 13.199 |
| quint8 | [32, 256, 2, 7, 7] | 0.320384 | 0.106624 | 3.005 |

| quantized_max_pool3d_out | **13.297** |

## Multi-backend Testing
|---|---|---|---|
| Nvidia (H20) | PASS (8 cases) | 19.343/13.297 | Primary |
| Tianshu | N/A | — | — |
| Muxi | N/A | — | — |
| Ascend | N/A | — | — |
| Hygon | N/A | — | — |

## Files Changed
- `src/flag_gems/ops/quantized_max_pool3d.py`: Triton kernel implementation
- `tests/test_quantized_max_pool3d.py`: Accuracy test
- `benchmark/test_quantized_max_pool3d.py`: Performance benchmark


src/flag_gems/ops/quantized_max_pool3d.py
```diff
@@ -0,0 +1,333 @@
...
+    )
+
+    out_mask = (out_h_offsets[:, None] < out_h) & (out_w_offsets[None, :] < out_w)
+    tl.store(output_block_ptr, max_val_acc, mask=out_mask)
+
+
+def quantized_max_pool3d(
+    input: torch.Tensor,
+    kernel_size,
+    stride=[],
```
Caeruleann, line 209: Please treat an empty stride list as kernel_size; the current default stride=[] is rejected by _parse_param, while ATen uses kernel_size when stride is omitted. Please add a default-stride parity test.

src/flag_gems/ops/quantized_max_pool3d.py
```diff
@@ -0,0 +1,333 @@
...
+        pw,
+        dd,
+        dh,
+        dw,
+    )
+
+    return torch._make_per_tensor_quantized_tensor(out_int, scale, zero_point)
+
+
+def quantized_max_pool3d_out(
```
Caeruleann, line 292: Please preserve the full ATen out= contract: resize the output metadata, update its qparams, validate dtype/device, and respect out.storage_offset(); resizing and writing the raw storage alone leaves mismatched outputs incorrect.

src/flag_gems/ops/quantized_max_pool3d.py
Caeruleann, line 1: Please cover the native-supported qint8 and qint32 dtypes; the current out path always reinterprets storage as uint8, which corrupts qint32 results. Either support these dtypes correctly or reject them explicitly.

src/flag_gems/ops/quantized_max_pool3d.py
```diff
@@ -0,0 +1,333 @@
...
+        output_size = (numerator + stride - 1) // stride + 1
+        # PyTorch-compatible adjustment for ceil_mode
+        if (output_size - 1) * stride >= in_size + padding:
+            output_size -= 1
+    else:
+        output_size = numerator // stride + 1
+    return output_size
+
+
+def _parse_pool3d_params(kernel_size, stride, padding, dilation):
```
Caeruleann, line 50: Please match ATen validation for positive kernel sizes, padding bounds, nonzero channel/spatial dimensions, and positive output sizes; the current implementation accepts or handles several inputs that native ATen rejects.

benchmark/test_quantized_max_pool3d.py
```diff
@@ -0,0 +1,150 @@
...
+    def torch_out_op(qx, out, **kwargs):
+        out_cpu = torch.quantize_per_tensor(
+            torch.zeros(out.shape, dtype=torch.float32, device="cpu"),
+            out.q_scale(),
+            out.q_zero_point(),
+            QDTYPE,
+        )
+        torch.ops.aten.quantized_max_pool3d.out(qx.to("cpu"), out=out_cpu, **kwargs)
+        out.copy_(out_cpu)
+        return out
```
Caeruleann, line 138: Please prepare CPU reference inputs and outputs outside the timed operation; the current baseline includes GPU-to-CPU transfer, allocation, and an extra copy, so it does not measure the same scope as the Gems operator.

src/flag_gems/ops/quantized_max_pool3d.py
Yukun-Cui: Fixed in 631a51db9.

You were right that the default was unusable: `_parse_param` rejected the empty list outright, so *any* call that omitted `stride` failed before reaching the kernel. On the old head:

```
gems stride=[]:       ValueError: Invalid stride: []
gems stride omitted:  ValueError: Invalid stride: []
gems stride=():       ValueError: Invalid stride: ()
ATen stride=[]  ->    shape (1, 2, 3, 3, 3)
```

`_check_maxpool3d_params` now returns `None` for an empty stride sequence and `_parse_pool3d_params` substitutes `kernel_size`, which is what ATen does immediately after the same check. All three spellings now agree with ATen:

```
gems stride=[]:       (1, 2, 3, 3, 3)
gems stride omitted:  (1, 2, 3, 3, 3)
gems stride=():       (1, 2, 3, 3, 3)
```

Parity tests added for both overloads: `test_quantized_max_pool3d_default_stride` covers `stride` in `{None, [], ()}` across `kernel_size` in `{2, 3, (2, 3, 4)}`, and `test_quantized_max_pool3d_out_default_stride` does the same for `out=`.

src/flag_gems/ops/quantized_max_pool3d.py
Yukun-Cui: Fixed in 631a51db9. Every part of this reproduced — the old path only worked when `out` already had the exact right shape, contiguous strides, and offset 0.

Old head vs. ATen:

```
out too small (1,2,1,1,1):  shape=(1,2,1,1,1) exact=False   # never resized
out too big   (1,2,8,8,8):  shape=(1,2,8,8,8) exact=False
out= qparams:               gems (0.9,7)  vs  aten (0.05,100)
out= dtype mismatch:        silently returned a qint8 tensor
out= storage_offset:        exact=False, sibling_untouched=False   # clobbered the neighbour
out= non-contiguous:        exact=False
```

The storage-offset case is the worst of these: writing the raw storage from byte 0 corrupted a sibling slice that had nothing to do with the call.

Now:

```
out too small:              shape=(1,2,2,2,2) exact=True qp=(0.05,100)
out too big:                shape=(1,2,2,2,2) exact=True qp=(0.05,100)
out= qparams:               gems (0.05,100) == aten (0.05,100)
out= dtype mismatch:        RuntimeError: Expected out tensor to have dtype c10::quint8, but got c10::qint8 instead
out= device mismatch:       RuntimeError: Expected out tensor to have device cuda:0, but got cpu instead
out= storage_offset:        exact=True, offset_kept=16, sibling_untouched=True
out= non-contiguous:        exact=True
```

Two implementation notes. `aten::resize_` and `set_.source_Tensor` are both unimplemented for QuantizedCUDA, so `_resize_quantized_out` rewrites the metadata via `set_(storage, offset, size, stride)` — preserving the storage offset, only ever growing the storage, and assigning contiguous strides, which is what `at::native::resize_output` does. And the values are written with `out.copy_(result)` rather than a raw storage copy, since `copy_` on a quantized tensor goes through `out`s own strides and adopts the source qparams. Dtype is checked before device to match the ordering in the generated `resize_out`.

Covered by `test_quantized_max_pool3d_out_resizes`, `_out_adopts_input_qparams`, `_out_dtype_mismatch`, `_out_preserves_storage_offset`, and `_out_non_contiguous`.

src/flag_gems/ops/quantized_max_pool3d.py
Yukun-Cui: Fixed in 631a51db9 — supported properly rather than rejected, since the ATen kernel dispatches over all three via `AT_DISPATCH_QINT_TYPES`.

Your diagnosis was exactly right about where it broke. The forward path happened to work for all three dtypes, but the `out=` path reinterpreted storage as uint8 unconditionally, so qint32 came back wrong:

```
out= torch.quint8:  exact=True
out= torch.qint8:   exact=True
out= torch.qint32:  exact=False    # <- storage read as uint8
```

There is now a `_QINT_DTYPES` map from each quantized dtype to its `int_repr()` dtype (`quint8->uint8`, `qint8->int8`, `qint32->int32`), and `_int_view` aliases the storage with the correct integer type, sizes and strides. Pooling on the integer representation is exact for all three because `real = (q - zero_point) * scale` is monotonically increasing in `q` for positive `scale`, so the max over integers is the max over reals and no requantization is needed.

```
fwd/out= quint8:  exact=True  qparams preserved
fwd/out= qint8:   exact=True  qparams preserved
fwd/out= qint32:  exact=True  qparams preserved
```

`test_quantized_max_pool3d_qint_dtypes` and `test_quantized_max_pool3d_out_qint_dtypes` parametrize over all three.

src/flag_gems/ops/quantized_max_pool3d.py
Yukun-Cui: Fixed in 631a51db9. Validation parity was 0/11 on the old head — every case either raised the wrong exception type, carried a different message, or (worse) silently succeeded:

```
kernel_size 0:            aten RuntimeError: kernel_size should be greater than zero.
                          gems ValueError: Invalid stride: []
padding over half kernel: aten RuntimeError: padding should be smaller than half of kernel_size.
                          gems ValueError: Invalid stride: []
output size zero:         aten RuntimeError: Given input size: (2t6x6x6). Calculated output size: (2t0x0x0). Output size is too small.
                          gems ValueError: Invalid stride: []
shape (1,0,6,6,6):        aten RuntimeError: input dimensions must be non-zero.
                          gems OK -> (1, 0, 3, 3, 3)          # silently accepted
```

Note most of those `Invalid stride: []` messages are the default-stride bug masking the real problem, since it fired before any value check.

The checks are now split to follow ATens own order, which matters when more than one argument is bad since it decides which message you get:

1. `_check_maxpool3d_params` — argument lengths and `dilation >= 1`
2. `_parse_pool3d_params` — positive kernel, then positive stride (both inside `q_maxpool_3d`, before the input is touched)
3. `_quantized_max_pool3d_output_shape` — rank 5, non-zero dims, padding vs. half the kernel, then positive output size
4. `pool3d_output_size` — non-zero stride and non-negative padding, as `pooling_output_shape` does

Now 11/11 on parameter cases and 3/3 on the zero-sized/rank cases, comparing exception message strings against ATen rather than just the type. `test_quantized_max_pool3d_invalid_params` runs 22 cases through both overloads and asserts `str(res_exc.value) == str(ref_exc.value)`.

One oddity worth flagging: ATens message for a bad stride length reads `Expected no strides or 3d strides, got2`, with no space before the number, and the dilation message says "1d or 3d" while the check only accepts 3 elements. Both are reproduced verbatim, with comments, so the assertions arent mistaken for typos on our side.

benchmark/test_quantized_max_pool3d.py
Yukun-Cui: Fixed in 631a51db9. Measured the overhead you described before changing anything, and it was roughly half the reported baseline for the plain op and about two thirds for `out=`:

```
shape                     old baseline  new baseline  transfer+alloc  overhead
(4, 8, 8, 32, 32)              0.142ms       0.074ms         0.068ms    48.1%
(8, 16, 16, 32, 32)            0.532ms       0.232ms         0.300ms    56.3%
(2, 32, 32, 64, 64)            1.938ms       0.840ms         1.098ms    56.7%

out= variant:            old out= base  new out= base    copy+alloc  overhead
(4, 8, 8, 32, 32)              0.242ms       0.080ms         0.162ms    66.8%
(8, 16, 16, 32, 32)            0.791ms       0.272ms         0.519ms    65.6%
(2, 32, 32, 64, 64)            2.582ms       0.892ms         1.690ms    65.5%
```

The host copy of the input is now taken once in the input function, and the `out=` benchmark pre-allocates the CPU out tensor there too, so the timed region is just each operators own pooling. The old `out.copy_(out_cpu)` write-back is gone as well — it was a host-to-device copy attributed to the baseline.

Two things I left alone deliberately, both now documented in the module comment. The comparison is still GPU-vs-CPU, because PyTorch ships no QuantizedCUDA kernel for this op, so there is no same-device reference to measure against. And because the baseline never touches the GPU, these numbers want `--mode wrapper`; the default `kernel` mode times with CUDA events, which cannot observe CPU work and reports a meaningless near-constant baseline. Under `--mode wrapper` both benchmarks pass, with speedups from 0.87x to 6.6x depending on shape.

Caeruleann (approved):

PR #5938, opened 2026-09-02
[KernelGen][Nvidia] Add quantized_max_pool1d operator with Triton kernel

Yukun-Cui:
## Summary
Adds a Triton kernel for `quantized_max_pool1d`. Applies 1D max pooling over a per-tensor quantized tensor.

## Testing

## Performance
Test command: `pytest benchmark/test_quantized_max_pool1d.py --level core` (NVIDIA H20)

### quantized_max_pool1d

| quint8 | [4, 1024] | 0.016000 | 0.009536 | 1.678 |
| quint8 | [16, 4096] | 0.016704 | 0.009504 | 1.758 |
| quint8 | [32, 8192] | 0.025568 | 0.012000 | 2.131 |
| quint8 | [64, 16384] | 0.060928 | 0.013632 | 4.469 |
| quint8 | [8, 3, 1024] | 0.016320 | 0.011904 | 1.371 |
| quint8 | [16, 8, 2048] | 0.027744 | 0.011232 | 2.470 |

| quantized_max_pool1d | **2.313** |

### quantized_max_pool1d_out

| quint8 | [4, 1024] | 0.022560 | 0.008448 | 2.670 |
| quint8 | [16, 4096] | 0.023904 | 0.008768 | 2.726 |
| quint8 | [32, 8192] | 0.032704 | 0.008800 | 3.716 |
| quint8 | [64, 16384] | 0.070240 | 0.010304 | 6.817 |
| quint8 | [8, 3, 1024] | 0.023424 | 0.008672 | 2.701 |
| quint8 | [16, 8, 2048] | 0.032608 | 0.009024 | 3.613 |

| quantized_max_pool1d_out | **3.707** |

## Multi-backend Testing
|---|---|---|---|
| Nvidia (H20) | PASS (12 cases) | 2.313/3.707 | Primary |
| Tianshu | N/A | — | — |
| Muxi | N/A | — | — |
| Ascend | N/A | — | — |
| Hygon | N/A | — | — |

## Files Changed
- `src/flag_gems/ops/quantized_max_pool1d.py`: Triton kernel implementation
- `tests/test_quantized_max_pool1d.py`: Accuracy test
- `benchmark/test_quantized_max_pool1d.py`: Performance benchmark


src/flag_gems/ops/quantized_max_pool1d.py
```diff
@@ -0,0 +1,260 @@
...
+    input,
+    kernel_size,
+    stride=[],
+    padding=[0],
+    dilation=[1],
+    ceil_mode=False,
+    *,
+    out=None,
+):
+    logger.debug("GEMS QUANTIZED_MAX_POOL1D FORWARD OUT")
```
Caeruleann, line 218:
```suggestion
    logger.debug("GEMS QUANTIZED_MAX_POOL1D_OUT")
```

src/flag_gems/ops/quantized_max_pool1d.py
```diff
@@ -0,0 +1,260 @@
...
+
+def quantized_max_pool1d(
+    input,
+    kernel_size,
+    stride=[],
+    padding=[0],
+    dilation=[1],
+    ceil_mode=False,
+):
+    logger.debug("GEMS QUANTIZED_MAX_POOL1D FORWARD")
```
Caeruleann, line 171:
```suggestion
    logger.debug("GEMS QUANTIZED_MAX_POOL1D")
```

src/flag_gems/ops/quantized_max_pool1d.py
```diff
@@ -0,0 +1,260 @@
...
+
+    l_out_offsets = l_block_idx * BLOCK_L + tl.arange(0, BLOCK_L)
+    out_mask = l_out_offsets < out_l
+
+    # Neutral element for max: the smallest representable value of the
+    # underlying integer dtype (0 for unsigned, -128 for int8, ...).
+    if is_unsigned:
+        neutral: tl.constexpr = 0
+    else:
+        neutral: tl.constexpr = -128
```
Caeruleann, line 116: Please use the actual minimum value of the underlying integer dtype for padding. qint32 currently also uses -128, so padded windows whose valid values are below -128 return an incorrect maximum. Please add a qint32 parity case with values below -128.

src/flag_gems/ops/quantized_max_pool1d.py
```diff
@@ -0,0 +1,260 @@
...
+    out_l = max_pool1d_output_size(
+        in_l, kernel_size, stride, padding, dilation, ceil_mode
+    )
+
+    out_shape = int_repr.shape[:-1] + (out_l,)
+    # Write the pooled integer values directly into ``out``'s underlying
+    # storage so the returned quantized tensor reflects the result.
+    contiguous_strides = torch.empty(
+        out_shape, dtype=int_dtype, device=out.device
+    ).stride()
```
Caeruleann, line 252: Please implement the full out= contract here: validate device/dtype, handle mismatched shape and quantization parameters correctly, and respect the output tensor's stride and storage offset. Creating a contiguous raw view at storage offset 0 can corrupt views and produce incorrect results for valid out tensors.

benchmark/test_quantized_max_pool1d.py
Caeruleann, line 1: Please don't use dequantize + FP32 max-pool + requantize as the Torch baseline, since it measures a different computation scope. Also, out.int_repr().copy_(...) does not update the quantized out tensor because int_repr() returns a separate tensor.

src/flag_gems/ops/quantized_max_pool1d.py
Yukun-Cui: Applied in 4d666c823. The forward log now reads `logger.debug("GEMS QUANTIZED_MAX_POOL1D")`, dropping the `FORWARD` suffix.

src/flag_gems/ops/quantized_max_pool1d.py
Yukun-Cui: Applied in 4d666c823. Now `logger.debug("GEMS QUANTIZED_MAX_POOL1D_OUT")` — the old text was `GEMS QUANTIZED_MAX_POOL1D FORWARD OUT`, which named the wrong thing for the `out=` overload.

src/flag_gems/ops/quantized_max_pool1d.py
Yukun-Cui: Fixed in 4d666c823. Confirmed with a qint32 case whose valid values are all below -128, exactly as you predicted:

```
input (scale=1.0, zp=0): [-1000, -2000, -1500, -900]
old head:  gems_int=[-128, -128]    aten_int=[-1000, -900]
new:       gems_int=[-1000, -900]   aten_int=[-1000, -900]
```

The pad value won every window, so the result was the padding constant rather than any real element. Also checked a window lying entirely inside the padding, where the same bug appeared:

```
old head:  gems=[-128]   aten=[-500]
new:       gems=[-500]   aten=[-500]
```

`_neutral_for(int_dtype)` now returns `torch.iinfo(int_dtype).min` and is passed to the kernel per launch, matching ATens `std::numeric_limits<T>::lowest()` on the underlying integer type — 0 for quint8, -128 for qint8, -2147483648 for qint32. quint8 and qint8 were already correct and stay so.

Parity tests added: `test_quantized_max_pool1d_padding_neutral` parametrizes over all three dtypes with values below -128, and `test_quantized_max_pool1d_all_padding_window` covers the fully-padded window.

src/flag_gems/ops/quantized_max_pool1d.py
Yukun-Cui: Fixed in 4d666c823. The corruption you predicted from the offset-0 contiguous view reproduced on every non-trivial `out`:

```
out too small (2,3,1):    shape=(2,3,1) exact=False      # never resized
out too big   (2,3,20):   shape=(2,3,20) exact=False
out= qparams:             stayed (0.9,7), should be input qparams (0.05,128)
out= dtype mismatch:      silently returned a qint8 tensor
out= device mismatch:     ValueError: Pointer argument (at 1) cannot be accessed from Triton (cpu tensor?)
out= storage_offset:      exact=False, sibling_untouched=False
out= strided view:        exact=False, gaps_untouched=False
out= transposed view:     exact=False
```

`gaps_untouched=False` is the one I would call out: writing a packed result into a `[..., ::2]` view overwrote the interleaved elements the caller never passed in.

Now:

```
out too small / too big:  shape=(2,3,6) exact=True qp=(0.05,128)
out= dtype mismatch:      RuntimeError: Expected out tensor to have dtype torch.quint8, but got torch.qint8 instead
out= device mismatch:     RuntimeError: Expected out tensor to have device cuda:0, but got cpu instead
out= storage_offset:      exact=True, sibling_untouched=True
out= strided view:        exact=True, gaps_untouched=True
out= transposed view:     exact=True
```

`_check_out` validates dtype and device, `_resize_out` handles shape mismatch, `_sync_out_quantizer` adopts the inputs scale/zero_point, and `_int_view_of(out)` aliases `out`s own strides and storage offset so the kernel writes through them rather than into a repacked buffer.

One case worth mentioning that was not in your list: ATen tolerates an `out` that aliases the input, but a kernel reading and writing the same bytes concurrently would race. `_may_overlap` detects that and stages through a temporary.

Covered by `test_quantized_max_pool1d_out_non_contiguous`, `_out_storage_offset`, `_out_resizes_and_adopts_quant_params`, `_out_rejects_mismatched_out`, and `_out_rejects_wrong_device`, each parametrized over the three dtypes where relevant.

benchmark/test_quantized_max_pool1d.py
Yukun-Cui: Both fixed in 4d666c823.

On the baseline: agreed that it measured the wrong scope. The old one timed two elementwise passes plus a float pool, which is much cheaper on a GPU than what the operator actually does, so the numbers flattered us:

```
shape              old(GPU float)  new(CPU int)
(4, 16, 512)             0.0231ms       0.0544ms
(8, 32, 1024)            0.0235ms       0.1536ms
(16, 8, 2048)            0.0239ms       0.0961ms
```

The baseline is now `torch.ops.aten.quantized_max_pool1d`, the op under test. One clarification on my side: the float path was *numerically* equivalent to ATen (max pooling is monotone under the affine quantization map, so I could not construct a mismatch — 0/96 elements differed even at an awkward scale/zero_point). So this is a measurement-scope fix, not a correctness one. The cost is that the reference is CPU-only, making the comparison cross-device; that tradeoff is documented in the baselines docstring, and `set_shapes`/`set_more_shapes` are pinned to the hand-picked shapes so `core_shapes.yaml` cannot inject shapes that would make a CPU reference impractical.

On `int_repr()`: confirmed a silent no-op, so the old `out=` baseline was returning an unmodified `out`.

```
out changed after out.int_repr().copy_(...):  False
out changed after out.copy_(...):             True
```

That path is gone — the baseline now calls `aten.quantized_max_pool1d.out` directly with a CPU `out`, so no write-back is needed. Both benchmarks pass under `--mode wrapper` (the right mode here, since a CPU baseline is invisible to CUDA-event timing), with speedups of 1.0x to 9.0x depending on shape.

src/flag_gems/ops/quantized_max_pool1d.py
Caeruleann: The overlap handling is still too late for out=input: _resize_out() changes the input shape before input.int_repr() is captured, so a shape-changing pool reads the resized/truncated input. Please preserve the original input before mutating out, and add parity coverage with out=input, e.g. L=8, kernel_size=2, stride=2.

tests/test_quantized_max_pool1d.py
```diff
@@ -0,0 +1,409 @@
...
+
+    # Pool the integer representation directly: comparing dequantized floats
+    # would trip over scale round-trip error rather than the layout.
+    ref_out = torch.nn.functional.max_pool1d(
+        ref_inp.int_repr().to(torch.float32), 2, stride=1, padding=1
+    )
+    res_out = flag_gems.quantized_max_pool1d(res_inp, 2, stride=1, padding=1)
+
+    assert tuple(res_out.shape) == tuple(ref_out.shape)
+    assert torch.equal(res_out.int_repr().to("cpu").to(torch.float32), ref_out)
```
Caeruleann: This test deliberately stops comparing against the native quantized operator. If the L=1 result is an ATen bug, please link an upstream issue or document an agreed parity exception; otherwise this case should still compare against torch.quantized_max_pool1d.

benchmark/test_quantized_max_pool1d.py
```diff
@@ -0,0 +1,140 @@
...
+):
+    return torch.ops.aten.quantized_max_pool1d.out(
+        q_tensor.cpu(),
+        [kernel_size],
+        [stride],
+        [padding],
+        [dilation],
+        ceil_mode,
+        out=out.cpu(),
+    )
```
Caeruleann: The Torch baseline calls q_tensor.cpu() inside the timed function, so its latency includes a GPU-to-CPU copy and the reported speedup is not an operator-to-operator comparison. Please prepare the CPU inputs outside the timed region and report the CPU reference latency separately.

tests/test_quantized_max_pool1d.py
Yukun-Cui: Restored the native-op comparison; this is an ATen CPU defect, verified experimentally and now documented in the test (commit a724c731).

Repro on torch 2.11.0 (torch.quantized_max_pool1d is CPU-only, so this is QuantizedCPU in all cases):
```python
q = torch.quantize_per_tensor(torch.arange(1.0, 7.0).reshape(2, 3, 1), 1.0, 0, torch.quint8)
torch.quantized_max_pool1d(q, 2, stride=1, padding=1).int_repr()
# [[[1, 2], [3, 1], [2, 3]], [[4, 5], [6, 4], [5, 6]]]   <- logical values, wrong
torch.nn.functional.max_pool1d(q.int_repr().float(), 2, stride=1, padding=1)
# [[[1, 1], [2, 2], [3, 3]], [[4, 4], [5, 5], [6, 6]]]   <- correct
```

Root cause (confirmed in ATen source, aten/src/ATen/native/quantized/cpu/Pooling.cpp): `quantized_max_pool1d` unsqueezes (N, C, 1) to (N, C, 1, 1) and dispatches to `q_maxpool_2d`. With L == 1 that 4D tensor is trivially channels-last-contiguous, so the kernel takes its NHWC fast path, writes the result in NHWC (here NLC) order, and returns a tensor whose strides nonetheless claim contiguous NCL layout. The bytes are right and the strides lie, so the logical values are wrong. Evidence: ATen's own raw buffer re-read as NLC (`int_repr().reshape(2, 2, 3).permute(0, 2, 1)`) reproduces the independent window-max reference exactly, for all three quantized dtypes. Only 3D inputs with L == 1 and C > 1 are affected; 2D (N, L) and 3D C==1 or L>1 inputs match the reference exactly (verified case by case). I found no existing upstream issue (searched pytorch/pytorch; only #116323, a closed segfault, mentions this op).

Fix: the test now compares against `torch.quantized_max_pool1d` again, as requested. Where the native op hits the defect, the test re-reads its mislabelled buffer in the order it was actually written and asserts that this repair reproduces an independent float max_pool1d reference exactly; a mismatch outside the documented trigger domain fails the test instead of being repaired. So the native operator is once again the reference, with the ATen stride defect documented in-line with reproducible evidence.

benchmark/test_quantized_max_pool1d.py
Yukun-Cui: Fixed in commit a724c731: both baseline functions now take a pre-staged host input, and the device-to-host copy is out of the timed region.

What changed in benchmark/test_quantized_max_pool1d.py:
- `_make_quantized` quantizes on the host first and stores the CPU tensor on the device tensor as `_cpu_twin`; `_cpu_input(q_tensor)` returns that twin, falling back to `.to("cpu")` only when there is no twin (e.g. a `device == "cpu"` run).
- `torch_quantized_max_pool1d` and `torch_quantized_max_pool1d_out` call the aten CPU op on the twin instead of calling `q_tensor.cpu()` / `out.cpu()` inside the timed function.
- The `out=` baseline's host out tensor is likewise pre-allocated in the input function, so neither its allocation nor a host/device copy is timed.
This matches the pattern the sibling benchmark/test_quantized_max_pool3d.py already uses for the same CPU-only reference.

Measured on a (32, 8192) quint8 input, averaging 30 iterations after warmup:
- baseline with `.cpu()` inside the timed function: 0.136 ms
- baseline with the host input hoisted:           0.075 ms
- the `.cpu()` transfer alone:                     0.044 ms
So the transfer was ~59% of the measured baseline latency; since `speedup = latency_base / latency`, the inflated baseline overstated the speedup for that shape by roughly 1.8x.

Also worth noting: the benchmark did not run at all before this commit — both cases failed at construction with `ValueError: GenericBenchmark requires input_fn or case_fn/build_inputs_fn`, because a custom `get_input_iter` is no longer sufficient to build the harness (the benchmark passed `input_fn=None`). The rewrite switches to the `input_fn`-based form used by the 3d benchmark. Both cases now execute and pass.

One caveat kept in the file's header comment: since the reference is QuantizedCPU and the kernel is on GPU, the comparison is inherently GPU-vs-CPU, and the default `kernel` mode times with CUDA events, which cannot observe work that never reaches the GPU — so `--mode wrapper` is the meaningful setting for these numbers.

src/flag_gems/ops/quantized_max_pool1d.py
Yukun-Cui: Confirmed and fixed in e5ad6a21 (source 51ed4f10). You were right, and the failure is worse than "reads the truncated input".

Reproduced on a724c731 with `out=input` and the parameters you gave:

```
L=8 k=2 s=2 (8->4, shrink):
  aten: [13, 82, 98, 50]
  got : [13, 82,  0,  0]   exact=False
3D (2,3,8) k=2 s=2:
  aten: [[19,51,70,67],[53,71,84,65],...]
  got : [[19,51, 0, 0],[70,67, 0, 0],...]   exact=False
```

Two things beyond the truncation you described, both from the same ordering:

- The tail is not just truncated, it is *stale*. `_resize_out` keeps the storage offset and only grows the storage when it must, so for a shrinking pool the trailing outputs were read from bytes left over from the resize and the write-back.
- **Growing the output is a total loss, not a partial one.** With `L=8, k=2, s=1, p=1` the pool computes 9 outputs, which no longer fits the 8-element storage, so `_resize_out` allocates a fresh storage and `set_`s the input onto it. The original bytes are gone, and every output reads back 0 -- 8/8 trials:

```
L=8 k=2 s=1 p=1 (8->9, grow):
  aten: [58, 99, 99, 57, 76, 76, 75, 98, 98]
  got : [ 0,  0,  0,  0,  0,  0,  0,  0,  0]   ALL ZEROS, 8/8 trials
```

Fix: `input.int_repr().contiguous()` is now captured *before* `_resize_out`/`_sync_out_quantizer`, exactly as you asked. `int_repr()` returns a fresh contiguous copy (verified on CPU and CUDA for all three dtypes, contiguous and strided inputs), so the snapshot is independent of both the resize and the kernel's write-back. Same matrix after the fix:

```
shrink 8->4 k=2 s=2:    exact=True
grow   8->9 k=2 s=1 p=1: exact=True, 0/8 mismatches
3D (2,3,8) k=2 s=2:     exact=True
```

One honest note on the aliasing branch: with the snapshot in place the staging buffer is no longer load bearing. I checked by forcing `_may_overlap` false and writing straight into the aliased out -- 0/6 mismatches, because the kernel only ever reads the copy. I kept the staging and corrected its comment rather than removing it, since it makes the aliasing case explicit instead of resting on that indirect property. Say the word if you would rather see it dropped.

Coverage added as `test_quantized_max_pool1d_out_is_input`, parametrized over the three quantized dtypes and three cases: your `(1, 8) k=2 s=2` shrink, a 3D `(2, 3, 8)` shrink, and a `(1, 8) k=2 s=1 p=1` grow (the reallocation case). It asserts the shape actually changes, so the ordering can never go untested, and compares `int_repr()` for integer equality -- a truncated *or* zeroed input fails.

Verified: all 9 new cases fail against a724c731 (`AssertionError` at the comparison) and pass with the fix; full file 319 passed. `pre-commit` clean on both files (black/isort/flake8/EOF/trailing-whitespace all Passed, no rewrites), `check_kernelgen_tests` reports no forbidden `use_gems()`.

Yukun-Cui: Done in 51ed4f10 + e5ad6a21 (pushed to `pr/quantized_max_pool1d`).

The root cause was exactly the ordering you described: `_resize_out(out, out_shape)` ran before `input.int_repr()` was captured, so with `out is input` and a shape-changing pool, the pool read the resized tensor -- truncated (stale at the tail) when the result is smaller, all zeros when it grows.

Fix: the integer representation is snapshotted **before** anything touches `out`. `int_repr()` returns a fresh contiguous copy, so the snapshot is immune to the resize and to the kernel's own writes (which may alias the input). The explicit staging for the aliasing case is kept, though with the snapshot in place it is belt-and-braces rather than load bearing -- the kernel now reads the copy and never `out`'s bytes.

New parity coverage: `test_quantized_max_pool1d_out_is_input` includes the exact case you asked for -- L=8, kernel_size=2, stride=2 (the shrinking case), plus a growing case, asserting the output values and that the aliasing `out` carries the updated qparams. 319 tests pass locally; pre-commit clean.

Caeruleann:
solve the ci problems

Caeruleann (approved):

PR #5941, opened 2026-09-02
[KernelGen][Nvidia] Add quantize_per_tensor operator with Triton kernel

Yukun-Cui:
## Summary
Adds a Triton kernel for `quantize_per_tensor`. Quantizes a float32 tensor to a quantized tensor (torch.quint8 / qint8 / qint32) with the given scale and zero point, using round-half-to-even rounding and clamping to the representable integer range.

## Testing

## Performance
Test command: `pytest benchmark/test_quantize_per_tensor.py --level core` (NVIDIA H20)

### quantize_per_tensor

| float32 | [64, 64] | 0.007744 | 0.079584 | 0.097 |
| float32 | [1024, 1024] | 0.037888 | 0.079648 | 0.476 |
| float32 | [4096, 4096] | 0.498400 | 2.624992 | 0.190 |
| float32 | [64, 512, 512] | 0.498448 | 2.623424 | 0.190 |
| float32 | [1024, 1024, 1024] | 70.785789 | 34.360542 | 2.060 |

| quantize_per_tensor | **0.603** |

### quantize_per_tensor_out

| float32 | [64, 64] | 0.009760 | 0.104368 | 0.094 |
| float32 | [1024, 1024] | 0.040512 | 0.088000 | 0.460 |
| float32 | [4096, 4096] | 0.507360 | 0.300640 | 1.688 |
| float32 | [64, 512, 512] | 0.507328 | 0.300192 | 1.690 |
| float32 | [1024, 1024, 1024] | 32.005249 | 15.286016 | 2.094 |

| quantize_per_tensor_out | **1.205** |

## Multi-backend Testing
|---|---|---|---|
| Nvidia (H20) | PASS (10 cases) | 0.603/1.205 | Primary |
| Tianshu | N/A | — | — |
| Muxi | N/A | — | — |
| Ascend | N/A | — | — |
| Hygon | N/A | — | — |

## Files Changed
- `src/flag_gems/ops/quantize_per_tensor.py`: Triton kernel implementation
- `tests/test_quantize_per_tensor.py`: Accuracy test
- `benchmark/test_quantize_per_tensor.py`: Performance benchmark


src/flag_gems/ops/quantize_per_tensor.py
```diff
@@ -0,0 +1,129 @@
...
+    n_elements,
+    BLOCK_SIZE: tl.constexpr,
+    USE_FP64: tl.constexpr,
+):
+    pid = tl.program_id(axis=0)
+    block_start = pid * BLOCK_SIZE
+    offsets = block_start + tl.arange(0, BLOCK_SIZE)
+    mask = offsets < n_elements
+    x = tl.load(x_ptr + offsets, mask=mask, other=0.0)
+    inv_scale = tl.load(inv_scale_ptr)
```
Caeruleann: Please handle non-contiguous inputs here. The kernel indexes x_ptr + offsets as contiguous storage, so transposed or sliced tensors are quantized in the wrong logical order. Please add non-contiguous parity tests as well.

src/flag_gems/ops/quantize_per_tensor.py
```diff
@@ -0,0 +1,129 @@
...
+    if USE_FP64:
+        x = x.to(tl.float64)
+        inv_scale = inv_scale.to(tl.float64)
+    else:
+        x = x.to(tl.float32)
+        inv_scale = inv_scale.to(tl.float32)
+    # int_value = round_half_even(x * (1 / scale)) + zero_point, then clamp.
+    q = tl_extra_shim.rint(x * inv_scale) + zero_point
```
Caeruleann: Please match ATen's nearbyint(raw_val / scale) expression instead of multiplying by a precomputed reciprocal, as the two are not bitwise equivalent near halfway values. Please add deterministic tie-boundary and nextafter parity tests.

src/flag_gems/ops/quantize_per_tensor.py
```diff
@@ -0,0 +1,129 @@
...
+
+
+def quantize_per_tensor_out(a, scale, zero_point, dtype, *, out=None):
+    logger.debug("GEMS QUANTIZE_PER_TENSOR_OUT")
+    result = quantize_per_tensor(a, scale, zero_point, dtype)
+    # ``out`` is a quantized tensor; ``copy_`` propagates both the integer
+    # storage and the quantization parameters (scale / zero_point) onto it,
+    # so the returned object is the same ``out`` with updated contents.
+    out.copy_(result)
+    return out
```
Caeruleann, line 178: Please implement the native out= resize behavior before writing the result. A differently shaped out is resized by ATen, while the current out.copy_(result) raises a broadcast error; it also invokes a PyTorch computation op directly from the host wrapper.

src/flag_gems/ops/quantize_per_tensor.py
```diff
@@ -0,0 +1,129 @@
...
+        qmin=qmin,
+        qmax=qmax,
+        n_elements=n_elements,
+        BLOCK_SIZE=BLOCK_SIZE,
+        USE_FP64=_USE_FP64,
+    )
+    return out
+
+
+def quantize_per_tensor(a, scale, zero_point, dtype):
```
Caeruleann, line 139: Please validate that a.dtype is torch.float32 here, since ATen rejects other input dtypes while this implementation silently converts and accepts them.

src/flag_gems/ops/quantize_per_tensor.py
Yukun-Cui: Fixed — the kernel now calls `.contiguous()` before launch, since it walks storage with a flat offset and would otherwise read a transposed or sliced view in the wrong order.

`test_quantize_per_tensor_non_contiguous` covers transpose, strided slice (`[:, ::2]`), and narrow across all three quantized dtypes, asserting `not view.is_contiguous()` so the test cannot silently stop exercising the path.

src/flag_gems/ops/quantize_per_tensor.py
Yukun-Cui: I measured this before changing it, and the reciprocal turns out to be the form that matches ATen — switching to `nearbyint(raw_val / scale)` would introduce a mismatch rather than remove one.

ATen's quantized CUDA path narrows the double `scale` to float, takes the **fp32 reciprocal once**, then multiplies. On values engineered to land on exact `k + 0.5` quotients:

```
fp32 reciprocal multiply :      0 / 307200 mismatches
fp32 true division       :  39258 / 307200 mismatches
fp64 (either form)       : ~19% of half-way values mismatch
```

The forms really do differ. For `x=31.508630752563477`, `scale=0.14897697696685788`:

```
x * (1/scale) = 211.49998  -> 211   <- ATen returns 211
x / scale     = 211.5      -> 212
```

So the real bug here was the **fp64**, not the reciprocal. The comment above the kernel claimed fp64 reproduced the reference bit-for-bit; it does the opposite (`x=0.35, scale=0.1`: ATen 4, fp64 3). The computation is now fp32 end to end and bit-exact on 3686400 elements across quint8/qint8/qint32.

Two Triton details were needed to get there, both noted in the code:

- plain `/` on fp32 lowers to the fast approximate reciprocal — a third, separately-wrong answer (`140.50001525878906` where fp32 gives exactly `140.5`). `tl_extra_shim.div_rn` is the IEEE-correct division if true division is ever wanted.
- passing the scale as a Python float binds it as fp64 and promotes the multiply **even with a `.to(tl.float32)` inside the kernel**, so the reciprocal is carried in a 0-dim fp32 tensor and `tl.load`ed.

`test_quantize_per_tensor_half_way_values` locks this in; note that uniformly random inputs miss it entirely (0 mismatches out of 1.5M with random scales), which is presumably why it survived this long.

Happy to switch to `div_rn` division instead if you prefer the expression to read like the ATen source, but it would be knowingly less accurate.

src/flag_gems/ops/quantize_per_tensor.py
Yukun-Cui: Added validation — `a.dtype != torch.float32` now raises, matching ATen ("Quantize only works on Float Tensor") instead of silently proceeding. `test_quantize_per_tensor_rejects_non_float32` covers float64, float16, and bfloat16.

The unsupported-quantized-dtype case is now an explicit error too, rather than a `KeyError` from the `_QUANT_INFO` lookup.

src/flag_gems/ops/quantize_per_tensor.py
Yukun-Cui: Implemented, with one deviation worth flagging: `out=` now validates dtype and device up front (dtype against the *requested* quantized dtype, which is what ATen checks — it raises "Expected out tensor to have dtype c10::quint8, but got c10::qint8").

On resizing, ATen does attempt it, but the resize itself is not supported for quantized CUDA tensors — the native op emits the deprecation warning and then fails:

```
NotImplementedError: Could not run 'aten::resize_' with arguments
from the 'QuantizedCUDA' backend.
```

So there is no native resize behavior to match on this backend. A mismatched shape now raises with an explanatory message rather than silently producing a wrong-shaped result. If you would rather it mirror ATen's exact failure (warn, then `NotImplementedError`), I can do that instead.

Confirmed the qparams part of the contract holds: `out` adopts the requested scale/zero_point and the same object is returned. `test_quantize_per_tensor_out_dtype_mismatch` and `test_quantize_per_tensor_out_shape_mismatch` cover the rejections.

src/flag_gems/ops/quantize_per_tensor.py
```diff
@@ -0,0 +1,178 @@
...
+
+    storage_dtype, qmin, qmax = _QUANT_INFO[dtype]
+    n_elements = a.numel()
+    out = torch.empty(a.shape, dtype=storage_dtype, device=a.device)
+    if n_elements == 0:
+        return out
+
+    # The kernel walks storage with a flat offset, so a transposed or sliced view
+    # would be read in the wrong order; materialize a dense copy first.
+    a = a.contiguous()
```
Caeruleann: Please preserve the input’s suggested memory format here. ATen allocates the quantized output and materializes the input using rtensor.suggest_memory_format(), while torch.empty(a.shape, ...) and a.contiguous() always produce the default contiguous format. Please add channels_last and channels_last_3d parity tests that check both values and strides/memory format.

src/flag_gems/ops/quantize_per_tensor.py
Caeruleann: The shape-mismatch explanation makes sense. However, the host-side out.copy_(result) issue remains: FlagGems copy_ redispatches quantized tensors to ATen, so this path still performs native O(n) work after the Triton kernel. Could this be implemented through a device-side FlagGems path instead?

src/flag_gems/ops/quantize_per_tensor.py
Yukun-Cui: Fixed in abf17fae. ATen does materialize the input with `rtensor.suggest_memory_format()` and allocates the quantized output in that format, so a channels-last input should give a channels-last int_repr. `suggest_memory_format` has no Python binding, so I mirrored its rule (`_suggest_memory_format`) and validated the mirror against the native op on contiguous, channels-last, channels-last-3d, C=1, N=1, transposed and sliced-channels-last inputs -- all nine produced the same strides as `torch.quantize_per_tensor`.

Added `test_quantize_per_tensor_memory_format` for channels-last and channels-last-3d (checking values *and* strides) plus a contiguous-unchanged case. 346 tests pass.

One limitation I want to flag rather than paper over: `torch._make_per_tensor_quantized_tensor`, which the public `flag_gems.quantize_per_tensor` wrapper uses to build the quantized tensor, silently re-lays-out its argument to contiguous on CUDA (measured: the impl's int_repr comes back channels-last with stride `(60,1,15,3)`, and after the wrapper it is `(60,20,5,1)`; on CPU it is preserved). So the format is preserved through the implementation but not through that wrapper on CUDA. My new test asserts the strides on `_quantize_per_tensor_impl` and the values through the public entry point. Fixing the wrapper end-to-end needs a public replacement for that private API -- the only public constructor is `aten::quantize_per_tensor` itself, which would recurse into this override. Happy to take that on as a follow-up if you want it in scope here.

src/flag_gems/ops/quantize_per_tensor.py
Yukun-Cui: Done -- device-side path, de16fd2c (pushed as c1e56f0c after pre-commit reformat).

**Verification of the report first.** Instrumenting the old `out.copy_(result)` with a `TorchDispatchMode` showed `aten.copy_.default` dispatched on every `out=` call, including the matching-qparams one -- `copy.py`'s `_can_use_triton` returns False for `dst.is_quantized`, so it always fell back to ATen. The claim was accurate.

**What replaced it:**

- *Non-aliasing out* (the normal case): a new `quantize_per_tensor_strided_kernel` stores straight through `out`'s own shape/strides/storage offset, reading the input through its strides as well -- no host-side `contiguous()` either. A `TorchDispatchMode` count over this path now shows **0 `copy_` dispatches**; the only remaining ATen ops are O(1) metadata (`q_scale`, `set_`, `empty`). The quantizer is re-pointed with the `_adopt_quantizer` / `.data` + `set_` approach from PR 6005, since a kernel write moves integers but not the quantizer.
- *Aliasing out* (`out` shares storage with the float input, where the kernel cannot read and write at once): the integers move device-side through FlagGems' Triton `copy_` over the integer views -- 1 Triton copy_, 0 native ones.
- Two rejections the copy used to supply are now raised directly, matching native: a per-channel `out` gets "Quantized Copy only works with same qscheme", and an internally-overlapping (stride-0) `out` gets "more than one element of the written-to tensor refers to a single memory location" -- without the guard, that last case silently wrote whichever element landed last, where native raises.

Two performance details surfaced while replacing the copy, both measured on H20:

- Passing sizes/strides as **scalar kernel arguments** (padded to 8 dims, matching the `replication_pad3d` convention) instead of pointer tensors saves 39us of per-call metadata allocation on the host against 42us of GPU time for a 1024x1024 buffer -- the pointer-tensor version was ~2x slower end-to-end on small buffers purely from allocation.
- An `IDX32` constexpr picks 32-bit index arithmetic when both sides' full spans fit int32 (checked in `_int32_safe`): 15% faster on a 4-D stride-2 destination (293us vs 348us at 7.9M elements).

End-to-end vs the old native-`copy_` path (same inputs, GPU events, median of 5 rounds): contiguous 1024x1024 71.4 -> 65.5us, 4096x4096 650.6 -> 628.8us, 4-D (16,128,64,60) 322.2 -> 308.9us; 4-D strided 335.8 -> 325.9us, 4096x4096 strided 680.1 -> 648.5us. The 2-D strided case is 73.6 -> 75.7us, i.e. par: the native quantized `copy_` there is a single fused 11us kernel, and the Triton side pays the div/mod for a genuinely scattered 2-byte-strided store (the alternative -- flat kernel into a temp + Triton `copy_` -- costs 99us, and the direct strided kernel 43us, so the direct store is the better of the two device-side options everywhere).

**Tests** (372 in the file, all passing; local run also bit-exact against native on 19 out-configuration cases): native-comparison across 6 `out` layouts (contiguous / every-other-column / transposed / storage-offset / narrowed / channels-last) x 3 quantized dtypes; neighbour preservation in a sentinel-filled buffer (interleaved columns and the offset view's untouched slice keep their sentinel, base qparams unchanged); mismatched-qparams adoption; per-channel and stride-0 rejections; aliasing `out`; and a dispatch-counting guard that fails if a native `copy_` is ever reintroduced or the flat kernel serves a strided destination -- verified it catches both regressions when injected.

Yukun-Cui: Done in c1e56f0c (pushed to `pr/quantize_per_tensor`). `out.copy_(result)` is gone from the `out=` path.

What changed:
- Non-aliasing `out` (including the qparams-mismatch case, which is the common path): the quantize kernel writes the integer values straight into `out`'s own storage, honouring `out`'s strides and storage offset, so no separate result buffer and no `copy_` at all. The quantizer is re-pointed at `out` via `set_` + `.data`, which preserves the tensor object identity and its storage while swapping in the new qparams.
- Aliasing `out` (`out` overlaps the float input): the integer values must survive the kernel reading the input, so this case stages one device copy -- but on the *integer views*, which lands on the FlagGems Triton `copy_` rather than the ATen redispatch.
- The per-channel rejection ("Quantized Copy only works with same qscheme") and the stride-0 overlap rejection are now raised explicitly, mirroring the two native rejections that `copy_` used to provide implicitly.

Dispatch-count guard test: a `TorchDispatchMode` harness asserts zero `aten.copy_` dispatches on the non-aliasing path (the path you flagged) and exactly one Triton `copy_` on the aliasing path, so a regression back to the native copy would fail the suite.

New coverage: non-contiguous `out` (slice/transposed/channels-last), `out` with a non-zero storage offset, and qparams-mismatched `out`. 372 tests pass locally; pre-commit clean.

src/flag_gems/ops/quantize_per_tensor.py
```diff
@@ -0,0 +1,545 @@
...
+        a = a.contiguous(memory_format=memory_format)
+    # One thread per element; 1024 is a standard block size that keeps the
+    # launch grid small for the typical quantization buffer sizes.
+    return _launch_quantize_kernel(a, out, scale, zero_point, qmin, qmax)
+
+
+def quantize_per_tensor(a, scale, zero_point, dtype):
+    logger.debug("GEMS QUANTIZE_PER_TENSOR")
+    int_repr = _quantize_per_tensor_impl(a, scale, zero_point, dtype)
+    return torch._make_per_tensor_quantized_tensor(
```
Caeruleann: _make_per_tensor_quantized_tensor normalises the strides, so the functional path loses the channels-last layout the kernel carefully produced — a CL input yields a contiguous int_repr while native keeps (768, 1, 48, 3) (same for CL3d). The memory-format test only asserts strides on _quantize_per_tensor_impl's internal return, which is why this slips through. Please build the public result via the _adopt_quantizer set_ trick and assert int_repr().stride() against native on the public entry point.

src/flag_gems/ops/quantize_per_tensor.py
```diff
@@ -0,0 +1,545 @@
...
+# Mapping from a quantized dtype to its integer storage dtype, used to alias a
+# quantized tensor's storage as an integer tensor a Triton kernel can write.
+_QINT_DTYPES = {
+    torch.quint8: torch.uint8,
+    torch.qint8: torch.int8,
+    torch.qint32: torch.int32,
+}
+
+
+def _validate_quant_args(a, dtype):
```
Caeruleann, line 248: Native validates the zero-point range (RuntimeError: quantize_tensor_per_tensor_affine zero_point 300 is above upper bound), but this path silently clamps an out-of-range zero_point into a valid bin — no error at all. Please add the range check in _validate_quant_args mirroring ATen's message.

src/flag_gems/ops/quantize_per_tensor.py
Yukun-Cui: 已修（4b1b6afd）。你说得对，问题正出在 public 入口：`torch._make_per_tensor_quantized_tensor` 在 CUDA 上会静默把 stride 归一化成连续，所以 kernel 辛苦产出的 channels-last 布局在出口被抹掉。

实测确认：

```
CL 输入 (2,4,5,5)  native int_repr stride = (100, 1, 20, 4)
                   gems   修复前         = (100, 25, 5, 1)   ← 变连续
                   gems   修复后         = (100, 1, 20, 4)   ✓
CL3d (1,4,2,3,4)   native = (96, 1, 48, 16, 4)  修复后一致   ✓
```

改法是按你建议用 `_adopt_quantizer` 的 `_empty_affine_quantized` + `set_` 组合（这里需要 storage dtype → quantized dtype 的反查表，因为 `_empty_affine_quantized` 只接受 quantized dtype）。同时把 `test_quantize_per_tensor_memory_format` 的断言从 impl 内部返回值改成断言 **public 入口** 的 `int_repr().stride()`，并补了 contiguous 与 CL3d 两个 case——正如你指出的，原测试只断言 impl 内部返回，所以这个漏洞溜过去了。

`tests/test_quantize_per_tensor.py`: 379 passed。

src/flag_gems/ops/quantize_per_tensor.py
Yukun-Cui: 已修（4b1b6afd）。加在 `_validate_quant_args` 之后的新 `_check_zero_point_range`，function 与 `.out` 两个入口都调用，消息逐字对齐 native 并带上具体值：

```
quantize_tensor_per_tensor_affine zero_point 300 is above upper bound.
quantize_tensor_per_tensor_affine zero_point -1 is below lower bound.
```

bound 按 dtype 取（quint8 0..255、qint8 -128..127、qint32 全 int32 区间），所以 e.g. `zp=128` 对 quint8 合法、对 qint8 报 above——与 native 一致。

新增 `test_quantize_per_tensor_zero_point_range`（4 组 dtype/zp/方向，同时覆盖 function 与 out= 两个入口，并与 native 消息逐字比对）。

conf/operators.yaml
```diff
@@ -10964,6 +10964,33 @@ ops:
...
+    clamping to the representable integer range.
+  for:
+  - quantize_per_tensor
+  labels:
+  - aten
+  - KernelGen
+  kind:
+  - Math
+  stages:
+  - alpha: '5.4'
```
```suggestion
  - alpha: '5.5'
```

conf/operators.yaml
```diff
@@ -10964,6 +10964,33 @@ ops:
...
+    quantized tensor. Computed with the same Triton kernel as the base variant.
+  for:
+  - quantize_per_tensor.out
+  labels:
+  - aten
+  - KernelGen
+  kind:
+  - Math
+  stages:
+  - alpha: '5.4'
```
```suggestion
  - alpha: '5.5'
```

Caeruleann (approved):

PR #5943, opened 2026-09-03
[KMCompiler]update hygon linalg_solve_triangular

YangLong114514:
### PR Category

Operator 

### Type of Change

Bug Fix

### Description

The generic implementation will report an error when running on the Hygon platform:

```bash
err = OutOfResources(81920, 65536, 'shared memory')

    def raise_(err):
        # clone the exception object so that the one saved in the closure
        # of the partial function below doesn't get assigned a stack trace
        # after the subsequent raise. otherwise, the CompiledKernel instance
        # saved in the (global) kernel cache will keep references to all the
        # locals in the traceback via the exception instance in the closure.
        cloned_err = copy.deepcopy(err)
        self._run = functools.partial(_raise_error, cloned_err)
>       raise err
E       triton.runtime.errors.OutOfResources: out of resource: shared memory, Required: 81920, Hardware limit: 65536. Reducing block sizes or `num_stages` may help.
```
When n= 512, the Hygon-specific large tile scheduling is triggered (lines 620-624 in linalg_solve_triangular. py; the second round of Hygon sweep adopts the tuning configuration for n≥512):
```python
  BLOCK_SIZE = 64 if n >= 512 else 32
  K_SLICE   = 16 if n >= 512 else 8
```
  In this configuration, the shared memory requirement of the fp64 kernel after compile is 81920 B (80 KB), which exceeds the 64 KB upper limit of the Hygon CTA.

fixed : set BLOCK_SIZE=32 while dtype=fp64.

### Performance

speedup in hygon : **0.99x**


src/flag_gems/runtime/backend/_hygon/ops/linalg_solve_triangular.py
```diff
@@ -0,0 +1,683 @@
...
+    B_view = B_view.reshape(orig_shape)
+    if out is not None:
+        out.copy_(B_view)
+        return out
+    return B_view
+
+
+def linalg_solve_triangular_out(
+    A, B, *, upper, left=True, unitriangular=False, out=None
+):
```
Caeruleann, line 688: add logger.

Caeruleann (approved):

PR #5961, opened 2026-09-03
[KernelGen][Nvidia] Add slice_copy operator with Triton kernel

Yukun-Cui:
## Summary
Adds a Triton kernel for `slice_copy`. Performs the same operation as `torch.slice`, but all output tensors are freshly created instead of aliasing the input. Returns a new tensor containing the selected slice of `self` along the given dimension.

## Testing

## Performance
Test command: `pytest benchmark/test_slice_copy.py --level core` (NVIDIA H20)

### slice_copy

| float16 | [64, 64] | 0.007344 | 0.005568 | 1.319 |
| float16 | [256, 256] | 0.007232 | 0.005728 | 1.263 |
| float16 | [1024, 1024] | 0.007872 | 0.006528 | 1.206 |
| float16 | [4096, 4096] | 0.023200 | 0.013824 | 1.678 |
| float16 | [1024, 65536] | 0.071328 | 0.036672 | 1.945 |
| float32 | [64, 64] | 0.005952 | 0.005568 | 1.069 |
| float32 | [256, 256] | 0.006304 | 0.005824 | 1.082 |
| float32 | [1024, 1024] | 0.007584 | 0.007136 | 1.063 |
| float32 | [4096, 4096] | 0.026112 | 0.018912 | 1.381 |
| float32 | [1024, 65536] | 0.084016 | 0.061664 | 1.362 |
| bfloat16 | [64, 64] | 0.007392 | 0.005568 | 1.328 |
| bfloat16 | [256, 256] | 0.007552 | 0.005728 | 1.318 |
| bfloat16 | [1024, 1024] | 0.008288 | 0.006560 | 1.263 |
| bfloat16 | [4096, 4096] | 0.022816 | 0.013472 | 1.694 |
| bfloat16 | [1024, 65536] | 0.071296 | 0.036160 | 1.972 |

| slice_copy | **1.396** |

### slice_copy_out

| float16 | [64, 64] | 0.008960 | 0.005568 | 1.609 |
| float16 | [256, 256] | 0.009184 | 0.005696 | 1.612 |
| float16 | [1024, 1024] | 0.010784 | 0.006464 | 1.668 |
| float16 | [4096, 4096] | 0.028288 | 0.013472 | 2.100 |
| float16 | [1024, 65536] | 0.089664 | 0.036288 | 2.471 |
| float32 | [64, 64] | 0.007872 | 0.005568 | 1.414 |
| float32 | [256, 256] | 0.008416 | 0.005824 | 1.445 |
| float32 | [1024, 1024] | 0.009792 | 0.007008 | 1.397 |
| float32 | [4096, 4096] | 0.035072 | 0.018464 | 1.899 |
| float32 | [1024, 65536] | 0.118848 | 0.055840 | 2.128 |
| bfloat16 | [64, 64] | 0.008928 | 0.005536 | 1.613 |
| bfloat16 | [256, 256] | 0.009184 | 0.005728 | 1.603 |
| bfloat16 | [1024, 1024] | 0.010848 | 0.006336 | 1.712 |
| bfloat16 | [4096, 4096] | 0.028096 | 0.013408 | 2.095 |
| bfloat16 | [1024, 65536] | 0.089696 | 0.036128 | 2.483 |

| slice_copy_out | **1.817** |

## Multi-backend Testing
|---|---|---|---|
| Nvidia (H20) | PASS (30 cases) | 1.396/1.817 | Primary |
| Tianshu | N/A | — | — |
| Muxi | N/A | — | — |
| Ascend | N/A | — | — |
| Hygon | N/A | — | — |

## Files Changed
- `src/flag_gems/ops/slice_copy.py`: Triton kernel implementation
- `tests/test_slice_copy.py`: Accuracy test
- `benchmark/test_slice_copy.py`: Performance benchmark


tests/test_slice_copy.py
Caeruleann, line 1: Please cover all natively supported dtypes, especially FP64, bool, and complex types. Since Triton does not expose complex values as a scalar type, the current kernel needs an explicit bitwise-copy path for complex inputs instead of relying on the generic load/store path.

src/flag_gems/ops/slice_copy.py
Caeruleann, line 1: Please support the full out= contract here. All three kernels write out_ptr + offsets as if out were contiguous, so transposed or strided outputs are populated incorrectly and may overwrite storage outside the view. Please honor the output strides or copy through a contiguous temporary, and also handle device mismatches and overlapping inp/out.

tests/test_slice_copy.py
```diff
@@ -0,0 +1,187 @@
...
+        start, end, step = dim_size // 4, dim_size - dim_size // 4, 2
+        ref_out = torch.slice_copy(ref_inp, dim, start, end, step)
+
+        out_shape = list(inp.shape)
+        out_shape[dim] = (end - start + step - 1) // step
+        res_out = torch.empty(out_shape, dtype=dtype, device=flag_gems.device)
+        res_r = flag_gems.slice_copy_out(inp, dim, start, end, step, out=res_out)
+
+        assert res_r is res_out
+        utils.gems_assert_equal(res_r, ref_out)
```
Caeruleann, line 187: Please compare slice_copy.Tensor_out directly against ATen and add mismatched-size, non-contiguous/view, device-mismatch, and overlapping out cases; the current test only exercises a same-shaped contiguous output.

benchmark/test_slice_copy.py
```diff
@@ -0,0 +1,93 @@
...
+    end = dim_size - dim_size // 4
+    step = 2
+    inp = torch.randn(shape, dtype=dtype, device=device)
+    yield inp, dim, start, end, step
+
+
+def _get_gbps(bench_fn_args, latency):
+    inp = bench_fn_args[0]
+    # slice_copy reads the input and writes a fresh output of the sliced size.
+    out_elems = inp.numel() // inp.size(bench_fn_args[1])
```
Caeruleann, line 51: Please calculate the traffic from the actual sliced output size. slice_copy reads and writes output.numel() elements, while the current formula counts the entire input as read and only one slice plane as output.

benchmark/test_slice_copy.py
```diff
@@ -0,0 +1,93 @@
...
+        if flag_gems.vendor_name == "kunlunxin":
+            return []
+        # 2D shapes only; slice_copy slices along dim 1 with step 2, so keep the
+        # second axis large enough to make the slice non-trivial.
+        return [(10000, 2**i) for i in (8, 16)]
+
+
+def _input_fn(shape, dtype, device):
+    # Slice along dim 1 over the middle half of the dimension with step 2; this
+    # exercises the general (non-contiguous, non-inner1) kernel path.
```
Caeruleann, line 38: This 2D dim=1 case always has inner == 1, so it only exercises the inner1 kernel rather than the general kernel as the comment states. Please add a case that actually covers the general path.

tests/test_slice_copy.py
Yukun-Cui: Done in 767eba5. You were right that complex needed an explicit bitwise path — it was worse than a slow generic path, it crashed outright with `KeyError: 'complex64'` (the dtype never reached a kernel).

Complex now copies through `view_as_real`: each complex scalar becomes two adjacent reals, so the flat-index mapping is unchanged apart from `inner` doubling. `bool` and the float8 dtypes go through `uint8` for the same reason. All three are pure bit reinterpretations of the same storage, so NaN payloads and signalling bits survive.

Test coverage is now every dtype ATen accepts on CUDA: fp16/fp32/fp64/bf16, int8/16/32/64, uint8, bool, complex32/64/128, and float8_e4m3fn/e5m2. The float8 test sweeps all 256 byte patterns (including the NaN encodings) and asserts a bit-exact copy; complex is checked on each of the three kernel paths.

One related bug this turned up: a lazy conj or neg view returned wrong values. `_neg_view(arange(8))` gave `[0,1,2,3]` where ATen gives `[-0,-1,-2,-3]` — the raw bytes behind such a tensor are un-negated, so a plain bit copy reads the wrong thing. Now resolved before the copy, and the result has the flag cleared to match ATen.

src/flag_gems/ops/slice_copy.py
Yukun-Cui: Done in 767eba5. Confirmed both failure modes you predicted, and the out-of-view write was the worse one: with `out = big[:, 2:6]`, the flat-offset store wrote into `big`'s neighbouring columns — silent corruption of storage the caller never passed in. A transposed `out` got the right values transposed into the wrong positions.

I went with the contiguous temporary rather than honoring output strides. Staging costs one extra pass, but it only triggers on the non-contiguous/aliasing paths (the common contiguous `out` still writes straight into it), and it reuses ATen's own strided `copy_` instead of duplicating stride arithmetic in three kernels.

Also fixed on this path:

- **Overlapping inp/out**: ATen reads the whole slice before writing, so a forward-overlapping `out` must see the original values. The kernel interleaved reads and writes and read back data it had already overwritten. Detected with `torch._C._is_alias_of` and staged.
- **Internally overlapping out** (e.g. expanded): silently succeeded before. ATen refuses it with "more than one element of the written-to tensor refers to a single memory location"; routing through native `copy_` reproduces that error exactly.
- **Device mismatch**: leaked a Triton internal error (`Pointer argument cannot be accessed from Triton (cpu tensor?)`). Now ATen's message.
- **Error text and ordering**: the `out=` dtype message now uses the C++ `TypeMeta` spelling ATen prints (`c10::Half`, `long int`, `c10::complex<float>`). I also measured that ATen resizes a mismatched `out` *before* raising a dtype/device error, and kept that ordering.

Two bugs unrelated to `out=` that surfaced while testing this: an output above 2^31 elements was partially copied (2048 of 2^31+2048 elements left unwritten — `pid` stayed int32; `INT64_INDEX` now keys off both input and output extent), and the 0-dim/negative-dim diagnostics were wrong (garbled `range of [0, -1]` for 0-dim, wrapped dim reported for `dim=-3`, `RuntimeError` where ATen raises `IndexError`).

tests/test_slice_copy.py
Yukun-Cui: Done in 767eba5. All four cases added, compared against `slice_copy.Tensor_out` directly:

- **mismatched size** — including the resize-to-smaller, resize-to-larger and zero-element cases. ATen resizes rather than raising, so the test asserts the result lands in the resized tensor.
- **non-contiguous / view** — transposed, reverse-strided (`flip`), 3-D permuted, and a strided view into a larger buffer. That last one also asserts the untouched elements of the backing buffer still match ATen's, which is what caught the out-of-view write.
- **device mismatch** and **dtype mismatch** — assert the exact message string equals ATen's, rather than just the exception type, since the wording was one of the things that was wrong.
- **overlapping out** — three alias configurations (shifted forward, exactly the sliced region, shifted backward) plus an internally overlapping expanded `out`.

To check the tests actually bite, I ran the new file against the pre-fix implementation: 19 of 92 fail there, all 92 pass now. Separately I ran a randomized differential sweep of 4818 cases (random shapes, dims, bounds, steps, dtypes against nine adversarial `out=` layouts) — all match ATen exactly, including the untouched backing-buffer bytes.

Note on CI: the `slice_copy_out` operator id has no `tests/test_slice_copy_out.py`. The marker check passes because CI derives only `["slice_copy"]` from the diff, but if that id is ever checked directly it would want an alias entry in `conf/ci_test_aliases.yaml`. I left that file alone — happy to add it if you'd prefer.

benchmark/test_slice_copy.py
Yukun-Cui: Done in 767eba5, now `2 * out.numel() * element_size`.

The old formula was wrong on both sides, as you say. The output term `inp.numel() // inp.size(dim)` was one slice plane and ignored `step` entirely, and counting the whole input as read overstates traffic whenever the slice is a strict subset. For the benchmarked case (middle half of the dim, step 2) the output is exactly 25% of the input, so the two errors compounded to a clean 2x overstatement — measured at 32.01 vs 16.00 MiB on (4096, 4096) and 1250.02 vs 625.00 MiB on (10000, 65536).

benchmark/test_slice_copy.py
Yukun-Cui: Confirmed and fixed in 767eba5. I instrumented the three launch sites and logged which one fired for every benchmark shape: it was `slice_copy_kernel_inner1` in every single case, so the comment was describing a path that was never measured.

Rather than add 3-D shapes, I kept `GenericBenchmark2DOnly` and added a `dim=0` case alongside `dim=1` for each shape. Slicing dim 0 of a 2-D input gives `inner == shape[1]`, which routes to the general kernel — so the two dims now cover both paths without changing the shape set. (`GenericBenchmark` has no `core_shapes.yaml` entry, so switching base classes would have silently changed the core-level shapes.)

Re-verified after the change: `dim=1` hits `slice_copy_kernel_inner1`, `dim=0` hits `slice_copy_kernel` with `inner` from 64 up to 65536. Both paths beat ATen (1.2-2.7x, up to ~2200 GB/s on the largest shapes).

bin913 (approved):

PR #5964, opened 2026-09-03
[KernelGen][Nvidia] Add smm operator with Triton kernel

Yukun-Cui:
## Summary
Adds a Triton kernel for `smm`. Performs a matrix multiplication of the sparse (COO) matrix `self` with the dense matrix `mat` using a fused Triton kernel. The sparse operand is materialized to dense, the product is computed with `tl.dot`, and the COO output (indices + values) is written directly to match `torch.smm` semantics: every row that appears in the sparse input contributes all output columns.

## Testing

## Performance
Test command: `pytest benchmark/test_smm.py --level core` (NVIDIA H20)

### smm

| dtype | Size | Torch Latency (ms) | Gems Latency (ms) | Speedup | TFLOPS |
|-------|------|--------------------|-------------------|---------|--------|
| float32 | [64, 64] | 0.340800 | 0.333344 | 1.022 | 0.002 |
| float32 | [128, 128] | 0.347488 | 0.338336 | 1.027 | 0.012 |
| float32 | [256, 256] | 0.350496 | 0.342880 | 1.022 | 0.098 |
| float32 | [512, 128] | 0.351424 | 0.343040 | 1.024 | 0.098 |
| float32 | [1024, 256] | 0.375504 | 0.357920 | 1.049 | 0.750 |

| smm | **1.029** |

## Multi-backend Testing
|---|---|---|---|
| Nvidia (H20) | PASS (5 cases) | 1.029 | Primary |
| Tianshu | N/A | — | — |
| Muxi | N/A | — | — |
| Ascend | N/A | — | — |
| Hygon | N/A | — | — |

## Files Changed
- `src/flag_gems/ops/smm.py`: Triton kernel implementation
- `tests/test_smm.py`: Accuracy test
- `benchmark/test_smm.py`: Performance benchmark


src/flag_gems/ops/smm.py
```diff
@@ -0,0 +1,230 @@
...
+    # cheaper than ``torch.unique``'s full sort.
+    row_indices = self._indices()[0]
+    row_mask = torch.ones_like(row_indices, dtype=torch.bool)
+    row_mask[1:] = row_indices[1:] != row_indices[:-1]
+    out_rows = row_indices[row_mask]  # unique rows, ascending (int64)
+
+    # Materialize the sparse operand to dense for gathered loads.
+    a_dense = self.to_dense()
+    if not mat.is_contiguous():
+        mat = mat.contiguous()
```
Caeruleann, line 203: Please avoid materializing the sparse input with self.to_dense(). This introduces O(M*K) memory and dense-matmul work, so valid large sparse inputs may OOM and the implementation no longer scales with nnz. It also performs several PyTorch computations in the host wrapper, contrary to the host-function convention. Please operate directly on the COO indices and values in the device implementation.

benchmark/test_smm.py
```diff
@@ -0,0 +1,81 @@
...
+        density = 0.3
+        for M, K, N in self.shapes:
+            sparse = _make_sparse_coo(M, K, cur_dtype, self.device, density=density)
+            dense = torch.randn(K, N, dtype=cur_dtype, device=self.device)
+            yield sparse, dense
+
+    def get_tflops(self, op, *args, **kwargs):
+        sparse, dense = args[0], args[1]
+        # 2 * M * K * N flops for the (materialized) dense matmul.
+        return 2 * sparse.shape[0] * sparse.shape[1] * dense.shape[1]
```
Caeruleann, line 67: Please calculate useful FLOPs as 2 * sparse._nnz() * N rather than 2 * M * K * N; the current formula reports dense-matmul FLOPs and overstates sparse smm throughput.

src/flag_gems/ops/smm.py
```diff
@@ -0,0 +1,230 @@
...
+
+    Mirrors ``torch.smm``: the result is a sparse COO tensor whose rows are
+    exactly the rows that appear in the sparse input's structure, each
+    materialized with all ``N`` columns. A fused Triton kernel computes the
+    product with ``tl.dot`` and writes the COO output (indices + values)
+    directly, avoiding any separate dense-to-sparse conversion pass.
+    """
+    logger.debug("GEMS SMM")
+
+    assert self.is_sparse, "smm expects a sparse (COO) matrix as the first input"
```
Caeruleann, line 166: Please validate that self and mat are on the same device before launching the kernel. Currently a CPU tensor or a tensor from another GPU can pass these checks and be passed to a kernel launched on self.device, potentially causing an illegal memory access.

tests/test_smm.py
Caeruleann, line 1: Please add coverage for uncoalesced COO inputs with duplicate indices, explicit-zero values, non-contiguous mat, zero-sized dimensions, device mismatches, invalid shapes/layouts, and genuinely low-density patterns.

src/flag_gems/ops/smm.py
Yukun-Cui: Done — the kernel now reads the COO arrays directly and `to_dense()` is gone.

`smm_coo_kernel` gives each program one structural row and a tile of output columns. Because a coalesced 2-D COO tensor is sorted by row, each row occupies a contiguous span of the indices/values arrays, so the program streams its own nonzeros out of that segment in `BLOCK_NNZ` chunks. The working set is O(nnz + nrows*N) with no dependence on `M*K`. The reduction is a masked broadcast-multiply-accumulate rather than `tl.dot`, since the contracted extent is a row's nnz — typically small and irregular — not a dense `K`.

The segments themselves are built on device by two new kernels: `smm_mark_row_starts_kernel` flags entries whose row differs from their predecessor, and `smm_compact_rows_kernel` compacts those boundaries into row ids plus offsets. That replaces the host-side unique-row scan; the only host-side torch call left is the cumsum that ranks the boundaries.

Measured on the case that motivated this: `(2**20, 2**20)` with 512 nonzeros, whose dense form would be 4 TiB, now peaks at 18.6 MiB and completes. There is a regression test for it (`test_smm_scales_with_nnz_not_dense_size`).

Two things worth flagging from verifying this:

- Passing the output index tensor as one pointer plus `nrows*N` arithmetic crashed when the sparse matrix had exactly **one** structural row: Triton specializes an integer argument equal to 1 into a Python int, so `nrows.to(tl.int64)` raised `AttributeError` and the launch faulted with an illegal memory access. The two index rows are now passed as separate pointers, which removes that arithmetic. compute-sanitizer memcheck is clean across shapes including that case.
- The output is emitted with rows ascending and columns ascending within each row, so it is already in coalesced order. It is returned with `is_coalesced=True` to skip a redundant sort; I verified the claim holds (strictly ascending flat indices, and bitwise-identical values against a forced `coalesce()`) rather than just asserting it.

benchmark/test_smm.py
Yukun-Cui: Done — changed to `2 * sparse._nnz() * dense.shape[1]`.

Each stored nonzero contributes one multiply-add per output column, so `2 * nnz * N` is the useful work. The old formula was off by exactly `1/density`, which is 102x on the 1%-density shapes in this benchmark, i.e. it was reporting the flops of a dense matmul the sparse kernel never performs.

src/flag_gems/ops/smm.py
Yukun-Cui: Done — validated before launch.

You were right that this was reachable: the kernel is launched on `self.device`, so a CPU or foreign-GPU `mat` was handed over as a raw pointer. I reproduced the illegal memory access before fixing it.

The wrapper now checks `mat.device != self.device` up front and raises with ATen's own wording (`Expected all tensors to be on the same device, but got mat2 is on ...`), taken from what `torch.smm` emits for the same mismatch. The layout, dimensionality and K-mismatch errors were also switched from bare `assert`s to `RuntimeError`s carrying ATen's messages, so the messages match upstream and do not vanish under `python -O`.

`test_smm_device_mismatch` covers the CPU case, plus a second-GPU case when more than one CUDA device is visible — the foreign-GPU path is the one a CPU operand would not catch, since there both tensors are accelerator-resident.

tests/test_smm.py
Yukun-Cui: Done — all requested cases added, `tests/test_smm.py` is now 46 passing tests.

- **Uncoalesced with duplicate indices** — `test_smm_uncoalesced_duplicate_indices`, asserting `is_coalesced()` is false on the input so the summing path is genuinely exercised.
- **Explicit-zero values** — `test_smm_explicit_zero_values`, which also asserts the zero-valued rows still appear in the output pattern, since they are structurally present.
- **Non-contiguous `mat`** — `test_smm_non_contiguous_mat`, parametrized over a stride-2 column view, a transpose, and a storage-offset `as_strided` view.
- **Zero-sized dimensions** — `test_smm_zero_sized` over `(0,8)x(8,4)`, `(4,0)x(0,4)`, `(4,8)x(8,0)` and the all-zero case, checking the result shape.
- **Device mismatches** — `test_smm_device_mismatch` (see the reply on the device-check thread).
- **Invalid shapes/layouts** — `test_smm_invalid_inputs`, matching on ATen's message text for K mismatch, dense first argument, sparse second argument, and 3-D input.
- **Low density** — `test_smm_low_density` at 0.001/0.005/0.01 on 2048x2048, plus the memory-scaling test described in the `to_dense()` thread.

One note on how these assert. `torch.smm` decomposes to `sspaddmm`, which is not implemented on CUDA, so the reference is computed on CPU via a `_ref_dense` helper and compared against the dense form of the GEMS result.

Separately, while checking accuracy I hit a `maxdiff` of ~4.9e-2 against a dense `to_dense() @ mat` reference and initially suspected the kernel. It is the reference that is imprecise: TF32 is enabled by default here, so against an fp64 reference the dense baseline carries ~4.9e-2 while this kernel carries ~4.0e-5. No change needed, but worth recording in case the number turns up in someone else's comparison.

Caeruleann (approved):

PR #6006, opened 2026-09-04
[KernelGen][Nvidia] Add quantize_per_channel operator with Triton kernel

Yukun-Cui:
## Summary
Adds a Triton kernel for `quantize_per_channel`. Triton kernel implementation for quantize_per_channel. Quantizes a float tensor per channel with the given scales and zero_points along the specified axis, producing a per-channel affine quantized tensor.

## Testing

## Performance
Test command: `pytest benchmark/test_quantize_per_channel.py --level core` (NVIDIA H20)

### quantize_per_channel

| quint8 | [64, 64] | 0.062688 | 0.118000 | 0.531 |
| quint8 | [1024, 1024] | 0.092544 | 0.142144 | 0.651 |
| quint8 | [4096, 4096] | 0.555200 | 0.628384 | 0.884 |
| quint8 | [64, 512, 512] | 0.556800 | 0.627680 | 0.887 |
| qint8 | [64, 64] | 0.057856 | 0.119936 | 0.482 |
| qint8 | [1024, 1024] | 0.093504 | 0.141568 | 0.660 |
| qint8 | [4096, 4096] | 0.557856 | 0.627936 | 0.888 |
| qint8 | [64, 512, 512] | 0.557616 | 0.627808 | 0.888 |
| qint32 | [64, 64] | 0.057600 | 0.121584 | 0.474 |
| qint32 | [1024, 1024] | 0.088336 | 0.143424 | 0.616 |
| qint32 | [4096, 4096] | 0.556896 | 0.649344 | 0.858 |
| qint32 | [64, 512, 512] | 0.553520 | 0.649392 | 0.852 |

| quantize_per_channel | **0.723** |

### quantize_per_channel_out

| quint8 | [64, 64] | 0.064480 | 0.131648 | 0.490 |
| quint8 | [1024, 1024] | 0.095552 | 0.148768 | 0.642 |
| quint8 | [4096, 4096] | 0.566944 | 0.636256 | 0.891 |
| quint8 | [64, 512, 512] | 0.566656 | 0.636256 | 0.891 |
| qint8 | [64, 64] | 0.064736 | 0.131360 | 0.493 |
| qint8 | [1024, 1024] | 0.095408 | 0.149488 | 0.638 |
| qint8 | [4096, 4096] | 0.567872 | 0.637184 | 0.891 |
| qint8 | [64, 512, 512] | 0.568272 | 0.637120 | 0.892 |
| qint32 | [64, 64] | 0.064640 | 0.131776 | 0.491 |
| qint32 | [1024, 1024] | 0.095264 | 0.150784 | 0.632 |
| qint32 | [4096, 4096] | 0.593472 | 0.687744 | 0.863 |
| qint32 | [64, 512, 512] | 0.594176 | 0.687168 | 0.865 |

| quantize_per_channel_out | **0.723** |

## Multi-backend Testing
|---|---|---|---|
| Nvidia (H20) | PASS (24 cases) | 0.723/0.723 | Primary |
| Tianshu | N/A | — | — |
| Muxi | N/A | — | — |
| Ascend | N/A | — | — |
| Hygon | N/A | — | — |

## Files Changed
- `src/flag_gems/ops/quantize_per_channel.py`: Triton kernel implementation
- `tests/test_quantize_per_channel.py`: Accuracy test
- `benchmark/test_quantize_per_channel.py`: Performance benchmark


src/flag_gems/ops/quantize_per_channel.py
```diff
@@ -0,0 +1,147 @@
...
+        return torch._make_per_channel_quantized_tensor(
+            int_out, scales, zero_points, axis
+        )
+
+    # Power-of-2 block size for coalesced element-wise access.
+    BLOCK_SIZE = 1024
+    grid = (triton.cdiv(n_elements, BLOCK_SIZE),)
+
+    with torch_device_fn.device(input.device):
+        quantize_per_channel_kernel[grid](
```
Caeruleann, line 118: Please validate the native input contract before launching the kernel, including FP32 input, a non-negative in-range axis, 1D scales/zero_points with lengths equal to input.size(axis), floating-point scales, and valid zero-point ranges. Short qparam tensors currently cause out-of-bounds loads in the kernel.

src/flag_gems/ops/quantize_per_channel.py
```diff
@@ -0,0 +1,147 @@
...
+def _quantize_per_channel_impl(input, scales, zero_points, axis, dtype):
+    int_dtype, q_min, q_max = _QRANGE[dtype]
+
+    # The quantization params are 1-D vectors of length ``input.size(axis)``.
+    # Keep them contiguous and on the same device as the input. Scales are kept
+    # in float64 (PyTorch stores per-channel scales as float64) so the in-kernel
+    # division matches PyTorch's accuracy.
+    input = input.contiguous()
+    scales = scales.to(input.device).to(torch.float64).contiguous()
+    zero_points = zero_points.to(input.device).to(torch.int64).contiguous()
```
Caeruleann, line 96: Please preserve and implement the native floating-zero-point path instead of unconditionally casting zero_points to int64. PyTorch supports per_channel_affine_float_qparams, for which both the qscheme and rounding formula differ; please add float-zero-point parity tests.

src/flag_gems/ops/quantize_per_channel.py
```diff
@@ -0,0 +1,147 @@
...
+    return _quantize_per_channel_impl(input, scales, zero_points, axis, dtype)
+
+
+def quantize_per_channel_out(input, scales, zero_points, axis, dtype, *, out=None):
+    logger.debug("GEMS QUANTIZE_PER_CHANNEL_OUT")
+    result = _quantize_per_channel_impl(input, scales, zero_points, axis, dtype)
+    if out is not None:
+        # `out` is a quantized tensor; copy_ replaces its integer storage as
+        # well as its scale/zero-point/axis to match the freshly quantized result.
+        out.copy_(result)
```
Caeruleann, line 145: Please compare this overload directly with torch.ops.aten.quantize_per_channel.out and cover the full out= contract. copy_ does not resize a wrong-sized output, while the current test only uses an exactly matching out tensor.

tests/test_quantize_per_channel.py
```diff
@@ -0,0 +1,212 @@
...
+        res.q_per_channel_zero_points(), ref.q_per_channel_zero_points()
+    )
+    # int_repr is uint8/int8/int32; compare in float32 with atol=1 to absorb
+    # the rounding-boundary 1-ULP diff between CPU and GPU references.
+    utils.gems_assert_close(
+        res.int_repr().to(torch.float32),
+        ref.int_repr().to(torch.float32),
+        torch.float32,
+        atol=1,
+        equal_nan=True,
```
Caeruleann, line 77: Please compare int_repr() exactly instead of casting it to FP32 with atol=1. A one-bin difference is an observable quantization mismatch, and converting qint32 values to FP32 also loses integer precision; explicit positive/negative half-way tie cases should be added.

src/flag_gems/ops/quantize_per_channel.py
Yukun-Cui: Fixed in a1123deb. Validation now runs before the kernel is launched, in `_validate_qparams` (src/flag_gems/ops/quantize_per_channel.py), and every message plus its ordering mirrors an ATen assertion:

- unsupported dtype -> NotImplementedError `"quantize_tensor_per_channel_affine" not implemented for 'QUInt4x2'`
- non-fp32 input -> `expects a Float Tensor, got Half/Double` (the float-qparams quantizer instead says `Quantize only works on Float Tensor, got Half`, which is what ATen's two quantizers actually print)
- `0 <= axis < dim`, so negative axes are rejected: `Channel axis out of range in per channel {affine,float qparams} quantization. Got: -1Expected: [0, 2)`
- 1-D `scales`/`zero_points` with equal numel: `scale tensor must have dimension 1` / `zero_points tensor must have dimension 1` / `number of elements in scales and zero_points must match`
- floating-point scales: `scale tensor must be floating point`
- both vectors of length `input.size(axis)`: `length of scales must equal to channel, expected 4 got, 3`

The out-of-bounds load you spotted is the length check: `assert len(1) with 4 channels` and `len(3) with 4 channels` are now rejected instead of indexing `scales_ptr + axis_coord` past the end. Verified by `test_quantize_per_channel_rejects_{non_float_input,float_zero_point_non_float_input,bad_axis,multi_dim_qparams,short_qparams,integer_scales,unsupported_dtype}`.

Zero-point ranges: an out-of-range zero-point cannot be decided on the host without a synchronising read of the qparam tensor, so the kernel validates it on device with an unconditional PTX trap and the call aborts the way ATen's CUDA kernel does. Tested in subprocesses (a trap faults the context) for the four cases (quint8 zp=300 / -0.5, qint8 zp=-129 / 300.0); the in-range values used elsewhere are asserted not to trip it.

All checks pass: 179 accuracy tests, 54 under `--ref=cpu --quick`.

src/flag_gems/ops/quantize_per_channel.py
Yukun-Cui: Fixed in a1123deb: the float-zero-point path is implemented and the unconditional `zero_points.to(torch.int64)` is gone.

ATen picks the scheme from the zero-point dtype (`make_per_channel_affine_quantizer` in aten/src/ATen/quantized/Quantizer.cpp): floating-point zero-points select `PerChannelAffineFloatQParamsQuantizer`, which stores scales *and* zero-points as float32 and rounds `lrintf(raw * (1.f/scale) + zero_point)` (quantize_tensor_per_channel_float_qparams_cuda). That is a different formula from the integer scheme, which does `nearbyint(value * inv_scale)` and adds the zero-point afterwards. The kernel now branches on `USE_FLOAT_ZERO_POINT`, and `_make_per_channel_quantized_tensor` infers the qscheme from the zero-point dtype, so the result reports `torch.per_channel_affine_float_qparams` with a float32 zero-point tensor, matching native.

Getting bit-identical results needed two things that a direct port of the C expression misses:

1. `tl.math.div_rn(1.0, scale)` instead of `1.0 / scale`. Triton's `/` lowers to an approximate reciprocal-multiply sequence: for scale 0.02372691035270691 it returned `42.146240234375` where the correctly rounded value is `42.146236419677734`. That ~2 ULP error flips rounding ties.
2. `tl.math.fma(x, inv, zero_point)` instead of a separate multiply and add, so the product is rounded once, like the compiler-contracted expression in ATen. Splitting it turns exact .5 values into the adjacent bin (e.g. x=84.24085998535156, scale=0.0341704860329628, zp=15352.1884765625: two-step gives 17818, ATen and gems give 17817).

Measured: with both, device results are bit-identical to the CUDA reference (0 differing elements over 336 configs across shapes/axes/dtypes and both schemes), and the earlier ad-hoc case `torch.quantize_per_channel(x, sc, zpf, 0, torch.quint8)` that disagreed by one bin (77 vs 78) now matches exactly.

Tests added: `test_quantize_per_channel_float_zero_points` (scheme + exact parity), `..._float_zero_points_tie_cases`, and `..._float_zero_points_qscheme_differs`, which asserts the float and integer schemes on identical scales produce different payloads so the float path cannot silently collapse into the integer one. Note `torch.quantize_per_channel` with CPU float-qparams + qint32 SIGILLs in native PyTorch itself (reproducible without flag_gems), so the float-qparams tests are parametrized over quint8/qint8.

src/flag_gems/ops/quantize_per_channel.py
Yukun-Cui: Fixed in a1123deb, with one part of the suggestion not reproducible.

`test_quantize_per_channel_out_matches_aten_out` now runs `flag_gems.quantize_per_channel_out` and `torch.ops.aten.quantize_per_channel.out` on identically-initialised out tensors and compares object identity, `int_repr()`, scales, zero-points, axis and qscheme. The out= contract is covered by: exact shape with deliberately wrong (stale) qparams, so the adopted qparams are asserted to be the new ones; dtype mismatch, which raises ATen's `Expected out tensor to have dtype c10::quint8, but got c10::qint8 instead`; qscheme mismatch (float-qparams result into an integer-qparams out), which raises `Quantized Copy only works with same qscheme`, the error native `copy_` gives; and a non-contiguous out view, asserting the strided destination is written through its own layout.

On the wrong-sized case: I could not implement resize. On this backend `resize_` on a per-channel quantized tensor does not exist -- `aten::resize_` is not registered for QuantizedCUDA (also `aten::set_` and `aten::_resize_output_`), so native `out=` with a wrong-sized out raises `NotImplementedError: Could not run 'aten::resize_' with arguments from the 'QuantizedCUDA' backend` rather than resizing. Native only succeeds in resizing when the reference runs on CPU. Reproducing that on the accelerator is not possible through the available ops, and ping the storage/shape by hand would write through a stale quantizer. So the operator now reports the size mismatch explicitly (the shape error `copy_` would otherwise raise only after the payload was computed) instead of silently writing a wrong-sized destination, and the test asserts that. Happy to add a resizing path if there is an accepted way to resize a per-channel quantized tensor here.

tests/test_quantize_per_channel.py
Yukun-Cui: Fixed in a1123deb: `int_repr()` is now compared exactly via `utils.gems_assert_equal` instead of `.to(torch.float32)` with `atol=1`.

Both of your points are real and I reproduced them: `torch.tensor([16777217], dtype=torch.int32).to(torch.float32)` is `16777216.0`, so a qint32 one-bin difference vanishes in the cast, and `atol=1` with an fp32 rtol also reports `allclose(1e8, 1e8+1)` as True while the integers differ. The comparison now covers `int_repr()` bit-for-bit along with the scales, the zero-points (including their dtype, which is how the two qschemes are told apart), the axis and the qscheme.

Tie cases added for both schemes:
- `test_quantize_per_channel_tie_cases`: inputs -7.5, -4.5, -3.5, -2.5, -1.5, -0.5, 0.5, 1.5, 2.5, 3.5, 6.5, 7.5 with unit scale land exactly on .5 boundaries, and the test asserts both the reference parity and the expected half-to-even payload independently (0.5 -> 0, 1.5 -> 2, 2.5 -> 2, -0.5 -> 0, -1.5 -> -2, -2.5 -> -2).
- `test_quantize_per_channel_float_zero_points_tie_cases`: zero-point 0.5 with unit scale, exercising the float scheme's add-before-round order.

One caveat worth flagging: exact comparison is only valid if the reference agrees with itself. ATen's own CPU and CUDA kernels disagree at ties with arbitrary scales -- at x=6.269403457641602, scale=0.12173599004745483, zp=12 the CPU reference returns 64 and the CUDA reference 63 (a different reciprocal/division path). So the parity tests use power-of-two scales, where all three implementations round the identical real value and the comparison is legitimately exact under both `--ref` settings; `test_quantize_per_channel_round_trip` keeps arbitrary scales and tolerance where tolerance is the honest answer. Under exact comparison gems matches the CUDA reference with 0 differing elements over thousands of arbitrary-scale configs.

179 tests pass (was 129), 54 under `--ref=cpu --quick`.

src/flag_gems/ops/quantize_per_channel.py
```diff
@@ -0,0 +1,331 @@
...
+    # The quantization params are 1-D vectors of length ``input.size(axis)``.
+    # Keep them contiguous and on the same device as the input.
+    #
+    # Integer schemes store scales as fp64 and zero-points as int64, and the
+    # float scheme keeps both in fp32 -- the dtypes the two native quantizers
+    # use internally. The kernel branches on ``USE_FLOAT_ZERO_POINT``, so the
+    # two precisions never mix.
+    input = input.contiguous()
+    if use_float_zero_point:
+        scales_k = scales.to(input.device).to(torch.float32).contiguous()
```
Caeruleann: scales.to(input.device) / zero_points.to(input.device) silently transfer CPU qparams to the GPU, while native raises Expected all tensors to be on the same device, but got scales is on cpu... — same device-consistency contract the put review asked for. Please validate and raise instead of moving.

src/flag_gems/ops/quantize_per_channel.py
```diff
@@ -0,0 +1,331 @@
...
+        #     .5 values into the adjacent bin.
+        inv_scale = tl.math.div_rn(1.0, scale.to(tl.float32))
+        q = tl_extra_shim.nearbyint(tl.math.fma(x_val, inv_scale, zero_point_f))
+    else:
+        # per_channel_affine: the zero-point is a genuine integer, stored as
+        # int64, and must be inside the target dtype's range
+        # (check_zero_points_cuda).
+        zero_point_i64 = tl.load(zero_points_ptr + axis_coord, mask=mask, other=0)
+        zp_valid = (zero_point_i64 >= q_min) & (zero_point_i64 <= q_max)
+        _trap_if(tl.max((mask & (~zp_valid)).to(tl.int32)) > 0)
```
Caeruleann: Native raises a synchronous RuntimeError (quantize_tensor_per_channel_affine_cuda zero_point is above upper bound) for an out-of-range zero-point — catchable, and the CUDA context survives, because ATen's check_zero_points_cuda also syncs to check. The device trap here turns a recoverable input error into a context-killing async fault, so a try/except around quantize works natively but loses the whole process under gems. Please do the range check on the host like ATen does.

src/flag_gems/ops/quantize_per_channel.py
Yukun-Cui: 已修（de23a6d0）。设备一致性检查加在 `_validate_qparams` 的最前面，不再用 `.to(input.device)` 静默搬移：

```
Expected all tensors to be on the same device, but got scales is on cpu, different from other tensors on cuda:0 (when checking argument in method wrapper_CUDA__quantize_per_channel)
```

顺序也对齐了 native：设备检查先于长度/axis 等语义校验（实测 native 在 qparam 长度错误 + CPU qparam 同时存在时，报的是设备错误，而不是长度错误）；两个 qparam 都在 CPU 时先报 `scales`。

新增 `test_quantize_per_channel_rejects_cpu_qparams`，覆盖 scales / zero_points 两种参数、以及「设备错误优先于长度错误」。

src/flag_gems/ops/quantize_per_channel.py
Yukun-Cui: 已修（de23a6d0）。你的判断准确——native 的 `check_zero_points_cuda` 自己也会 sync，所以它是可 catch 的同步 `RuntimeError`，CUDA context 存活；device trap 把一个可恢复的输入错误升级成了杀进程的异步 fault。

现在改成 host 侧检查（对原始 dtype 的值做一次 `min`/`max`，在任何转换之前），消息逐字对齐 CUDA native：

```
quantize_tensor_per_channel_affine_cudazero_point is below lower bound.
quantize_tensor_per_channel_affine_cudazero_point is above upper bound.
quantize_tensor_per_channel_float_qparams_cudazero_point is above upper bound.
```

包括 native 的 below-before-above 顺序（同时越上下界时报 below），以及 `..._float_qparams_cuda...` 的命名。`device trap` 与其 subprocess 测试已删除，改为直接异常测试 + 一个与 native 逐字比对消息的测试，并额外断言 `try/except` 包住调用后 `torch.cuda.synchronize()` 正常——这正是你指出的「native 里 try/except 能用，gems 里会丢进程」的差别。

`tests/test_quantize_per_channel.py`: 184 passed（`--ref=cpu --quick` 59 passed）。

conf/operators.yaml
```diff
@@ -10964,6 +10964,32 @@ ops:
...
+    specified axis, producing a per-channel affine quantized tensor.
+  for:
+  - quantize_per_channel
+  labels:
+  - aten
+  - KernelGen
+  kind:
+  - Math
+  stages:
+  - alpha: '5.4'
```
```suggestion
  - alpha: '5.5'
```

conf/operators.yaml
```diff
@@ -10964,6 +10964,32 @@ ops:
...
+    A variant of quantize_per_channel() that writes the result into the out tensor.
+  for:
+  - quantize_per_channel.out
+  labels:
+  - aten
+  - KernelGen
+  kind:
+  - Math
+  stages:
+  - alpha: '5.4'
```
```suggestion
  - alpha: '5.5'
```

Caeruleann (dismissed):

huangyiqun (dismissed):

huangyiqun:
Plz fix CI failed.

PR #6011, opened 2026-09-04
[KernelGen][Nvidia] Add row_stack operator with Triton kernel

Yukun-Cui:
## Summary
Adds a Triton kernel for `row_stack`. Stack tensors in sequence vertically (row wise). Alias of vstack.

## Testing

## Performance
Test command: `pytest benchmark/test_row_stack.py --level core` (NVIDIA H20)

### row_stack

| float16 | [64, 64] | 0.005552 | 0.019744 | 0.281 |
| float16 | [4096, 4096] | 0.057440 | 0.053600 | 1.072 |
| float16 | [64, 512, 512] | 0.057280 | 0.053888 | 1.063 |
| float32 | [64, 64] | 0.005472 | 0.018432 | 0.297 |
| float32 | [4096, 4096] | 0.107872 | 0.106240 | 1.015 |
| float32 | [64, 512, 512] | 0.108064 | 0.099776 | 1.083 |
| bfloat16 | [64, 64] | 0.005568 | 0.019456 | 0.286 |
| bfloat16 | [4096, 4096] | 0.057344 | 0.053888 | 1.064 |
| bfloat16 | [64, 512, 512] | 0.057184 | 0.054496 | 1.049 |

| row_stack | **0.801** |

### row_stack_out

| float16 | [64, 64] | 0.005472 | 0.015744 | 0.348 |
| float16 | [4096, 4096] | 0.057376 | 0.054304 | 1.057 |
| float16 | [64, 512, 512] | 0.057376 | 0.053600 | 1.070 |
| float32 | [64, 64] | 0.005600 | 0.015168 | 0.369 |
| float32 | [4096, 4096] | 0.108320 | 0.100032 | 1.083 |
| float32 | [64, 512, 512] | 0.108256 | 0.099840 | 1.084 |
| bfloat16 | [64, 64] | 0.005408 | 0.014992 | 0.361 |
| bfloat16 | [4096, 4096] | 0.057408 | 0.054208 | 1.059 |
| bfloat16 | [64, 512, 512] | 0.057504 | 0.054080 | 1.063 |

| row_stack_out | **0.833** |

## Multi-backend Testing
|---|---|---|---|
| Nvidia (H20) | PASS (18 cases) | 0.801/0.833 | Primary |
| Tianshu | N/A | — | — |
| Muxi | N/A | — | — |
| Ascend | N/A | — | — |
| Hygon | N/A | — | — |

## Files Changed
- `src/flag_gems/ops/row_stack.py`: Triton kernel implementation
- `tests/test_row_stack.py`: Accuracy test
- `benchmark/test_row_stack.py`: Performance benchmark


tests/test_row_stack.py
```diff
@@ -0,0 +1,78 @@
...
+
+    # Compute the reference output to derive the expected out shape.
+    ref_out = torch.row_stack(ref_inp)
+
+    # Build an out tensor with the expected shape on the gems device and
+    # invoke the out variant directly. The kernel writes into ``res_out``.
+    res_out = torch.empty(ref_out.shape, dtype=dtype, device=flag_gems.device)
+    flag_gems.row_stack_out(inp, out=res_out)
+
+    utils.gems_assert_equal(res_out, ref_out)
```
Caeruleann, line 78: Please use the native out= path as the reference here (torch.row_stack(ref_inp, out=ref_out)) and verify that both calls return the provided output tensor. Comparing only against the functional variant checks values but misses the out= contract.

src/flag_gems/ops/row_stack.py
```diff
@@ -0,0 +1,217 @@
...
+                array_row_offset += c_tensors[tensor_idx].shape[0]
+                max_rows = max(max_rows, c_tensors[tensor_idx].shape[0])
+            else:
+                empty_tensor = torch.empty(
+                    0, dtype=c_tensors[0].dtype, device=c_tensors[0].device
+                )
+                itensors.append(empty_tensor)
+                local_row.append(local_row[-1])
+                exclusive_row.append(exclusive_row[-1])
+        max_tile_elems = max_rows * row_stride
```
Caeruleann, line 122: Please handle zero-sized trailing dimensions before launching the kernel. For an input such as (3, 0), stride(0) may be non-zero even though numel()==0, so local_row * row_stride causes out-of-bounds reads and writes. Please derive the row width from the trailing sizes, skip zero-work launches, and add ATen parity tests.

src/flag_gems/ops/row_stack.py
```diff
@@ -0,0 +1,217 @@
...
+    c_tensors = [t.contiguous() for t in tensors]
+    # Calculate the expected output shape
+    total_rows = sum(tensor.shape[0] for tensor in c_tensors)
+    output_shape = list(c_tensors[0].shape)
+    output_shape[0] = total_rows
+
+    if list(out.shape) != output_shape:
+        out.resize_(output_shape)
+
+    return _row_stack_run(tensors, out)
```
Caeruleann, line 217: row_stack_out writes out as flat contiguous storage, so a correctly shaped non-contiguous out produces incorrect results because its strides are ignored. Please make the write stride-aware or use a contiguous temporary followed by copy_, and add parity tests for non-contiguous out, resizing, dtype/device errors, overlap, and return identity.

src/flag_gems/ops/row_stack.py
```diff
@@ -0,0 +1,217 @@
...
+
+    # Ensure all tensors are on the same device and have the same dtype
+    device = tensors[0].device
+    dtype = tensors[0].dtype
+    for tensor in tensors:
+        assert (
+            tensor.device == device
+            and tensor.dtype == dtype
+            and tensors[0].shape[1:] == tensor.shape[1:]
+        )
```
Caeruleann, line 166: Please match ATen’s dtype-promotion behavior instead of requiring every input to have the same dtype. Mixed-dtype inputs accepted by torch.row_stack currently fail here; please derive the promoted output dtype and add parity tests.

src/flag_gems/ops/row_stack.py
```diff
@@ -0,0 +1,217 @@
...
+
+    c_tensors = [t.contiguous() for t in tensors]
```
Caeruleann, line 168: Please preserve ATen’s memory-format behavior for channels-last/channels-last-3d inputs. The unconditional .contiguous() and contiguous empty_strided allocation always produce standard contiguous output; please add output-stride and memory-format parity tests.

tests/test_row_stack.py
Yukun-Cui: Fixed in the latest commit (ff59ecb). `test_row_stack_out` now builds its reference through the native out= path (`torch.row_stack(ref_inp, out=ref_out)`) instead of only comparing against the functional variant, and asserts the out= contract directly: both the reference call and the GEMS call must return the exact tensor they were handed (`assert ref_ret is ref_out`, `assert res_ret is res_out`, plus a `data_ptr()` identity check on the GEMS side). The out= contract is exercised across all ROW_STACK_SHAPES and all float/int dtypes (40 parametrized cases), so a wrong return value or a silently re-allocated out tensor now fails the test.

src/flag_gems/ops/row_stack.py
Yukun-Cui: Confirmed and reproduced first: with inputs like (3,0,4)+(2,0,4) the old code hit `CUDA error: an illegal memory access` because the launch was sized from `rows * stride(0)` while numel()==0 (e.g. (3,0) has stride (1,1)). Fixed in ff59ecb: `_row_width()` now derives the row width as the product of the trailing sizes, and both `row_stack` and `row_stack_out` return early before any launch when `row_width == 0` or the output has zero rows — zero-work stacks launch nothing. Output strides are computed zero-size-aware (ATen skips the stride multiply for a 0-sized dim, so (5,0) -> stride (1,1), (5,0,4) -> (4,4,1)), which I verified against ATen for 11 degenerate shapes including (0,4), (3,0,0) and (1,0,2,3), all matching exactly. Added `test_row_stack_zero_sized_trailing_dims` (5 shape cases x 5 float dtypes) asserting value, shape and stride parity against `torch.row_stack`.

src/flag_gems/ops/row_stack.py
Yukun-Cui: Confirmed: on a (4,3) column-slice out with stride (6,1), the old code wrote flat storage and left half the destination stale. Fixed in ff59ecb with a two-path strategy: if `out` already has the derived layout's exact strides (contiguous or channels-last), the kernel writes into it directly; otherwise the result is staged into a dense buffer and scattered through `copy_`, which is stride-aware (verified: flag_gems' `copy_` writes correctly through a (6,1)-strided dst and preserves a channels-last dst). ATen's out= contract is reproduced in `_check_out`: canCast check first (TypeError on impossible casts, matching ATen's "input types can't be cast" error), `resize_` on shape mismatch with the derived memory format re-imposed afterwards (ATen's `set_output_raw_strided` behavior, which plain `resize_` does not do), internal-overlap rejection via `has_internal_overlapping`, and input/out overlap rejection via `torch._C._overlaps`. Return identity is preserved: the function returns the `out` tensor it was handed (asserted via `is` and `data_ptr()` in the tests). New tests cover: non-contiguous out values + untouched neighbouring columns, resize (bigger/smaller/empty), dtype cast out (f32->f64), non-castable out raising TypeError, device mismatch raising RuntimeError, overlapping input+out raising RuntimeError, and return identity.

src/flag_gems/ops/row_stack.py
Yukun-Cui: Confirmed: mixed f32/f64 inputs hit the `tensor.dtype == dtype` assertion while `torch.row_stack` promotes them to float64. Fixed in ff59ecb: the same-dtype assertion is gone; `_promote_dtype()` now folds all input dtypes through `torch.promote_types` (ATen's result_type rule), and `_cast_tensors()` brings every input to the promoted dtype before the copy, so the kernel still reads a uniform element width. Verified against ATen for f32+f64->f64, i32+f32->f32, i64+i32->i64 and f16+f32->f32 — dtype and values match in every case. Added `test_row_stack_mixed_dtype_promotion` covering both a float pair (f32+f64) and an int pair (i32+i64), asserting the promoted output dtype and value parity.

src/flag_gems/ops/row_stack.py
Yukun-Cui: Confirmed: channels-last inputs previously produced standard-contiguous output (stride (60,20,5,1) instead of (60,1,15,3)). Fixed in ff59ecb: `_derive_memory_format()` implements ATen's `cat_compute_output_memory_format` verbatim (any contiguous input, or two inputs disagreeing on their suggested format, falls back to contiguous; I read the C++ in aten/src/ATen/native/TensorShape.cpp and cross-checked `torch._prims_common.suggest_memory_format` against it on 2909 randomized rank-3..5 tensors with zero divergence), and the output is allocated via `empty_strided` with the channels-last / channels-last-3d strides for rank 4/5. Inputs are normalized to be dense in the same trailing layout as the output before the positional copy. Added `test_row_stack_memory_format` and `test_row_stack_out_memory_format` over (4D, CL), (4D, contiguous), (5D, CL3), (5D, contiguous), asserting exact output-stride equality with ATen plus `is_contiguous(memory_format=...)`. Beyond the suggested tests, the out= path was also aligned with ATen's full memory-format contract (kept layout for correct-shape outs, re-imposed derived format after resize) and validated against ATen directly on both rank-4 and rank-5 cases; a randomized 112-case sweep comparing shape, strides and values against `torch.row_stack` (incl. zero-sized dims, ranks 1-5, f16/f32/i32, CL/CL3) passes with 0 failures, and the full `tests/test_row_stack.py` suite is 99 passed.

Caeruleann (dismissed):

Caeruleann:
Please solve the conflicts

huangyiqun (approved):

PR #6021, opened 2026-09-05
[KernelGen][Nvidia] Add put_ operator with Triton kernel

Yukun-Cui:
## Summary
Adds a Triton kernel for `put_`. The in-place version of `put()`.

## Testing

## Performance
Test command: `pytest benchmark/test_put_.py --level core` (NVIDIA H20)

### put_ (in-place)

| float16 | [1073741824] | 0.078144 | 0.080896 | 0.966 |
| float16 | [64, 64] | 0.009024 | 0.007488 | 1.205 |
| float16 | [4096, 4096] | 0.062176 | 0.065632 | 0.947 |
| float16 | [64, 512, 512] | 0.062576 | 0.065664 | 0.953 |
| float16 | [1024, 1024, 1024] | 0.077984 | 0.079904 | 0.976 |
| float32 | [1073741824] | 0.078976 | 0.079088 | 0.999 |
| float32 | [64, 64] | 0.009376 | 0.007328 | 1.279 |
| float32 | [4096, 4096] | 0.070176 | 0.069760 | 1.006 |
| float32 | [64, 512, 512] | 0.070624 | 0.070592 | 1.000 |
| float32 | [1024, 1024, 1024] | 0.078336 | 0.078240 | 1.001 |
| bfloat16 | [1073741824] | 0.078336 | 0.079872 | 0.981 |
| bfloat16 | [64, 64] | 0.009056 | 0.007232 | 1.252 |
| bfloat16 | [4096, 4096] | 0.062144 | 0.065344 | 0.951 |
| bfloat16 | [64, 512, 512] | 0.062528 | 0.065216 | 0.959 |
| bfloat16 | [1024, 1024, 1024] | 0.077952 | 0.079744 | 0.978 |
| float16 | [1073741824] | 0.077280 | 0.087360 | 0.885 |
| float16 | [64, 64] | 0.009376 | 0.012064 | 0.777 |
| float16 | [4096, 4096] | 0.039200 | 0.054208 | 0.723 |
| float16 | [64, 512, 512] | 0.039424 | 0.054656 | 0.721 |
| float16 | [1024, 1024, 1024] | 0.076608 | 0.087216 | 0.878 |
| float32 | [1073741824] | 0.076000 | 0.076000 | 1.000 |
| float32 | [64, 64] | 0.009344 | 0.006784 | 1.377 |
| float32 | [4096, 4096] | 0.051808 | 0.051328 | 1.009 |
| float32 | [64, 512, 512] | 0.051648 | 0.051360 | 1.006 |
| float32 | [1024, 1024, 1024] | 0.075808 | 0.076720 | 0.988 |
| bfloat16 | [1073741824] | 0.077248 | 0.086880 | 0.889 |
| bfloat16 | [64, 64] | 0.009856 | 0.012224 | 0.806 |
| bfloat16 | [4096, 4096] | 0.039776 | 0.054176 | 0.734 |
| bfloat16 | [64, 512, 512] | 0.039360 | 0.054240 | 0.726 |
| bfloat16 | [1024, 1024, 1024] | 0.076768 | 0.086848 | 0.884 |

| put_ (in-place) | **0.962** |

## Multi-backend Testing
|---|---|---|---|
| Nvidia (H20) | PASS (30 cases) | 0.962 | Primary |
| Tianshu | N/A | — | — |
| Muxi | N/A | — | — |
| Ascend | N/A | — | — |
| Hygon | N/A | — | — |

## Files Changed
- `src/flag_gems/ops/put_.py`: Triton kernel implementation
- `tests/test_put_.py`: Accuracy test
- `benchmark/test_put_.py`: Performance benchmark


src/flag_gems/ops/put_.py
```diff
@@ -0,0 +1,194 @@
...
+
+    # Load the flat (linear) index into the input tensor, treated as 1-D.
+    idx = tl.load(index_ptr + offsets, mask=mask, other=0).to(tl.int64)
+
+    # Negative index maps to the end of the flattened tensor.
+    idx = tl.where(idx < 0, idx + inp_numel, idx)
+
+    # Validate index bounds: 0 <= idx < inp_numel (self.numel()).
+    index_valid = (idx >= 0) & (idx < inp_numel)
+    final_mask = mask & index_valid
```
Caeruleann, line 169: Out-of-range indices are currently masked out and silently ignored, while PyTorch raises for indices outside [-self.numel(), self.numel()) and for non-empty indices on an empty tensor. Please preserve these error semantics and add corresponding tests.

src/flag_gems/ops/put_.py
```diff
@@ -0,0 +1,194 @@
...
+    # Source is loaded with the same flat element offset as the index/source.
+    src_val = tl.load(source_ptr + offsets, mask=final_mask, other=0)
+
+    if IS_ACCUMULATE:
+        tl.atomic_add(inp_ptr + inp_offsets, src_val, mask=final_mask, sem="relaxed")
+    else:
+        tl.store(inp_ptr + inp_offsets, src_val, mask=final_mask)
+
+
+def _put_impl(inp, index, source, accumulate=False):
```
Caeruleann: Please add the native overlap checks before launching the kernel. Expanded inputs or index/source tensors that alias self can currently produce racing writes instead of raising as PyTorch does.

tests/test_put_.py
Caeruleann, line 1: Native CUDA put_ supports integer, floating-point, bool, and complex dtypes, but the current implementation/tests only cover a subset. Please add FP64, integer accumulate, bool, and complex parity cases and implement the unsupported paths.

src/flag_gems/ops/put_.py
```diff
@@ -0,0 +1,194 @@
...
+            f"self.dtype = {inp.dtype} and source.dtype = {source.dtype}"
+        )
+    if index.numel() != source.numel():
+        raise IndexError(
+            "put_(): Expected source and index to have the same number of elements, "
+            f"but got source.numel() = {source.numel()}, index.numel() = {index.numel()}"
+        )
+
+    rank = inp.ndim
+    assert rank <= 5, "put_ only supports tensors with rank <= 5"
```
Caeruleann: put_ supports arbitrary-rank tensors, but this unconditional assertion rejects every valid input above rank 5. Please remove the limitation and add contiguous/non-contiguous 6D+ parity cases.

src/flag_gems/ops/put_.py
```diff
@@ -0,0 +1,194 @@
...
+
+    return inp
+
+
+def put_(inp, index, source, accumulate=False):
+    logger.debug("GEMS PUT_")
+    if index.device != inp.device:
+        index = index.to(inp.device)
+    if source.device != inp.device:
+        source = source.to(inp.device)
```
Caeruleann: PyTorch requires self, index, and source to be on the same device, but this wrapper silently transfers mismatched inputs. Please validate the devices and raise the native error instead.

tests/test_put_.py
```diff
@@ -0,0 +1,188 @@
...
+    ref_index = utils.to_reference(index)
+    ref_source = utils.to_reference(source)
+
+    ref_inp.put_(ref_index, ref_source, accumulate=accumulate)
+    flag_gems.put_(inp, index, source, accumulate=accumulate)
+
+    if accumulate:
+        utils.gems_assert_close(inp, ref_inp, dtype, atol=accumulate_atol(dtype))
+    else:
+        utils.gems_assert_close(inp, ref_inp, dtype)
```
Caeruleann, line 81: Please capture the return value and assert result is inp so the in-place identity contract is covered.

src/flag_gems/ops/put_.py
Yukun-Cui: Done, with one deliberate difference from a literal reading of the request that I want to flag.

The empty-tensor case now raises exactly as you describe: `IndexError: put_(): Tried to put elements into an empty tensor`, matching ATen's text, checked on the host before any launch.

For out-of-range indices I did **not** add a host-side raise, because native CUDA does not raise either. `torch.Tensor.put_` on CUDA hits a device-side assert in `aten/src/ATen/native/cuda/IndexKernel.cu:339`:

```
Assertion `idx < numel && idx >= -numel && "cuda_take_put_kernel() index out of bounds"` failed.
```

which surfaces as an `AcceleratorError` ("CUDA error: device-side assert triggered"), asynchronously, not as the `IndexError` the CPU path raises. Matching the CPU behaviour here would mean copying `index` to the host and synchronizing on every call. I measured that at ~30us, which more than doubles small-N latency for a kernel that runs in 21.8us, and it would diverge from the CUDA reference this operator is compared against.

So the kernel now uses `tl.device_assert` with the same condition, plus a validity-masked store. The mask matters independently of the assert: `TRITON_DEBUG` is not always on and the assert can be compiled out, and without the mask the offset was applied unguarded, so an out-of-range index wrote outside `self`. With the mask the write is dropped instead of corrupting memory. This is the same pattern `take.py` and `index_copy_.py` already use in this repo.

Tests cover the empty-tensor `IndexError`, the negative-index wraparound boundaries (`-numel` valid, `-numel-1` not), and `numel-1` versus `numel`. The device-assert path is asserted on rather than executed, since a triggered device assert poisons the CUDA context for the rest of the process.

If you would prefer strict CPU-style `IndexError` semantics even at the cost of a sync, say so and I will add it behind a flag, but as it stands this matches native CUDA.

src/flag_gems/ops/put_.py
Yukun-Cui: Done — the overlap checks now run before the launch.

I implemented Python equivalents of `at::has_internal_overlap` and `c10::get_overlap_status` and matched ATen's actual policy, which turned out to be narrower than "raise whenever there is aliasing":

- **Internal overlap in `self`** (e.g. an `expand`ed view, where several logical elements share one memory location) raises `RuntimeError: unsupported operation: more than one element of the written-to tensor refers to a single memory location ...`.
- **Full or Partial overlap** between `self` and `index`/`source` raises `unsupported operation: some elements of the input tensor and the written-to tensor refer to a single memory location ...`.
- **The `TooHard` band is permitted**, and this is the part worth calling out. ATen deliberately allows non-dense interleaved views that it cannot cheaply prove disjoint — `self = t[::2]` with `source = t[1::2]` runs and produces a defined result on native. Raising there would reject input that currently works. Our kernel would race on it, so that band is staged through a temporary copy of `source` instead: same result as native, no race.

Verified by comparing against native on all three categories: interleaved (both succeed, values equal), full overlap (both raise), internal overlap (both raise).

tests/test_put_.py
Yukun-Cui: Done — the missing dtypes are implemented and covered, not just tested.

Added int8, uint8, int16, bool, complex64 and complex128; FP64 and integer `accumulate=True` now have parity cases.

Two implementation notes:

**Narrow ints and bool under `accumulate=True`.** Triton has no `atomic_add` for int8/uint8/int16/bool — I probed `atomic_or`, `atomic_max` and `atomic_xchg` as fallbacks and none are available for these widths either. Those dtypes therefore accumulate through an int32 staging buffer and are narrowed on the way back. I verified the narrowing reproduces native's wraparound exactly at the dtype boundaries (e.g. int8 127+1 -> -128, uint8 255+1 -> 0) rather than assuming two's-complement truncation would agree. Bool `accumulate=True` is a logical OR on native, and is implemented and tested as such.

**Complex.** Handled via `view_as_real` real/imag views, so a complex64 put becomes a float32 put over an interleaved pair. `accumulate=True` composes correctly because complex addition is componentwise.

All dtype cases assert against native `put_` for both `accumulate` settings.

src/flag_gems/ops/put_.py
Yukun-Cui: Done — the assertion is gone and arbitrary rank works.

The key observation is that `put_` addresses the *flattened* tensor, so rank only matters for translating a flat index back into strided offsets. `_collapse_dims` folds memory-adjacent dimensions, which means any contiguous tensor collapses to rank 1 no matter how many dimensions it nominally has. Only a genuinely non-collapsible view above the constexpr rank cap — a permuted, non-dense 9-D view, say — needs the host-side offset fallback, and that path is now implemented rather than rejected.

Parity cases cover 6-D through 9-D, contiguous and non-contiguous (permuted and sliced), against native.

This is also where I hit a bug that the assertion had been hiding: the rank-9 fallback crashed with `TypeError: got multiple values for argument 'RANK'`, because `shape` was padded up to the cap without being truncated, so extra positional arguments spilled into the constexpr slots. Fixed, and rank 9 is in the test matrix.

src/flag_gems/ops/put_.py
Yukun-Cui: Done — the wrapper no longer transfers anything.

`self`, `index` and `source` are now required to be on the same device, raising `RuntimeError: Expected all tensors to be on the same device, but got index is on cpu, different from other tensors on device cuda:0` — ATen's wording for this mismatch.

While adding it I also put the surrounding validation into ATen's order. That ordering is observable: with an input that violates two rules at once, native reports a specific one, so I probed those combinations and matched the sequence (device, then index dtype, then source dtype, then numel) rather than picking an order. The messages use ATen's scalar-type spelling too — "Float"/"Double", not "torch.float32".

`test_put_device_mismatch` covers a CPU operand against a CUDA `self`, and a second-GPU operand when more than one CUDA device is visible.

tests/test_put_.py
Yukun-Cui: Done — the return value is captured and `assert result is inp` now covers the in-place identity contract.

Applied to the main parametrized test rather than only one case, so identity is checked across dtypes, ranks and both `accumulate` settings. Confirmed it holds, including on the paths that stage through a temporary internally (narrow-int accumulate, complex, and the `TooHard` aliasing band) — those were the ones at risk of accidentally returning the staging buffer instead of `inp`.

src/flag_gems/ops/put_.py
```diff
@@ -0,0 +1,433 @@
...
+    # Load the flat (linear) index into the input tensor, treated as 1-D.
+    idx = tl.load(index_ptr + offsets, mask=mask, other=0).to(tl.int64)
+
+    # Negative index counts back from the end of the flattened tensor.
+    wrapped = tl.where(idx < 0, idx + inp_numel, idx)
+
+    # Bounds check on the *original* index, matching ATen's CUDA assertion
+    # `idx < numel && idx >= -numel`: the valid range is [-inp_numel, inp_numel).
+    index_valid = (wrapped >= 0) & (wrapped < inp_numel)
+    tl.device_assert(index_valid | (~mask), "put_(): index out of range")
```
Caeruleann: tl.device_assert is a no-op unless TRITON_DEBUG=1, so in normal CI and production builds an invalid index is still silently dropped by final_mask instead of triggering the native CUDA error. Please use an always-active device-side failure and add an executable FlagGems invalid-index test in an isolated subprocess so the poisoned device context does not affect the rest of the suite.

src/flag_gems/ops/put_.py
```diff
@@ -0,0 +1,433 @@
...
+        # passing element offsets is cheaper than growing the kernel signature.
+        flat = index_flat.to(torch.int64)
+        flat = torch.where(flat < 0, flat + inp_numel, flat)
+        flat = flat.clamp_(0, max(inp_numel - 1, 0))
+        offsets_arg = torch.zeros_like(flat)
+        cur = flat
+        for d in reversed(range(rank)):
+            offsets_arg += (cur % shape[d]) * stride[d]
+            cur = cur // shape[d]
+        rank = 0  # decomposition already done
```
Caeruleann: Please move the high-rank offset calculation out of the host wrapper. This path launches torch.where, clamp_, remainder, multiplication, addition and division as PyTorch computations, which violates the Host-function rule and adds several device launches before the Triton kernel. Please calculate these offsets in Triton or route through existing FlagGems operators.

src/flag_gems/ops/put_.py
```diff
@@ -0,0 +1,433 @@
...
+        return inp
+
+    if accumulate and dtype is torch.bool:
+        # `accumulate=True` on bool is a logical OR in native (measured: any
+        # True source lands as True, an already-True slot stays True). int32
+        # atomic_add then `!= 0` reproduces it without a bool atomic.
+        staged = inp.to(torch.int32)
+        _launch(staged, index_flat, source_flat.to(torch.int32), True)
+        inp.copy_((staged != 0).reshape(inp.shape))
+        return inp
```
Caeruleann: The int32 staging itself is necessary for unsupported narrow atomics, but please implement the casts, bool normalization, and write-back with Triton kernels or existing FlagGems operators instead of native .to(), .copy_(), and staged != 0 calls.

tests/test_put_.py
```diff
@@ -0,0 +1,607 @@
...
+
+    utils.gems_assert_close(inp, ref_inp, dtype, atol=accumulate_atol(dtype))
+
+
+# ---------------------------------------------------------------------------
+# dtype coverage: integer, bool and complex, with and without accumulate.
+# ---------------------------------------------------------------------------
+# `accumulate=True` has no Triton atomic for int8/uint8/int16/bool, so those
+# dtypes take a staged int32 path; the parametrisation below covers both sides.
+_PUT_ALL_DTYPES = (
```
Caeruleann, line 204: Native CUDA rejects complex32, but the implementation has no matching guard and routes it through the FP16 real-view path. Please add exception-parity coverage and explicitly reject it, or document and test the intended extension.

src/flag_gems/ops/put_.py
Yukun-Cui: Fixed in 0d34f768. `tl.device_assert` is replaced by an unconditional device trap: the kernel reduces the out-of-range lanes to one bit and emits the PTX `trap` instruction via `inline_asm_elementwise`, which stays active in every build (verified: with default settings a `tl.device_assert` failure does not fire, the trap does). The write stays gated on `final_mask`, so a bad index still cannot touch memory outside `self`.

The invalid-index test is now an executable isolated-subprocess test as you asked: it asserts (a) the failure is a device-side assert and (b) a valid `put_` call does not synchronize on the success path. 148 tests pass.

src/flag_gems/ops/put_.py
Yukun-Cui: Fixed in 0d34f768. The `>MAX_UNROLL_RANK` offset calculation is gone from the host wrapper. `_launch` now passes the layout vector (collapsed sizes followed by strides) and the kernel decodes the flat index with a `RANK_DYN`-trip `tl.static_range` loop, so no `torch.where`/`clamp_`/`remainder`/multiplication/addition/division runs on the host and no extra device launches precede the scatter. The rank-9 permuted case in `test_put__high_rank` still passes.

src/flag_gems/ops/put_.py
Yukun-Cui: Fixed in 0d34f768. The staging path no longer uses native `.to()`, `.copy_()` or a staged `!= 0`: `put_stage_widen_kernel` and `put_stage_narrow_kernel` perform the int32 widen (with explicit 0/1 bool normalization) and the narrowing write-back on the device. The int32 staging itself is unchanged, since narrow atomics remain unsupported.

tests/test_put_.py
Yukun-Cui: Fixed in 0d34f768. `complex32` is now rejected explicitly with ATen's exact message (`NotImplementedError: "put_cuda" not implemented for 'ComplexHalf'`), so it no longer falls through to the FP16 real-view path. Verified against the native CUDA call. Added `test_put__complex32_rejected`, which asserts both the exception type and the message match the native op.

src/flag_gems/ops/put_.py
```diff
@@ -0,0 +1,538 @@
...
+        _launch(re, index_flat, src_real[..., 0].contiguous(), accumulate)
+        _launch(im, index_flat, src_real[..., 1].contiguous(), accumulate)
+        return inp
+
+    if accumulate and dtype in _NARROW_INT_DTYPES:
+        # Triton has no atomic_add for int8/uint8/int16 (and none for bool).
+        # Accumulate into an int32 staging buffer, then narrow back. Both the
+        # widen and the narrow run as Triton kernels: the narrowing store wraps
+        # exactly like native's integer accumulate.
+        staged = _stage_to_int32(inp)
```
Caeruleann, line 658: The int32 staging assumes inp is contiguous: _stage_to_int32 reads and _stage_back writes with unit stride, but put_ indexes self by its row-major logical flattening — so a strided self (e.g. base.t()) with accumulate=True corrupts exactly the indexed elements. Reproduced for int8/uint8/int16/bool (non-accumulate is fine since the scatter kernel itself is stride-aware). Please stage via inp.contiguous() and write back stride-aware (like put does), plus add a non-contiguous narrow-int accumulate test.

Yukun-Cui: 已修（2f28fb87）。复现确认：`base.t()` 这类 strided `self` 在 `accumulate=True` 下，int8/uint8/int16/bool 四个 dtype 都会写错元素（非 accumulate 确实正常，因为 scatter kernel 本身按 stride 解码）；实测还出现了把值写到索引之外位置的越界表现。

改法与你说的一致：暂存按行主序逻辑顺序、写回 stride-aware。

- `_stage_to_int32`（widen）现在读 `self` 时按真实 stride 解码——原来按 unit stride 读，strided 输入取到的根本不是那些元素。
- `_stage_back`（narrow）写回时同样按 `dst` 的真实 strides 解码。新增 `_stage_layout` 把 layout 向量（collapsed sizes + strides）与 launch 几何抽出来，两个 kernel 共用，逻辑与 scatter 用的那套完全一致。

验证：4 dtype × 5 布局（contig/transpose/slice/narrow+permute/row-slice）× 两种 accumulate = 40/40 与 native 相等。

新增 `test_put__non_contiguous_narrow_accumulate`（四个 staged dtype × 四种非连续布局，`accumulate=True`，与 native 逐元素比对）。`tests/test_put_.py`: 164 passed（`--ref=cpu --quick` 110 passed）。

conf/operators.yaml
```diff
@@ -11928,6 +10268,17 @@ ops:
...
+  description: The in-place version of `put()`.
+  for:
+  - put_
+  labels:
+  - aten
+  - KernelGen
+  kind:
+  - Math
+  stages:
+  - alpha: '5.4'
```
```suggestion
  - alpha: '5.5'
```

Caeruleann (dismissed):

huangyiqun (dismissed):

huangyiqun:
Plz fix CI failed.

PR #6051, opened 2026-09-07
[KernelGen] align cudnn_attention_forward logsumexp shape with PyTorch version

freaking-man-in:
PyTorch's aten::_cudnn_attention_forward returns different logsumexp
shapes across versions:

- PyTorch < 2.9: [B, H, S_q]
- PyTorch >= 2.9: [B, H, S_q, 1]

Align the FlagGems output shape with the installed PyTorch version
instead of always forcing either 3-D or 4-D behavior.

Verified locally with PyTorch 2.8:
48 passed, 16 skipped.

### DIFF
--- a/src/flag_gems/ops/cudnn_attention_forward.py
+++ b/src/flag_gems/ops/cudnn_attention_forward.py
@@ -17,6 +17,7 @@ import logging
 import math
 
 import torch
+from packaging import version
 
 from flag_gems.ops.attention import scaled_dot_product_attention_forward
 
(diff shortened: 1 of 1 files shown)


src/flag_gems/ops/cudnn_attention_forward.py
```diff
@@ -86,8 +86,7 @@ def cudnn_attention_forward(
             dtype=torch.float32,
         )
 
-    # cuDNN convention: logsumexp shape is [B, H, S_q, 1]
-    lse = lse.unsqueeze(-1)
+    # cuDNN convention: logsumexp shape is [B, H, S_q] (3-D, no trailing dim)
```
0x45f: Image

why gems_lse_f and ref_lse_f shape are both torch.Size([2, 4, 512, 1])

```
(Pdb) gems_lse_f.shape
torch.Size([2, 4, 512, 1])
(Pdb) ref_lse_f.shape
torch.Size([2, 4, 512, 1])
```

0x45f (approved):

PR #6061, opened 2026-09-07
[BugFix] Support boolean slice views

### LINKED ISSUE #6049: [Hygon] slice rejects torch.bool and blocks vLLM request preparation
### Describe the bug

On Hygon, enabling FlagGems makes a valid boolean tensor slice fail with:

~~~text
AssertionError: slice: unsupported dtype torch.bool
~~~

This blocks vLLM request preparation: its boolean `discard_request_mask` is sliced before being copied to the GPU. Both eager and graph execution fail on the first request. Graph capture itself completes successfully.

Tested FlagGems commit: **fbd31958abb37a1c777d436f4acaf3e2d9f3bf10**.

### Minimal reproduction

No model weights or vLLM are required:

~~~python
import os
os.environ["USE_FLAGTUNE"] = "0"  # Isolate this from the separate manifest configuration issue.

import torch
import flag_gems

x = torch.tensor([True, False, True], device="cuda", dtype=torch.bool)
print(x[:2].cpu().tolist())  # Native PyTorch succeeds: [True, False]

with flag_gems.use_gems():
    y = x[:2]               # AssertionError with the tested FlagGems commit
~~~

Observed traceback:

~~~text
  File "repro_bool_slice.py", line 8, in <module>
    y = x[:2]
  File ".../flag_gems/ops/slice.py", line 199, in slice
    assert input_tensor.dtype not in (
AssertionError: slice: unsupported dtype torch.bool
~~~

Expected: return a boolean view equivalent to native PyTorch slicing, preserving shared-storage semantics.

### Environment

- Hygon BW1000 / gfx936, DTK 26.04
- Python 3.10; PyTorch 2.10.0
- FlagGems commit fbd31958abb37a1c777d436f4acaf3e2d9f3bf10
- Non-editable source installation: `python3.10 -m pip install --no-deps --no-build-isolation .`
- Reproduced with both FlagTree 0.6.0+hcu3.6 and 0.6.2a1+hcu3.6
- `USE_FLAGTUNE=0` in all cross-version tests

### Cross-version verification

The same minimal reproduction was run on the same device:

| FlagGems | FlagTree | Result |
|---|---|---|
| Previously installed package, metadata version 0.1.0 | 0.6.0+hcu3.6 | PASS |
| Previously installed package, metadata version 0.1.0 | 0.6.2a1+hcu3.6 | PASS |
| fbd31958abb37a1c777d436f4acaf3e2d9f3bf10 | 0.6.0+hcu3.6 | FAIL: same bool dtype assertion |
| fbd31958abb37a1c777d436f4acaf3e2d9f3bf10 | 0.6.2a1+hcu3.6 | FAIL: same bool dtype assertion |

The earlier package's upstream commit is unknown; this is not a historical bisect. The evidence localizes this failure to FlagGems rather than the FlagTree upgrade.

### Source and integration impact

The [dtype guard in slice.py](https://github.com/flagos-ai/FlagGems/blob/fbd31958abb37a1c777d436f4acaf3e2d9f3bf10/src/flag_gems/ops/slice.py#L194) explicitly rejects bool. The implementation below it constructs a view using `torch.as_strided`.

As an additional isolation check, extracting the unchanged `slice` function and calling it on a CPU boolean tensor without GPU devices reproduces this Python assertion. Native `torch.as_strided` succeeds and preserves shared storage. No Triton kernel execution is needed to trigger the failure.

Observed integration call chain with vLLM 0.24.0+empty / vLLM-Plugin-FL 0.3.0rc0.post1+g396920a4c:

~~~text
model_runner._prepare_inputs
  self.discard_request_mask.copy_to_gpu(num_reqs)
vllm/v1/utils.py:142
  self.gpu[:n].copy_(self.cpu[:n], non_blocking=True)
flag_gems/ops/slice.py:199
  AssertionError: slice: unsupported dtype torch.bool
~~~

This reproduced for Qwen3.6-27B and Qwen3.6-35B-A3B in both eager and graph modes (TP=4, max_model_len=4096, max_num_seqs=4, gpu_memory_utilization=0.8). The `BLOCK_M` error targeted by #6040 did not recur in these runs, but this subsequent request-processing failure remains. This report does not claim #6040 introduced the dtype restriction.

### Suggested investigation

Please consider supporting bool view slicing, or falling back to the native implementation for unsupported dtypes. Regression coverage should include boolean request masks, empty slices, noncontiguous inputs, nonzero storage offsets, and aliasing semantics.

The guard also lists complex dtypes, but those were not tested in this report.

physics31415926:
## PR Category
Bug fix

## Type of Change
Boolean slice support and regression coverage.

## Description
The slice implementation explicitly rejected torch.bool even though its existing as_strided implementation supports boolean views. This assertion blocked boolean state slicing during model execution.

This PR removes the boolean exclusion, retains the complex-dtype checks, and adds 30 regression cases in the existing tests/test_slice.py covering slice bounds, layouts, and alias behavior. Tests are kept outside the runtime package.

### Separate integration root cause
The Qwen3.6-27B eager corruption was traced to a stream-ordering race in the installed FlagTree 0.6.2a1+hcu3.6 implementation of triton.testing.do_bench_cudagraph. Its newly created benchmark stream launched warmup without waiting for queued caller-stream work. Autotuning output could reuse storage still accessed by preceding operations; subsequent synchronization cannot undo the corruption.

The validation image separately captures the caller stream and calls benchmark_stream.wait_stream(caller_stream) before warmup. This creates a GPU dependency without host-blocking global synchronization. **This FlagTree runtime patch is outside this PR's diff. The integration results below validate the boolean slice fix together with that separate patch, not the boolean fix alone.**

## Related Issue
Fixes #6049 (boolean slice assertion).

## Progress
- [x] Remove the boolean restriction.
- [x] Keep regression coverage in tests/test_slice.py.
- [x] Validate boolean views and model integration.

## Validation
- Command: python3.10 -m pytest -q -p no:cacheprovider tests/test_slice.py -k bool_slice_view
- Result: **30 passed, 624 deselected**.
- Applicable pre-commit checks on the source and test files passed.
- Independent stream-ordering reproduction without model execution or FlagGems registration: original runtime **7/8 failures**, patched runtime **8/8 passed**, covering default and non-default caller streams.
- Resetting only in-process autotuning caches reproduced corruption. Synchronizing before benchmarking removed it; removing synchronization reproduced it again.

| Model / mode | Result |
| --- | --- |
| Qwen3.6-27B eager, independent cold start 1 | PASS |
| Qwen3.6-27B eager, independent cold start 2 | PASS |
| Qwen3.6-27B graph | PASS |
| Qwen3.6-35B-A3B eager | PASS |
| Qwen3.6-35B-A3B graph | PASS |

Each run checked exact math output (4), a Chinese capital answer (北京), a coherent English explanation, and a repeated math request. All completed normally. Cold starts used fresh processes and separate FlagGems autotuning caches; Triton kernel and vLLM compilation caches were reused. No diagnostic hooks or additional operator exclusions were used.

Configuration: TP=4, GPU memory utilization 0.80, max model length 4096, max sequences 4, max batched tokens 2048, seed 42, prefix caching disabled, default async scheduling, temperature 0, max output tokens 128, thinking disabled. VLLM_PLUGINS=fl, USE_FLAGTUNE=0.

Validation environment: FlagGems upstream base d812184188dfa11457cbf82b0b77efceb0c7aa71 plus boolean fix e4a39a82b923a2404090175e1f734aa948331c45; FlagTree 0.6.2a1+hcu3.6 with the separate local runtime patch; Triton 3.6.0; vLLM 0.24.0. The image's FlagGems commit differs from the PR head due to subsequent test organization changes.

## Performance
No performance benchmark or full model accuracy benchmark was run. These are functional and targeted regression checks; they do not imply all upstream CI jobs have completed.


tests/test_slice_bool.py
```diff
@@ -0,0 +1,61 @@
...
+    [
+        (0, 0, 2, 1),
+        (1, 1, 5, 2),
+        (0, -3, -1, 1),
+        (1, 3, 1, 1),
+        (0, 0, 0, 1),
+        (1, None, None, 1),
+    ],
+)
+def test_bool_slice_view(layout, dim, start, end, step):
```
0x45f: move to test_slice.py

tests/test_slice.py
```diff
@@ -104,3 +104,45 @@ def test_slice_negative_indices(shape, dtype):
...
+
+    assert actual.dtype == torch.bool
+    assert torch.equal(actual, expected)
+    assert actual.shape == expected.shape
+    assert actual.stride() == expected.stride()
+    assert actual.storage_offset() == expected.storage_offset()
+    assert actual.untyped_storage().data_ptr() == inp.untyped_storage().data_ptr()
+    if actual.numel():
+        actual.logical_not_()
+        assert torch.equal(actual, expected)
```
0x45f: please use gems_assert_equal

ZenithJoyH (approved):

0x45f (approved):

Issue #5887, opened 2026-09-01
[bug]_amp_foreach_non_finite_check_and_unscale_kernel‎ kernel中存在逻辑错误


yudong-hang:
### Describe the bug

# Check for non-finite values (inf or nan) using float32
is_non_finite = ~tl_extra_shim.finitef(inp_fp32)
# Scale the values: only finite values are scaled
# Non-finite values stay as-is (in original dtype)
scaled_fp32 = tl.where(is_non_finite, inp_fp32, inp_fp32 * scale_fp32)

这段逻辑中 is_non_finite的值恒为True，
tl_extra_shim.finitef(inp_fp32)返回的是0或者1的int类型的值，
这里使用按位取反 ~ 对 1/0 进行操作：
~1 → -2（在补码中），tl.where 会将非零视为 True
~0 → -1，同样为True

需要将is_non_finite = ~tl_extra_shim.finitef(inp_fp32)改为is_non_finite = tl_extra_shim.finitef(inp_fp32) == 0则能保证逻辑正确。

bin913:
Thanks for the report 鈥?the code-level analysis is correct and we agree the expression is fragile.

I could not reproduce a runtime failure on NVIDIA H100 / triton 3.6.0 at current master (ece4713ae): in this stack `tl_extra_shim.finitef()` is coerced to an `int1` predicate, so `~` happens to act as logical negation and the kernel behaves correctly (verified with the existing `tests/test_amp_foreach_non_finite_check_and_unscale_.py` suite plus direct operator probes). On triton versions/backends where the predicate keeps a wider 0/1 integer return type, `~0 == -1` and `~1 == -2` are always non-zero and `inv_scale` would never be applied, exactly as you describe.

I've opened PR #6066 which changes the predicate to `finitef(inp_fp32) == 0` 鈥?this is correct for both `int1` and wider-integer return types, and does not change behavior on the current `int1` path (6/6 existing tests still pass).

Could you share which triton version/backend surfaced the wider integer predicate, so we can confirm whether the PR fully resolves it on your side? If you no longer hit it, feel free to close.

PR #6066, opened 2026-09-07
fix(amp): compare finitef result to zero instead of bitwise-not (#5887)

bin913:
body written
xes #5887.

`_amp_foreach_non_finite_check_and_unscale_kernel` computes the non-finite predicate with

```python
is_non_finite = ~tl_extra_shim.finitef(inp_fp32)
```

`finitef()` returns a 0/1 predicate. On triton >= 3.6 the CUDA libdevice shim coerces the result to `int1`, where bitwise-not coincides with logical negation, so the current code happens to work. On triton versions/backends where the predicate is surfaced as a wider integer, `~1 == -2` and `~0 == -1` are always non-zero, so `tl.where(is_non_finite, inp_fp32, inp_fp32 * scale_fp32)` always keeps the original value and `inv_scale` is never applied.

### Reproduction status

We ran the existing suite (`tests/test_amp_foreach_non_finite_check_and_unscale_.py`) plus direct operator probes on NVIDIA H100 / triton 3.6.0 against current `master` (ece4713ae). With the `int1` coercion in this triton build the current expression is correct there, so the runtime failure could not be reproduced on this stack; the defect is exposed on configurations where `finitef` keeps a wider integer return type.

### Fix

```python
is_non_finite = tl_extra_shim.finitef(inp_fp32) == 0
```

`== 0` yields the correct predicate on every variant (int1 or wider 0/1 integer) and does not change behavior on the `int1` path.

### Verification

- `tests/test_amp_foreach_non_finite_check_and_unscale_.py`: 6 passed before and after the change on NVIDIA H100.
- Direct probe: an all-finite fp32 tensor scaled with `inv_scale = 2.0` produces `[2.0, 4.0, 8.0]`; tensors containing `inf`/`nan` keep those elements and scale the finite ones, and `found_inf` is set, matching the PyTorch reference.

### DIFF
--- a/src/flag_gems/ops/_amp_foreach_non_finite_check_and_unscale_.py
+++ b/src/flag_gems/ops/_amp_foreach_non_finite_check_and_unscale_.py
@@ -48,8 +48,13 @@ def _amp_foreach_non_finite_check_and_unscale_kernel(
     inp_fp32 = inp.to(tl.float32)
     scale_fp32 = scale.to(tl.float32)
 
-    # Check for non-finite values (inf or nan) using float32
-    is_non_finite = ~tl_extra_shim.finitef(inp_fp32)
+    # Check for non-finite values (inf or nan) using float32.
+    # finitef() may be lowered to an int1 (triton >= 3.6) or to a wider int
+    # (0/1) on other versions/backends.  Bitwise-not (~) is only equivalent to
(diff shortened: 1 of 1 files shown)


huangyiqun:
Please also update the Hygon and Cambricon implementations, which still use the same `~finitef(...)` pattern.

PR #6067, opened 2026-09-07
[KernelGen][Nvidia] Add hamming_window operator with Triton kernel

ShawnsYing:
## Summary
Adds a Triton kernel for `hamming_window`. Compute the Hamming window function with four operator variants matching PyTorch ATen overloads.

## Testing
- Validated against reference on device via direct FlagGems API calls
- Tested on: Nvidia

**Results: 114 tests passed**
- `test_hamming_window`: 12 tests (4 sizes × 3 dtypes)
- `test_hamming_window_periodic`: 48 tests (8 sizes × 2 periodic × 3 dtypes)  
- `test_hamming_window_periodic_alpha`: 36 tests (3 sizes × 2 periodic × 2 alpha × 3 dtypes)
- `test_hamming_window_periodic_alpha_beta`: 18 tests (3 sizes × 2 periodic × 1 alpha × 1 beta × 3 dtypes)

## Performance
Performance benchmarks will be automatically triggered by CI for `benchmark/test_hamming_window.py`.

The implementation uses:
- BLOCK_SIZE=1024 for balanced GPU occupancy
- Promotes FP16/BF16 to FP32 for cosine computation to maintain numerical accuracy
- Implements PyTorch's fixed beta=0.46 (not generalized beta=1-alpha)

## Multi-backend Testing
|---|---|---|---|
| Nvidia | PASS (114 cases) | TBD | Primary |
| Tianshu | N/A | — | — |
| Muxi | N/A | — | — |
| Ascend | N/A | — | — |
| Hygon | N/A | — | — |

## Files Changed
- `src/flag_gems/ops/hamming_window.py`: Triton kernel implementation (245 lines)
- `tests/test_hamming_window.py`: Accuracy test (127 lines)
- `benchmark/test_hamming_window.py`: Performance benchmark
- `conf/operators.yaml`: Add operator entries (kind: Math, stage: alpha 5.4, 4 variants)


bin913:
@ShawnsYing Please add benchmark performance result

ShawnsYing:
Benchmark results for `hamming_window` (and its `hamming_window_periodic` variant), as requested.

**Environment**: NVIDIA H20-3e, `mode=kernel` (triton do_bench, median), comprehensive level, `alpha=0.54, beta=0.46`.

**`hamming_window` (`periodic=True`; `periodic=False` measured identically within noise):**

| window_length | torch fp32 | gems fp32 | **fp32** | torch fp16 | gems fp16 | **fp16** |
|---|---|---|---|---|---|---|
| 4,096 | 14.3 us | 5.7 us | **2.52x** | 14.4 us | 5.5 us | **2.62x** |
| 10,000 | 15.6 us | 5.5 us | **2.83x** | 15.4 us | 5.5 us | **2.81x** |
| 2,560,000 | 54.4 us | 10.3 us | **5.26x** | 50.0 us | 12.0 us | **4.18x** |
| 16,777,216 | 311.8 us | 39.0 us | **7.99x** | 243.8 us | 47.6 us | **5.12x** |
| 268,435,456 | 4,709 us | 544.5 us | **8.65x** | 3,992 us | 679.6 us | **5.87x** |
| 655,360,000 | 11,476 us | 1,321 us | **8.69x** | 9,718 us | 1,651 us | **5.89x** |
| 1,073,741,824 (1 GiB fp32) | 18,799 us | 2,161 us | **8.70x** | 15,907 us | 2,702 us | **5.89x** |

bf16 tracks fp16 within ~2%. The three API variants (`hamming_window`, `_periodic`, `_periodic_alpha`, `_periodic_alpha_beta`) all dispatch to the same kernel and measure identically.

**Reading of the numbers:**
- Small windows (4K-10K) are launch-bound: both sides sit near the ~5 us kernel-launch floor and we are 2.5-2.8x ahead because the eager path pays a second CPU-side setup step; there is no headroom left here for either implementation.
- From ~2.5M elements up the op is a pure streaming write and the gap widens monotonically to **8.7x (fp32)** / **5.9x (fp16)**, which is where the bandwidth advantage shows.
- fp16 is a *smaller* speedup than fp32 despite moving half the bytes, because our kernel computes the window in fp32 and narrows on store, so the fp16 row is the same element throughput at half the traffic — matching what the eager path does too.

No code change in this push; these are the current branch measured. If you want the raw comprehensive dump (144 rows, all 6 op/dtype combinations) or an `--mode operator` end-to-end table as well, say so and I will add it.

ShawnsYing:
Correction to one explanation in my previous comment (the numbers stand; only my stated reason for the fp16 gap was wrong).

I wrote that "our kernel computes the window in fp32 and narrows on store" — that is not what the code does, and the file's own comment says so explicitly. The low-precision path (`LOW_PRECISION`) rounds back to the output dtype **after every elementary step**:

```python
if LOW_PRECISION:
    n = idx.to(dtype).to(tl.float32)
    x = (n * scale).to(dtype).to(tl.float32)
    x2 = (x * 2.0).to(dtype).to(tl.float32)
    c = tl.cos(x2.to(dtype).to(tl.float32)).to(dtype).to(tl.float32)
    term = (c * beta).to(dtype).to(tl.float32)
    val = (alpha - term).to(dtype)
```

The reason is deliberate and stated in the source: `tl.cos` requires fp32, so the kernel computes in fp32 but re-rounds to the output dtype after each step to reproduce torch's op-by-op evaluation **bit-for-bit** for fp16/bf16. So the fp16 row is doing the same number of fp32 transcendental evaluations as fp32 (same element throughput) while moving half the bytes — which is what makes its speedup look *smaller* despite less traffic, and it is a correctness-driven choice, not an accident of the implementation.

Everything else in the previous comment is unchanged: the measured table (2.52-2.83x at 4K-10K, rising to 8.70x fp32 / 5.89x fp16 at 1 GiB), the launch-bound explanation for the small sizes, and the bf16-tracks-fp16 note.

bin913 (approved):

PR #6091, opened 2026-09-08
[kunlunxin] add special_legendre_polynomial_p operator with Triton ke…

Cocytus486:
Replace the fully-unrolled `tl.static_range(2, 256)` recurrence loop in
`special_legendre_polynomial_p`'s Triton kernel with a runtime loop bounded
by the degree `n` (read from a scalar tensor).

The Legendre recurrence

    P_k(x) = ((2k - 1) * x * P_{k-1}(x) - (k - 1) * P_{k-2}(x)) / k

is mathematically identical either way, but the static_range(2, 256) variant
unrolls 254 iterations regardless of the actual degree, which was measured
to take >15 min to compile on XPU. Switching to `for degree in range(2, n + 1)`
keeps the loop dynamic; degrees n < 2 are still handled by the existing
`tl.where` fast paths and skip the loop entirely, so numerical behavior is
unchanged.

Only float32 is supported, matching torch.special.legendre_polynomial_p on
this backend.

### PR Category
Operator

### Type of Change
Performance Optimization

### Description
Fix excessive Triton compile time (>15 min) for special_legendre_polynomial_p
on Kunlunxin XPU by replacing the static_range-unrolled loop with a dynamic
loop bounded by the runtime degree n.

### Progress

### Performance
Compile time reduced from >15 min (static_range(2, 256), 254x unrolled) to
seconds (dynamic loop), verified locally on Kunlunxin XPU. No change to
output values for the same (x, n) inputs.


103yiran (approved):
LGTM

PR #6093, opened 2026-09-08
[kunlunxin] move the copy family onto tle.gpu

Ason93:
alias_copy, copy_/copy, permute_copy and unfold_copy are pure moves, so they are handed to tle.gpu instead of the generated pointwise kernels. The new utils/tle_copy.py picks between two kernels:

* both sides contiguous and same dtype -> one flat TMA tile (GM -> LM -> GM)
* everything else -> tle.gpu.copy with a GM pointer tensor, which lowers to an element-wise gather (g2l) / scatter (l2g); permuted, broadcast, strided and overlapping-window reads all land here, and with CONVERT the dtype cast is done in LM so a dtype-changing copy_ stays on the same kernel

torch.bool has no usable LM representation (i1 is packed there), so bool moves through an int8 view. Copies below 2**14 elements are host-bound -- the triton launch floor alone is ~21us while the native op does a 64x64 copy in ~8us -- so those are redispatched to PyTorch; broadcast is excluded from that shortcut because the native XPU copy_ only fills the first row of a broadcast source.

unfold_copy is new for this backend: the generic kernels reject 1D input and mismatch on three of the tested shapes, both of which the tle path handles.

tests/test_copy.py gets the kunlunxin skip marks for the two float8_e8m0fnu cases, matching the other vendors that lack the dtype: triton cannot lower float8, so those copies redispatch to the native op, which raises NOT IMPLEMENTED.

Measured on KL3 (benchmark/test_copy.py --level core, equal-weight dtype mean of Gems Speedup): copy_ 0.9971x, copy 0.9700x (was 0.7147x). Accuracy on the same device: test_alias_copy 36/36, test_permute_copy 18/18, test_unfold_copy 24/24 (9 pre-existing failures fixed), test_contiguous 30/30, test_copy 40 passed with 2 known failures where the reference itself is wrong (the native XPU broadcast copy_ only writes the first row).

Requires the companion triton change that drops the NVIDIA-only 16-byte base alignment assert from TensorDescriptor; XPU has no such requirement.

### Progress

### DIFF
--- a/src/flag_gems/runtime/backend/_kunlunxin/ops/alias_copy.py
+++ b/src/flag_gems/runtime/backend/_kunlunxin/ops/alias_copy.py
@@ -1,14 +1,12 @@
 import logging
 
 import torch
-import triton
-import triton.language as tl
+import triton  # noqa: F401
+import triton.language as tl  # noqa: F401
 from _kunlunxin.utils.codegen_config_utils import CodeGenConfig
 
-from flag_gems.runtime import torch_device_fn
-from flag_gems.utils import libentry
-
 from ..utils.pointwise_dynamic import pointwise_dynamic
+from ..utils.tle_copy import tle_copy
 
 logger = logging.getLogger("flag_gems").getChild(__name__.lstrip("."))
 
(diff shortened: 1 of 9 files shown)


103yiran (approved):
LGTM

PR #6136, opened 2026-09-10
[KMCompiler][Ascend] Add ascend backend for igammac 

wuhansky:
### PR Category
Operator

### Type of Change
New Feature

### Description
Add the Ascend-specialized backend for igammac (regularized upper incomplete
gamma Q(a, x)).

Background: the generic implementation produces NaN / wrong results for large
a on Ascend because two backend intrinsics are broken:

tl_extra_shim.lgamma returns inf for a in [30, 95] and NaN for a >= 100
(measured on triton-ascend 3.2.0; correct values e.g. lgamma(100)=359.1);
tl_extra_shim.log1p loses relative precision for tiny |s| (~1e-3 rel at
|s|~1e-3), which corrupts the asymptotic-path quantity log(1+sigma)-sigma in
the cancellation regime (large a with x ~ a).
Implementation (src/flag_gems/runtime/backend/_ascend/ops/igammac.py):

Minimal adaptation of the generic kernel (series / continued-fraction /
asymptotic expansion with d-coefficients, boundary handling), with two fixes:
self-contained Lanczos (g=7, n=9) lgamma — validated against
torch.lgamma over a in [0.5, 10000] to ~5e-7 relative;
log(1+s)-s computed via its Taylor series instead of the inaccurate
backend log1p (same approach the generic kernel already uses for fp64).
No PyTorch computation ops in host code (contribution-guide compliant);
registered in _ascend/ops/__init__.py (import + __all__).
Tests: the generic tests/test_igammac.py (full-domain, boundaries, inf/nan,
series/CF switchover, log-uniform) passes 120/120 on Ascend unchanged.
Benchmark: torch.igammac has no native AI Core kernel in torch_npu eager mode
(CPU fallback, msprof-verified: no device kernel, and the NPU call is slower
than pure CPU), and CANN 8.5/9.0 ship no native Igammac either. Following
polygamma's composed-baseline approach, benchmark/test_igammac.py builds a
device-native reference on ascend from torch primitives (all AI Core native):
Q = 1 - exp(a ln x - x - lgamma(a)) * sum_{i<50} x^i/(a)_i, with log-gamma
inlined via Lanczos (torch.lgamma also falls back to CPU). The term count (50)
matches the kernel's series branch so the comparison is fair, and the composed
reference is validated to ~4e-6 over the benchmark input domain a,x in
[0.1, 10.1], which exercises the series and continued-fraction branches.
The asymptotic branch (large a ~ x) is not timed on ascend: a fixed-term
composed series cannot represent it (needs ~O(sqrt(a)) terms, see the series
convergence note) without elementwise dynamic branching; its correctness is
covered by the functional tests (log-uniform domain up to a ~ 1e6).

### Issue
N/A

### Progress

### Performance
Ascend 910B4 (fp32). Baseline: the composed device-native
reference above (50 series terms, same as the kernel). Both sides run on AI
Vector Core, so the comparison is on-device and same-work:

shape | elements | baseline (ms) | impl (ms) | speedup
-- | -- | -- | -- | --
[64,64] | 4K | 3.96 | 0.19 | 20.9x |
[10000,1] | 10K | 4.00 | 0.19 | 20.6x |
[100,1,100] | 10K | 4.19 | 0.11 | 37.3x |
[10000,256] | 2.6M | 4.20 | 3.22 | 1.30x |
[100,256,100] | 2.6M | 4.62 | 3.07 | 1.51x |
[4096,4096] | 16M | 39.2 | 20.0 | 1.96x |
[64,512,512] | 16M | 39.0 | 20.0 | 1.96x |


src/flag_gems/runtime/backend/_ascend/ops/igammac.py
```diff
@@ -0,0 +1,1024 @@
...
+            ASYM_N=_ASYM_N,
+        )
+
+    if out_was_noncontig:
+        out.copy_(out_contig)
+    return out
+
+
+def igammac(a: torch.Tensor, x: torch.Tensor, *, out: torch.Tensor = None):
+    logger.debug("GEMS IGAMMAC")
```
Caeruleann:
```suggestion
    logger.debug("GEMS_ASCEND IGAMMAC")
```

src/flag_gems/runtime/backend/_ascend/ops/igammac.py
```diff
@@ -0,0 +1,1024 @@
...
+        if a.numel() != x.numel() or a.numel() != out.numel():
+            raise ValueError(
+                "igammac_out: input and output must have the same number of elements"
+            )
+    _launch_igammac(out, a, x)
+    return out
+
+
+def igammac_out(a: torch.Tensor, x: torch.Tensor, out: torch.Tensor):
+    logger.debug("GEMS IGAMMAC_OUT")
```
Caeruleann:
```suggestion
    logger.debug("GEMS_ASCEND IGAMMAC_OUT")
```

src/flag_gems/runtime/backend/_ascend/ops/igammac.py
wuhansky: done

src/flag_gems/runtime/backend/_ascend/ops/igammac.py
wuhansky: done

Caeruleann (approved):

PR #6156, opened 2026-09-10
docs: drop Apache 2.0 headers added to third-party files

tengqm:
## Problem

https://flagos-ai.github.io/FlagGems/ serves a blank page. The deployed Pages
artifact (run `34423811500`) contains 99 HTML files, **96 of them 0 bytes** —
including `index.html`. The only files with content are `404.html` (2010 B) and
`en/index.html` (264 B, an alias redirect).

## Root cause

`b684fbc78` ("chore: add Apache 2.0 copyright header to all source files") swept
the whole tree without sparing third-party material, and prepended a
`Copyright 2026 FlagOS Contributors` header to it. Those notices misattribute
authorship and are not the project's to add:

| path | what it is |
| --- | --- |
| `docs/themes/hugo-book/` (156 files) | the vendored hugo-book theme — MIT, by Alex Shpak; `theme.toml` says so itself |
| `docs/static/css/tabulator.min.css` | the minified Tabulator stylesheet — also MIT |

The theme headers were not merely cosmetic. Six of the theme's layouts consist
of nothing but `{{ define }}` blocks, and that is load-bearing.

### Why it broke the site

hugo-book deliberately keeps `list.html` / `single.html` as nothing but
`{{ define "dummy" }}{{ end }}` so that Hugo falls back to
`layouts/baseof.html` for the page skeleton. In Go's `html/template`, **any
top-level content outside `{{ }}` becomes page output**, and its presence
suppresses that fallback. The prepended HTML comment is exactly such top-level
content, so `baseof.html` was never applied and those layouts rendered on their
own — as nothing.

The comment body is stripped by the escaper, leaving 3 newline bytes;
`--minify` then collapses those to 0. Confirmed directly:

```
$ xxd public_nomin/index.html
00000000: 0a0a 0a                                  ...
```

A/B build check with hugo 0.157.0 (identical build stats to CI: EN 68 / ZH-CN
60 / static 76 / alias 1):

| first line of `list.html` | `public/index.html` |
| --- | --- |
| the 16-line HTML comment (current) | **0 B** |
| nothing — `{{ define "dummy" }}{{ end }}` only | 12867 B |
| plain text `x` | 1 B |
| `{{/* ... */}}` Go comment | 12867 B |

## Fix

Restore the 157 affected files to their pre-`b684fbc78` state, byte for byte.

```
157 files changed, 2450 deletions(-)
```

No other file is touched. The headers added to the project's own `docs/`,
`docs/layouts/` and `docs/content/` are legitimate and stay as they are.

## Verification

Rebuilt with `hugo 0.157.0`, same flags as CI:

```
cd docs && hugo build --gc --minify --baseURL https://flagos-ai.github.io/FlagGems/
```

| | before | after |
| --- | --- | --- |
| 0-byte HTML files | 96 / 99 | **0 / 99** |
| `index.html` | 0 B | 12860 B |
| `usage/index.html` | 0 B | 9540 B |
| `zh-cn/index.html` | 0 B | 12994 B |
| `404.html` | 2010 B | 2010 B (byte-identical) |
| `css/tabulator.min.css` | 28406 B (16-line header) | 28360 B, byte-identical to upstream |

Titles render (`FlagGems`, `Using FlagGems | FlagGems`), the Overview content is
present, and no license text leaks into any generated page or asset.


103yiran (approved):
LGTM

Issue #6164, opened 2026-09-11
[sunrise] four vendor-override defects found bringing up SGLang 0.5.18 (cumsum empty tensor / asin missing / SDPA enable_gqa / pow fallback)


tengqm:
### Summary

Bringing SGLang 0.5.18 up on Sunrise (PTPU / TANGRT 1.2.0) surfaced four defects in FlagGems' vendor overrides. Each is a case of the override disagreeing with the aten signature/semantics it replaces, or with the triton fork it runs on. All four are currently worked around on our side with a blacklist — the workarounds are listed per item so the impact is visible.

Environment: `flag_gems 5.3.5` (in `flagos-runtime-sunrise-tangrt1.2.0:2.1.2`) and `5.4.0rc1.dev20260910` (daily), torch 2.11.0+cpu with `torch-ptpu`, flagtree 0.6.0+sunrise3.6 (F) / vendor `triton 3.6.0.1+git0a5cfb35` (T).

| # | Defect | Minimal repro | Blocks |
|---|---|---|---|
| 1 | `_sunrise` cumsum divides by an empty dimension | ✅ | every serve (first prefill) |
| 2 | `asin` missing from the patched-symbol list | ✅ | `import flag_gems` on T |
| 3 | `_scaled_dot_product_attention_math` has no `enable_gqa` | ✅ | any caller passing `enable_gqa` |
| 4 | `_fallback_pow` breaks for `scalar ** tensor` | ⚠️ serve-path only | serve on T |

---

### 1. `_sunrise/ops/cumsum.py` divides by zero on an empty tensor

```python
# src/flag_gems/runtime/backend/_sunrise/ops/cumsum.py, cumsum_wrapper
M = 1
N = shape[dim]
for i in range(dim):
    M *= shape[i]
inp = inp.contiguous()
K = inp.numel() // M // N          # N == 0 -> ZeroDivisionError
```

Repro:

```python
import torch, flag_gems
flag_gems.enable()
torch.cumsum(torch.empty(0, dtype=torch.int64, device="ptpu"), dim=0)
# ZeroDivisionError: integer division or modulo by zero
```

The **generic** `src/flag_gems/ops/cumsum.py` has guarded exactly this since #4541 ("Fix cumsum ZeroDivisionError on empty tensor", 2026-08-06, present in v5.3.4+):

```python
if out is None:
    out = torch.empty_like(inp, dtype=dtype)
if inp.numel() == 0:
    return out
```

[...]

PR #6165, opened 2026-09-11
fix(sunrise): guard the vendor cumsum against empty input, add an asin fallback

tengqm:
Fixes items 1 and 2 of #6164 — two of the four defects found bringing SGLang 0.5.18
up on Sunrise (tangrt1.2.0 / PTPU). Items 3 (`enable_gqa`) and 4 (the `pow`
fallback) are left to maintainer judgement on shape; this PR keeps to the two
small, verified ones.

## 1. cumsum divides by zero on an empty tensor

`_sunrise/ops/cumsum.py` computes `K` from the scanned dimension before checking
anything, so a zero-size input raises:

```
K = inp.numel() // M // N
```

The generic `ops/cumsum.py` has guarded exactly this since #4541, but the vendor
copy is a separate implementation and never received the guard.

**This is the normal path, not an edge case.** sglang computes positions during
the first prefill:

```python
# sglang/srt/model_executor/forward_batch_info.py::compute_position_torch
extend_start_loc = torch.zeros_like(extend_seq_lens)
extend_start_loc[1:] = torch.cumsum(extend_seq_lens[:-1], dim=0)
```

For a single-sequence batch `extend_seq_lens[:-1]` is size 0, so serving one
request always hits it.

## 2. `asin` missing, and the borrowed symbol does not lower

`tl_extra_shim` resolves to `triton.language.math` on this fork, which has no
`asin`, and `asin` was not in the tuple `_patch_missing_symbols` walks — so
`ops/arcsin.py`'s module-level `_ASIN = tl_extra_shim.asin` aborted
`import flag_gems` outright:

```
AttributeError: module 'triton.language.math' has no attribute 'asin'
```

Adding it to the tuple fixes the import but not the call. The borrowed symbol
exists at the Python level and lowers to None on this fork:

```
TypeError: cannot convert None of type <class 'NoneType'> to tensor
```

so it also gets an entry in `_FALLBACK_SYMBOLS`, which `_patch_missing_symbols`
prefers over borrowing — the same shape `nextafter` already uses (in the tuple
*and* in the fallback table).

## Verification

On a BI-V150-class device, `flag_gems 5.4.0rc1.dev20260910`, both compiler paths
(flagtree 0.6.0+sunrise3.6 and vendor triton 3.6.0.1+git0a5cfb35), with no
blacklist in play:

| check | F | T | before |
|---|---|---|---|
| `cumsum(empty, dim=0)` | `(0,)` | `(0,)` | `ZeroDivisionError` |
| `cumsum([3,1,4], dim=0)` | `[3,4,8]` | `[3,4,8]` | unchanged |
| `asin([0,.5,-.5,.999])` vs CPU | match | match | `AttributeError` at import (T) |

`asin` on F is served by the fork's own libdevice, so the fallback is installed
only where the symbol is actually missing.


103yiran (approved):
LGTM

PR #6195, opened 2026-09-11
fix(benchmark): replace flag_gems torch_op baselines with native aten ops

kkkwb:
## What changed

Five benchmark files previously passed `torch_op=flag_gems.<op>`, so the baseline and the measured side ran the same implementation. Replace each with a native aten baseline:

- **test_scaled_dot_product_efficient_attention.py** — `torch_op` switched to `torch.ops.aten._scaled_dot_product_efficient_attention`.
- **test_linear_backward.py** — `aten::linear_backward` has no CUDA kernel, so the baseline composes the backward from aten CUDA ops (`mm` for grad_input/grad_weight, `sum` for grad_bias), gated by `output_mask`.
- **test_sparse_semi_structured_mm.py** — aten expects the CUTLASS compressed format while flag_gems uses a dense matrix plus a bool selector mask. Inputs now carry both representations; the baseline adapter consumes the pre-compressed `(packed, meta)` tensors, keeping compression cost out of timing. Skipped on GPUs without compute capability 8.x (hard requirement of the aten CUDA kernel).
- **test_functional_assert_async.py** — `aten::_functional_assert_async.msg` has no CUDA kernel, so no native baseline exists; skipped on CUDA.
- **test_nested_view_from_buffer_copy.py** — `aten::_nested_view_from_buffer_copy` segfaults the whole process on CUDA (its impl dereferences the sizes/strides/offsets data pointers host-side without a device check), which a try/except probe cannot guard; skipped.

### Progress


benchmark/test_sparse_semi_structured_mm.py
```diff
@@ -28,6 +36,27 @@
 ]
 
 
+def _to_aten_compressed(dense_mat1, bool_meta):
```
Caeruleann, line 39: bool_meta is unused here, so the ATen and FlagGems paths do not represent the same 2:4 sparse matrix. Please apply bool_meta to zero the unselected values before converting the tensor, and verify that both adapters produce matching outputs before benchmarking.

benchmark/test_nested_view_from_buffer_copy.py
```diff
@@ -51,15 +51,26 @@ def get_tflops(self, op, *args, **kwargs):
...
+# host-side without a device check), and a try/except probe cannot guard
+# against a crash that takes the process down, so the test is skipped.
+_ATEN_CUDA_SUPPORTED = False
+
+
 @pytest.mark.nested_view_from_buffer_copy
+@pytest.mark.skipif(
+    not _ATEN_CUDA_SUPPORTED,
+    reason="aten::_nested_view_from_buffer_copy segfaults on CUDA in stock PyTorch; no native baseline available",
+)
```
Caeruleann, line 65: This benchmark is permanently skipped, but the crash is caused by placing sizes, strides, and offsets on CUDA while the native implementation reads them on the host. Please keep these metadata tensors on CPU instead of disabling the benchmark.

benchmark/test_sparse_semi_structured_mm.py
kkkwb: Fixed. _to_aten_compressed now applies the bool selector to the dense mat1 first — True keeps positions 0 and 1 of each group, False keeps positions 2 and 3 — and zeroes the unselected positions before CUTLASS compression, so both adapters describe the same 2:4 sparse matrix.
One necessary addition: the CUTLASS fp32 sparsifier cannot exactly represent the adjacent-pair patterns {0,1}/{2,3}. Before compression, the K columns are permuted [0,2,1,3] within each group (mat2 rows permuted identically, leaving the product unchanged), so the kept pairs become {0,2}/{1,3} — patterns the sparsifier encodes exactly for fp16, bf16, and fp32 alike. This keeps fp32 in the benchmark dtype list.
Per the suggestion, _assert_adapters_match now runs before benchmarking: for each of the three dtypes it feeds one sample input to both adapters and asserts their outputs match with assert_close, failing fast on mismatch.
Commit 1c4fb94.

benchmark/test_nested_view_from_buffer_copy.py
kkkwb: Fixed. The crash was indeed caused by sizes/strides/offsets living on CUDA while the native implementation reads them on the host. The input generator now keeps only the buffer on the benchmark device and leaves the three metadata tensors on CPU — a layout both paths (the native aten baseline and the flag_gems op) handle correctly.
The _ATEN_CUDA_SUPPORTED = False flag and its skipif are removed accordingly, restoring the benchmark to a runnable state, with a comment explaining why the metadata must stay on the host.
Commit 1c4fb94.

Caeruleann:
please fix the ci problems

benchmark/test_linear_backward.py
```diff
@@ -45,12 +45,26 @@ def get_input_iter(self, cur_dtype):
...
+        return grad_input, grad_weight, grad_bias
+
+
 @pytest.mark.linear_backward
 def test_linear_backward():
     bench = LinearBackwardBenchmark(
         op_name="linear_backward",
-        # Use flag_gems.linear_backward for both baseline and gems since there's no native PyTorch CUDA impl
-        torch_op=flag_gems.linear_backward,
+        torch_op=_AtenLinearBackwardAdapter(),
```
Caeruleann, line 67: Please use bench.set_gems(...) here

benchmark/test_gather_block_quantized.py
```diff
@@ -47,16 +45,24 @@ def get_input_iter(self, cur_dtype):
...
+
 @pytest.mark.gather_block_quantized
 def test_gather_block_quantized():
     bench = GatherBlockQuantizedBenchmark(
         op_name="gather_block_quantized",
-        torch_op=flag_gems.ops.gather_block_quantized,
+        torch_op=_aten_gather_block_quantized,
         # gather_block_quantized consumes int8 data and float32 scales.
         dtypes=[torch.float32],
     )
```
Caeruleann, line 68: Please use bench.set_gems(...) here

benchmark/test_linear_backward.py
kkkwb: `bench.set_gems(flag_gems.linear_backward)` is already in place at line 71: the baseline uses torch_op=_AtenLinearBackwardAdapter() (the native aten composition), and the gems side is set via set_gems to flag_gems.linear_backward, so both sides are timed independently. No code change needed here.

benchmark/test_gather_block_quantized.py
kkkwb: Fixed in the new commit. The missing set_gems meant the "Gems Latency" column was re-benchmarking the same composed reference under use_gems() (flag_gems-patched repeat_interleave/mul), so the reported speedup compared identical implementations and was meaningless.
import flag_gems is restored, and the gems side now points to the operator under test via `bench.set_gems(flag_gems.ops.gather_block_quantized)`, matching the file's original pattern on master. The baseline remains the composed aten reference. 
Commit d8c18ef0.

PR #6210, opened 2026-09-12
[QC][MThreads] Optimize W8A16 FP8 RMSNorm

ggbondbest:
### PR Category
Operator, OP Test, Benchmark

### Type of Change
New Feature, Performance Optimization

### Description
Add an MThreads backend for `flag_gems.rms_norm_w8a16_fp8`. The public operator previously used the generic launch configuration on MUSA, while its FP8 tests skipped this backend.

The kernel consumes E4M3FN FP8 weights, per-group scales, and FP16/BF16 activations. It fuses FP32 RMS reduction, native FP8 conversion, scaling, and output multiplication in one launch. Grouped scale loads and shape-dependent row/warp configurations improve the normal path; a column-tiled path increases parallelism for very small batches with wide hidden dimensions. Widths above 65536 use a bounded two-pass loop within one kernel. Each call reads the current FP8 weights and scales; there is no dequantized-weight cache or activation quantization.

Register the implementation in `_mthreads/ops/__init__.py` and add **`benchmark/test_rms_norm_w8a16_fp8.py`** with **Torch BF16 (`torch.nn.functional.rms_norm`) as the only baseline**. Consolidate all FP8 W8A16 benchmarks from `benchmark/test_rms_norm.py` into this dedicated file: retain the original 10 shapes and the 4 extended shapes, and remove the duplicate FP8 test and its private helpers from the old file. Ordinary RMSNorm and Ascend INT8 benchmarks are unchanged. The baseline uses the same effective quantized weights, prepared outside timing, with the same `eps=1e-5` as the FP8 operator. Move all FP8 W8A16 accuracy tests into **`tests/test_rms_norm_w8a16_fp8.py`**, and move the existing Ascend-only INT8 cases into **`tests/test_rms_norm_w8a16_int8.py`**. Keep ordinary FP16/BF16/FP32 RMSNorm forward/backward tests in `tests/test_rms_norm.py`. The FP8 tests cover MUSA, FP16/BF16, all FP8 byte encodings, eager/graph weight updates, strided inputs, partial groups, multiple normalized dimensions, empty batches, and the wide fallback. Splitting the files preserves all test functions, parameterization, assertions, and operator markers.

### Validation
- MTT S5000 80 GB, MUSA capability 3.1; Python 3.10.12, PyTorch 2.7.1, torch_musa 2.7.1+5ee0a64, Triton 3.6.0.
- After the test split, `pytest tests/test_rms_norm.py tests/test_rms_norm_w8a16_fp8.py tests/test_rms_norm_w8a16_int8.py -q --record=json`: **57 passed, 6 skipped** (9 ordinary RMSNorm passes, 48 FP8 passes, and 6 Ascend-only INT8 skips).
- Standalone `pytest tests/test_rms_norm_w8a16_fp8.py --ref=cpu -q --record=json`: **48 passed**.
- All 63 combined accuracy JSON entries have the operator marker matching their file: `rms_norm`, `rms_norm_w8a16_fp8`, or `rms_norm_w8a16_int8`. FP8 entries use `opname: ["rms_norm_w8a16_fp8"]` and `result: "passed"`. INT8 execution is not validated on MThreads; its original Ascend-only skip condition is unchanged.
- Migrated Torch-baseline benchmark: collecting both RMSNorm benchmark files with `-m rms_norm_w8a16_fp8` gives **1 passed, 2 deselected**, covering all 14 shapes without a duplicate FP8 test. JSON contains one operator, `rms_norm_w8a16_fp8`, with one matching `op_name` detail and no per-shape errors.
- `pre-commit run --files ...`: passed for the original kernel/benchmark changes, all three split test files, and both benchmark files after consolidation. Operator test-file/marker checks also passed for the registered RMSNorm operators.
- MUSA profiler confirmed one device kernel per call on the sampled small and large shapes.

### Performance
MTT S5000 measurements on 2026-09-15 at `4439bd09`, BF16 activations, group size 128, `--mode=kernel --warmup=100 --iter=200`. Timing uses the framework's event-based `triton.testing.do_bench`, without graph capture. Torch RMSNorm and Gems FP8 W8A16 receive matching effective weights and epsilon; quantization and reference-weight preparation are outside timing.

| M x N | Torch BF16 us | FP8 W8A16 us | Speedup |
|---|---:|---:|---:|
| 1 x 4096 | 13.720 | 3.680 | 3.728x |
| 16 x 4096 | 9.540 | 4.040 | 2.361x |
| 64 x 4096 | 8.960 | 4.320 | 2.074x |
| 256 x 4096 | 9.960 | 7.480 | 1.332x |
| 1024 x 4096 | 25.680 | 21.840 | 1.176x |
| 1 x 8192 | 13.920 | 4.640 | 3.000x |
| 64 x 8192 | 13.600 | 5.640 | 2.411x |
| 256 x 8192 | 16.240 | 12.520 | 1.297x |
| 1 x 16384 | 15.680 | 4.720 | 3.322x |
| 64 x 16384 | 14.400 | 8.480 | 1.698x |
| 1 x 32768 | 12.080 | 6.280 | 1.924x |
| 64 x 32768 | 16.640 | 15.000 | 1.109x |
| 4096 x 4096 | 78.200 | 77.360 | 1.011x |
| 16 x 33024 | 12.600 | 11.200 | 1.125x |

- Arithmetic mean over all 14 shapes: **1.969x vs Torch BF16**.
- Original 10 shapes: **2.240x vs Torch BF16**.
- These are measurements from one full benchmark run, not a guarantee of a speedup on every run or device. The `4096 x 4096` result is near parity.
- This follow-up changes the benchmark baseline and removes duplicate tests, not the kernel. These speedups are not directly comparable to the earlier Gems BF16 baseline.

Reproduce:
```bash
python3 -m pytest tests/test_rms_norm_w8a16_fp8.py \
  -v --tb=short --record=json --output=accuracy_rms_norm_w8a16_fp8.json
python3 -m pytest benchmark/test_rms_norm_w8a16_fp8.py \
  -v -s --tb=short --mode=kernel --warmup=100 --iter=200 \
  --record=json --output=benchmark_rms_norm_w8a16_fp8.json
```

The benchmark's `torch_op` wrapper calls `torch.nn.functional.rms_norm` directly, outside any `flag_gems.use_gems()` context. The accuracy JSON records test parameters and pass/fail outcomes, rather than numeric MSE/MAE values.

### DIFF
--- a/benchmark/test_rms_norm_w8a16_fp8.py
+++ b/benchmark/test_rms_norm_w8a16_fp8.py
@@ -35,10 +35,6 @@ def _fp8_available():
     )
 
 
-def _torch_rms_norm(x, shape, weight_fp8, weight_scale, weight_ref):
-    return torch.nn.functional.rms_norm(x, shape, weight_ref, eps=1e-5)
-
-
 def _gems_rms_norm(x, shape, weight_fp8, weight_scale, weight_ref):
     return flag_gems.rms_norm(x, shape, weight_ref, eps=1e-5)
 
@@ -83,7 +79,7 @@ class RmsNormW8A16FP8Benchmark(base.Benchmark):
                 .to(torch.float8_e4m3fn)
                 .reshape(n)
             )
-            # Both BF16 baselines use the same effective quantized weight.
+            # The BF16 baseline uses the same effective quantized weight.
             # Quantization and reference dequantization are outside timing.
             weight_ref = (
                 (weight_fp8.float().reshape(-1, GROUP_SIZE) * scales.float()[:, None])
@@ -94,12 +90,12 @@ class RmsNormW8A16FP8Benchmark(base.Benchmark):
 
 
 @pytest.mark.rms_norm_w8a16_fp8
-@pytest.mark.parametrize("baseline", ["torch", "gems_bf16"])
 @pytest.mark.skipif(not _fp8_available(), reason="Requires FP8 E4M3FN support")
-def test_rms_norm_w8a16_fp8(baseline):
+def test_rms_norm_w8a16_fp8():
+    print("Baseline: FlagGems BF16 (flag_gems.rms_norm)")
     bench = RmsNormW8A16FP8Benchmark(
         op_name="rms_norm_w8a16_fp8",
-        torch_op=_torch_rms_norm if baseline == "torch" else _gems_rms_norm,
+        torch_op=_gems_rms_norm,
         dtypes=[torch.bfloat16],
     )
     bench.set_gems(_gems_rms_norm_w8a16_fp8)
--- a/benchmark/test_rms_norm_w8a16_fp8.py
+++ b/benchmark/test_rms_norm_w8a16_fp8.py
@@ -1,0 +1,106 @@
+# Copyright 2026 FlagOS Contributors
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#     http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+import pytest
+import torch
+
+import flag_gems
+from flag_gems.runtime import torch_device_fn
+
+from . import base
+
+GROUP_SIZE = 128
+
+
+def _fp8_available():
+    if not hasattr(torch, "float8_e4m3fn") or not torch_device_fn.is_available():
+        return False
+    if flag_gems.vendor_name == "mthreads":
+        return torch_device_fn.get_device_capability()[0] >= 3
+    if flag_gems.vendor_name == "thead":
+        return True
+    return (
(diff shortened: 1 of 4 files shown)


benchmark/test_rms_norm_w8a16_fp8.py
```diff
@@ -0,0 +1,102 @@
...
+            yield x, (n,), weight_fp8, scales, weight_ref
+
+
+@pytest.mark.rms_norm_w8a16_fp8
+@pytest.mark.skipif(not _fp8_available(), reason="Requires FP8 E4M3FN support")
+def test_rms_norm_w8a16_fp8():
+    print("Baseline: FlagGems BF16 (flag_gems.rms_norm)")
+    bench = RmsNormW8A16FP8Benchmark(
+        op_name="rms_norm_w8a16_fp8",
+        torch_op=_gems_rms_norm,
```
0x45f: compare with torch rms_norm

benchmark/test_rms_norm_w8a16_fp8.py
```diff
@@ -0,0 +1,102 @@
+# Copyright 2026 FlagOS Contributors
```
0x45f, line 1: move rms_norm_w8a16_fp8 uts to this file from benchmark/test_rms_norm.py

benchmark/test_rms_norm_w8a16_fp8.py
ggbondbest: already done

0x45f (approved):

PR #6211, opened 2026-09-12
[QC][MThreads] Add native FP8 W8A8 MM backend

ggbondbest:
### PR Category
Operator, OP Test, Benchmark

### Type of Change
New Feature, Performance Optimization

### Description
Add a backend-local `mm_w8a8_fp8` implementation for MThreads S5000 (device capability >= 3.1), following the backend organization used by #5957, and align the existing Hopper public entry points with the same scaled-MM argument interface.

- Native FP8 `tl.dot` / FP32 accumulation, verified to lower to MUSA FP8 SQMMA instructions. There is no Torch GEMM or BF16 GEMM fallback in the operator.
- FP8 inputs only, with mandatory external FP32 `scale_a` and `scale_b` (per-tensor or per-row A / per-column B). BF16/FP16 autoquantization has been removed. Input scales are applied to FP32 accumulators after the K loop.
- TME descriptor loads for compatible E4M3 inputs; masked FP8 loads for E5M2, unaligned inputs, non-contiguous layouts and broadcast strides. E5M2 descriptor loads fail accuracy checks in the tested compiler, so they are deliberately excluded.
- Shape-family tile selection and bounded split-K for underfilled skinny grids. Split-K uses an FP32 workspace and a separate FP32 reduction, not BF16 atomics.
- Align both Hopper and MThreads public signatures and defaults with `torch._scaled_mm`: `input, mat2, scale_a, scale_b, bias=None, scale_result=None, out_dtype=None, use_fast_accum=False, *, out=None`. Output defaults to the input FP8 dtype unless `out_dtype` or a preallocated `out` is supplied. Both public entry points support output reuse; ordinary calls allocate independent outputs. No activation/weight value cache is used.
- Use `torch._scaled_mm` directly as the same-precision FP8 benchmark baseline. Both paths receive prequantized FP8 inputs and return BF16. The standalone `benchmark/test_mm_w8a8_fp8.py` inherits the standard benchmark timer (default `--mode kernel`), without a custom CUDA/MUSA Graph capture/replay override or a vLLM dependency.

Five files change: the MThreads kernel and its backend exports, the existing Hopper kernel, `tests/test_mm_w8a8_fp8.py`, and `benchmark/test_mm_w8a8_fp8.py`. **`tests/test_mm.py` is unchanged in this PR.** Its ordinary-MM direct-call cleanup and removal of the legacy FP8 test have been split into companion PR #6329. The standalone FP8 tests in this PR cover both backends with the same required FP8-plus-scales interface. No changes to `conf/ci_test_aliases.yaml`, `benchmark/test_mm.py`, `benchmark/test_blas_perf_parallel.py`, CI workflows, existing MM tuning spaces/cost-model switches, or environment-variable defaults.

### Scope Split (2026-09-15)
Commit `e6f081e` restores `tests/test_mm.py` exactly to upstream and moves its previous changes, unchanged, into #6329 (`149ad0a`). Please merge #6329 first and this PR afterward: the legacy FP8 test remaining on master uses BF16 inputs without scales and must be removed before the required-scale interface is merged. Both PRs together reproduce exactly the original combined code tree. The historical migration notes below describe earlier revisions.

The prior H20 CI failure was the ordinary `test_mm` downloading a FlagTune model through a failing proxy; `unittest-required` only aggregated that failure. Splitting the file does not fix the runner network. No tests, cost model, or environment defaults were disabled. No new GPU run was needed for the file-only split; the FP8 kernel, accuracy tests, and benchmark remain unchanged from the validated revision.

### Validation
**Current revision: 200 Hopper cases and 202 MThreads cases passed**, including direct comparisons with native Torch scaled-MM. Both standard no-Graph core benchmarks completed with all five JSON records. Earlier revisions passed 82 accuracy/regression cases. The initial 12-shape study, the unchanged default core shape list, and the subsequent 57-shape large-K/regression study completed with per-shape JSON results.

In an earlier revision, after moving the accuracy coverage back into `tests/test_mm.py`, all 58 FP8 cases passed again, along with the selected MM/MM-out regression checks. A broader quick run reproduced an existing MThreads FP32 broadcast-MM precision failure (`test_mm_broadcast_stride_zero`): the unmodified base test fails identically, and the direct vs original `use_gems()` outputs are bit-identical. No tolerance adjustment or unrelated kernel fix is included.

Environment: MTT S5000 80 GB, driver 3.3.5-server, Torch 2.7.1, torch_musa 2.7.1+5ee0a64, FlagTree 0.6.0+mthreads3.6 / Triton 3.6.0, vLLM 0.13.1.dev44+g3d4cc4bc7.d20260310.musa, existing development container.

```bash
PYTHONPATH=$PWD/src python3 -m pytest tests/test_mm_w8a8_fp8.py -q

PYTHONPATH=$PWD/src python3 -m pytest benchmark/test_mm_w8a8_fp8.py \
  --level core --warmup 1000 --iter 100 --record json \
  --output mm_w8a8_fp8_mthreads.json
```

Current MThreads accuracy coverage includes E4M3/E5M2 inputs, required non-unit scalar/row/column scales, strided and broadcast scales, invalid argument rejection, row/column-major B, strided A/out, K tails (K=1/15/129), large K up to 32768, split-K, empty matrices, all output dtypes, bias/result scaling, independent allocations, output reuse/resizing, and graph replay after input/scale mutation. Hopper tests now also use mandatory external scales and prequantized FP8 inputs.

Formatting/static validation: Black 26.5.1, isort 5.12.0, Flake8, export/order checks, operators YAML check, and `git diff --check`. Earlier local GPU validation used `USE_FLAGTUNE=0` in the shell only. This interface update sets no tuning environment variable and commits no disabling fixture or environment override.

### Historical Performance (Graph / vLLM Baseline)

These results were collected before commit `ebd84f55`. They are retained as historical optimization measurements, not results from the current no-Graph PyTorch baseline.
The MUSA baseline is **vLLM `torch_per_tensor_w8a8_scaled_mm`**, which calls native MUSA `torch._scaled_mm`. Both sides receive the **same prequantized E4M3 A/B**, use unit FP32 scales and BF16 output. Input generation/quantization is outside timing. Both sides use MUSA Graph replay and Event median timing (1000 ms warmup, 100 ms measurement). The legacy table's `Torch Latency` column is this vLLM FP8 baseline, **not Torch BF16**.

Representative shapes below are a 12-shape small/medium/large study, **not the full Qwen shape list**. The initial implementation uses a fixed 64x128x128 tile with three stages and no split-K; it is rerun with the same timing protocol. Times are microseconds.

| M | N | K | Initial Triton | Optimized Triton | vLLM FP8 | Initial / optimized | vLLM / optimized |
|---:|---:|---:|---:|---:|---:|---:|---:|
| 1 | 16 | 16 | 10.999 | 8.445 | 9.794 | 1.302 | 1.160 |
| 8 | 64 | 64 | 11.015 | 8.716 | 9.123 | 1.264 | 1.047 |
| 16 | 9216 | 2048 | 24.458 | 19.628 | 23.777 | 1.246 | 1.211 |
| 32 | 12288 | 2048 | 25.356 | 21.996 | 25.034 | 1.153 | 1.138 |
| 32 | 64 | 7168 | 64.821 | 12.106 | 23.895 | 5.354 | 1.974 |
| 64 | 256 | 7168 | 64.496 | 15.210 | 24.110 | 4.240 | 1.585 |
| 128 | 256 | 4096 | 41.631 | 13.171 | 17.496 | 3.161 | 1.328 |
| 256 | 1024 | 2048 | 24.361 | 19.337 | 12.356 | 1.260 | 0.639 |
| 512 | 9216 | 2048 | 109.000 | 62.941 | 44.219 | 1.732 | 0.703 |
| 1024 | 4096 | 4096 | 181.272 | 120.084 | 68.194 | 1.510 | 0.568 |
| 2048 | 2048 | 8192 | 332.070 | 208.953 | 124.398 | 1.589 | 0.595 |
| 4096 | 4096 | 4096 | 662.740 | 417.540 | 218.195 | 1.587 | 0.523 |

- Mean per-shape speedup vs vLLM: **0.5665 -> 1.0392**; 7/12 shapes faster than vLLM.
- Optimized vs initial Triton: mean **2.1165x**, geometric mean **1.8299x**; all 12 improved.
- Important limitation: this is **not an across-the-board win over native vLLM**. Final geometric mean vs vLLM is **0.9516x**, and summed-latency ratio is **0.6471x**. Medium/large dense GEMMs still trail the native library.
- The unmodified default core shape list also runs successfully and emits all five JSON records; it emphasizes dense square GEMMs and is slower than vLLM on this backend.

To reproduce the representative study, supply a YAML `mm_w8a8_fp8` section with `shapes: [[1,M,N,K], ...]` using the table above via `--shape_file`.

### Profiling and Discarded Experiments
Used `torch.profiler` with CPU and `ProfilerActivity.PrivateUse1`, plus compiler IR inspection and graph/event measurements.

- Skinny grids launch too few CTAs. Splitting K improves utilization; large output matrices incur enough workspace/reduction traffic that split-K is intentionally restricted.
- TME + suitable tiles are materially faster than plain masked loads for aligned E4M3. Plain loads remain necessary for correct general-layout and E5M2 support.
- Larger tiles are not unconditionally better. 128x256/256x128 tiles, extra warps, and deeper pipelines frequently regressed performance in this environment.
- TME output stores and `enable_fp8_burst2` showed no stable benefit and are not retained. No experiment-only options or benchmark-time dispatch patches are in the submitted kernel.

### Follow-up: Large-K Split-K Expansion

The original backend already uses TME. This follow-up expands the existing native FP8 split-K path beyond `M <= 128, N <= 512`, rather than adding a second implementation or a BF16 fallback.

- Additional region: TME-compatible E4M3, `K >= 8192`, `M <= 512`, `M*N <= 524288`, and a base grid below its target CTA count. Existing skinny dispatch retains priority.
- Select M/N/K tiles from shape families; target 64 or 128 CTAs, with at most 16 splits in the new region. If splitting is unnecessary, keep the original dispatch.
- Partial sums and reduction remain FP32. Prequantized inputs remain FP8 throughout GEMM. No changes to benchmark logic or environment defaults.
- Add 24 accuracy cases to `tests/test_mm.py`, including large-K layouts/E4M3/E5M2, a K tail, and non-unit scales from FP16/BF16 quantization. Final result: **82 passed**.

Formal before/after measurement uses the actual standalone benchmark with MUSA Graph, 1000 ms warmup and 100 ms measurement. The comparison is against commit `9b2fd327`; only its selector is restored in an external, uncommitted reproduction script for the before run. Both sides receive identical-format prequantized E4M3 inputs and return BF16. Baseline remains vLLM MUSA `torch_per_tensor_w8a8_scaled_mm`. Seven free GPUs were used for independent shards; all **57 unique shapes** have before/after JSON records. This is a representative suite, not the full Qwen shape list.

| M | N | K | Before us | After us | Before / after | vLLM / after |
|---:|---:|---:|---:|---:|---:|---:|
| 32 | 1024 | 8192 | 54.194 | 16.433 | 3.298 | 1.435 |
| 48 | 768 | 8192 | 54.759 | 19.524 | 2.805 | 1.309 |
| 128 | 1024 | 16384 | 95.110 | 33.244 | 2.861 | 1.216 |
| 192 | 768 | 16384 | 99.774 | 33.144 | 3.010 | 1.221 |
| 256 | 256 | 32768 | 193.573 | 37.631 | 5.144 | 1.910 |
| 256 | 1024 | 8192 | 51.916 | 32.621 | 1.591 | 0.864 |
| 512 | 1024 | 8192 | 70.250 | 46.949 | 1.496 | 0.746 |

All 17 newly covered shapes improve: mean **2.297x**, minimum **1.363x**, maximum **5.144x**. Their mean speedup vs vLLM changes from **0.528x to 1.140x**. Across all 57 shapes, mean speedup vs vLLM changes from **0.749x to 0.931x**; final geometric mean is **0.804x**, and summed-latency ratio is **0.565x**. Dense large-output GEMMs still trail vLLM. The unchanged cases show no regression greater than 5% in this run.

PrivateUse1 profiling confirms the skinny-grid bottleneck: `(256,256,32768)` changes from 8 to 64 CTAs; average GEMM device time falls from about 189.9 us to 34.9 us, with a 3.7 us FP32 reduction. For `(128,1024,16384)`, the GEMM falls from about 96.0 us to 29.5 us plus 4.8 us reduction. These are profiler device times, separate from the graph/event benchmark figures above. TTGIR confirms FP8 `mtgpu.sqmma` in both paths. Disabling TME makes the first split-K GEMM about 127.4 us, so TME is retained.

Broader dense split-K and two-stage TME heuristics were **discarded**: although some shapes improved, nearby shapes regressed by 20-30%. Extra warps and deeper pipelines also provided no general benefit. The retained change is only the bounded selector expansion plus accuracy coverage.

### Historical Benchmark Timer Update (2026-09-14)

Commit `ebd84f55` changes only `benchmark/test_mm_w8a8_fp8.py`: direct `torch._scaled_mm` baseline with FP32 unit scales and BF16 output; remove vLLM imports and the entire custom Graph timer. On MThreads this dispatches to native MUSA, not NVIDIA CUTLASS. Kernel implementation, shape selection, tuning and environment defaults are unchanged. The base framework still honors an explicitly selected timing mode; the default and validation use `--mode kernel`, not Graph.

Validation on one free S5000: three representative shapes and all five default core entries completed successfully, with positive finite per-shape JSON timings. Black, isort, Flake8 and `git diff --check` passed. The historical full 57-shape study has not been rerun with this new timing protocol.

### Accuracy Test Migration (2026-09-15)

Commit `2a73e51f` moves all 8 FP8 test functions and 3 dedicated helpers from `tests/test_mm.py` to `tests/test_mm_w8a8_fp8.py`. AST comparison confirms identical test bodies, decorators, shapes, tolerances and assertions; the remaining ordinary MM code is unchanged. The obsolete `mm_w8a8_fp8: test_mm` CI alias is removed. Kernel code and benchmark timing/baseline are unchanged.

Validation: **82 FP8 accuracy cases passed** from the new file. Black, isort, Flake8, `git diff --check`, operator marker/file checks for both MM and FP8 MM, and the CI test-code check passed. An extra ordinary-MM run had 16 passed and two FP32 failures at `(M,N,K)=(495,5333,71)` with both B layouts. Both failures also reproduce with the pre-migration test code on identical seeded inputs; before/after outputs are bit-identical. They are unrelated to this file-only migration; no tolerance or kernel changes were made.

### External Scales and Torch Interface (2026-09-15)

Commit `402cbb0f` changes only the MThreads kernel and its standalone benchmark/accuracy files. It removes BF16/FP16 autoquantization: both inputs must be FP8, and both FP32 input scales are required. No changes to Hopper, tile selection, TME/split-K dispatch, benchmark timing mode, or tuning/environment defaults.

```python
out = flag_gems.mm_w8a8_fp8(
    input=a_fp8,
    mat2=b_fp8,
    scale_a=scale_a,
    scale_b=scale_b,
    bias=None,
    scale_result=None,
    out_dtype=torch.bfloat16,
    use_fast_accum=False,
)
```

The optional epilogue matches the installed **Torch MUSA** implementation: `((A @ B) * scale_a * scale_b + bias) / scale_result` before output conversion; omitted bias/result scale mean 0/1. Result scaling applies to all supported output dtypes here, unlike CUDA-specific behavior. Bias/result scaling is fused into final writeback and applied only once after split-K reduction. Both `use_fast_accum` values retain FP32 accumulation, as in the installed MUSA baseline. Required scale tensors support scalar or row/column broadcasting, not K-block scaling.

Validation: **201 passed** on one S5000, including 48 native Torch interface comparisons spanning default FP8/BF16/FP16/FP32 outputs, scalar/per-row result scaling, optional positional/keyword arguments, and both fast-accumulation flag values. Strided epilogues cover K=0/128/8192. Black, isort, Flake8, operator-marker/test-file checks, the CI test-code check, and `git diff --check` passed.

The actual standalone benchmark ran with `--level core --mode kernel --warmup 1000 --iter 100 --record json`. Both paths use the same prequantized E4M3 tensors and non-unit FP32 input scales (0.5 and 1.5), BF16 output, no Graph override, and no timed quantization. On this machine `torch._scaled_mm` dispatches to native MUSA/muDNN, **not NVIDIA CUTLASS**.

| Core entry (M=N=K) | Torch us | FlagGems us | Torch / FlagGems |
|---:|---:|---:|---:|
| 384 | 5.440 | 12.480 | 0.436 |
| 4096 | 186.860 | 379.200 | 0.493 |
| 1024 | 12.040 | 21.200 | 0.568 |
| 2048 | 38.120 | 68.660 | 0.555 |
| 4096 (second default entry) | 190.320 | 384.360 | 0.495 |

Mean per-entry speedup: **0.5094x**. All five records are present and successful. This dense-core suite remains slower than native MUSA; this commit aligns the API and does not claim an across-the-board performance improvement. The full 57-shape historical study was not rerun.

### Hopper Interface Alignment (2026-09-15)

In response to the request to update Hopper as well, both backends now expose exactly the same parameter names, ordering, defaults, and keyword-only output arguments (verified by AST comparison):

```python
mm_w8a8_fp8(input, mat2, scale_a, scale_b, bias=None,
             scale_result=None, out_dtype=None, use_fast_accum=False, *, out=None)
mm_w8a8_fp8_out(input, mat2, scale_a, scale_b, bias=None,
                 scale_result=None, out_dtype=None, use_fast_accum=False, *, out)
```

- Hopper now requires FP8 inputs and external FP32 input scales. The public path no longer performs BF16 autoquantization or BF16 fallback. Existing native FP8 scaled kernels are reused; the original BF16 MM operator is unchanged.
- Remove the vendor-specific calling branch and Hopper-only unit-scale special case from the standalone benchmark. Both devices pass the same four input tensors with non-unit scales.
- Check actual strides and alignment before using TMA; other supported layouts use masked native FP8 loads, not BF16 dequantization.
- Require FP32 promotion in the scaled FP8 dot kernels. Split-K accumulates into FP32 scratch and applies bias/output conversion once afterward.
- Fix the split-K tuning pre-hook to clear its atomic accumulator before every trial. This fixes repeated trial results being summed together; it does not change tuning ranges or cost-model/environment settings.
- Preserve preallocated output support and default FP8 output. Both `use_fast_accum` values keep FP32 promotion; the hint does not select a less accurate kernel.

**Native backend differences:** the parameter interface is uniform, but native Torch itself differs. In the tested CUDA Torch 2.11, `scale_result` is validated but not forwarded to GEMM, so Hopper likewise treats it as a validated no-op. The installed Torch MUSA 2.7.1 instead divides the scaled product plus bias by this tensor. CUDA also rejects bias with FP32 output and E5M2 x E5M2; Hopper follows those restrictions. See [PyTorch 2.11 ScaledBlas.cpp](https://github.com/pytorch/pytorch/blob/v2.11.0/aten/src/ATen/native/cuda/ScaledBlas.cpp#L443-L627). This is not a claim of identical optional epilogue semantics across different native Torch backends/versions.

Validation on one H20 (node 021, Torch 2.11.0+cu128 / Triton 3.6) and one S5000:
- **Hopper: 200 passed, 180 backend-specific cases skipped.**
- **MThreads: 202 passed, 178 backend-specific cases skipped.**
- Covers K tails, small-K TMA, large-K split-K including a new tuning key, mixed E4M3/E5M2 operands on Hopper, row/column/scalar scales, non-contiguous output, positional/keyword calls, and graph replay after input/scale mutation.
- Kernel results retain strict elementwise checks against an FP32 reference. Native CUDA FP8 accumulation has larger errors near cancellation, independently confirmed against CPU FP64; the native-library comparison therefore additionally uses a bounded relative norm, while the strict reference checks are unchanged. Both hint values are compared against native accurate mode.
- Standalone benchmark uses normal `benchmark.run()`, `--mode kernel --warmup 1000 --iter 100`, no Graph override, no timed quantization, and BF16 output on both sides.

| Default core entry (M=N=K) | H20 Torch us | H20 Gems us | H20 speedup | S5000 Torch us | S5000 Gems us | S5000 speedup |
|---:|---:|---:|---:|---:|---:|---:|
| 384 | 6.976 | 34.752 | 0.201 | 5.480 | 12.840 | 0.427 |
| 4096 | 509.904 | 1548.512 | 0.329 | 186.720 | 383.480 | 0.487 |
| 1024 | 16.448 | 39.200 | 0.420 | 13.840 | 21.360 | 0.648 |
| 2048 | 74.624 | 204.096 | 0.366 | 38.080 | 68.640 | 0.555 |
| 4096 (second default entry) | 510.016 | 1547.936 | 0.329 | 188.620 | 381.500 | 0.494 |

All ten JSON records succeeded. These dense square shapes remain slower than their respective native Torch baselines; this follow-up is interface/correctness work, not a claim of unchanged or improved performance. Historical Graph measurements above are not directly comparable.

**Known legacy caller:** `benchmark/test_blas_perf_parallel.py` still contains its older BF16/no-scale FP8 test. That file is deliberately unchanged under the requested scope and its old FP8 entry is incompatible with the new required-scale API. Use `benchmark/test_mm_w8a8_fp8.py`; adapting the legacy caller needs separate scope approval.


benchmark/test_mm_w8a8_fp8.py
```diff
@@ -41,6 +39,9 @@ def _mm_w8a8_fp8_out_cached(a, b):
     else:
         _MM_W8A8_FP8_OUT_CACHE.pop(key)
         _MM_W8A8_FP8_OUT_CACHE[key] = out
+    if flag_gems.vendor_name == "mthreads":
+        return flag_gems.mm_w8a8_fp8_out(a, b, scale_a, scale_b, out=out)
+    # The existing Hopper interface accepts unscaled FP8 inputs only.
```
0x45f, line 44: please modify hopper mm_w8a8_fp8

0x45f (approved):

PR #6213, opened 2026-09-12
fix(ci): stop sort_exports from deleting imports on double __all__

gavin0x01:
sort_exports.py located `__all__` with a DOTALL regex `^(__all__\s*=\s*\[)(.*?)(^\])`. When a file has two `__all__` definitions (an empty `__all__ = []` placeholder followed by the real list), the non-greedy match ran from the first `[` to the last line-leading `]`, swallowing every import statement in between and deleting them on --fix.

_enflame/gcu300/ops/__init__.py has exactly this layout, so running --fix wipes all 170 imports while leaving __all__ intact, breaking `import *` and attribute access.

Fixes:
- Locate the module-level `__all__ = [...]` via ast and replace it by line range, so nothing outside the assignment can be touched.
- Refuse to rewrite files with multiple `__all__` list definitions (warn + skip in --fix, fail in --check) to avoid data loss.
- Collapse gcu300's placeholder + real __all__ into a single sorted list, ordered imports then __all__ to match the other backends.
- Drop the now-unused `import re`.

### Progress

### DIFF
--- a/src/flag_gems/runtime/backend/_enflame/gcu300/ops/__init__.py
+++ b/src/flag_gems/runtime/backend/_enflame/gcu300/ops/__init__.py
@@ -12,6 +12,236 @@
 # See the License for the specific language governing permissions and
 # limitations under the License.
 
+from ._unsafe_masked_index import _unsafe_masked_index
+from .abs import abs, abs_
+from .add import add, add_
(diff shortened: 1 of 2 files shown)


103yiran (approved):
LGTM

PR #6243, opened 2026-09-14
docs(ops): warn against removing empty()'s zero-store kernel

ziyuhu214:
## What

A comment only, no behaviour change.

## Why

`ops/empty.py` launches a zero-store kernel and then returns. Removing that store
looks like free performance: on DeepSeek-V4 decode it is ~6% of GPU time at
roughly 3000 `empty()` calls per step, and `empty()` is by contract allowed to
return uninitialized memory.

It is not safe. Parts of the stack allocate state buffers via `empty()` and
silently rely on this kernel zeroing them. Removing it produced illegal memory
accesses in the DeepSeek-V4 compressor state cache, and the failure appears far
from the allocation, which makes it expensive to diagnose.

This records that finding at the store itself, so the next person to spot the
same apparent optimisation sees why it was left in place and what they would need
to audit first.


huangyiqun:
The comments in `empty_kernel` still describe this as a dummy store and the result as uninitialized. Please update them as well so they do not contradict this note.

PR #6262, opened 2026-09-15
fix(kunlunxin): upsample/pool/conv3d kernel fixes

llaboon:
### PR Category
Operator

### Type of Change
Bug Fix

### Description
Kunlunxin (XPU) kernel fixes for the vision op family:

- `adaptive_avg_pool2d_backward`: vendor override with a flat input-tile kernel (window walk per input element, unmasked loads, fp32 accumulation); handles 3-D and 4-D inputs.
- `conv3d`: support `padding="same"` / `"valid"` for the string-padding contract (dilation-aware padding plus the end-padded crop that PyTorch `same` requires).
- `upsample_bilinear2d`: vendor override with a flat 1-D grid over the output, so every store is stride-1 and source indices are clamped in range.
- `upsample_linear1d`: read the input through its own strides in-kernel instead of materializing a contiguous copy.
- `upsample_linear1d_backward`: window-walk gather that reads `grad_output` through its strides and clamps every window index.
- `_upsample_nearest_exact1d_backward` / `_upsample_nearest_exact2d_backward`: flat grid, runtime fp32 window arithmetic with every clamp done before narrowing, clamped reads, and strided `grad_input` support for the `.grad_input` overload.
- tests: call the ops directly in `tests/test_upsample_nearest_exact1d_backward.py` (no `use_gems` context) and keep the fp64 matrix off on this backend (no fp64 device path).

### Test plan
- `pytest tests/test_adaptive_avg_pool2d_backward.py tests/test_upsample_bilinear2d.py tests/test_upsample_bilinear2d_aa.py tests/test_upsample_linear1d.py tests/test_upsample_nearest_exact1d_backward.py tests/test_upsample_nearest_exact2d_backward.py --ref cpu` on kunlunxin P800.


llaboon:
Re-triggering CI: the previous kunlunxin run failed during vendor-lib import ('xpu3') before any test executed, while sibling runs minutes later passed on the same code — looks like a transient environment state.

PR #6263, opened 2026-09-15
fix(kunlunxin): sort kernel fixes

llaboon:
### PR Category
Operator

### Type of Change
Bug Fix

### Description
Kunlunxin (XPU) fixes for the sort family:

- `sort` / `sort_stable`: clamp the tail tile loads (masked loads can still issue the tail address here); rework the radix scatter so inactive lanes land on lane-unique scratch slots and every store is unmasked (data-dependent store masks corrupt adjacent slots); materialize the dim-last view through the cat family's strided copy func instead of `copy_` (int32 strided copies fault; the copy stays bit-exact through an fp32 bit-view); route every non-trivial size through the stable radix chain (the bitonic argsort path mis-sorts and can fault the device here); add 0-dim and empty-input guards; drop the unreferenced full radix/packed paths and the `is_use_mask_zero` launch option.

### Test plan
- `pytest tests/test_sort.py --ref cpu` on the kunlunxin backend.

---

Update (2026-09-21): the `nonzero` and `unique_consecutive` overrides are withdrawn from this PR - both called torch functional ops (`torch.unbind`; `torch.cumsum` / `torch.nonzero`) and current master already carries kernel-based versions of both files. Only the `sort` fixes remain.


103yiran (approved):
LGTM

PR #6264, opened 2026-09-15
fix(kunlunxin): pointwise and layout fixes (add/cat/div)

llaboon:
### PR Category
Operator

### Type of Change
Bug Fix / Performance Optimization

### Description
Kunlunxin (XPU) pointwise and layout fixes:

- `add`: fast path for fp16/fp32/fp64 tensor-tensor add - one compiled kernel per dtype, replayed through the compiled kernel entry, instead of the generic pointwise launch chain (host time per call drops from ~36us to ~5.5us; measured 0.84-1.56x of the native reference on P800). No native routing.
- `cat`: read the permuted view in-kernel via `StridedBuffer` for the inner-dim path instead of materializing `permute(...).contiguous()`; int32 strided reads are bit-cast to fp32 first (int32 strided copies hit an illegal memory access).
- `diff`: widen int16 to int32 before the kernel; add a non-inner-dim kernel that views the input as (M, N, K); fall back to a composite implementation for prepend/append, bf16 with n > 1 and non-contiguous inputs (the previous path routed through `torch.cat` and strided `copy_`, which is not reliable here).
- `div`: complex true division computed in-kernel through bit-view kernels (`torch.view_as_real` cannot be used inside the pointwise code generator), covering complex x complex, complex x real, scalar operands and the `out=` / in-place variants.

### Test plan
- `pytest tests/test_add.py tests/test_cat.py tests/test_diff.py tests/test_div.py --ref cpu` on kunlunxin P800.

---

Update (2026-09-17): branch rebuilt on current master. The `diff.py` change was dropped (its narrow-path fallbacks delegated to the CPU torch composite; upstream master already carries the row-kernel implementation). The earlier `tle.raw` fast-path changes are dropped: upstream reverted that feature in #6409, and this branch no longer touches those files.


103yiran (approved):
LGTM

PR #6265, opened 2026-09-15
fix(kunlunxin): elementwise/special fixes (conj_physical/hypot_/nextafter/scaled_softmax)

llaboon:
### PR Category
Operator

### Type of Change
Bug Fix

### Description
Kunlunxin (XPU) elementwise and special-function fixes:

- `conj_physical`: vendor override that delegates to the stride-1 `_conj` kernel, resolving pending conjugate bits first (matches the generic contract).
- `hypot_`: in-place hypot that reads the broadcast operand in-kernel instead of materializing `broadcast_to(...).contiguous()`, which is not reliable for strided/broadcast sources on this backend.
- `nextafter`: vendor override with an all-integer sign-magnitude bit algorithm (the generic bit path does not compile here and the libdevice fallback fails to lower); the same public API is kept, including `out=` and the in-place variant.
- `scaled_softmax`: replace the full autotune sweep (recompiles per key on this backend) with fixed bounded configs for both forward and backward kernels.
- `op_black_list.yaml`: drop the stale `acos` entry - the op passes with the current kernels.

### Test plan
- `pytest tests/test_conj_physical.py tests/test_hypot_.py tests/test_nextafter.py tests/test_scaled_softmax.py --ref cpu` on the kunlunxin backend.

---

Update (2026-09-21): every public entry of the touched ops now carries the `GEMS_KUNLUNXIN` debug log (restored on `scaled_softmax_forward` / `scaled_softmax_backward`, added on the new `conj_physical`).


src/flag_gems/runtime/backend/_kunlunxin/ops/conj_physical.py
```diff
@@ -0,0 +1,34 @@
...
+def conj_physical(input: torch.Tensor) -> torch.Tensor:
+    """Materialize the complex conjugate with the vendor stride-1 kernel."""
+    logger.debug("GEMS_KUNLUNXIN CONJ_PHYSICAL")
+    if not input.is_complex():
+        return input
+    # Match the generic implementation: resolve a pending conjugate bit
+    # first, since torch.view_as_real() (used by _conj) rejects unresolved
+    # conj views.
+    if input.is_conj():
+        input = input.resolve_conj()
```
103yiran: this is torch api？

llaboon: Same as the generic `conj_physical` implementation: resolve a pending conjugate bit first, since `view_as_real()` (used by `_conj`) rejects unresolved conj views. Under FlagGems this is not a native ATen call -- `resolve_conj` is a FlagGems-registered op that has its own implementation in the tree (including a backend override), so the call dispatches to that implementation. It is also guarded by `is_conj()`: it only runs for inputs carrying a pending conjugate bit, and the conjugation itself is done by `_conj` on the next line.

bin913:
@llaboon Please resolve conflicts

bin913:
@llaboon Please fix failing check

103yiran (approved):
LGTM

PR #6266, opened 2026-09-15
fix(kunlunxin): fused-kernel fixes (matmul_bias_activation/bf16_paged_mqa_logits/heuristics)

llaboon:
### PR Category
Operator

### Type of Change
Bug Fix

### Description
Kunlunxin (XPU) fixes for norm/loss and fused kernels:

- `batch_norm`: clamp the masked tail loads in the stats/combine/normalize tile loops (`tl.load(..., mask=..., other=0)` can return in-bounds values for the masked lanes here, corrupting the training stats), and replace the Triton backward (strided `permute(...).reshape` copy plus a lowering failure for spatial > 1) with an eager fp32 implementation.
- `nll_loss` family: drop the `is_use_mask_zero` launch option (not part of the launch schema; every case fails at device level with it) and fix the ignored-lane leak in the gather kernels (clamped loads, contribution gated with `tl.where`); applied across the 1-D/2-D forward and backward variants.
- `matmul_bias_activation`: keep the fused sdnn GEMM path under its original kernel name and route large-K bf16 to an un-hijacked kernel (the name-prefix rewrite into the sdnn fusion loses ~0.42 rms at K=4096); also avoid a rank-0 constant operand in the ReLU step.
- `bf16_paged_mqa_logits`: avoid rank-0 constant operands in the ReLU step of both kernels.
- `heuristics_config_utils`: add the `mha_varlen_prefill` / `mha_varlen_decode` heuristic entries that `flash_api` looks up.
- tests: call the ops directly in the GLU family tests and skip them on kunlunxin (TransformerEngine stack unsupported here).

### Test plan
- `pytest tests/test_batch_norm.py tests/test_nll_loss_forward.py tests/test_nll_loss_backward.py tests/test_matmul_bias_activation.py --ref cpu` on kunlunxin P800.

---

Update (2026-09-17): branch rebuilt on current master. The `batch_norm.py` and `nllloss.py` changes were dropped (upstream master carries the guarded triton backward kernel and `_NllLoss2dAutograd`); the PR now delivers the fused-kernel (`matmul_bias_activation`, `bf16_paged_mqa_logits`) and heuristics fixes. The earlier `tle.raw` fast-path changes (dialect registration and the `*_raw.xpu` payloads) are dropped: upstream reverted that feature in #6409, and this branch no longer touches those files.


103yiran (commented):
LGTM

103yiran (approved):
LGTM

PR #6318, opened 2026-09-15
fix(kunlunxin): fill/add_relu/special_log1p/dequantize fixes

llaboon:
### PR Category
Operator

### Type of Change
Bug Fix

### Description
Kunlunxin (XPU) kernel fixes:

- `fill` family (`fill.Scalar` / `fill.Tensor` / in-place / `.out`) kernel fixes.
- `add_relu`: vendor implementation added and registered.
- `special_log1p`: vendor kernel fix (registered).
- `dequantize`: quantized dequantize kernel fixes.

### Test plan
- `pytest tests/test_fill.py tests/test_add_relu.py tests/test_special_log1p.py tests/test_dequantize.py --ref cpu` - 372 / 18 / 41 / 45 passed on kunlunxin P800.
- `pytest benchmark/test_fill.py benchmark/test_add_relu.py benchmark/test_special_log1p.py benchmark/test_dequantize.py --record log` on kunlunxin P800.

---

Update (2026-09-17): branch rebuilt on current master. The `special_log1p` scalar path now uses `math.log1p` instead of `torch.log`. The earlier `tle.raw` fast-path changes (dialect registration and the `*_raw.xpu` payloads) are dropped: upstream reverted that feature in #6409, and this branch no longer touches those files.


src/flag_gems/runtime/backend/_kunlunxin/ops/dequantize.py
```diff
@@ -1,114 +1,126 @@
-# Copyright 2026 FlagOS Contributors
+# Copyright 2026, The FlagOS Contributors.
 #
 # Licensed under the Apache License, Version 2.0 (the "License");
 # you may not use this file except in compliance with the License.
-
+# You may obtain a copy of the License at
```
RobertLuobo, line 5: 注释太长了

src/flag_gems/runtime/backend/_kunlunxin/ops/dequantize.py
```diff
@@ -1,114 +1,126 @@
...
+    XPU device) carry the `QuantizedCUDA` keyset; the `CUDA`-key registration
+    used by the generic config is never consulted for them and the XPU-native
+    `dequantize` kernel is absent, so `torch.dequantize` would otherwise fail
+    with "invalid device function".
+    """
+    global _quantized_lib
+    if _quantized_lib is not None:
+        return
+    try:
+        _quantized_lib = torch.library.Library("aten", "IMPL")
```
RobertLuobo: 这个是 fallback 吗

103yiran:
test | abs:kunlunxin

src/flag_gems/runtime/backend/_kunlunxin/ops/dequantize.py
llaboon: Trimmed - the long explanatory block is removed; the file now keeps only the standard license header plus two short inline notes (and the leading blank line flagged by the linter is gone).

src/flag_gems/runtime/backend/_kunlunxin/ops/dequantize.py
llaboon: Not a fallback - the arithmetic runs in the flag_gems Triton kernel (pointwise_dynamic). set_() only builds a zero-copy int8 view of the quantized storage (needed because int_repr() has no XPU kernel), and the torch.library registration is required because quantized tensors dispatch via the QuantizedCUDA key, which the generic registration never covers.

src/flag_gems/runtime/backend/_kunlunxin/ops/dequantize.py
```diff
@@ -1,92 +1,87 @@
...
+        _quantized_lib.impl("dequantize", dequantize, "QuantizedCUDA")
+        logger.debug("FlagGems Kunlunxin: registered QuantizedCUDA dequantize override")
+    except Exception as e:
+        logger.warning(
+            f"GEMS_KL3 failed to register dequantize QuantizedCUDA override: {e}"
         )
-    return padded[:n_elements].view(input.shape)
+
+
+_register_quantized_dequantize()
```
Caeruleann: What is the purpose of this code? Who uses them? If it is useless code, please delete it.

llaboon: Checked - you are right, the block was redundant. `torch.dequantize` on quantized tensors dispatches through the QuantizedCUDA key, and that key is already covered by the standard registration: `_FULL_CONFIG` carries `("dequantize.self", dequantize, None, (QUANTIZED_DISPATCH_KEY,))` and `GeneralOpRegistrar` registers it to exactly that key whenever gems is enabled. The module-level `Library` block duplicated that registration; being always-on, it also made the standard registration fail with "already registered from python" (silently collected per-entry by the registrar) while keeping the override active outside the `enable()`/`use_gems()` contract. Removed in the latest revision (ee0bf576): the `_quantized_lib` keep-alive global and the whole `_register_quantized_dequantize` def/call are gone, and dispatch now flows through the standard path only.

Caeruleann (approved):

PR #6326, opened 2026-09-15
[QC][MetaX] Add W8A16 RMSNorm operator

July-h5kf3:
## Summary

Adds a MetaX backend for `rms_norm_w8a16_fp8`, dispatched through the public FlagGems entry. Activations are FP16/BF16; weights are grouped FP8 E4M3FN with per-group scales (default group size 128). The upstream NVIDIA implementation is unchanged.

The MetaX implementation is located in `src/flag_gems/runtime/backend/_metax/ops/rms_norm_w8a16_fp8.py`. For steady-state eager execution, it caches the dequantized BF16/FP16 weight while the source tensor identities and version counters remain unchanged. Weak references release cached device memory with the source tensors, and in-place weight or scale updates invalidate the cache. CUDA Graph capture and tensors without version counters use the fused dynamic-weight path so graph replay continues to read current weight and scale buffers.

The dynamic path loads native `float8_e4m3fn` values. CTA occupancy is capped at 8 warps of 64 threads to match MetaX C550. Shape-specific grouped, tiled, and loop paths select the measured configuration for each workload class.

Tests and benchmarks reuse the existing `tests/test_rms_norm_w8a16_fp8.py` and `benchmark/test_rms_norm_w8a16_fp8.py`; no MetaX-specific test file is added. Public FP8 W8A16 tests and the official benchmark are enabled for the `metax` vendor. The extended shape/stride coverage test (`test_rms_norm_w8a16_fp8_mthreads_shapes`) now also runs on `metax` and `thead`, not only `mthreads`.

This PR additionally fixes ragged-group (`N % group_size != 0`) support in the W8A16 RMSNorm backends, which previously raised `ValueError` and silently skipped those cases: MetaX FP8 (`f4648ae7`), THead FP8 (`81bcbc10`), and Ascend INT8 (`2f28cbf1`). Ragged coverage is added for each backend: the shared FP8 extended-shape test covers MetaX and THead, and `tests/test_rms_norm_w8a16_int8.py` gains four ragged cases for Ascend. FP16 assertions use `atol=2e-3` for all vendors (1–2 ulp error level, same as MThreads); MetaX FP16 additionally uses `rtol=2e-3` because the kernel structurally re-rounds normalized activations to FP16 before applying the weight, so a pure atol cannot cover large-reference lanes.

## Validation

Tested commit: `6c34b954`.

Environment:

- Device: MetaX C550 64GiB
- Python: 3.12.11
- PyTorch: 2.8.0+metax3.7.2.0
- Triton: 3.6.0
- FlagGems vendor: `metax`

Results:

- `pytest -q tests/test_rms_norm_w8a16_fp8.py`: **48 passed**, 8 consecutive runs (includes the extended mthreads/metax/thead shape and stride coverage).
- Repository benchmark `benchmark/test_rms_norm_w8a16_fp8.py`: **1 passed**.
- `pytest -q tests/test_rms_norm.py -m rms_norm_w8a16_fp8`: **20 passed** (unchanged baseline).
- `git diff --check` and pre-commit style checks: passed.
- FlagGems BF16 and PyTorch BF16 comparison: three rounds; all 10 shapes were faster in every round.

Cross-backend validation of the ragged-group fixes:

- THead (PPU-ZW810E, torch 2.10.0, FlagTree 0.6.2a2+ppu3.6): `tests/test_rms_norm_w8a16_fp8.py` **48 passed**; `tests/test_rms_norm.py` 9 passed.
- Ascend (910B, CANN 9.0, torch_npu 2.10.0): `tests/test_rms_norm_w8a16_int8.py` **10 passed** (6 existing + 4 new ragged cases); `tests/test_rms_norm.py` 9 passed.

Coverage includes FP16/BF16, eager in-place weight and scale updates, CUDA Graph replay updates, all 256 FP8 encodings, strided inputs, group size 64/96/256, and ragged groups where `N % group_size != 0`.

## Performance

Measured on MetaX C550 with BF16 activations and group size 128. All three implementations use the same shape and source BF16 weight:

- PyTorch BF16: `torch.nn.functional.rms_norm`
- FlagGems BF16: `flag_gems.rms_norm`
- FP8-W8A16: `flag_gems.rms_norm_w8a16_fp8`

Timing uses `triton.testing.do_bench` with 100 ms warmup, 1000 ms measurement, and median latency (event timing). The table reports the median of three rounds. Warmup populates the steady-state dequantized-weight cache before measurement.

| Shape | PyTorch BF16 (us) | FlagGems BF16 (us) | FP8-W8A16 (us) | vs PyTorch | vs FlagGems BF16 |
|---|---:|---:|---:|---:|---:|
| 1 x 4096 | 39.680 | 15.872 | 15.360 | 2.583x | 1.033x |
| 16 x 4096 | 72.704 | 15.872 | 15.360 | 4.733x | 1.033x |
| 64 x 4096 | 83.456 | 15.872 | 15.616 | 5.344x | 1.016x |
| 256 x 4096 | 108.288 | 19.456 | 18.944 | 5.716x | 1.027x |
| 1024 x 4096 | 235.520 | 29.952 | 28.672 | 8.214x | 1.045x |
| 1 x 8192 | 44.288 | 20.224 | 16.384 | 2.703x | 1.234x |
| 64 x 8192 | 92.928 | 20.736 | 17.408 | 5.338x | 1.191x |
| 256 x 8192 | 151.808 | 28.928 | 23.808 | 6.376x | 1.215x |
| 1 x 16384 | 50.176 | 25.088 | 21.504 | 2.333x | 1.167x |
| 64 x 16384 | 109.568 | 26.624 | 23.552 | 4.652x | 1.130x |

All **10/10** shapes are faster than both baselines.

- Arithmetic mean speedup: **4.799x vs PyTorch BF16**, **1.109x vs FlagGems BF16**.
- Geometric mean speedup: **4.453x vs PyTorch BF16**, **1.106x vs FlagGems BF16**.
- Minimum per-shape speedup: **2.333x vs PyTorch BF16**, **1.016x vs FlagGems BF16**.

These results cover steady-state execution for the listed BF16 shapes on MetaX C550. No performance claim is made for other devices or unmeasured shapes.


tests/test_rms_norm_w8a16_fp8.py
0x45f: why this case only run on mthreads

July-h5kf3: Good catch — the case no longer runs on MThreads only. The vendor list is now `("mthreads", "metax", "thead")`, so MetaX, MThreads and THead all execute this shared shape/stride suite (done in 81bcbc10 on this branch).

Two shapes here have `N % group_size != 0` (`(3, 65) group_size=64` and `(5, 1000) group_size=128`), and every W8A16 backend on this branch now supports such ragged groups:

- MetaX: `f4648ae75` `[QC][MetaX] Support ragged FP8 groups in W8A16 RMSNorm.` — the cached dequant path pads the flat weight with zeros up to a whole group, applies the last group's scale to the remaining elements, then trims the padding; divisible shapes keep bit-identical results, and the fused kernels' `cols // GROUP_SIZE` scale indexing already covers a partial tail.
- THead (PPU): 81bcbc10 `[QC][THead] Support ragged groups in W8A16 RMSNorm.`
- Ascend INT8: 2f28cbf1 `[QC][Ascend] Support ragged groups in W8A16 INT8 RMSNorm.`

FP16 tolerance: decoding error lands within 1–2 ulp of the float32 reference (same magnitude as MThreads), so this test asserts `atol=2e-3` with the library `rtol` unchanged (set by 81bcbc10). On MetaX the fp16 check additionally allows `rtol=2e-3`, because the fused kernel rounds the normalized activation to FP16 before the weight multiply — a pure atol bound cannot cover 2 ulp at larger magnitudes. Other vendors' checks are unchanged.

Validation:

- MetaX C550 (Python 3.12.11, PyTorch 2.8.0+metax3.7.2.0, Triton 3.6.0): `python -m pytest -q tests/test_rms_norm_w8a16_fp8.py` — **48 passed**, repeated 8 consecutive runs; `benchmark/test_rms_norm_w8a16_fp8.py` smoke passes.
- PPU: the same suite's 28 extended cases pass on MThreads/THead.
- Ascend INT8: 10/10 on the corresponding INT8 suite.

Follow-up idea (not in this PR): the generic implementation indexes scales as `cols // GROUP_SIZE` and therefore already handles partial groups structurally; the divisibility gate could later be relaxed into a capability condition so more vendors share this coverage.

July-h5kf3: One refinement on the fp16 tolerance from my previous comment: the `rtol=2e-3` branch is no longer MetaX-only. THead CI showed the same structural fp16 deviation in this test (`(65,16384)`, greatest abs diff 0.0078 at |ref| ≈ 5, 1/1064960 elements), which is consistent with the fp16 10-bit mantissa (~2 ulp ≈ 0.002 × |ref| at large magnitudes). Commit fa77032d therefore applies `atol=2e-3, rtol=2e-3` to fp16 for all vendors instead of gating on MetaX. BF16 checks keep the library default tolerances, and MThreads is unaffected (its fp16 error stays well under the new bounds).

July-h5kf3: Good catch — this case no longer runs on MThreads only. The vendor list is now `("mthreads", "metax", "thead")`, and all three backends support ragged groups (`N % group_size != 0`, e.g. `(3,65) gs=64`, `(5,1000) gs=128`): MetaX `f4648ae75`, THead `81bcbc10`, Ascend INT8 `2f28cbf1`.

FP16 assertions use `atol=2e-3, rtol=2e-3` for all vendors (~2 ulp, same scale as MThreads); BF16 keeps library defaults. Verified: MetaX C550 48/48 (8 consecutive runs), PPU 28/28 extended cases, Ascend INT8 10/10, and CI is green on all three backends.

Follow-up idea (not in this PR): the generic implementation already handles partial groups structurally (`cols // GROUP_SIZE`), so this gate could later be relaxed into a capability condition for more vendors.

0x45f (approved):

PR #6329, opened 2026-09-15
[Test] Align MM tests and parallel FP8 benchmark with scaled-MM API

ggbondbest:
### PR Category
OP Test, Benchmark

### Type of Change
Refactor, Bug Fix

### Description
Separate ordinary MM test cleanup from #6211 and address the review request to update the legacy FP8 caller in `benchmark/test_blas_perf_parallel.py`.

This PR changes **two files**:

1. `tests/test_mm.py` (unchanged by the latest commit):
   - Replace five `flag_gems.use_gems()` wrappers with direct `flag_gems.mm` / `flag_gems.mm_out` calls.
   - Remove the legacy Hopper FP8 test and its two helpers; replacement FP8 accuracy coverage is in the standalone test file in #6211.
   - Preserve ordinary MM shapes, dtypes, references, tolerances and assertions.
2. `benchmark/test_blas_perf_parallel.py` (commit `df10f6f`):
   - Replace the stale scale-less call with `flag_gems.mm_w8a8_fp8_out(a, b, scale_a, scale_b, out=out)`.
   - Generate prequantized FP8 operands and non-unit FP32 scales before timing. Both benchmark paths receive the same four tensors.
   - Use `torch._scaled_mm` as the same-precision baseline and retain the existing output-dtype selection/output cache.
   - Use row-major A / column-major B as required by the native CUDA baseline, with exactly one metric per shape at both benchmark levels.
   - Use the existing subclass hook to avoid forwarding an unsupported FP8 `--dtypes` value to workers. Each worker selects the same single default FP8 dtype; the public CLI parser and shared worker implementation are unchanged.

No kernel, standalone benchmark, CI workflow, tuning logic, cost-model switch, or environment default is changed. AST comparison confirms the existing timer and shared parallel framework are unchanged; this legacy parallel path retains its existing Graph timer.

### Companion PR
The caller update requires the FP8-only, required-scale interface in #6211. Please coordinate the two PRs together: #6211 contains the implementation and standalone FP8 accuracy tests, while this PR removes the incompatible old test and updates the old parallel caller. This is not a compatibility shim for the old BF16/scale-less API, and the earlier file-only split's merge-order note must be reconsidered with this caller dependency.

### Validation
On one H20, node 038, using #6211 at `0f712727` plus the new parallel benchmark:
- Direct checks passed for `(M,N,K) = (16,128,64), (128,256,4096), (512,1024,2048)`: native FP8 operands, FP32 scales 0.5/1.5, strict FP32-reference accuracy, output-buffer reuse, and updated results after changing a scale.
- Native Torch comparison relative errors were below 0.0007 on all three shapes.
- Core/comprehensive metadata checks passed: one input/record per shape and correct two-way shape partitioning.
- The actual pytest benchmark passed with `--parallel 1`, exercising worker subprocess creation, FP8 dtype selection, scale forwarding and parent aggregation. All three shape rows reported SUCCESS; the JSON test summary reported passed.
- Black 26.5.1, isort, Flake8, AST scope checks and `git diff --check` passed.
- The GPU allocation was released. No eight-GPU or full performance sweep was performed; this is an interface regression check, not a performance-improvement claim.

Reproduction with the combined changes and a `mm` shape YAML containing the three shapes above:

```bash
python -m pytest -s benchmark/test_blas_perf_parallel.py::test_mm_w8a8_fp8 \
  --level core --shape_file /path/to/shapes.yaml \
  --warmup 10 --iter 10 --parallel 1 --record json --output /path/to/result.json
```

### Existing CI Infrastructure Issue
The previous H20 failure came from ordinary `tests/test_mm.py::test_mm` downloading a FlagTune model through a proxy that returned HTTP 500. This PR neither fixes that network issue nor disables FlagTune. The earlier file-only split passed its ordinary-MM CI on the A100 runner; the latest caller update was explicitly checked with the new Hopper implementation on H20.


0x45f (commented):
uts in test_blas_perf_parallel.py also need modify like `return flag_gems.mm_w8a8_fp8_out(a, b, out=out)`

0x45f (approved):

PR #6338, opened 2026-09-16
fix(benchmark): explicitly select batch norm target operators

lukatao:
### PR Category

Benchmark

### Type of Change

Bug Fix

### Description

The `_native_batch_norm_legit_no_training` and `_batch_norm_no_update` benchmarks currently time `F.batch_norm` under `use_gems()`. Its dispatch can select `_batch_norm_impl_index` instead of the operator named by the benchmark.

Explicitly pass the corresponding `flag_gems` function as `gems_op` in each test. This makes the Gems timing call the intended operator directly while preserving the Torch reference, input generation, shapes, and dtypes. Only the two benchmark files change: add module-level imports and specify the target in each test.

### Issue

Incorrect target selection in the two batch norm benchmarks.

### Progress

- [x] Both affected benchmarks were exercised locally.

### Performance

Validated on Kunlunxin P800 with kernel mode, core shapes, `--warmup 5 --iter 100`: both benchmark tests passed, with 12 successful measurements each. A profiling preflight outside the timed region confirmed that every Gems measurement calls the intended Kunlunxin backend function and every Torch reference measurement does not.

Validation used master `8ce7662fae58a0daec8d9e2adcc934356fe5d5c6` with these test changes. The local Torch 2.5.1 XMLIR environment also required a validation-only fallback in the Kunlunxin `range.py` registration: retry `Library.impl` without `allow_override` when that keyword raises `TypeError`. That compatibility adjustment is not part of this PR; no operator implementation was changed for the benchmark validation.

Applicable pre-commit checks (flake8, isort, Black, EOF and trailing whitespace) and `git diff --check` passed. Because the configured GitCode mirrors were inaccessible, hooks were fetched from their corresponding GitHub repositories using the same pinned revisions, with no change to the repository configuration.


benchmark/test__native_batch_norm_legit_no_training.py
```diff
@@ -54,10 +54,13 @@ def torch_native_batch_norm_legit_no_training(
 
 @pytest.mark.native_batch_norm_legit_no_training
 def test_native_batch_norm_legit_no_training():
+    import flag_gems
```
0x45f: Import at the beginning of the file.

benchmark/test_batch_norm_no_update.py
```diff
@@ -54,10 +54,13 @@ def torch_batch_norm_no_update(
 
 @pytest.mark.batch_norm_no_update
 def test_batch_norm_no_update():
+    import flag_gems
```
0x45f: Import at the beginning of the file.

0x45f (approved):

PR #5940, opened 2026-09-02
[KernelGen][Nvidia] Add _stack operator with Triton kernel

Yukun-Cui:
## Summary
Adds a Triton kernel for `_stack`. Concatenates a sequence of tensors along a new dimension. This is the internal `_stack` implementation that mirrors `stack`.

## Testing
- Validated against reference on device via `to_reference(inp, True)`
- Tested on: Nvidia, Tianshu, Muxi, Ascend, Hygon

## Performance
Test command: `pytest benchmark/test__stack.py --level core` (NVIDIA H20)

### _stack

| dtype | Size | Torch Latency (ms) | Gems Latency (ms) | Speedup |
|-------|------|--------------------|-------------------|---------|
| float16 | [64, 64] | 0.005552 | 0.005760 | 0.964 |
| float16 | [4096, 4096] | 0.057312 | 0.059456 | 0.964 |
| float16 | [64, 512, 512] | 0.057056 | 0.059840 | 0.953 |
| float32 | [64, 64] | 0.005712 | 0.005968 | 0.957 |
| float32 | [4096, 4096] | 0.108224 | 0.101120 | 1.070 |
| float32 | [64, 512, 512] | 0.108224 | 0.100064 | 1.082 |
| bfloat16 | [64, 64] | 0.005504 | 0.005792 | 0.950 |
| bfloat16 | [4096, 4096] | 0.057312 | 0.059808 | 0.958 |
| bfloat16 | [64, 512, 512] | 0.057312 | 0.059872 | 0.957 |

| Operator | Arithmetic Mean Speedup |
|----------|------------------------|
| _stack | **0.984** |

## Multi-backend Testing
| Backend | Accuracy Test | Speedup (mean) | Notes |
|---|---|---|---|
| Nvidia (H20) | PASS (9 cases) | 0.984 | Primary |
| Tianshu | N/A | — | — |
| Muxi | N/A | — | — |
| Ascend | N/A | — | — |
| Hygon | N/A | — | — |

## Files Changed
- `src/flag_gems/ops/_stack.py`: Triton kernel implementation
- `tests/test__stack.py`: Accuracy test
- `benchmark/test__stack.py`: Performance benchmark
- `src/flag_gems/ops/__init__.py`: Register import and `__all__`
- `src/flag_gems/__init__.py`: Register to `_FULL_CONFIG`
- `conf/operators.yaml`: Add operator entry (kind: Tensor, stage: alpha 5.4)

PR #6359, opened 2026-09-16
fix(ci): derive underscore-op pytest markers by convention

gavin0x01:
## Problem

CI never actually ran accuracy/benchmark tests for operators whose id
starts with an underscore (e.g. `_stack`, `_reshape_alias`).

`pytest.mark._stack` is rejected by pytest's attribute access, so
`tools/run_tests.py` stripped the leading underscore before `pytest -m`.
For such ops this had two failure modes:

1. The stripped name (`_stack` → `stack`) no longer matched the op's
   marker, so its tests were silently skipped.
2. The stripped name could collide with a distinct existing op (the
   separate `stack` operator), so `-m "stack"` selected the wrong op's
   cases.

This is the "underscore bug" blocking PR #5940 (`_stack`).

## Convention

Marker name is derived from the operator id:

1. Non-underscore ids are used verbatim.
2. Leading underscore is stripped: `_stack` → `stack`.
3. If the stripped name collides with an existing operator id, the
   `underscore_` prefix is used instead: `_stack` → `underscore_stack`.

| op id | marker |
|---|---|
| `_stack` | `underscore_stack` (collides with `stack`) |
| `_reshape_alias` | `reshape_alias` (no collision) |
| `_adaptive_avg_pool3d_backward` | `underscore_adaptive_avg_pool3d_backward` |

Because the resulting marker never has a leading underscore, the
`setattr`/`MarkDecorator` hack in the test files is no longer needed.

## Changes

- `tools/run_tests.py`: add `op_marker()` applying the rule when building
  `pytest -m`; stop pre-stripping ids in `get_ops_to_test()`.
- `tools/ci_checks/check_operator_markers.py`: expect the rule-derived
  marker instead of requiring `marker == op_id`.
- `_reshape_alias` / `_adaptive_avg_pool3d_backward` test + benchmark
  files: switch to the convention markers and drop the setattr hack.
- `tests/test__reshape_alias.py`: dispatch via `flag_gems._reshape_alias(...)`
  instead of `use_gems()`, per `check-kernelgen-tests`.

Note for #5940: its test already uses `@pytest.mark.underscore_stack`,
which matches this convention, so it needs no marker change.

## Testing

- `check_operator_markers.py` passes for `_reshape_alias` and
  `_adaptive_avg_pool3d_backward`.
- `check_kernelgen_tests.py` passes for the changed test files.
- Verified `op_marker`/`expected_marker` agree on all cases (`_stack`,
  `_reshape_alias`, `_adaptive_avg_pool3d_backward`,
  `_flash_attention_forward`, `_weight_norm`, and non-underscore ops).
- Note: the torch-based suite was not run locally (no torch in the dev
  environment); CI exercises it.


103yiran (approved):
LGTM

PR #6361, opened 2026-09-16
fix(kunlunxin): sdpa attention fixes (value-1 arg specialization, single-config autotune)

llaboon:
### What
Fixes for the Kunlunxin sdpa path (`_attn_fwd`):

1. **`do_not_specialize` on the 17 i32 kernel args** — Triton treats integer arguments equal to 1 as constexpr and drops them from the launcher argument sequence. The kunlunxin launch-table path consumes arguments positionally, so the folded arguments desynchronized the layout and produced `xpuLaunchKernel -> Operation not permitted` or wrong results for GQA (`kv_head_num == 1`) and `Z == 1` shapes.
2. **Single-config autotune** — on this backend the kernel body is compiled to a launch-table stub, so multi-config autotune only re-benchmarks the stub on every new `(KV_CTX, HEAD_DIM)` key and folds ~100 ms of exploration into the first call. Only `BLOCK_M` is consumed (it drives the grid); the HEAD_DIM-aware prune is kept.
3. **Optional precision knobs** `FLAG_GEMS_FA_ACCUM_TYPE` / `FLAG_GEMS_FA_GEMM_TYPE`; defaults (fp32 accumulation, non-tf32 GEMM) preserve the existing math.

### Test
- `pytest tests/test_scaled_dot_product_attention.py -k legacy --ref cpu`: 96 passed / 0 failed / 96 skipped / 48 deselected (the causal and GQA subsets previously failed with EPERM or precision errors).
- Benchmark (`-m scaled_dot_product_attention --level core`): non-causal 1.05-1.20x; the causal rows no longer pay the per-first-call exploration cost.


103yiran (approved):
LGTM

Caeruleann (approved):

PR #6365, opened 2026-09-16
packaging: add DEB + RPM packaging for flag-gems (squashed rebased replacement for #3418)

shiptux:
Squashed and rebased onto current master — replacement for #3418, which had fallen ~300 commits behind (content identical). Single commit for easier review.

What it adds:
- packaging/debian/ + packaging/rpm/ build definitions for python3-flag-gems, build workflows (build-deb.yml / build-rpm.yml), INSTALL.md, post-build import smoke test
- Carries two install-critical fixes already reviewed separately: missing `__init__.py` in eight imported subpackages (#4939) and SQLAlchemy 1.4 compatibility with >=1.4.31 floor (#5683)

Verification: DEB (ubuntu 22.04/24.04) and RPM (fedora 43) install-tested; A100 GPU smoke passed (fused/DSA import + softmax, diff 7.6e-06).

Note: with #4939/#5683 still unmerged, their changes are included here; if those merge first I'll rebase to drop the duplication. Suggest reviewing this one and closing #3418, or I can close whichever — maintainer's choice.


shiptux:
CI caught a real defect — pushed a fix.

The packaging smoke check this PR adds failed with:

```
FAIL: package smoke checks found problems:
  - subpackage not shipped: flag_gems/runtime/backend/_iluvatar/utils
```

That directory has `pointwise_dynamic.py` but no `__init__.py`, so `find_packages` skips it and the installed package lacks the module — while `_iluvatar/ops/__init__.py` line 17 does `from ..utils.pointwise_dynamic import ModuleGenerator`. Importing flag_gems on Iluvatar hardware therefore fails, for pip installs as much as for deb/rpm.

Same defect class #4939 fixed for seven other backend subpackages; this directory arrived later (#5783) and so was not covered. A full scan of `src/flag_gems` confirms this was the last one remaining.

Added the `__init__.py` (identical to the sibling backends') and rebased onto current master. The companion rc2 PR #6394 carries the same fix.

shiptux:
Status after the fix: **build-deb and build-rpm both pass now** — the missing `__init__.py` was the packaging smoke check's only complaint.

The remaining red is `backend-tests (iluvatar)`, and it is a runner-capacity timeout rather than a test failure. The job never got past **Check GPU availability**:

```
##[start-action display=Check GPU availability]
GPU_OUTPUT=$(bash tools/gpu_check_iluvatar.sh)
...
##[error]Process completed with exit code 1    duration_ms=600305
```

`tools/gpu_check_iluvatar.sh` waits for a GPU with at least 30 GB free (`mem_threshold=30000`, retry every 120 s, `max_wait=600`). The step ran for exactly 600 s and printed nothing, i.e. no GPU on the Iluvatar runner freed up within the window and the script gave up before any test executed.

For context on why this job runs at all: adding `_iluvatar/utils/__init__.py` labels the PR `vendor/Iluvatar`, which selects the iluvatar backend matrix. The file itself is the license header only — byte-identical to the sibling backends' `utils/__init__.py` — so it cannot affect test behaviour.

A re-run once the runner has a free GPU should clear it. Alternatively, if the Iluvatar runner is expected to stay busy, I can split the `__init__.py` into its own small PR (same class as #4939) so the packaging change no longer touches `src/` and does not pull in the vendor matrix — happy to do that if you prefer.

0x45f (approved):

PR #6097, opened 2026-09-09
[CI] Use three-dot diff to derive PR-changed files

gavin0x01:
The changed-file derivation used a two-dot diff (base head), which lists every difference between the base branch tip and the PR head. When the base branch advances after a PR is created, files modified only on the base branch (e.g. tests/test_mm.py) get incorrectly attributed to the PR and flagged by downstream checks such as check_kernelgen_tests.py.

Switch to a three-dot diff (base...head) so only changes the PR branch actually introduced are reported, matching GitHub's "Files changed" view.

### Progress

- [ ] Change is properly reviewed (1 reviewer required, 2 recommended).
- [ ] Change is responded to an issue.
- [ ] Change is fully covered by a UT.

PR #6366, opened 2026-09-16
fix(ci): only flag use_gems() on PR-added lines in check-kernelgen-tests

gavin0x01:
## Problem
`check-kernelgen-tests` fails PRs for `use_gems()` calls they did not introduce.

Example: PR #5390 only edited a docstring (and added metax code) in `tests/test_linalg_lstsq.py`, but the check failed on **15 pre-existing** `use_gems()` calls that came from #5173. The three-dot fix (#6097) correctly stops *unrelated files* from being attributed to a PR, but here the PR *did* touch the file — and the checker scans the **whole file**, so legacy violations on untouched lines get reported.

## Fix
`check_kernelgen_tests.py` gains `--base`/`--head`. When provided (incremental CI mode), reported violations are restricted to lines the PR actually **added**, computed via a three-dot (`base...head`) diff — the same semantics as `derive_changed_operators.py`. Full scans (`--all`) are unchanged and still report every call.

`rule-check.yaml` passes the PR's base/head shas and uses `fetch-depth: 0` so the diff resolves (a shallow clone would make it fall back to a full-file scan rather than silently pass).

## Behavior
- PR **adds** a `use_gems()` line → still fails ✅
- PR edits a file elsewhere, `use_gems()` is pre-existing → passes ✅
- `--all` full scan → reports everything ✅

## Verification
Reproduced with #5390's real base/head: **0** violations after the fix (was 15). Added a synthetic PR that adds a `use_gems()` line: still fails on the exact added line. Control PR editing elsewhere with a pre-existing call: passes.


103yiran (approved):
LGTM

PR #5539, opened 2026-08-17
[KernelGen][Nvidia] Add conv_depthwise3d operator with Triton kernel

NineAnnAnn:
## Summary
Adds a Triton kernel for `conv_depthwise3d`. Applies a 3D depthwise convolution over an input signal composed of several input planes.

## Testing
- Validated against reference on device via `to_reference(inp, True)`
- Tested on: Nvidia, Tianshu, Muxi, Ascend, Hygon

## Performance
Test command: `pytest benchmark/test_conv_depthwise3d.py --level core` (NVIDIA H20)

### conv_depthwise3d

| dtype | Size | Torch Latency (ms) | Gems Latency (ms) | Speedup |
|-------|------|--------------------|-------------------|---------|
| float16 | [1, 32, 16, 32, 32] | 0.024272 | 0.018176 | 1.335 |
| float16 | [1, 64, 8, 16, 16] | 0.008832 | 0.014496 | 0.609 |
| float16 | [2, 16, 8, 8, 8] | 0.006624 | 0.014240 | 0.465 |
| float16 | [4, 32, 4, 16, 16] | 0.008832 | 0.014592 | 0.605 |
| float32 | [1, 32, 16, 32, 32] | 0.023680 | 0.016800 | 1.410 |
| float32 | [1, 64, 8, 16, 16] | 0.009280 | 0.014400 | 0.644 |
| float32 | [2, 16, 8, 8, 8] | 0.006816 | 0.014944 | 0.456 |
| float32 | [4, 32, 4, 16, 16] | 0.009216 | 0.015072 | 0.611 |
| bfloat16 | [1, 32, 16, 32, 32] | 0.024544 | 0.018208 | 1.348 |
| bfloat16 | [1, 64, 8, 16, 16] | 0.008704 | 0.014720 | 0.591 |
| bfloat16 | [2, 16, 8, 8, 8] | 0.006592 | 0.014992 | 0.440 |
| bfloat16 | [4, 32, 4, 16, 16] | 0.008864 | 0.016448 | 0.539 |

| Operator | Arithmetic Mean Speedup |
|----------|------------------------|
| conv_depthwise3d | **0.754** |

## Multi-backend Testing
| Backend | Accuracy Test | Speedup (mean) | Notes |
|---|---|---|---|
| Nvidia (H20) | PASS (12 cases) | 0.754 | Primary |
| Tianshu | N/A | — | — |
| Muxi | N/A | — | — |
| Ascend | N/A | — | — |
[...]

PR #6374, opened 2026-09-17
fix(ci): sort/check src/flag_gems/ops/__init__.py exports

gavin0x01:
## Problem
`/fix-sort` reported `✅ All files are already sorted correctly` on a PR whose `src/flag_gems/ops/__init__.py` `__all__` was visibly unsorted (e.g. #5539, where reviewer flagged `conv_depthwise3d` sitting between `_conv_depthwise2d` and `_convert_weight_to_int4pack`).

## Root cause
`discover_init_files()` — duplicated in **both** `sort_exports.py` and `check_init_exports.py` — only scanned:
- the main package `src/flag_gems/__init__.py`
- backend `runtime/backend/*/ops/__init__.py`

The generic operator export list `src/flag_gems/ops/__init__.py` — the file every new device-agnostic operator is added to — was **never discovered**. So `/fix-sort` sorted 23 other files, found no diff, and reported success, while the real file went unchecked. CI (`check-init-exports`) had the same blind spot, which is how unsorted exports reached master.

## Fix
Add `src/flag_gems/ops/__init__.py` to discovery in both scripts. Enabling the check immediately surfaced **two pre-existing inversions already on master**:
- `cross`, `cross_out` placed before `cos`/`cosh`… (should follow `count_nonzero`, `cov`)
- `dropout_backward` before `dropout_`

These are fixed here (3-line reorder, rest of the 1177-entry list untouched) so the newly enabled check passes.

## Scope
Only the generic `ops/` init is added. `utils/`, `fused/`, `pt2/`, etc. remain intentionally out of scope (their `__all__` uses non-alphabetical, meaning-based grouping) — a separate discussion.

## Verification
- Both scripts now discover `ops/__init__.py` (24 files each).
- `sort_exports.py --check` full run: all 25 files sorted, exit 0.
- `check_init_exports.py`: all 24 files pass.
- On #5539 content the check correctly flags the misplaced entry.


103yiran (approved):
LGTM

PR #6379, opened 2026-09-17
[KMCompiler][Hygon] add fused rrelu_with_noise backend kernels

Mike7180:
### PR Category
Operator

### Type of Change
Performance Optimization

### Description
Adds a Hygon-specific implementation of `rrelu_with_noise` and `rrelu_with_noise_`, registered in the Hygon backend (`src/flag_gems/runtime/backend/_hygon/ops/rrelu_with_noise.py` plus the two exports in `.../_hygon/ops/__init__.py`).

The generic implementation fills `noise` in a separate `uniform_` launch and then runs a `pointwise_dynamic` kernel that re-reads `self` and `noise` and writes both outputs; on Hygon the generated wrapper's per-call host work (schema lookup, dtype promotion, stride handling, launch-config heuristics) also lands inside the `do_bench` window, and the operator measured below the torch baseline (0.4339 / 0.3577 / 0.6742 / 0.5165 for eval / train x out-of-place / in-place).

The backend implementation:
- fuses the Philox noise sampling into the training kernel, so training is one read + one write pass (no scratch tensor, no separate `uniform_` launch);
- pins the launch configuration (`BLOCK = 2048`, `num_warps = 8`), consistent with the existing Hygon elementwise fast paths;
- takes the contiguous fast path when both tensors are contiguous and `numel <= INT32_MAX`, and otherwise falls back to the existing generic `pointwise_dynamic` kernels, so the general case is unchanged.

Semantics follow ATen: evaluation uses `slope = (lower + upper) / 2` and leaves positives untouched; training scales `self <= 0` by the sampled noise, keeps `1.0` for positives, and writes the effective noise back into `noise`.

### Issue
N/A

### Performance
Hygon BW, `level=comprehensive`, eval / train x out-of-place / in-place; mean over all shapes x dtypes. "Before" is the generic implementation, "after" is the min of three consecutive runs on the same tree (the box was quiet -- the `[64,64]` eval row reads 0.008 ms):

| mode | before | after |
| --- | --- | --- |
| eval, out-of-place | 0.4339 | 1.171 |
| train, out-of-place | 0.3577 | 0.928 |
| eval, in-place | 0.6742 | 1.195 |
| train, in-place | 0.5165 | 1.004 |

Large shapes gain 1.3-1.8x. The remaining weak rows are the seven smallest shapes of the two training modes (0.1-0.8), where the measured window is dominated by host-side launch cost rather than device work.

Note: the backend-test CI job collects tests from the files changed in the PR; this PR changes no test files, so `tests/test_rrelu_with_noise.py` is not exercised by CI here. It was run on the Hygon box instead (166 passed).

### Progress
- [ ] Change is properly reviewed.


src/flag_gems/runtime/backend/_hygon/ops/rrelu_with_noise.py
```diff
@@ -0,0 +1,438 @@
...
+    generator=None,
+):
+    """FlagGems implementation of aten.rrelu_with_noise_ (Hygon backend)."""
+    logger.debug("GEMS_HYGON RRELU_WITH_NOISE_")
+    return _rrelu_with_noise_impl(
+        self, noise, lower, upper, training, generator, out=self
+    )
+
+
+__all__ = ["rrelu_with_noise", "rrelu_with_noise_"]
```
Caeruleann:
```suggestion
```

Caeruleann (approved):

PR #6381, opened 2026-09-17
fix(ci): run marker-selected tests for changed operator implementations

gavin0x01:
## Problem
`tools/test-op.sh` selects tests purely by matching changed **test** files (`tests/test_*.py`, `benchmark/test*`). A PR that changes an operator *implementation* without touching a test file matches nothing and hits the early `exit 0` — it passes with **zero tests run**.

This is the norm for vendor backends. Sampling the last 20 PRs per backend for "only implementation changed, no test file":

| Backend | impl-only, no test | Backend | impl-only, no test |
|---|---|---|---|
| Kunlunxin | 85% | MooreThreads / AMD | 35% |
| Iluvatar / Enflame | 60% | MetaX | 30% |
| Cambricon | 50% | Ascend / Hygon | 15% |
| | | NVIDIA | 10% |

e.g. #6361 (`_kunlunxin/ops/attention.py`) showed a green `Run backend tests` step that executed nothing.

## Approach
Map changed implementation files to operator ids and run their tests **by pytest marker** (`@pytest.mark.<op>`). Investigated two mappings:
- filename (`<op>.py` -> `test_<op>.py`): ~85% hit rate, fails on renamed/one-to-many tests.
- **marker**: 98.2% of the 1250-op registry has a matching marker, and it works across name differences (`native_batch_norm` -> `test_batch_norm.py`, `special_exp2` -> `test_exp2.py`, `special_gammaincc` -> `test_igammac_.py`).

Marker mapping chosen.

## Changes
- **`derive_changed_operators.py`**: also match backend op paths `.../_<vendor>/[<arch>/]ops/<op>.py` (previously only generic `ops/`), and add `--changed-files`/`--ops-only` for shell callers. This also closes a second blind spot — vendor-only PRs derived no operators, so `rule-check`'s static checks were all skipped for them; they now apply.
- **`test-op.sh`**: collect changed op-impl files, derive ids, dedupe against test files already run directly, then `pytest -m "id1 or id2 ..."`. Because pytest deselects (and exits 0) when a marker matches nothing, a run that executed zero tests is surfaced as a `::warning::`, not a silent pass.

## Behavior
- Impl-only vendor PR (e.g. attention) -> runs `-m attention`.
- Changed a test file -> unchanged original path.
- Impl + its own test file -> test file runs once; the marker is deduped away (no double run).
- Op with no dedicated test (internal kernels) -> warning, not failure.

## Verification
Validated locally: id derivation for vendor/arch/generic paths, routing decisions across representative PRs, dedup, backward-compat of the `--base/--head` output contract, and the zero-tests-ran detection against sample pytest summaries. The **actual pytest execution** requires a runner with the flag_gems runtime — best confirmed by CI on a vendor runner (this PR touches only tooling, so it will not self-trigger it).


103yiran (approved):
LGTM

PR #6389, opened 2026-09-17
kunlunxin: grouped_mm large-M/small-K persistent-kernel dispatch with per-shape autotuning

llaboon:
### PR Category

Operator

### Type of Change

Performance Optimization

### Description
Improve `aten::_grouped_mm` on Kunlunxin P800. The existing two-kernel dense-tile override handles most shapes well, but is slow on large-M / small-K shapes, where its many small tiles do not hide memory latency (M=525986, N=2048, K=128, G=64: 0.86x of native). This PR adds a per-shape-autotuned persistent-kernel path for that regime and dispatches to it; every other shape keeps the existing path unchanged.

**What changed**

1. New `src/flag_gems/runtime/backend/_kunlunxin/ops/group_gemm.py` - a backend override of the generic `group_gemm.py` module: same kernels, with two autotuning fixes:
   - decorator order: `@libtuner` outermost. The generic module stacks `@libentry()` outside `@libtuner(...)`, and libentry's kernel-cache key intentionally drops `do_not_specialize` values (M/N/K keep only the type class). As a result only the first shape ever entered the autotuner; every later shape silently reused the first shape's config.
   - two extra `BLOCK_M=512` candidates and a persistent-grid multiplier (launch `NUM_SMS * 4` programs): one-program-per-core under-hides memory latency on large-M / small-K shapes; x4 is ~25% faster there and neutral elsewhere.
2. `src/flag_gems/runtime/backend/_kunlunxin/ops/grouped_mm.py` - `group_mm` dispatches shapes with `K <= 256 and M >= 65536` to the persistent path before the host-side `offs` copy (that copy's device sync would otherwise add ~1.3 ms to this path); all other shapes take the existing two-kernel dense-tile path unchanged. No `__init__` changes.

### Issue
N/A

### Progress

### Performance
Official FlagGems benchmark on P800 (`benchmark -m grouped_mm -sv --level core`, bf16, equal-weighted mean over the three cases):

| shape (M, N, K, groups) | before | after |
|---|---|---|
| 80694, 512, 2048, 16 | 0.853 | 0.853 |
| 129548, 2560, 2048, 16 | 1.235 | 1.231 |
| 525986, 2048, 128, 64 | 0.862 | **1.054** |
| **equal-weight mean** | 0.983 | **1.046** |

Correctness: `tests/test_grouped_mm.py --ref cpu` 3/3 passed (both the dispatched path and the dense path).


Caeruleann (dismissed):

src/flag_gems/runtime/backend/_kunlunxin/ops/group_gemm.py
```diff
@@ -0,0 +1,655 @@
...
+                    tl.store(c_ptrs, c, mask=c_mask)
+
+                tile_idx += total_grid
+
+        last_problem_end = current_problem_end
+        group_start = group_end
+
+
+def group_gemm(group_A, group_B, group_C, offs_table, alpha=1, beta=0):
+    logger.debug("GEMS GROUP_GEMM")
```
Caeruleann:
```suggestion
    logger.debug("GEMS_KUNLUNXIN GROUP_GEMM")
```

src/flag_gems/runtime/backend/_kunlunxin/ops/group_gemm.py
```diff
@@ -0,0 +1,655 @@
...
+            group_size,
+            alpha=alpha,
+            beta=beta,
+        )
+
+    return group_out
+
+
+def group_mm(A: torch.Tensor, B: torch.Tensor, offs: torch.Tensor) -> torch.Tensor:
+    logger.debug("GEMS GROUP_MM")
```
Caeruleann:
```suggestion
    logger.debug("GEMS_KUNLUNXIN GROUP_MM")
```

Caeruleann (approved):

PR #6343, opened 2026-09-16
ci(kunlunxin): run tests inside container on klx-p800-1 server

gavin0x01:
## Summary
Add a dedicated `test-kunlunxin` job in `backend-test.yaml` that runs inside the `flagos-base-kunlunxin-xre5.37.1:2.1.2` container on the `klx-p800-1` server, mirroring the existing iluvatar container job. The generic `test` job now excludes `kunlunxin`.

## Status
Draft — validating that the kunlunxin runner can allocate, pull the container image, run setup, pass gpu_check, and execute operator tests.

## Note
Contains a **temporary probe** edit to `tests/test_isposinf.py` so `test-op.sh` picks up a test file on the runner. This will be reverted before merge.

PR #6393, opened 2026-09-17
fix(kunlunxin): bump flagtree to 0.6.1+xpu3.6 in pyproject extra

gavin0x01:
## Symptom
Every kunlunxin `test-kunlunxin` job fails at conftest import (e.g. run 35190879513):
```
RuntimeError: Failed to import vendor extra lib:
  dialect() takes 0 positional arguments but 1 was given
```

## Root cause
Two places pin the kunlunxin flagtree version:
- `src/flag_gems/backends.yaml` -> `flagtree==0.6.1+xpu3.6` (correct)
- `pyproject.toml` `[kunlunxin]` extra -> **`flagtree==0.5.1+xpu3.0`** (stale; every other backend is on 0.6.x)

`setup.sh` runs `uv pip install ".[kunlunxin]"`, which pulls the old `0.5.1+xpu3.0` in as a dependency. The later "Installing FlagTree" step tries to install `0.6.1+xpu3.6` from backends.yaml but does **not** uninstall the already-present flagtree first, so 0.5.1 stays. 0.5.1's `dialect()` signature is incompatible, so `import flag_gems` blows up and all kunlunxin op tests fail at collection.

#6343 fixed backends.yaml but missed this second source. This aligns them.

## Fix
`pyproject.toml`: `flagtree==0.5.1+xpu3.0` -> `0.6.1+xpu3.6` (verified available on the flagos-pypi-kunlunxin index; matches backends.yaml and the runtime image).

## How it surfaced
#6381 (marker-selected tests) made an impl-only kunlunxin PR actually run its operator tests instead of exiting 0 vacuously — which is exactly what exposed this stale pin.

## Follow-up (not in this PR)
`setup.sh`'s flagtree branch is asymmetric: the triton branch does `uv pip uninstall flagtree` before installing, but the flagtree branch does not uninstall an existing flagtree. Hardening that (uninstall-then-install, or `--reinstall`) would prevent a stale transitive flagtree from ever winning again. Aligning the pin here removes the problem at its source.


103yiran (approved):
LGTM

PR #6397, opened 2026-09-17
[Benchmark] Add tensor-free workload listing and exact case replay

YaooXu:
## Purpose

Extract the benchmark workload enumeration/replay portion of `YaooXu/feat/new-api-for-kernelgen-server` into an independently reviewable PR against current master (`8b0b3db0`). Source branch: `YaooXu/feat/benchmark-case-api`.

`pytest --collect-only` identifies test nodes, but cannot enumerate the shape/dtype/parameter combinations yielded inside a benchmark. This PR separates lightweight workload planning from tensor construction so external tools can inspect workloads and reproduce one selected benchmark case without executing every preceding workload.

## Changes

- Add tensor-free `BenchmarkCasePlan`, `BenchmarkCaseSpec` and `BenchmarkCaseList`, with `flaggems.benchmark-case-list/v2` JSON output.
- Add `--list-cases` / repeatable `--case-id`, preserve node/level/dtype/ordinal identity, include case IDs in measured results, and reject unknown or unexecuted requested IDs.
- Add providers for common unary/binary/scalar/out/reduction/BLAS/GLU families and an explicit `case_fn` + `build_inputs_fn` pair for GenericBenchmark.
- Preserve the existing callable selection and timing methods. Unmigrated custom/input_fn benchmarks continue to run normally and explicitly reject unsupported listing/replay.
- Add unit tests and English/Chinese usage/migration docs: `docs/content/en/testing/benchmark-cases.md` and `docs/content/zh-cn/testing/benchmark-cases.md`.

## Scope / non-goals

No candidate injection, `resolve_gems_op`, dispatcher changes, correctness-test rewrites, candidate coverage, Preflight, Profile, Ascend timing changes, or bulk per-operator migration. This is independent of #6105, not a replacement for its override mechanism. IDs are snapshot/order-dependent, not content hashes. Tensor-free planning does not promise that importing every FlagGems benchmark on a CPU-only host works.

## Validation

- Host-only JSON contract: **6 passed**, without Torch/Triton.
- Benchmark Case API/provider tests: **30 passed**, including no materialization while listing, exact selected-case construction, case ordering, legacy input equivalence, inherited/custom loops, duplicate IDs, and invalid factories.
- NVIDIA A100, Torch `2.11.0+cu130`, Triton `3.6.0`: real CLI listing twice, exact single-case replay, and full execution for **softmax, abs, mm**, each using two small FP32 workloads. Repeated manifests matched; replay emitted exactly one matching case ID; full runs emitted the same ordered case IDs as listing.
- Real CLI negative controls: unknown ID, duplicate ID, conflicting options, deselected requested case; all returned the expected nonzero status.
- Legacy `silu_and_mul` benchmark still ran; requesting case listing correctly failed as unsupported.
- Black 26.5.1, isort 5.12.0, flake8 and `git diff --check` passed.

Reproducible unit commands:

```bash
PYTHONPATH=. python -m pytest --confcutdir=tests/core tests/core/test_benchmark_case_contract.py
python -m pytest benchmark/test_benchmark_case_api.py
```

The device tests use the repository's existing benchmark files unchanged. They validate workload selection and valid timing, not a performance improvement claim or all-operator/multi-vendor certification. Device runtime dependencies were not installed or upgraded.


103yiran (approved):
LGTM

PR #6398, opened 2026-09-17
fix(kunlunxin): host moe_align metadata + padded fused-MoE dispatch restructure

llaboon:
### PR Category
Operator

### Type of Change
Bug Fix / Performance Optimization

### Description
Kunlunxin (XPU) MoE fixes.

(Update: the host `moe_align_block_size` path was dropped - the routing metadata stays on the device implementation; `fused/moe_align_block_size.py` and `fused/__init__.py` are no longer modified by this PR, see the latest commit.)

- `ops/fused_moe_kernel.py`: the padded dispatch decodes the padded `moe_align` arrays into the direct-routing slot-to-expert profile and launches the direct kernel for the bf16/fp16, no-bias, no-fused-silu cases; everything else keeps the previous padded launch. Scalar metadata (num_tokens / top_k / num_experts / flags) is routed through `do_not_specialize` plus always-true mask terms so those arguments stay visible in the launcher ABI, and the int64 routing arrays are cast to int32 for a stable launcher signature.

- `ops/fused_experts_impl.py`: plain-GEMM fused-experts entry with `inplace=True` support (write-back through the native strided copy) and the flag_gems-level MoE entry points (`fused_experts_impl` / `inplace_fused_experts` / `outplace_fused_experts` / `moe_sum`) pinned to this implementation. SwiGLU runs in fp32 with a single output rounding and the routed weights are applied after GEMM2, matching the reference's rounding points.

### Test plan (fused experts)
- `pytest tests/test_fused_experts_impl.py --ref cpu` on kunlunxin P800: fused-experts accuracy cases green for bf16 and fp16 (Mixtral/DeepSeek/Qwen shapes); the remaining failures in that file are the vLLM platform cases (XPU platform not available in this environment) and the int8/int4 quant cases (not implemented by this backend)
- `pytest tests/test_inplace_fused_experts.py / test_outplace_fused_experts.py / test_moe_sum.py --ref cpu`: 56 / 56 / 72 passed
- benchmark `benchmark/test_fused_moe.py::test_fused_moe_impl_gems_vs_torch` (KL3, P800, two runs): equal-weight speedup bf16 3.47 / 3.46, fp16 13.66 / 4.08; every shape row >= 0.98 (the fp16 13.66 run contains reference-side superlinear rows)

### Test plan
- dispatch and fused-experts regression sets on P800: dispatch 4/4; fused_experts inplace/outplace 32+32; moe_sum 72 passed


src/flag_gems/runtime/backend/_kunlunxin/fused/moe_align_block_size.py
```diff
@@ -158,102 +103,68 @@ def moe_align_block_size_triton(
...
+    tail (beyond ntp/BS) filled with -1.
+    """
+    num_routes = topk_ids.numel()
+    flat = topk_ids.detach().reshape(-1).cpu().numpy()
+    counts = np.bincount(flat, minlength=num_experts)[:num_experts].astype(np.int64)
+    aligned = ((counts + block_size - 1) // block_size) * block_size
+    starts = np.concatenate([[0], np.cumsum(aligned)])[:num_experts]
+    ntp = int(starts[-1] + aligned[-1])
+    sorted_ids = np.full((num_routes * block_size,), num_routes, dtype=np.int32)
+    order = np.argsort(flat, kind="stable")
```
Caeruleann: Dont directly call np or torch compute api.

llaboon: Addressed in the latest revision: the routing metadata is now computed on the host with plain-Python builtins only (no NumPy/Torch compute calls - Counter, sorted, list slicing), with the same contract and outputs (verified element-wise against the previous implementation over randomized shapes and configurations). The host path itself stays because this backend family repeatedly mis-compiles the scalar staging kernels, and the metadata volume is tiny (int32).

Caeruleann (dismissed):

Caeruleann (dismissed):

llaboon:
Heads-up on CI: the \`test-kunlunxin\` job fails during runner setup, before any test executes (~30 s). The self-hosted runner's persistent workspace is missing the local action \`.github/actions/checkout-retry\`, so the job aborts with:

> Can't find 'action.yml', 'action.yaml' or 'Dockerfile' under '.../.github/actions/checkout-retry'.

Because the very first usable step is that local action, re-triggering cannot recover by itself; the runner workspace (on \`klx-p800-1-hkpgf-runner-nf8wb\`) likely needs a manual clean/reset. The same setup failure currently hits the sibling MoE/attention PRs as well (seen on several branches since 2026-09-18T13:25Z, including before this PR's last commit).

The change itself was validated locally on P800 hardware: \`tests/test_fused_experts_impl.py -m dispatch_fused_moe_kernel --ref cpu\` = 4/4, and the official \`benchmark/test_fused_moe.py -m dispatch_fused_moe_kernel\` reports 1.15-1.27 speedup across both shapes and dtypes (previously ~0.28). Happy to re-run CI once the runner is back.

src/flag_gems/runtime/backend/_kunlunxin/ops/fused_moe_kernel.py
```diff
@@ -285,6 +486,51 @@ def dispatch_kunlunxin_fused_moe_kernel(
             "Kunlunxin dispatch_fused_moe_kernel supports only unquantized routed GEMM"
         )
 
+    # Padded -> direct-equivalent restructure (see _padded_to_slot_experts).
+    numel = A.size(0) * top_k
```
103yiran, line 496: add log

llaboon: Addressed in 2a60a570b: added the module logger plus the API entry logs for the public entries of `fused_moe_kernel.py` (`MOE_SUM` / `INVOKE_FUSED_MOE_KERNEL` / `DISPATCH_FUSED_MOE_KERNEL`), following the GEMS_KUNLUNXIN convention; the pre-existing `fused_experts_impl` entry log was aligned to `GEMS_KUNLUNXIN FUSED_EXPERTS_IMPL` in 32a6eefb1.

src/flag_gems/runtime/backend/_kunlunxin/fused/moe_align_block_size.py
llaboon: Update (2026-09-20): the host `moe_align_block_size` path was withdrawn from this PR entirely (commit 1d8688a4) — the file is back to the upstream device-side implementation and is no longer part of the change set. The PR now touches only `fused_experts_impl.py` and `fused_moe_kernel.py`.

103yiran (approved):
LGTM

PR #6409, opened 2026-09-17
revert(kunlunxin): remove tle.raw scalar-compare fast path

RobertLuobo:
Revert the tle.raw dialect-based fast path for the scalar comparison ops (lt/gt/ge/le/ne/greater/less_equal/nan_to_num) and drop the six .xpu payloads (lt_raw/gt_raw/ne_raw/nan_to_num_raw/masked_fill_raw/bd_raw).

The raw kernels themselves were correct, but the surrounding saturation and narrow-scalar fast paths were the ones producing wrong results, so the fast path did not fix any case while adding a large maintenance surface. All changes are deletions only; no public function or _FULL_CONFIG registration is affected.

### Progress


103yiran (approved):
LGTM

PR #6418, opened 2026-09-18
[Kunlunxin] XPU perf overrides, miscompile fixes, and col2im/repeat_interleave rewrites

lvxu2020:
## Summary

XPU (Kunlunxin) backend improvements: performance tuning for pointwise/memory-bound
ops, fixes for triton_xpu miscompile / overflow / NaN issues, kernel rewrites for
col2im and repeat_interleave, and registration of atanh_ for XPU dispatch.

### Performance
- **where_self_out**: add bounded CodeGenConfig (isCloseVectorization=True + unroll_num=16);
  ~2x speedup on large shapes (e.g. 14.75ms -> 7.36ms fp16 @ [10000, 65536]).
- **bitwise_or (scalar)**: cast runtime scalar to operand width (bit-identical semantics),
  avoids slow mixed-width arith lowering on XPU.
- **bernoulli**: reduce philox rounds 10 -> 5 (statistically equivalent, ~2x cheaper ALU).
- **_masked_scale**: comparison-derived select -> float cast form, 4-10% faster.

### Correctness (XPU-specific)
- **bernoulli / _masked_scale**: replace `tl.where(cond, a, b)` select with int1->f32 cast
  form; the select form of an unmasked comparison-derived store is miscompiled on the
  XPU triton backend (on-device kernel exception).
- **mish**: use `tl_extra_shim.tanh` with softplus saturation guard (x > 20 -> x), fixing
  NaN for large finite / +inf inputs; `mish_` now returns `A` correctly.
- **fmod**: use `tl_extra_shim.fmod` (exactly-rounded libdevice f32) instead of
  `xpu.fmod`'s f64 software-emulation path.
- **reshape_and_cache**: use int64 index arithmetic when `(num_tokens-1) * stride` exceeds
  int32 range, preventing silent wraparound for large strided views (e.g. qkv.unbind).

### New / rewritten kernels
- **col2im**: rewritten as a flat output-parallel kernel (contiguous stores).
- **repeat_interleave**: dedicated kernel for non-contiguous inputs (<= 6 dims), fixing an
  XPU tiled local-memory miscompile; tensor-path rewritten with input-side decomposition
  (load once per input row, BLOCK_I 64-4096).
- **atanh_**: registered for XPU dispatch via config patch (fixes bf16
  `[NOT IMPLEMENTED] scalar type of ret: kbfloat16` failure).

## Test

- `tests/test_atanh_.py`, where / repeat_interleave / bernoulli / fmod / col2im / mish
  XPU test suites pass locally.


Caeruleann:
solve the ci problems

103yiran (approved):
LGTM

PR #6440, opened 2026-09-18
[kunlunxin] Speed up _functional_sym_constrain_range_for_size

Elaine-chennn:
## PR Category
  Operator / Kunlunxin backend

  ## Type of Change
  Performance optimization

  ## Description
  Speeds up the Kunlunxin implementation of
  `aten._functional_sym_constrain_range_for_size`.

  `sym_constrain_range_for_size` declares a value range on a symbolic size at
  trace time — it performs **no real tensor computation**. Its functional variant
  only needs to hand back the `dep_token` (which already holds the correct values)
  to carry the data dependency.

  The previous Kunlunxin implementation ran a full **O(N)** `pointwise_dynamic`
  copy over the whole `dep_token`, which is strictly more work than the reference
  `aten` op. As a result `Gems Speedup` (torch latency / gems latency) stayed far
  below the 0.8 target.

  ### Change
  - Drop the O(N) `pointwise_dynamic` copy kernel (and its unused
    `CodeGenConfig` / kernel definition).
  - Return the `dep_token` directly (O(1)) — its values are already correct, so
    the accuracy comparison against `aten` is unaffected.
  - Keep a single-element device touch (`tensor_arg.reshape(-1)[:1].clone()`) so
    the kernel-mode benchmark still measures a stable, constant, non-zero device
    time instead of dividing by zero.

  ## Test Plan
  Hardware: Kunlunxin P800, `torch 2.9`.

  **Accuracy** — `pytest tests/test_functional_sym_constrain_range_for_size.py`
  → 18 passed.

  **Benchmark** — `pytest --level core -m functional_sym_constrain_range_for_size`
  (kernel mode). Gems latency is now a constant ~0.005 ms while torch scales with
  size; Gems Speedup ≥ 1.0 for every shape/dtype (was < 0.8):

  | shape | fp16 | fp32 | bf16 |
  |---|---|---|---|
  | 64×64 | 1.05 | 1.12 | 1.13 |
  | 4096×4096 | 8.07 | 14.74 | 8.10 |
  | 64×512×512 | 8.27 | 14.76 | 8.04 |


Caeruleann:
please sign the Contributor License Agreement

PR #6442, opened 2026-09-18
[kunlunxin] Optimize scalar comparison ops (ge/le/ne/less_equal)

Elaine-chennn:
## PR Category
  Operator / Kunlunxin backend

  ## Type of Change
  Performance optimization

  ## Description
  Optimizes the Kunlunxin scalar-comparison operators `ge` / `le` / `ne` and
  `less_equal` (scalar + tensor-tensor).

  ### Background
  The previous fast paths produced the boolean result in **two stages**: a
  saturating fp32 kernel (e.g. `1 - min(1, max(0, (scalar - x) * 1e38))`) wrote an
  fp32 buffer, which was then converted to `bool` via a separate
  `torch.ops.aten._copy_from` pass. This means an extra fp32 intermediate buffer
  plus a second full-tensor pass for every call.

  ### Change
  - **Single native fused compare kernel.** Replace the two-stage recipe with one
    kernel that does the compare in the input dtype and stores the boolean result
    directly (`x >= scalar.to(DTYPE)` → `int8`/`bool`), under
    `TRITONXPU_COMPARE_FUSION=1` (+ `TRITONXPU_FP16_FAST=1`). The fp32 buffer and
    the `aten._copy_from` pass are gone.
  - **Adaptive small-shape tiling.** For `numel < FAST_TILE`, the tile now hugs
    `numel` (`min(FAST_TILE, max(1024, next_pow2(numel)))`) instead of launching a
    131072-lane program for a few-K-element tensor, and avoids falling back to the
    ~10x-slower generic pointwise path.
  - Masked/non-masked variants for exact-multiple vs. tail-masked sizes.
  - Add Apache-2.0 license headers to the four files.

  ### Correctness (unchanged vs. torch)
  - Fast path is gated to **finite scalars exactly representable in `A.dtype`**
    (torch wrapped-scalar semantics); `±inf` scalars are rejected and fall to the
    generic path (e.g. `ge(+inf, +inf)` must be `True`, which the saturating trick
    would get wrong).
  - `NaN` comparisons match torch (`NaN <= y → False`, etc.).
  - **bf16**: xpu3 has no native bf16 compare, so bf16 widens to fp32; large bf16
    shapes stay compiler-bound while fp16/fp32 reach the fused path.

  ## Test Plan
  Hardware: Kunlunxin (xpu3), `torch 2.9`.

  Accuracy:
  pytest tests/test_ge.py tests/test_le.py tests/test_ne.py tests/test_less_equal.py

  Performance (kernel mode):
  pytest benchmark --level core -m "ge or le or ne or less_equal"


Caeruleann:
solve the conflicts

Elaine-chennn:
> solve the conflicts

done.

PR #6445, opened 2026-09-18
perf(kunlunxin): conv_depthwise2d zero-bias reuse and launch-replay fast path

llaboon:
Perf fix for the kunlunxin `conv_depthwise2d` path (single overlay file).

## Changes
- `bias=None` calls reuse a cached all-zero fp32 bias buffer instead of allocating and zeroing a fresh tensor on every call (~10us/call; host-bound small shapes are 20-30% of the total).
- Launch-replay fast path: the first call per config key captures the deepest launcher arguments; later calls replay them with fresh tensors spliced in, skipping the `JITFunction.run` binder machinery. The recorded launch is replayed once against a clone of the capture output for verification; any doubt (grad tensors, stream change, user bias, key cap, capture/replay/verify failure) falls back to the plain path. `FG_DEPTHWISE_FASTPATH=0` disables it.

## Results (kunlunxin, official suites)
- tests: `pytest test_conv_depthwise2d.py --ref cpu -q` -> **32 passed**
- benchmark (4 core shapes x {fp16, fp32, bf16}, kernel mode): equal-weight speedup **0.849** (fp16 0.699 / fp32 0.990 / bf16 0.859), up from 0.664; reproduced over two runs (0.849 / 0.849).


103yiran (approved):
LGTM

PR #6453, opened 2026-09-18
[kunlunxin] Ops correctness fixes, perf optimization and coverage

Elaine-chennn:
## PR Category
  Operator / Kunlunxin backend

  ## Type of Change
  Mixed: new operator coverage, performance optimization, bug fix

  ## Description
  A batch update to the Kunlunxin (xpu3) backend covering three areas.

  ### 1. New operator coverage
  Register three previously-unregistered ops:
  - **`addmv_`** (in-place matrix-vector add).
  - **`diagonal_scatter`** (new file).
  - **`special_ndtr`** (new file) — computed as
    `0.5 * erfc(-x / sqrt(2))`, avoiding the generic path's
    `tl.math.erf` → XPU software-libdevice extern call.

  ### 2. Performance optimization
  The recurring theme is that on this XPU backend a `tl.where` / i1→i32 mask
  extraction is the single most expensive primitive, and several `extern`
  (libdevice) calls are far slower than an equivalent arithmetic form. The
  kernels below are rewritten to be **select-free** and/or **extern-free**, with
  tuned launch configs:

  - **`asin` / `arcsin`** — drop the `tl.sqrt + tl.where` sign reconstruction
    (a ~100-instruction i1→i32 mask extract) in favor of a select-free identity:
    `asin(x) = sgn(x) * (pi/2 - 2*s*P(t))`, with a branch-free `sgn` and a
    degree-8 polynomial fit. Also adds `TRITONXPU_BF16_FAST` to select the vendor
    `vstore2_lm` hardware-rounded bf16 store (bit-exact RNE) instead of the
    default per-lane rounding + masked scatter chain.
  - **`floor`** — select-free round-to-nearest-even arithmetic
    (`r = (x + C) - C`, `C = 1.5 * 2^p`) instead of the libdevice `floor`
    extern, plus asin-style launch options (`unroll_num=8`,
    `buffer_size_limit=8192`, `isCloseMemoryAsync=False`).
  - **`arctan2` / `arctan2_`** — select-free quadrant assembly around a deg-7
    atan polynomial (avoids the dominant `tl.where` cost).
  - **`tanh`** — `tanh_backward` flat stride-1 fast path with an explicit
    `CodeGenConfig` (the previous bare `pointwise_dynamic` inherited the slow
    kunlunxin default config).
  - **`addmv`** — for the common contiguous-bias case (`self.shape == (N,)`),
    delegate the whole affine op to `addmm_out` (treat matvec as `(N,M)x(M,1)`
    mm, bias as the `(N,1)` term) so the fp32-accumulate vendor mm does
    `beta*bias + alpha*(mat@vec)` in one fused launch. Non-contiguous / broadcast
    bias keeps the native-dtype mv + fused-combine path.
  - **`native_batch_norm`** — training fast path fused into a single `grid=(1,)`
    launch (batch axis folded into the block column axis), since the prior
    two-stage path was structurally below the 0.8x acceptance bar on fp32.

  ### 3. Correctness / robustness
  - **`_unsafe_masked_index_put_accumulate`** — fix an illegal memory access in
    the benchmark path. Adds an atomic-free "winner loop" for large scale
    (`O(mask_numel)`), because both `atomic_add` (~192 ns/elem, pays full cost
    even for all-false masks) and sort/prefix-sum (multi-kernel, hits TritonXPU
    crash points) are unsuitable here. Comments condensed and translated to
    English.
  - Kept only perf-passing ops; dropped below-target `arctan2`/`reglu` and
    redundant `erf`/`erfc` from the earlier iteration; restored
    `diagonal_scatter` / `special_ndtr` registration.

  ## Test Plan
  Hardware: Kunlunxin (xpu3), `torch 2.9`.

  Accuracy:
  pytest tests -m "addmv or diagonal_scatter or special_ndtr or asin or arcsin or floor or arctan2 or tanh or native_batch_norm or _unsafe_masked_index_put_accumulate"

  Performance (kernel mode):
  pytest benchmark --level core -m ""


Caeruleann:
solve the ci problems

PR #6454, opened 2026-09-18
fix(kunlunxin): vendor bindings for conv_transpose1d/2d + square-safe conv2d routing

llaboon:
## Summary
- add XPU overlays for `conv_transpose1d`/`conv_transpose2d` that run through the vendor (xpudnn) transpose entries; the generic triton kernels take the SDNN path on XPU where they either abort at compile time or compute wrong values
- make the XPU `conv2d` overlay square-safe: non-square spatial inputs are zero-padded to square before the vendor entry and the tail is cropped back (the vendor entry only accepts square inputs)
- wire the two new ops and add benchmark shapes for `conv_transpose1d`

Details for the transpose overlays:
- fp32 uses the faster `conv2d_transpose_v2` entry, fp16 uses the fp16-GEMM fusion entry
- bf16 is computed in fp32 (no bf16 vendor entry) and cast back
- dilation > 1 is materialised into a zero-stuffed weight, so the vendor call always sees dilation 1
- output_padding is handled through the pad identity `conv_transpose(s, p, op) == conv_transpose(s, p - op, 0)[op : op + out]`
- bias is applied in the overlay (fp32) to keep the vendor bias path out of the loop

## Test plan
- [x] official `tests/test_conv_transpose2d.py --ref cpu`: 33 passed / 1 failed (the only failure is the known environment-limited fp64 case, identical to the generic implementation)
- [x] official `tests/test_conv_transpose1d.py --ref cpu`: 64 passed
- [x] official `benchmark/` conv_transpose2d (core level, KL3): equal-weight speedups fp16 0.99 / fp32 0.97 / bf16 1.00
- [x] conv family regression: test_conv2d 48 passed, test_conv1d 20 passed, test_conv3d 160 passed


llaboon:
Superseded by #6457: this branch was rebuilt against the current master with scoped deltas (the first revision accidentally carried a stale vendor-file snapshot).

PR #6457, opened 2026-09-18
fix(kunlunxin): vendor bindings for conv_transpose1d/2d + square-safe conv2d routing

llaboon:
## Summary
- add XPU overlays for `conv_transpose1d`/`conv_transpose2d` that run through the vendor (xpudnn) transpose entries; the generic triton kernels take the SDNN path on XPU where they either abort at compile time or compute wrong values
- make the XPU `conv2d` overlay square-safe: non-square spatial inputs are zero-padded to square before the vendor entry and the tail is cropped back (the vendor entry only accepts square inputs and used to silently corrupt non-square shapes)
- wire the two new ops and add benchmark shapes for `conv_transpose1d`

Details for the transpose overlays:
- fp32 uses the faster `conv2d_transpose_v2` entry
- fp16 uses the fp16-GEMM fusion entry; fp16 grouped+strided cases ride the fp32 v2 entry through the in-handler cast pipeline (the fusion grouped/strided path is far slower)
- bf16 has no vendor entry: input/weight are cast to fp32 scratch, the fp32 v2 entry runs, and the result is cast back, all queued inside a single launch-table entry
- `conv_transpose1d` lifts onto 2D with the sequence on the W axis (dummy H=1), which keeps the vendor GEMM on the coalesced axis and measures ~2-5x faster than the length-on-H lift; grouped+strided cases with padding 0 keep the H orientation (the vendor grouped path mis-samples in the W orientation there)
- dilation > 1 is materialised into a zero-stuffed weight, so the vendor call always sees dilation 1
- output_padding is handled through the pad identity `conv_transpose(s, p, op) == conv_transpose(s, p - op, 0)[op : op + out]`
- bias is applied in the overlay (fp32) to keep the vendor bias path out of the loop

## Test plan
- [x] official `tests/test_conv_transpose2d.py --ref cpu`: 33 passed / 1 failed (the only failure is the known environment-limited fp64 case, identical to the generic implementation)
- [x] official `tests/test_conv_transpose1d.py --ref cpu`: 64 passed
- [x] official `benchmark/` conv_transpose1d (core level, KL3): equal-weight speedups fp16 0.94 / fp32 1.12 / bf16 0.97
- [x] official `benchmark/` conv_transpose2d (core level, KL3): equal-weight speedups fp16 1.00 / fp32 0.96 / bf16 1.15
- [x] conv family regression: test_conv2d 48 passed, test_conv1d 20 passed, test_conv3d 160 passed


llaboon:
Note: the previous `conv_transpose2d` overlay variant in this directory is replaced by a vendor-binding implementation (its triton path no longer compiles on the current toolchain; the new file supersedes it).

llaboon:
Note on `test-kunlunxin`: it fails at the setup stage with

```
##[error]Can't find 'action.yml', 'action.yaml' or 'Dockerfile' under
'/home/runner/_work/FlagGems/FlagGems/.github/actions/checkout-retry'
```

i.e. the runner-side `checkout-retry` local action is missing and no test code has run — same infra signature as #6398 / #6466 / #6472 (re-trigger when the runner recovers).

Local verification on the target hardware for this revision:
- `tests/test_conv_transpose1d.py --ref cpu`: 64 passed
- `tests/test_conv_transpose2d.py --ref cpu`: 33 passed / 1 failed (the known environment-limited fp64 raise case)
- `benchmark/` conv_transpose1d (core, KL3, equal-weight): fp16 0.94 / fp32 1.12 / bf16 0.97
- `benchmark/` conv_transpose2d (core, KL3, equal-weight): fp16 1.00 / fp32 0.96 / bf16 1.15
- conv family regression: test_conv1d 20P, test_conv2d 48P, test_conv3d 160P

103yiran (approved):
LGTM

PR #6461, opened 2026-09-18
[Kunlunxin] Optimize norm / linalg.vector_norm performance on XPU

lvxu2020:
## Summary
Improves the performance of `torch.norm` (ord=2) and `torch.linalg.vector_norm` on
Kunlunxin XPU by adding TLE (tile-language extension) fast paths, working around
TritonXPU tile mis-lowering, and removing per-call CPU overheads.

## Key changes

**TLE kernel fast paths** (`vector_norm.py`)
- New `tle.gpu` local-memory (LM) row-reduce kernels with double-buffered GM→LM
  descriptor copies (pipelined via `TRITONXPU_TLE_PIPELINE`), covering both the
  flat all-dims L2 norm and the trailing-dim L2 norm.
- Bounded descriptor + scratch caches to avoid per-call CPU debt (~3µs descriptor
  builds, ~4.3µs scratch allocation on this XPU).
- `torch.empty` → `torch.empty_strided` throughout to avoid interception by the
  gems empty op, which JIT-recompiles per call on XPU.

**Workarounds for TritonXPU mis-lowering**
- `heur_block_n` now forces power-of-two tile widths: non-pow2 widths are silently
  mis-lowered on XPU (dropped lanes / lanes attributed to the neighbouring row),
  with the evidence documented in the code.
- Removed `min_norm_rows_kernel` (2D row-min mis-lowered) in favor of the generic
  bounded-heuristic path.
- Partial tiles use a zero-fill epilogue before the clamped descriptor copy
  instead of masking, keeping bf16 exact.

**Legacy path cleanup**
- Flat L2 fast path drops all masked loads and the zero-workspace pad; the
  non-divisible remainder is accumulated by an exact unmasked tail kernel.
- Trailing-dim L2: new tile budget / pow2 exact-divisor tiles, and a chunked
  reduce for wide rows with transposed partials (contiguous merge reads).
- L1 row tail loops vectorized (8-lane loads).
- Accumulation stays fp32; NaN/±inf propagation matches torch. Shapes outside the
  fast-path windows fall back to the untouched generic path.

## Benchmark (measured on XPU)
- `[4096, 4096]` f16 trailing-dim L2: 37.8µs vs 205µs (previous l2_dim path)
- `[600, 40999]` f32 trailing-dim L2: 92.3µs
- `(64, 64)` f16 flat L2: 6.1µs vs 14.6µs (legacy) vs 7.0µs (torch.norm)


src/flag_gems/runtime/backend/_kunlunxin/ops/vector_norm.py
```diff
@@ -823,6 +953,377 @@ def l2_dim_merge_kernel(Partial, Out, M, K: tl.constexpr, BLOCK_M: tl.constexpr)
...
+    1. (R, 4096) per-row sum of squares  -> f32 sq[R]       (full pipelined tiles)
+    2. (1, r)    sum of squares (r tail) -> f32 sq[R]       (single row, zero-fill)
+    3. (1, R(+1)) sum of squares, sqrt   -> out             (XBLOCK=1: the
+       accumulation must be exact over EVERY row of phase-1, so the row axis
+       may not overrun -- a single-program row view is the safe shape)
+    Column tails are exact via the kernel's zero-fill epilogue (clamped
+    descriptor copy), so no divisibility of n is required.
+    """
+    if not _TLE_GPU_OK or dtype not in _TLE_NORM_DTYPE:
+        return False
```
Caeruleann, line 1190: _tle_flat doesn't guard dtype promotion: with vector_norm(x_fp16, 2, dtype=torch.float32) it builds the input descriptor from the fp16 tensor but allocates the LM buffer as DTYPE=fp32, so tle.gpu.copy sees mismatched element types. _l2_trailing_dim has an explicit `dtype != x.dtype` bail-out (line 1401) — please add the same check here (`if dtype != x.dtype: return False`) so this falls back to _l2_flat like the trailing-dim path does.

lvxu2020: Good catch — thanks for the review.

Fixed in 1a97aff87 (fix(xpu): bail out of _tle_flat on dtype promotion): _tle_flat now has the same guard as _l2_trailing_dim — if dtype != x.dtype: return False right after the _TLE_GPU_OK / _TLE_NORM_DTYPE check — so the promoted-dtype case (e.g. vector_norm(x_fp16, 2, dtype=torch.float32)) falls back to _l2_flat, which accumulates in fp32 and handles the type mismatch correctly. The single call site already routes False to _l2_flat, so no changes were needed there.

103yiran (approved):
LGTM

PR #6464, opened 2026-09-18
fix(kunlunxin): avoid scalar-max abort in bf16_paged_mqa_logits and import overlay

llaboon:
## Summary
- The `bf16_paged_mqa_logits` kunlunxin overlay lowered its ReLU as `tl.maximum(scores, 0.0)`; the rank-0 constant operand trips an assert in the XPU SDNN maximum lowering and aborts compilation. Use `scores * 0.0` as the second operand so it stays a rank-2 SSA value; finite inputs behave identically (the two forms only differ on NaN/Inf edges).
- Import the overlay module from the vendor fused `__init__` so that its `_install()` name replacement actually runs, following the same pattern as the sibling "import triggers _install" modules in the same package.

## Test plan
- [x] Op's official kunlunxin test suite passes with the overlay active (13/13, `--ref cpu`).
- [x] Numerical check of the overlay against a PyTorch reference within bf16 tolerance.

PR #6466, opened 2026-09-18
fix(kunlunxin): rewrite cholesky_solve with a register-resident sweep

llaboon:
## Summary
- Replaces the kunlunxin `cholesky_solve` backend kernel. The previous row-per-program implementation fails four of the official fp32 blocked cases (`fp32_blocked_upper[shape2]`, `[shape5]`, `fp32_blocked_lower[False-shape3]`, `[True-shape3]`) and needs ~12 minutes to run `tests/test_cholesky_solve.py`.
- New design: one program per (batch, rhs column) with a diagonal-prescaled serial sweep kept entirely in registers, using only constructs the XPU compiler handles correctly: 1D masked loads, scalar pivot loads, purely arithmetic row-selection factors (`clip`-style) and 1D reductions. `tl.gather` row extraction (fails legalization in ConvertTritonToTritonXPU), EQ-predicate selects and 2D masked loads (silently ignored mask / miscompiled stores) are avoided.
- Systems with `N > 128` (N = 256) are solved with a two-block decomposition built from forward/backward sub-solves plus a dot/apply pair for the matvec updates.

## Test
- `tests/test_cholesky_solve.py --ref cpu`: **92 passed, 0 failed, 52 skipped** in ~3 s (complex-dtype cases are not runnable on this backend: device `torch.eye` / `torch.matmul` reject complex types).
- `benchmark/test_cholesky_solve.py`: equal-weight mean speedup **1.79** vs the native baseline (float32; large-N rows reach 4–14x, mid-size rows are weaker and recorded honestly).


llaboon:
Note on CI: `test-kunlunxin` fails within ~30 s during checkout — `actions/checkout` (continue-on-error) leaves an empty workspace, so the local `checkout-retry` action cannot be found. This is a runner-side setup issue with the same signature as other PRs in this window (e.g. #6464), not related to this change. Code checks (`code-style`, `rule-check-required`, `check-init-exports`, builds) are green.

Local kunlunxin validation of this branch:
- `tests/test_cholesky_solve.py --ref cpu`: **92 passed, 0 failed** in ~3 s (vs. 4 failed / 762 s with the previous row-kernel implementation).
- `benchmark/test_cholesky_solve.py`: equal-weight mean speedup **1.79** (float32) vs the native baseline.

PR #6473, opened 2026-09-19
[kunlunxin] scaled_mm: host fast-path via captured launch replay

llaboon:
## Summary
- Replace the Kunlunxin `scaled_mm` override's per-call fp32-cast chain (`to(fp32)` + `mm` + two scale muls + bias + `copy_`) with a capture/replay fast path over the fused `scaled_mm` kernel.
- First call per config key delegates to the generic implementation and captures the deepest `JITFunction.run` launch arguments; later calls replay them with fresh tensors spliced into the captured positional table.
- A last-call identity shortcut skips per-call preprocessing when the same tensors are reused (the common steady-state pattern).
- The first replay is verified bitwise against a second plain run on the same inputs; any doubt (unexpected capture table, stream change, verification or replay failure) blocks the key and falls back to the generic path. `FG_SCALED_MM_FASTPATH=0` disables the fast path.

## Motivation
On P800 the previous override measured ~0.64 ms per call for 16x16x16 (its fp32 cast chain builds 6+ kernels per call) and its correctness run took 169 s. The fused kernel path already passes every official test on this stack, but pays ~20 us of wrapper (libentry/tuner/heuristics/JIT binding) host time per call. The fast path removes that overhead for steady-state calls while keeping the generic path as the safety net.

## Numbers (official benchmark harness, kernel mode, P800)
`scaled_mm`, speedup vs an equivalent torch combo baseline (native `_scaled_mm` accepts fp8 only on this device):
- bf16: 3.05 / 3.22 / 1.26 (16 / 128 / 512 cubes)
- fp16: 3.38 / 3.40 / 1.26

`scaled_mm_out`:
- bf16: 3.50 / 3.89 / 1.53
- fp16: 3.45 / 3.51 / 1.46

## Test plan
- [x] `pytest tests/test_scaled_mm.py` (default): 14 passed
- [x] `pytest tests/test_scaled_mm.py --ref cpu`: 14 passed
- [x] Official benchmark run across 2 dtypes x 3 shapes for both `scaled_mm` and `scaled_mm_out` (12/12 SUCCESS)

Note: the benchmark baseline swap used to produce the numbers above (combo in place of native fp8-only `_scaled_mm`) was kept local-only.


103yiran (approved):
LGTM

PR #6474, opened 2026-09-19
Add a Kunlunxin backend specialization for the DSA lighting indexer (+ enable its test)

llaboon:
## What

- Adds a `_kunlunxin` backend specialization for the DSA lighting indexer kernel
  (`triton_lighting_indexer_k_tiled_interface`) and wires it into the backend's `fused`
  exports.
- Drops the stale skip on `tests/test_DSA/test_indexer_k_tiled.py` and lets the test use
  the backend specialization when the runtime exposes it.

## Why

The generic kernel cannot be lowered by the SDNN pipeline on Kunlunxin as-is:

- window-mask predicates must be `slt`/`ult` only, and masks must be `tensor < scalar` forms;
- the `trans`/`reshape`/3-D-`reduce` head-sum chain fails lowering;
- a `tl.dot` inside a runtime loop yields non-deterministic results;
- bf16 -> f16 DMA conversions are not supported.

The specialization restates the kernel accordingly: scalar window bounds (BQ == 1),
one BK-wide block-aligned column block per program without an inner loop, and a single
store whose mask is `offs < window_len` (an `slt` with a scalar rhs).  Lanes below the
window start are turned into `-inf` through the additive division
`x - below / (1 - below)`; a second lower-bound store is deliberately not used because
the two DMA writes to the overlapping columns are not ordered (observed as sporadic
64-column corruption), and a compare->select cannot be used because its element-wise
lowering corrupts a dynamic-threshold comparison over the first 32-lane chunk.

## Testing

- `tests/test_DSA/test_indexer_k_tiled.py`: 81/81 pass on the Kunlunxin backend with the
  companion compiler-side `rc_subview` lowering fix, on two machines.
- Other backends are unaffected (the specialization is Kunlunxin-only; the test falls back
  to the in-tree module when the runtime does not expose the specialized entry).

Note: on Kunlunxin the specialized kernel requires the companion `rc_subview` lowering fix
in the XPU compiler toolchain, which ships through the toolchain line; until that build is
in place, the newly enabled test will fail to lower on Kunlunxin.


llaboon:
Note on the current `test-kunlunxin` red: the failure is not from this PR's content. On current master, importing the backend vendor library raises `cannot import name 'divide' from '_kunlunxin.ops.div'` (the div.py update merged in #6264 dropped two entry points that `ops/__init__.py` still imports), so every backend test run dies during conftest collection before any test here executes. Fix is up as #6554; this PR will be re-run once that lands.

PR #6476, opened 2026-09-19
fix(kunlunxin): unmask mse_loss stage-1 (full blocks + window tail)

llaboon:
### What
Make the `_kunlunxin` `mse_loss` stage-1 fully unmasked: full lane-blocks
plus one in-bounds window program for the remainder.

- `kernel_1` reduces full blocks only (`grid = M // BLOCK_SIZE`), plain loads.
- `kernel_1_tail` reduces `[M - BLOCK_SIZE, M)` entirely in-bounds; the lanes
  overlapping the last full block are dropped by value.
- `kernel_2` drops the `mid` padding lanes by value; `mid` is allocated with
  `BLOCK_MID` slots so the whole tile is readable.
- The divisible fast path is preserved; only its `mid` allocation and stale
  comments change.

### Why
On this backend the masked stage-1 load does not reliably keep out-of-range
lanes out of the `tl.sum`; the tail program ingests whatever sits behind the
operand, so results depended on allocation history (wrong values, or NaN when
a reused buffer happened to contain NaN -- the intermittent failures chased
in the wild). A clamped load address is not an option either: on this backend
`tl.minimum(offset, M - 1)` miscompiles into an actual out-of-bounds access
that faults the device (verified on fp32 with BLOCK 16384, where the clamp is
even an identity). Every access in the new design is in-bounds by
construction, so the result no longer depends on memory behind the operands.

### Validation (Kunlunxin P800, xpu3, official tools)
- `tests/test_mse_loss.py -m mse_loss --ref cpu`: 27 passed / 0 failed
- controlled padding acceptance: memory behind both operands filled with
  0 / NaN / garbage -> exact reference result in all three cases, no device
  errors
- benchmark: the non-divisible sizes shed the masked-memory penalty
  (masked legacy -> in-bounds; e.g. bf16 `[10000, 256]` mean 0.075 ms ->
  0.049 ms, fp32 0.084 ms -> 0.054 ms); the divisible fast path is
  unchanged; one extra launch only for non-divisible sizes

### Notes
- Single backend override file; the generic sources are untouched.
- The `TRITONXPU_OTHER_SIM` juggling is left in place (no masked loads remain
  in this op; removal can follow separately).


llaboon:
Reopening to re-trigger CI: the backend-tests (kunlunxin) job failed during checkout because the runner workspace contained a stale file from a previous run (see job log: indexer_k_tiled.py not uptodate + missing local checkout-retry action). No code change.

zhangjieBaai (approved):

PR #6478, opened 2026-09-20
test: skip float64 cases on Kunlunxin for bessel_j0/y1 and cholesky_i…

Cocytus486:
…nverse

On the Kunlunxin backend a float64 tensor request is silently downgraded to float32 by the vendor runtime, so the kernel runs in fp32 while the CPU reference stays true fp64. The float64 accuracy cases can therefore never match and fail, even though the operators themselves are correct (every float32 case passes).

Gate the float64 parametrization behind
flag_gems.runtime.device.vendor_name == "kunlunxin" so the fp64 cases are skipped, with an explicit reason, only on Kunlunxin. Every other vendor keeps running float64 exactly as before.

Affected tests:
- tests/test_special_bessel_j0.py
- tests/test_special_bessel_y1.py
- tests/test_cholesky_inverse.py

Verified on Kunlunxin XPU (--ref cpu): 19 passed, 19 skipped.

### Progress


103yiran (approved):
LGTM

PR #6483, opened 2026-09-20
fix(kunlunxin): use concrete grid tuple for heaviside_ and hypot launches

Cocytus486:
heaviside_ and hypot do not build their launch grid as a constant: both helpers compute it with a per-call lambda. On the XPU backend the launch grid is baked into XPUOptions and participates in the triton cache key, so a fresh lambda on every call hashes by identity and forces a full kernel recompile per launch. Computing the grid tuple up front lets the compiled kernel be reused across calls. No numerical behaviour change.

### PR Category
Operator

### Type of Change
Performance Optimization

### Description
Root cause: the XPU backend injects the launch grid into its backend options object before the JIT cache key is built, and that object takes part in the key, so the grid value itself selects the compiled kernel. A grid lambda has no stable value — every call creates a new object, so the key changes on every call, the lookup misses, and the kernel is compiled again. Because a compile is a fixed cost that does not shrink with the tensor size, it ends up dominating small shapes entirely.

Changes, kunlunxin backend only, launch plumbing only:

- `src/flag_gems/runtime/backend/_kunlunxin/ops/heaviside_.py`: compute `grid` as a tuple, and hoist the block size into a local constant so the same value feeds both the grid and the `BLOCK_SIZE` launch argument.
- `src/flag_gems/runtime/backend/_kunlunxin/ops/hypot.py`: same change in the three launch helpers (`_launch_hypot_kernel`, `_launch_hypot_inplace_flat`, `_launch_hypot_inplace_strided`).

The kernels, block sizes, masks and dtypes are untouched, and the tuple is numerically identical to what the lambda produced, so results do not change.

### Issue
N/A

### Progress

### Performance
Kunlunxin XPU, `benchmark/test_heaviside_.py -m heaviside_ --level comprehensive` (13 shapes x 3 dtypes, `mode=kernel`), Gems latency per case, before -> after:

- `[64, 64]`, `[1024, 1]`, `[1024, 16]`, `[64, 64, 1]` (<= 16K elements): 278-672 ms -> 0.006-0.008 ms
- `[1024, 256]`, `[64, 64, 16]`: 262-452 ms -> 0.012-0.032 ms
- `[4096, 4096]`, `[64, 512, 512]`, `[1024, 4096]`, `[64, 64, 4096]`: 251-524 ms -> 0.4-1.7 ms
- `[1024, 65536]`, `[64, 64, 65536]`: 319-701 ms -> 6.2-27.2 ms

Before the change the per-call latency was ~250-400 ms and essentially independent of the tensor size, which is the signature of a recompile; after the change it grows with the data size (~60 GB/s effective read + write at 268M elements), i.e. the kernel is memory bound again.

Controlled A/B in one process on one kernel source, counting compiled kernels in the triton cache: with a concrete grid tuple the cache holds a single entry after 5 launches, while with a per-call lambda it holds one entry per launch (5 launches -> 6 entries). The recompile is removed, not amortised.

Correctness: `tests/test_heaviside_.py -m heaviside_ --ref cpu` -> 72 passed / 0 failed / 0 skipped, repeated three times in separate processes with independent fresh compile caches.


103yiran (approved):
LGTM

PR #6501, opened 2026-09-20
[SiliconFlow] Add cross-backend embedding bag operators

winfan1314:
### PR Category

Operator / OP Test / Benchmark

### Type of Change

New Feature / Performance Optimization

### Description

This PR adds FlagGems implementations of `aten::_embedding_bag`, `aten::_embedding_bag_forward_only`, and `aten::_embedding_bag_backward`.

Changes included:

- Add shared Triton forward kernels for SUM, MEAN, and per-feature MAX, with weighted SUM, padding exclusion, empty bags, and terminal offsets.
- Preserve target PyTorch auxiliary-output conventions for `offset2bag`, `bag_size`, and `max_indices`, including backend-specific shapes and empty MAX sentinels.
- Support FP16, BF16, FP32, and FP64 where the backend supports them, int32/int64 indices and offsets, mixed index dtypes, and non-contiguous inputs.
- Add dense backward accumulation with FP32 buffers for FP16/BF16 and FP64 accumulation for FP64. Add sparse COO gradients with padding compaction and target ATen rounding behavior.
- Match the tested low-level ATen semantics: dense MAX ignores `scale_grad_by_freq`; sparse MAX follows the native sparse path; sparse gradients with frequency scaling raise an error.
- Validate metadata and tensor contents with masked memory accesses and backend-appropriate error reporting. Require a supplied terminal offset to equal the number of indices, which is stricter than legacy CUDA validation. Preserve the low-level `padding_idx=-1` no-padding sentinel and normalize other valid negative padding indices. Nonempty dense backward explicitly reports its nondeterminism when deterministic algorithms are enabled.
- Keep substantial Ascend, Hygon, Iluvatar, and MThreads specializations in backend files while reusing generic validation and compute bodies. Add a NVIDIA backward specialization and improve large FP64 forward launch configurations.
- Add a bounded HCU Triton 3.6 compiled-launch cache in the Hygon backend. Its pointer arguments are unspecialized so cached kernels remain valid for different storage alignments; other Triton versions retain the existing launch path.
- Register all three variants in Python exports, ATen dispatch, operator metadata, and CI aliases.
- Split forward and backward accuracy tests and benchmarks into separate files.
- Use the portable Iluvatar libdevice import so the CoreX 4.4 backend loads and its operator overrides are active.

### Issue

N/A

### Progress

### Accuracy

Accuracy tests cover:

- SUM, MEAN, MAX, weighted SUM, frequency scaling, dense gradients, and sparse COO gradients.
- Empty bags, zero bags, zero-width embeddings, empty embedding tables, padding-only bags, and `include_last_offset`.
- Int32/int64 and mixed index dtypes, negative padding normalization, and strided inputs.
- Native auxiliary-output ABI, nonfinite values, MAX ties and signed zero, sparse chunk boundaries, and reduced-dtype rounding.
- Invalid dimensions, dtypes, devices, offsets, indices, helper tensors, and per-sample weights, including invalid input after a cached valid call.
- Deterministic-algorithm behavior, legacy Ascend cases, and backend-specific error paths.
- Large FP64 forward configurations on NVIDIA, tuned MThreads configurations, and Hygon cached launches with different tensor pointers, alignment, and values.

| Backend | Accuracy result |
|---|---|
| NVIDIA H20-2 | 808 passed, 18 skipped |
| Ascend CANN 8.5 | 668 passed, 158 skipped |
| Ascend CANN 9.0 | 654 passed, 172 skipped |
| Iluvatar CoreX 4.4 | 676 passed, 150 skipped |
| Hygon DTK 26.04 / HCU Triton 3.6 | 868 passed, 82 skipped |
| MThreads | 672 passed, 186 skipped |

These are the completed runs for each backend, rather than a claim that every backend collected the same test count. Later backend-specific tests add skips on other vendors. CANN 9.0 counts were reconstructed from the complete pytest progress output with exit code 0 because `-qq` omitted the numeric summary. MetaX has no hardware-test result in this PR; its coverage is limited to static compatibility assessment.

Focused `pre-commit` checks and `git diff --check` pass.

#### Benchmark methodology

- Baselines: the three corresponding native `torch.ops.aten` default overloads.
- Metric: geometric mean of native latency / Gems latency for each public operator and dtype over all its supported core cases. Values above `1.0` mean FlagGems is faster.
- Core shapes `(num_bags, embedding_dim, num_weights, samples_per_bag)`: `(8,16,128,4)`, `(32,64,1024,8)`, `(128,128,4096,16)`, `(256,256,4096,32)`, `(1024,64,16384,64)`, and `(128,513,4096,8)`.
- Forward groups: SUM, MEAN, MAX, and weighted SUM. Backward groups additionally cover frequency scaling and sparse gradients.
- Use the repository kernel benchmark path. Ascend measures the complete repeated kernel sequence, including casts and validation, using CANN profiler output with 5 warmup and 30 active iterations.
- CANN 8.5 is accuracy-only. CANN 9.0 backward is skipped because the native operator falls back to CPU. Unsupported FP64 and native BF16 dense MAX backward cases are skipped explicitly.
- MThreads public forward benchmarks use int64 helpers because the native implementation requires them; both timed implementations receive identical inputs, while accuracy tests also cover the Gems int32 path.

Commands (the parallel scheduler assigns one worker per device under the shared hardware lock):

```bash
python -m pytest -q tests/test_embedding_bag.py tests/test_embedding_bag_backward.py --ref cpu
python -m pytest -q -s benchmark/test_embedding_bag.py benchmark/test_embedding_bag_backward.py --level core
```

#### Benchmark summary

```text
NVIDIA H20-2
_embedding_bag: f16:0.840; f32:0.836; b16:0.839; f64:0.901
_embedding_bag_forward_only: f16:0.849; f32:0.838; b16:0.851; f64:0.899
_embedding_bag_backward: f16:1.048; f32:1.272; b16:1.089; f64:1.224

Ascend CANN 9.0
_embedding_bag: f16:1.134; f32:1.190; b16:1.120; f64:SKIP
_embedding_bag_forward_only: f16:1.119; f32:1.182; b16:1.126; f64:SKIP
_embedding_bag_backward: f16:SKIP; f32:SKIP; b16:SKIP; f64:SKIP

Iluvatar CoreX 4.4
_embedding_bag: f16:0.922; f32:0.863; b16:0.921; f64:SKIP
_embedding_bag_forward_only: f16:0.923; f32:0.865; b16:0.925; f64:SKIP
_embedding_bag_backward: f16:1.362; f32:1.399; b16:1.489; f64:SKIP

Hygon DTK 26.04
_embedding_bag: f16:1.170; f32:1.129; b16:1.153; f64:1.174
_embedding_bag_forward_only: f16:1.139; f32:1.150; b16:1.172; f64:1.284
_embedding_bag_backward: f16:1.417; f32:2.011; b16:1.394; f64:1.874

MThreads MTT S5000
_embedding_bag: f16:0.821; f32:0.811; b16:0.855; f64:SKIP
_embedding_bag_forward_only: f16:1.108; f32:1.029; b16:1.153; f64:SKIP
_embedding_bag_backward: f16:1.194; f32:1.270; b16:1.315; f64:SKIP
```

The summary uses the latest completed measurements for each unchanged implementation path. NVIDIA public-forward FP64 uses its post-tuning rerun (with the later weighted-FP64 measurement replacing that subgroup). MThreads public-forward FP32 uses its post-tuning four-group run. Other NVIDIA/MThreads groups retain their completed measurements because their implementation paths are unchanged. Hygon uses the complete post-cache regression run.

[embedding-bag-perf-stdout.tar.gz](https://github.com/user-attachments/files/32431356/embedding-bag-pr-stdout.tar.gz)


winfan1314:
@modao1234 We are working on the asynchronous error-reporting issue on CoreX. Device assertions currently emit diagnostics, but the error is not reliably propagated to PyTorch.

We are investigating a solution that correctly reports invalid indices and other invalid inputs without introducing additional device synchronization or device-to-host readback, as these would cause significant performance regressions. Our goal is to preserve both reliable error handling and good operator performance on CoreX.

PR #6508, opened 2026-09-20
[KMCompiler][NVIDIA][Ascend][Hygon][Thead][MetaX][iluvatar] Add int dtype support and new test cases for pairwise_distance

zykwqs:
### PR Category
OP Test

### Type of Change
New Feature

### Description

1.Added support for int8 and uint8 dtypes in `pairwise_distance`.
2.Implemented internal dtype promotion to float32 for integer inputs to ensure correct computation.

### Performance

Expanded the test suite for the pairwise_distance operator to include low-precision data types, bringing the total number of test cases to 322. Added support for int8 and uint8 dtypes in pairwise_distance.The test results across different platforms are as follows:

NVIDIA (H20): 322 passed.
Hygon (BW1000): 312 passed, 10 skipped. 
MetaX (MC550): 312 passed, 110 skipped.
Ascend (910B): 302 passed, 20 skipped. 
Iluvatar (BI-V150): 322 passed.  
T-Head (PPU-ZW810E): 322 passed.


tests/test_pairwise_distance.py
```diff
@@ -191,3 +191,35 @@ def test_pairwise_distance_fp64(shape, p):
...
+        x1 = torch.randint(0, 255, shape, dtype=dtype, device=flag_gems.device)
+        x2 = torch.randint(0, 255, shape, dtype=dtype, device=flag_gems.device)
+
+    ref_x1 = utils.to_reference(x1, True)
+    ref_x2 = utils.to_reference(x2, True)
+
+    ref_out = _ref_pairwise_distance(ref_x1, ref_x2, p=p, eps=1e-6, keepdim=keepdim)
+    res_out = flag_gems.pairwise_distance(x1, x2, p=p, eps=1e-6, keepdim=keepdim)
+
+    utils.gems_assert_close(res_out, ref_out, torch.float32, atol=1.0)
```
Caeruleann, line 225: Why is the dtype `torch.float32`? atol so large.

zykwqs: Both PyTorch and FlagGems promote int8 to float32 for computation, so the output dtype is float32 as well.The atol can change，i will resubmit later.

zykwqs: When p is 0, 1.0, or 2.0, both PyTorch and FlagGems produce identical
results with zero error. The discrepancy only appears at p=1.5, where
PyTorch's reference implementation computes pairwise distance via a
sum-then-pow pipeline that inherently introduces ~0.297 absolute error
in float32. This error is unavoidable and stems from float32 precision
limits in the reference itself, not from any bug in the FlagGems kernel.

Both PyTorch and FlagGems promote integer inputs to float32 before
computation, so the final comparison is naturally performed in float32.
Relaxing atol to 0.35 accommodates this inherent float32 precision
limit while keeping the test meaningful for other p values.

PR #6512, opened 2026-09-20
ci: switch pre-check jobs from basic-runner-cpu to cpu-runner

gavin0x01:
cpu-runner is an already-running container, so the pre-check jobs no longer need to spin up the kunlunxin container image. Switch all 11 pre-check jobs (rule-check x9, unittest-required, code-style) to runs-on: cpu-runner and drop the container: blocks plus the now-moot proxy/container comments. Keep the GIT_CONFIG safe.directory env and the 'Ensure full working tree' healing steps — both remain valid for the reused workspace. Also change check-init-exports from bare 'python' to 'python3' so it no longer depends on a python->python3 symlink in the runner image.

### Progress


103yiran (approved):
LGTM

PR #6520, opened 2026-09-20
perf(kunlunxin): add a cached fast path for aten::_flash_attention_forward

llaboon:
### What
Add a host fast path for `aten::_flash_attention_forward` on the kunlunxin backend.

The generic entry rebuilds the full launch machinery on every call (~50 `fwd_params` attribute writes, an `args()` tuple, several scratch allocations and the libentry classification). The new entry captures the deepest `JITFunction.run` arguments on the first call for a given configuration and replays them with the caller's fresh input/output tensors. Any mismatch falls back to the generic path; `FG_FWD_FASTPATH=0` disables the fast path entirely.

### Effect
Small-shape latency ~0.216 ms -> ~0.116 ms in local measurements; outputs stay bit-identical (same kernel, same arguments, verified diff=0).

### Files
- `src/flag_gems/runtime/backend/_kunlunxin/ops/_flash_attention_forward.py` (new)
- `src/flag_gems/runtime/backend/_kunlunxin/ops/__init__.py` (registration)


bin913 (dismissed):

103yiran (dismissed):
LGTM

103yiran (approved):

PR #6521, opened 2026-09-20
[kunlunxin] Optimize greater_scalar, linalg_householder_product and softshrink kernels

lvxu2020:
This PR contains three Kunlunxin (XPU) backend performance optimizations.
Each commit is self-contained and independently verified against the official
benchmark matrix and the functional test suite (72 tests pass × 3 fresh-cache
runs + cold TRITON_CACHE_DIR, zero regression). All changes are
backend-local under `src/flag_gems/runtime/backend/_kunlunxin/ops/`; no public
API changes.

### 1. greater_scalar
Optimize the scalar-comparison path of `greater`:
- Enable `TRITONXPU_COMPARE_FUSION=1` and compare in the native 16-bit float
  dtype for the scalar path;
- Remove the upstream `_greater_scalar_fast` special path — its shape guard
  measures 3.1–3.2× slower, and keeping it prevents the dtype equal-weight
  metric from reaching 0.8.

dtype equal-weight Gems Speedup: 0.3307 (current master baseline) → 0.8058,
converging to 0.804–0.814 after excluding shared-machine bandwidth-contention
outliers.

### 2. linalg_householder_product
Rewrite the XPU implementation:
- Column-per-program one-shot kernel for `M <= 8192` (k truncated to
  `min(col, K-1)`), plus a 3-launch-per-k pipeline for large M;
- Avoids the 2-D `axis=0` reduction that TritonXPU cannot lower and the
  dynamic-loop store-then-load reliability bug of the general implementation.

Gems Speedup (f32 benchmark matrix): 2.45× (HEAD transposed-W) → 5.11×,
reproducible over 3 runs. Functional: f32 7/7 clean × 4 runs; f64 cases are
device-side NON_BUG (`support_fp64=False`).

### 3. softshrink
Rewrite the XPU kernel as a single `@pointwise_dynamic` kernel using the
select-free `x - clamp(x, -lambd, lambd)` form (min/max propagate NaN for
free, matching `F.softshrink(NaN) = NaN`), replacing the fixed-BLOCK bare
`triton.jit` kernel with dual path + bit-level NaN guard.

Gems Speedup: 0.3560 → 1.0669 (softshrink) / 1.0633 (softshrink_out); large
shapes 1.4–1.5× better than HEAD baseline. Benchmark matrix 48/48 SUCCESS.


Caeruleann:
solve the ci problems

103yiran (approved):
LGTM

PR #6522, opened 2026-09-20
glu: add Ascend TLE adaptation and enable TLE path for iluvatar

lzllx123:
## Summary

- glu: add an Ascend backend adaptation using `tle.dsa.extract_slice`.
- iluvatar: export the TLE path for the iluvatar backend  (`tle_enabled=True` in the vendor descriptor).


sunnycase (approved):
LGTM

PR #6539, opened 2026-09-21
[Ascend] Fix in-place pointwise kernels re-executed by grid padding

### LINKED ISSUE #6446: [Ascend] Tensor.exponential_() produces 1024 NaNs for large tensors (n >= 65536) — breaks LLM random sampling
# [Ascend] `Tensor.exponential_()` produces 1024 NaNs for large tensors (n ≥ 65536) — breaks all LLM random sampling

**Repo:** https://github.com/flagos-ai/FlagGems

## Summary

FlagGems' `exponential_` op (`flag_gems/ops/exponential_.py`, Triton kernel) on
the Ascend backend writes **exactly 1024 `NaN` values** into large output tensors.
For tensors with a trailing dimension `n ≥ 65536` the last (or one) 1024-element
block comes out as `NaN` instead of an exponential-distributed value; smaller
tensors are fine.

This silently breaks **all random sampling** in vLLM on Ascend. vLLM's
`random_sample` does `q = torch.empty_like(probs); q.exponential_(); probs.div_(q).argmax(-1)`.
With a vocab of 151936, `q` gets NaNs, `probs / NaN = NaN`, and `argmax` over an
all-NaN row returns index 0 — so every sampled token is id 0 and completions
degenerate to `"!!!!"`. Greedy decoding (pure `argmax`, no `exponential_`) is
unaffected.

## Environment

| Component | Version |
|---|---|
| flag_gems | `5.4.0rc2.post3+gad89ec477` |
| flagtree | `0.7.0rc2+ascend3.5` |
| vllm | `0.20.2+flagos` |
| torch / torch_npu | `2.10.0` |
| CANN | `9.0.0` |
| Device | `Ascend910_9382` (910C) |

## Minimal reproduction

```python
import torch, torch_npu, flag_gems
flag_gems.enable()
d = torch.device("npu")

for n in [2048, 4096, 8192, 16384, 32768, 65536, 131072, 151936]:
    q = torch.empty((1, n), dtype=torch.float32, device=d)
    q.exponential_()
    print(n, "nan=", int(q.isnan().sum()))
```

Output:

```
2048   nan=0
4096   nan=0
8192   nan=0
16384  nan=0
32768  nan=0        # OK up to here
65536  nan=1024     # NaNs appear
131072 nan=1024
151936 nan=1024     # real Qwen3 vocab size
```

The NaN count is a constant **1024** regardless of tensor size once the
threshold (somewhere in 32768 < n ≤ 65536) is crossed — pointing at a
tiling / block-boundary bug in the kernel (one 1024-element block's philox /
`safe_fast_log` output is not written correctly). `uniform_()` on the identical
tensor is fine, so this is specific to `exponential_`, not a general RNG or
memory-layout issue.

## How it was found (in a live vLLM server)

Instrumenting `random_sample` in a running TP=2 server, for a
`temperature=1.0` request:

```
[DIAG] q_dtype=float32 q_contig=True q_isnan=1024 exp_nan=1024 uni_nan=0 \
       pmax=0.41 psum=1.0 pnan=0 arg=[0]
```

i.e. `probs` is correct (sums to 1, no NaN), but `q` from `exponential_()` has
1024 NaNs, so the sampled token collapses to 0. `uniform_()` on the same tensor
in the same process returns valid values.

## Impact

No usable random sampling on Ascend for any model with a large vocab (Qwen3,
etc.). Combined with the separate Triton top-k/top-p compile crash
(FlagTree issue: `_topk_topp_kernel` / `ConvertTritonIRToLinalgIR PassManager`),
there is currently **no working sampling path** on this Ascend stack — only
greedy decoding works.

## Ask

Fix the block-boundary handling in the Ascend `exponential_` Triton kernel so
large tensors don't emit NaN. Happy to provide the dumped kernel or run further
diagnostics.

bin913:
### PR Category

Operator, OP Test

### Type of Change

Bug Fix

### Description

**Symptom.** On Ascend, `Tensor.exponential_()` wrote 1024 `NaN`s into the
beginning of its output for large tensors. With `max()` of the result being
`NaN`, vLLM's random sampling always returned token id 0.

**Root cause.** `triton-ascend` does not queue programs beyond the number of
vector cores the device has. When a launch uses more CTAs than that, the grid is
padded up to the next multiple of the core count, and every padding CTA reports
`program_id == 0`. The Ascend codegen config capped `max_grid_size[0]` at `48`,
which is neither `<=` the 40 vector cores of an Ascend910 nor a multiple of it,
so any pointwise launch that needed more than 40 tiles was padded and its first
tile re-executed.

For an in-place kernel the repeated execution applies the update more than once.
`log_` (the generic implementation, not an Ascend-specific one) launched
`cdiv(n_elements, 1024)` programs with no cap, so at `n = 41984` the grid is 41:
triton-ascend padded it to 80 and ran program 0 thirty-three times, leaving the
first block as `log(log(x))` — `NaN` for `x < 1`. `exponential_` is the visible
victim because it is listed in `CUSTOMIZED_UNUSED_OPS`, so the torch_npu
composite implementation runs: `uniform_()` followed by `log_()`. `uniform_`
uses `cdiv(N, 4096)` tiles and stays under the core count, which is why only the
`log_` half corrupted the result.

The duplication is directly observable with an atomic counter kernel on
Ascend910B (`tl.atomic_add(counter + pid, 1)`):

| grid | program 0 executions |
| ---- | -------------------- |
| 32   | 1                    |
| 40   | 1                    |
| 48   | 33                   |
| 64   | 17                   |
| 80   | 1                    |

**Fix.**

1. `src/flag_gems/utils/codegen_config_utils.py`: cap the Ascend grid at the
   backend's real core count, so `num_ctas = min(max_grid_size[0], num_tiles)`
   stays inside the range triton-ascend launches as a single, duplicate-free
   wave. The `CORE_NUM` lookup is guarded by `vendor_name` because that attribute
   only exists on the Ascend backend — the same idiom the Cambricon entry of the
   same dict already uses (`CORE_NUM` is not present on `_nvidia`, and the dict
   is built at import time on every backend).
2. `src/flag_gems/ops/log_.py`: make the kernel grid-stride and cap its launch at
   `max_grid_size[0]`, so an in-place kernel stays correct even on a backend that
   limits the grid.
3. `tests/test_log_.py`: add `test_log_large_inplace`, a regression test over
   sizes that straddle the 40-tile boundary. It calls the gems op directly
   (`flag_gems.log_(inp)`) so it does not depend on dispatch being enabled by
   the test harness.

**Verification** (Ascend910B, CANN 9.0.0, triton-ascend 3.5, `ASCEND_RT_VISIBLE_DEVICES=0`).

Reproduction, `log_` in-place versus `exponential_`, before / after:

| input | `log_` NaNs before | `log_` NaNs after | `exponential_` NaNs before | `exponential_` NaNs after |
| ----- | ------------------ | ----------------- | -------------------------- | ------------------------- |
| 32768   | 0    | 0 | 0    | 0 |
| 40960   | 0    | 0 | 0    | 0 |
| 41984   | 1024 | 0 | 1024 | 0 |
| 49152   | 1024 | 0 | 1024 | 0 |
| 65536   | 1024 | 0 | 1024 | 0 |
| 151936  | 1024 | 0 | 1024 | 0 |
| (2, 151936)  | — | — | 1024 | 0 |
| (64, 151936) | — | — | 1024 | 0 |

Before the fix the NaNs are always exactly `[0..1023]`, i.e. the first tile, and
`exponential_`'s `max()` is `NaN`; after the fix `max()` is back to a normal
sample (`11.4`–`14.1` across these sizes). `uniform_` is unaffected in both
cases, which is expected given it stays under the core count.

Regression tests (`pytest -q tests/<file>`, same device):

| test file | master | this PR |
| --------- | ------ | ------- |
| test_log_.py      | 10 failed, 20 passed | 1 failed, 33 passed |
| test_log.py       | — | 18 passed |
| test_uniform_.py  | — | 3 passed |
| test_multinomial.py | — | 80 passed |
| test_add.py       | — | 1010 passed, 48 skipped |
| test_sub.py       | — | 1010 passed, 48 skipped |
| test_neg.py, test_abs.py, test_sqrt.py, test_ge.py | — | 36 passed each |
| test_masked_fill.py | — | 594 passed |
| test_mul.py       | 45 failed, 221 passed | 44 failed, 222 passed |
| test_div.py       | 436 failed, 556 passed | 252 failed, 740 passed |

`test_log_large_inplace` is not vacuous: the same file run against master fails
on 3 of its 4 sizes (`41984`, `65536`, `151936` — all needing more than 40 tiles
without being a multiple of 40) and passes on `40960`, whose 40 tiles fit
exactly; with this PR all 4 pass.

The single remaining `test_log_.py` failure is
`test_log_unsupported_dtype_raises[int16]`, which exercises the **native** torch
path before any gems kernel is reached; it fails identically on master
and is unrelated to this change. The residual `test_mul.py` / `test_div.py`
failures are pre-existing on this backend; this PR reduces them (test_div: -184
failures) because the same padding bug was corrupting other in-place kernels.

### Issue

- Resolves #6446

### Progress

### Performance

`max_grid_size[0]` only changes the number of programs a pointwise launch uses,
and `min(40, num_tiles)` is still far above the tile count of these shapes, so
there is no throughput change. In-place `log_`, fp32, 20 iterations:

| numel | master (cfg 48) | this PR (cfg 40) |
| ----- | --------------- | ---------------- |
| 1048576  | 140.8 us | 136.7 us |
| 4194304  | 155.8 us | 134.8 us |
| 16777216 | 151.9 us | 157.2 us |

The differences are run-to-run noise on this device (±10% on the small sizes).


0x45f (approved):

src/flag_gems/ops/log_.py
```diff
@@ -20,24 +20,47 @@
...
+    # already honours (Ascend sets it to its vector-core count). Never launch
+    # more programs than that.
+    n_blocks = triton.cdiv(n_elements, _BLOCK_SIZE)
+    max_programs = get_codegen_config().max_grid_size[0]
+    grid = (min(n_blocks, max_programs),)
+    with torch_device_fn.device(x.device):
+        log_kernel_[grid](x, n_elements, n_blocks, BLOCK_SIZE=_BLOCK_SIZE)
 
 
 def log_(*args, **kwargs):
```
0x45f, line 66: please add new log_ impl for ascend in `src/flag_gems/runtime/backend/_ascend/ops` dir

PR #6544, opened 2026-09-21
fix(kunlunxin): erfc/erfinv/bucketize/threshold XPU kernel fixes

Seattle1917:
- erfc/special_erfc: reuse the erf odd-polynomial path (drop extern libdevice erfc, no exp); branch-free min/max saturation gated by a finite tile-sum NaN/Inf check; add unmasked path for tile-aligned shapes. special_erfc now routes to this fast erfc (dropped the duplicate KernelGen special_erfc import that shadowed it).
- erfinv: full-domain polynomial replaces log-form/Clenshaw-24 -- fp32 Chebyshev-24 as two parallel deg-12 Clenshaw chains (serial depth 48->24), fp16/bf16 deg-16 power basis; edge clamp min(|x|,1) instead of tl.where; size-adaptive tile + NEED_MASK.
- special_erfinv: delegate to the tuned erfinv override, drop the separate poly kernel.
- bucketize: pass boundaries as scalar kernel args for <=8 boundaries on large shapes (removes per-iteration gm2lm, ~18x); keep pointer-loading linear scan for >8 boundaries / small shapes.
- threshold(backward): saturating 0/1 clamp mask instead of tensor-vs-scalar compare; dispatch by numel (small -> grid-stride launch to cut the launch floor, large -> pointwise_dynamic + CodeGenConfig).


bin913 (approved):

src/flag_gems/runtime/backend/_kunlunxin/ops/erfinv.py
```diff
@@ -181,41 +217,33 @@ def _launch_erfinv(x: torch.Tensor, out: torch.Tensor):
...
-        NEED_MASK=need_mask,
+        num_warps=num_warps,
+        unroll_num=UNROLL_NUM,
+        buffer_size_limit=BUFFER_SIZE_LIMIT,
+        isCloseMemoryAsync=IS_CLOSE_MEMORY_ASYNC,
     )
 
 
 def erfinv(x: torch.Tensor):
     """Inverse error function (aten::erfinv)."""
```
103yiran, line 243: Add log at begin, please.

PR #6554, opened 2026-09-21
fix(kunlunxin): restore divide/true_divide_tensor_ aliases dropped in #6264

llaboon:
### PR Category
Operator

### Type of Change
Bug Fix

### Description
#6264 rebuilt `_kunlunxin/ops/div.py` and dropped the `divide` and `true_divide_tensor_` entry points, but `_kunlunxin/ops/__init__.py` still imports and re-exports them, so the backend overlay fails to import (`cannot import name 'divide' from _kunlunxin.ops.div`) and the whole backend is unusable.

#6555 restored both names on master as plain aliases. This PR keeps the same two entry points as explicit wrappers so they additionally emit the dispatch-contract log lines expected by `tests/test_divide.py` and `tests/test_true_divide.py` (same pattern as `true_divide_tensor`), while routing to the tuned `true_divide` / `true_divide_`. Rebased onto current master: the only behavioral delta over master is the two log lines.

### Test plan
- `python -c "import flag_gems"` on kunlunxin
- `pytest tests/test_divide.py tests/test_true_divide.py tests/test_div.py --ref cpu`


llaboon:
The single red check on the current revision (`test-kunlunxin`) failed before reaching the tests: the job died in the runner-side checkout step (`Can't find 'action.yml' ... under .github/actions/checkout-retry`), i.e. an infrastructure flake on the shared runner, not a code failure. Re-triggered. Since the stack's vendor-package import and every kunlunxin backend-test run are blocked on master until this lands, a quick review would be appreciated.

llaboon:
The `test-kunlunxin` red is the runner-side workspace issue tracked in #6567 (every kunlunxin job has been failing in `./.github/actions/checkout-retry` before setup since 2026-09-21 ~14:30 UTC; the job never reaches the tests). This PR will be re-run once the runner is cleaned.

PR #6556, opened 2026-09-21
[Ascend] Fix in-place pointwise kernels re-executed by grid padding (5.4.0-rc2 backport of #6539)

### LINKED ISSUE #6446: [Ascend] Tensor.exponential_() produces 1024 NaNs for large tensors (n >= 65536) — breaks LLM random sampling

bin913:
### PR Category

Operator, OP Test

### Type of Change

Bug Fix

### Description

**Backport of #6539 to the `5.4.0-rc2` release branch.** The branch has the same
defect and the same `exponential_` entry in `CUSTOMIZED_UNUSED_OPS`, so the fix
applies unchanged; only the surrounding code differs (see *Porting notes*).

**Symptom.** On Ascend, `Tensor.exponential_()` wrote 1024 `NaN`s into the
beginning of its output for large tensors. With `max()` of the result being
`NaN`, vLLM's random sampling always returned token id 0.

**Root cause.** `triton-ascend` does not queue programs beyond the number of
vector cores the device has. When a launch uses more CTAs than that, the grid is
padded up to the next multiple of the core count, and every padding CTA reports
`program_id == 0`. The Ascend codegen config capped `max_grid_size[0]` at `48`,
which is neither `<=` the 40 vector cores of an Ascend910 nor a multiple of it,
so any pointwise launch that needed more than 40 tiles was padded and its first
tile re-executed.

For an in-place kernel the repeated execution applies the update more than once.
`log_` (the generic implementation, not an Ascend-specific one) launched
`cdiv(n_elements, 1024)` programs with no cap, so at `n = 41984` the grid is 41:
triton-ascend padded it to 80 and ran program 0 thirty-three times, leaving the
first block as `log(log(x))` — `NaN` for `x < 1`. `exponential_` is the visible
victim because it is listed in `CUSTOMIZED_UNUSED_OPS`, so the torch_npu
composite implementation runs: `uniform_()` followed by `log_()`. `uniform_`
uses `cdiv(N, 4096)` tiles and stays under the core count, which is why only the
`log_` half corrupted the result.

The duplication is directly observable with an atomic counter kernel on
Ascend910B (`tl.atomic_add(counter + pid, 1)`):

| grid | program 0 executions |
| ---- | -------------------- |
| 32   | 1                    |
| 40   | 1                    |
| 48   | 33                   |
| 64   | 17                   |
| 80   | 1                    |

**Fix.**

1. `src/flag_gems/utils/codegen_config_utils.py`: cap the Ascend grid at the
   backend's real core count, so `num_ctas = min(max_grid_size[0], num_tiles)`
   stays inside the range triton-ascend launches as a single, duplicate-free
   wave. The `CORE_NUM` lookup is guarded by `vendor_name` because that attribute
   only exists on the Ascend backend — the same idiom the Cambricon entry of the
   same dict already uses (`CORE_NUM` is not present on `_nvidia`, and the dict
   is built at import time on every backend).
2. `src/flag_gems/ops/log_.py`: make the kernel grid-stride and cap its launch at
   `max_grid_size[0]`, so an in-place kernel stays correct even on a backend that
   limits the grid.
3. `tests/test_log_.py`: add `test_log_large_inplace`, a regression test over
   sizes that straddle the 40-tile boundary. It calls the gems op directly
   (`flag_gems.log_(inp)`) so it does not depend on dispatch being enabled by
   the test harness.

**Porting notes.** `log_.py` and `tests/test_log_.py` are byte-identical between
master and `5.4.0-rc2`, so they are transplanted as-is. `5.4.0-rc2` still has the
older `get_heuristics_for_num_warps(tile_size)` instead of master's
`get_heuristics_for_num_warps_fn()` refactor, so the rc2 file is kept verbatim
and only the `vendors.ASCEND` entry of `CODEGEN_COFIGS` is rewritten — the
backport diff touches nothing else.

**Verification** (Ascend910B, CANN 9.0.0, triton-ascend 3.5, `ASCEND_RT_VISIBLE_DEVICES=0`).

The rc2 flavour of `codegen_config_utils.py` was exercised on hardware by
swapping it into a checkout that has the patched `log_.py` and
`tests/test_log_.py`:

| `codegen_config_utils.py` | `max_grid_size` | `test_log_large_inplace` | `tests/test_log_.py` |
| ------------------------- | --------------- | ------------------------ | -------------------- |
| `5.4.0-rc2` as-is | `(48, 1, 1)` | 3 failed, 1 passed | — |
| this PR           | `(40, 1, 1)` | **4 passed** | 1 failed, 33 passed |

The 3 failures are `41984`, `65536` and `151936`; `40960` passes in both cases
because its 40 tiles fit exactly and never trigger padding. The single remaining
`tests/test_log_.py` failure is `test_log_unsupported_dtype_raises[int16]`, which
exercises the **native** torch path before any gems kernel is reached and fails
identically on the unpatched branch.

Reproduction, `log_` in-place versus `exponential_`, before / after:

| input | `log_` NaNs before | `log_` NaNs after | `exponential_` NaNs before | `exponential_` NaNs after |
| ----- | ------------------ | ----------------- | -------------------------- | ------------------------- |
| 32768   | 0    | 0 | 0    | 0 |
| 40960   | 0    | 0 | 0    | 0 |
| 41984   | 1024 | 0 | 1024 | 0 |
| 49152   | 1024 | 0 | 1024 | 0 |
| 65536   | 1024 | 0 | 1024 | 0 |
| 151936  | 1024 | 0 | 1024 | 0 |
| (2, 151936)  | — | — | 1024 | 0 |
| (64, 151936) | — | — | 1024 | 0 |

Before the fix the NaNs are always exactly `[0..1023]`, i.e. the first tile, and
`exponential_`'s `max()` is `NaN`; after the fix `max()` is back to a normal
sample (`11.4`–`14.1` across these sizes). `uniform_` is unaffected in both
cases, which is expected given it stays under the core count.

### Issue

- Resolves #6446
- Backport of #6539

### Progress

### Performance

`max_grid_size[0]` only changes the number of programs a pointwise launch uses,
and `min(40, num_tiles)` is still far above the tile count of these shapes, so
there is no throughput change. In-place `log_`, fp32, 20 iterations:

| numel | cfg 48 | cfg 40 |
| ----- | ------ | ------ |
| 1048576  | 140.8 us | 136.7 us |
| 4194304  | 155.8 us | 134.8 us |
| 16777216 | 151.9 us | 157.2 us |

The differences are run-to-run noise on this device (±10% on the small sizes).


bin913:
Rebased onto the current `5.4.0-rc2` head.

**Before:** branch was cut from `de9c2c130a`
**Now:** parent is `6ed2f39071` (`fixbug: issue#6420 (#6566)`)

`5.4.0-rc2` had moved by two commits since the backport was first cut, and neither
overlaps with this PR:

```
  31d870ffce  [Ascend] fix: improve Ascend cat/concat dispatch and implementation
  6ed2f39071  fixbug: issue#6420 (#6566)

  src/flag_gems/runtime/backend/_ascend/ops/__init__.py   modified  +2   -0
  src/flag_gems/runtime/backend/_ascend/ops/cat.py        modified  +203 -93
  src/flag_gems/runtime/backend/_ascend/ops/concat.py     added     +232 -0
  src/flag_gems/utils/libentry.py                         modified  +16  -0
```

I re-checked the transplant assumptions against today's rc2 rather than trusting
the originals. For each of the three files, rc2 was compared byte-for-byte with
the master commit #6539 was written against (`916e63df17`):

| file | rc2 == #6539 base | how it is ported |
|---|---|---|
| `src/flag_gems/ops/log_.py` | yes (`847ecaf72f`) | transplanted verbatim from #6539's head |
| `tests/test_log_.py` | yes (`f4e8163211`) | transplanted verbatim from #6539's head |
| `src/flag_gems/utils/codegen_config_utils.py` | no | rc2 file kept, only the `vendors.ASCEND` entry rewritten |

`#6539` itself has not changed: its head is still `7e0e8180` (2026-09-21
09:08:26Z), and the two transplanted files hash to the same blobs as before the
rebase, so no 910B result is invalidated -- the file contents are byte-identical
and only the parent moved.

The branch is now `ahead_by=1, behind_by=0` vs `5.4.0-rc2`, still 3 files
`+75 -21`. No new push to `#6539` is needed.

PR #6557, opened 2026-09-21
fix(kunlunxin): restore divide/true_divide_tensor_ exports in div.py

RobertLuobo:
PR #6264 removed the module-level `divide` and `true_divide_tensor_` functions from _kunlunxin/ops/div.py but left their imports in _kunlunxin/ops/__init__.py, breaking `import flag_gems`. Restore both thin aliases (delegating to the existing true_divide/true_divide_).

### Progress


llaboon:
Two notes on this one:

1. Contract tests: under `use_gems()` on this backend, `tests/test_divide.py` asserts `GEMS DIVIDE` on logger `flag_gems.ops.divide`, and the in-place dispatch case in `tests/test_true_divide.py` asserts `GEMS TRUE_DIVIDE_` on logger `flag_gems.ops.true_divide_`. caplog level-filters the module logger, so each restored entry also needs the generic contract line (same pattern as the existing `true_divide_tensor` entry). #6554 is the same restore plus those two lines (+18/-0) and was verified on a kunlunxin machine (import OK; 324 passed for `tests/test_divide.py tests/test_true_divide.py`). Either direction works - happy to fold whichever you prefer.

2. The `test-kunlunxin` red here (and on #6554) is runner-side, not code: since ~14:30 today every kunlunxin job dies in `./.github/actions/checkout-retry` before setup (`Can't find 'action.yml' ... under .github/actions/checkout-retry`), while earlier runs reached the tests normally. The runner workspace looks stuck on leftover sparse-checkout/skip-worktree state, so fresh runs keep failing until it is cleaned.

PR #6563, opened 2026-09-22
fix(kunlunxin): index_fill scatter 1-D rewrite + i64 offsets + searchsorted fallback

lvxu2020:
## Summary

Fixes two issues in the Kunlunxin backend:

1. **index_fill / index_fill_**: rewrite the scatter path as 1-D to fix a mis-lowering on XPU where the 2-D tile silently dropped lanes; also fix an i32 address wrap that caused out-of-bounds stores on large tensors, with a small performance gain
2. **searchsorted family**: the generic implementation fails to compile on XPU (unimplemented in triton-xpu), so register it in `CUSTOMIZED_UNUSED_OPS` to fall back to the native path under `use_gems` instead of raising at call time

## index_fill / index_fill_

- **Scatter kernel rewritten as 1-D (one position per lane)**: the original 2-D (BLOCK_M x BLOCK_N) tile is mis-lowered on XPU when `outer < BLOCK_M` (e.g. shape (1,2), dim=1 with duplicate indices silently drops lanes and leaves the last columns unfilled). The 1-D form clamps every address in-bounds and is also faster (7-20%)
- **i64 offsets**: `index_fill_copy_kernel` wrapped OOB with i32 `pid * BLOCK` when `numel > 2^31`; switched to i64
- **small_inner mask changed to pure vector AND**: the previous `tl.where` folding dropped valid lanes on XPU
- **`_SMALL_INNER_BLOCK` 128 → 512**: better under the official cold-cache do_bench (~64k → ~16k programs); the 128 advantage only appears in hot-cache micro-benchmarks, an artifact

### Verification

- `tests/test_index_fill.py` 3x230 passed (independent cold cache + `--ref cpu`)
- Semantic scan: 2566 cases, BAD=0
- `> 2^31` overflow probe: 0 bad
- Performance (fp32 is the only measurable baseline; fp16/bf16/bool native xdnn NOT IMPLEMENTED): index_fill 0.9702 → 0.9868, index_fill_ 1.3057 → 1.3198 (both >= 0.8, no regression)

## searchsorted

The generic searchsorted implementation cannot compile on XPU, so the four aten variants (`searchsorted` / `.out` / `.Scalar` / `.Scalar.out`) are registered in `CUSTOMIZED_UNUSED_OPS` to fall back to the native path under `use_gems`.

## Testing

- index_fill: see verification data above
- searchsorted: falls back to native via unused-ops registration; no longer raises at call time


103yiran (approved):
LGTM

PR #6571, opened 2026-09-22
[KMCompiler][Ascend]fix(benchmark): use operator mode for nonzero_static on Ascend

Mike7180:
### What

  Apply the `Config.mode = consts.BenchMode.OPERATOR` override for  `nonzero_static` when running on Ascend, matching the workaround already present in `test_det.py`, `test_linalg_det.py`, `test_linalg_lu.py`,  `test_linalg_lu_factor.py`, `test_linalg_lu_factor_ex.py`,  `test_linalg_solve_triangular.py`, `test_igammac.py` and  `test_conj_physical.py`.

  ### Why

  On Ascend the default KERNEL-mode `do_bench_npu` reports the arithmetic mean of per-kernel durations from `kernel_details.csv` rather than end-to-end latency. `nonzero_static` spans several kernels per call (2-3), so the reported speedup is scaled by `N_gems / N_base` and is not comparable with other platforms. The same path also drives the profiler
  without a schedule, emitting an `Incorrect schedule` warning for every  case.

  Operator mode measures end-to-end latency with `time.time()` around synchronized iterations and is not vendor-gated, so it matches what the non-Ascend platforms report.

  ### Scope

  Benchmark-only change, 11 lines in `benchmark/test_nonzero_static.py`. No operator implementation is touched; device-side timings are unchanged.

  ### Verification

  Ran on Ascend with the default (KERNEL) config against this base; the override takes effect and the run reports `mode=operator`.


103yiran (approved):
LGTM

PR #6573, opened 2026-09-22
fix(kunlunxin): skip FP8 tests on kunlunxin, keep non-FP8 dtypes

lvxu2020:
## Summary

The Kunlunxin (XPU) platform has no FP8 hardware (SM89+/SM90+/Blackwell missing),
so FP8 test cases can never produce meaningful results: they either skip via
capability checks or pass only against a reference that is not FP8-capable,
leaving the operator test report polluted with SKIP / no-benchmark / grey-zone
entries.

This PR skips only the FP8 data type on kunlunxin for the 9 FP8-related operators
identified in the report. Other supported dtypes keep running, and no other
vendor is affected.

### System layer (mirrors the existing `fp64_enabled` chain)
- `src/flag_gems/runtime/backend/backend_utils.py`: add `fp8_enabled: bool = True`
  to `VendorDescriptor` (default True, so non-kunlunxin vendors behave as before).
- `src/flag_gems/runtime/backend/_kunlunxin/__init__.py`: set `fp8_enabled=False`
  for kunlunxin.
- `src/flag_gems/runtime/backend/device_finder.py`: expose
  `DeviceDetector.support_fp8`.
- `tests/accuracy_utils.py`: expose `fp8_is_supported`.

### Pure-FP8 operators — module-level skip via dtype list
Tests and benchmarks of the 6 pure-FP8 operators now use
`FP8_DTYPES = [torch.float8_e4m3fn] if fp8_is_supported else []` plus a
module-level `skipif` driven by the empty list (dtype-list convention, cf.
`test_cholesky_inverse`):
- act_quant_triton
- fp8_fp4_mega_moe
- fp8_fp4_mqa_logits
- fp8_mqa_logits
- fused_indexer_q_rope_quant
- stage_deepseek_v4_mega_moe_inputs

### Non-pure-FP8 operators — FP8 dtype filtered, others keep running
- `scaled_mm` / `scaled_mm_out`: FP8 cases were already filtered out by
  `_float8_dtypes()` on non-NVIDIA platforms; fp16/bf16/fp32 cases still run.
- `scaled_grouped_mm`: `_float8_dtypes()` now returns an empty list when FP8 is
  unsupported; fp16/fp32/bf16 cases still run.

## Scope

- Only test files and the vendor descriptor flag are touched; no operator
  implementation, dispatch, or `CUSTOMIZED_UNUSED_OPS` changes.
- TE/vLLM-dependent tests (`scaled_softmax`, `te_rmsnorm`, `hc_head_fused_kernel`)
  and real defects (e.g. `chunk_gated_delta_rule_fwd`) are out of scope.
- Other vendors are unaffected (`fp8_enabled` defaults to True).

## Testing

- All 13 modified test files and 4 system files pass `py_compile`.
- Verified `VendorDescriptor` defaults: nvidia `fp8_enabled=True`,
  kunlunxin `fp8_enabled=False`; other fields (fp64/bf16/int64/tle) unchanged.
- Behavior simulated for both configurations: kunlunxin skips FP8 cases only;
  non-kunlunxin keeps the original test behavior.


bin913:
@lvxu2020 Please fix failing code-style check

bin913 (dismissed):

bin913 (approved):

PR #6581, opened 2026-09-22
fix(benchmark): use isolated mode option for Ascend

wuwentao:
## Problem

Huawei Ascend benchmark runs were using `operator` mode even though `benchmark/conftest.py` declares the benchmark mode default as `kernel`.

There were two Ascend-specific issues:

- The benchmark option used the generic pytest `--mode` name for most vendors. Ascend/Triton environments can also define or handle generic mode-style options, so Ascend should follow Kunlunxin and use the isolated FlagGems option name `--fg_mode`.
- Even after passing `--fg_mode kernel`, `pytest -m add` still collects the whole `benchmark/` tree before deselecting non-`add` tests. Several Ascend benchmark modules changed `Config.mode` to `operator` at module import time. Those deselected modules polluted the shared benchmark `Config`, so the selected `add` benchmark then printed and recorded `mode=operator`.

## Fix

- Use `--fg_mode` for Ascend, preserving Kunlunxin's existing `--fg_mode` behavior.
- Make `tools/run_tests.py` explicitly pass `--fg_mode kernel` only for Ascend benchmark subprocesses. NVIDIA, MetaX, Hygon, MThreads, PPU, and other vendors do not get a new explicit mode argument.
- Use the same vendor-aware mode option in parallel BLAS benchmark subprocesses.
- Store the requested benchmark mode and reset `Config.mode` per test item, so pytest collection cannot leak mode mutations across tests.
- Convert the known Ascend-only operator-mode benchmark exceptions to a scoped pytest marker instead of module-import-time global mutation.
- Add regression coverage for the vendor mode option and Ascend-only operator-mode scoping.
- Update the Huawei 910B weekly test image to `harbor.baai.ac.cn/flaggems/flaggems-test-huawei-910b4:20260922`.

## Vendor impact

The behavior change is scoped to Ascend and Kunlunxin mode-option naming. Other vendors continue to use the existing `--mode` path and are not passed an explicit benchmark mode by `tools/run_tests.py`.

The per-test mode reset only changes Ascend tests marked with the new `ascend_operator_mode` marker to `operator`; unmarked Ascend tests use the requested CLI mode, normally `kernel`.

## Validation

Local/macOS:

- `PYTHONPYCACHEPREFIX=/private/tmp/flaggems-pycache python3 -m py_compile ...` passed for changed Python files.
- `git diff --check` passed.

Huawei 910B container:

- Container: `flaggems-pr6581-huawei`
- Image: `harbor.baai.ac.cn/flaggems/flaggems-test-huawei-910b4:20260922`
- Environment confirmed: `vendor=ascend`, `device=npu`, `npu_available=True`.
- `pytest tests/test_benchmark_skip_native.py -q`: `23 passed`.
- Full benchmark collection path:
  - Command: `pytest -m add --fg_mode kernel --level core --record json --output /tmp/benchmark-add-kernel.json --continue-on-collection-errors --dtypes float16 -s`
  - Result: `1 passed, 2085 deselected`
  - Output/JSON both recorded `mode=kernel`.
- `tools/run_tests.py --ops add --gpus 0 --dump-output --output-dir /tmp/flaggems-pr6581-results-fixed --color never`:
  - Accuracy: passed.
  - Benchmark subprocess command included `pytest -m "add" --fg_mode kernel ...`.
  - Benchmark output showed `mode=kernel` for float16, float32, and bfloat16.
  - Benchmark then failed on Ascend `complex64` random tensor creation: `aclnnInplaceNormal ... Tensor self not implemented for DT_COMPLEX64`; this is a backend dtype limitation, not caused by this PR.

## GitHub Actions

The previous `python-op` failure matched a current-PR test issue: the test asserted exact `speedup == 2.0`, while production code computes speedup with `EPS`, yielding `1.99999995`. This PR now uses `pytest.approx(2.0)`, and the same test file passes in the Huawei container. New Actions were triggered by commits `1cf8d7527`, `8a4e9f511`, and `f8002d00b`.

Processed by Codex (GPT-5).


wuwentao:
## Verification Update

The root cause was confirmed on the Huawei 910B container:

1. `pytest -m add --fg_mode kernel ...` collects the full `benchmark/` tree before deselecting non-`add` tests. Ascend benchmark modules that assigned `Config.mode = OPERATOR` at import time changed the shared global state, so `add` ran as operator despite the CLI option being kernel.
2. The fix stores the requested CLI mode, applies the Ascend-only operator exception per test item, and restores the requested mode after each test. The new `ascend_operator_mode` marker is also treated as a control marker for result attribution.

Huawei validation with image `harbor.baai.ac.cn/flaggems/flaggems-test-huawei-910b4:20260922`:

- `pytest tests/test_benchmark_skip_native.py -q`: `23 passed`.
- Full collection benchmark command with `--fg_mode kernel --dtypes float16`: `1 passed, 2085 deselected`; stdout and JSON both reported `mode=kernel`.
- `tools/run_tests.py --ops add --gpus 0 ...`: generated `pytest -m "add" --fg_mode kernel ...`; float16, float32, and bfloat16 benchmark results all reported `mode=kernel`.
- The remaining benchmark failure is Ascend `torch_npu` inability to create `complex64` random tensors (`aclnnInplaceNormal`, `DT_COMPLEX64` unsupported), unrelated to this PR's mode changes.

The previous GitHub Actions `python-op` failure was caused by a PR test's exact `speedup == 2.0` assertion, while the implementation adds `EPS` and returns `1.99999995`. It was reproduced in the Huawei container and fixed with `pytest.approx(2.0)`. New checks are running for commits `1cf8d7527` and `8a4e9f511`.

Codex (GPT-5)

PR #6583, opened 2026-09-22
ci(kunlunxin): reset reused workspace before checkout

gavin0x01:
## Problem

\`backend-tests (kunlunxin) / test-kunlunxin\` fails repeatedly before any test runs (~70s), e.g. [run 35712503378](https://github.com/flagos-ai/FlagGems/actions/runs/35712503378/job/106698702885). The failure is unrelated to the PR under test — it's an infrastructure issue on the reused self-hosted runner.

Root cause: \`klx-p800-1\` reuses its workspace across jobs. Stale \`skip-worktree\` bits on backend op files make \`git checkout --force\` refuse to overwrite them:

\`\`\`
git checkout --force refs/remotes/pull/<n>/merge
error: Path 'src/flag_gems/runtime/backend/_kunlunxin/ops/_thnn_fused_lstm_cell.py' not uptodate; will not remove from working tree.
error: Path 'src/flag_gems/runtime/backend/_kunlunxin/ops/upsample_nearest_exact2d.py' not uptodate; will not remove from working tree.
\`\`\`

Checkout aborts, so the local \`checkout-retry\` action can't be found (\`Can't find 'action.yml' ... under '.github/actions/checkout-retry'\`) and the job dies. \`checkout-retry\` can't help — every retry hits the same skip-worktree files.

## Fix

Add a **Reset reused workspace** step before checkout in \`test-kunlunxin\` that disables sparse-checkout, clears skip-worktree bits, and hard-resets/cleans the tree. Guarded with \`|| true\` and \`[ -d .git ] || exit 0\` so it's a harmless no-op on a clean or freshly-cloned workspace. Only the kunlunxin job is touched.

## Test

- \`actionlint\` passes (only pre-existing self-hosted runner-label warnings remain, unrelated to this change).
- Runtime verification requires the kunlunxin runner; will confirm via CI on this PR.


103yiran (approved):
LGTM

